PR #759, opened 2018-12-21
Library_Engine: Updates to library namespaces - structural sections

rwemay:
No test script - use CreateData components to test, if successful it should look like the below. New issue to be raised on asymmetric profiles. 

image


epignatelli:
This does not remove the old ones to me, I can't include it in the installer yet.

PR #774, opened 2019-01-18
Fixes #770 - (De)Serialize generic method parameters constraints

### LINKED ISSUE #770: Serialiser_Engine: Problem with (De)Serialising methods with generics
There is a problem (de)serialising methods with generics as inputs and outputs.

As an example:

```
        public static T SomeMethodName<T>(this T obj)
        {
            //Do stuff
        }
```

This makes the components referencing those kinds of method fail to read in correctly in Grasshopper.

When debugging, the code fails to match the input parameters on the following rows:

https://github.com/BHoM/BHoM_Engine/blob/5c26468ee5afc22d21100e4fc93d6423fd6313ad/Serialiser_Engine/Objects/BsonSerializers/MethodBaseSerializer.cs#L137-L139

where the parameters fail to match. The serialised part has the type `System.Object`, while the method has `T`.

Problem might lay deeper though, as have not looked to closely on this part of the code.

epignatelli:
Fixes #770 
The problem was that generic types with constraints were not handled properly. They were hardly serialised to their constraint instead of a richer generic. All the necessary methods were implemented but not used.
This pr:
- Gets parameters or its constraints if the method is generic in both serialisation and deserialisation routines.
- Exposes publicly as Create method and renames `RestoreMethod` to `MethodBase` as per compliance
- Exposes publicly `GenericTypeConstraint` in the Query Class (there might be an argument for moving this to the `Modify` class

We also have a weird duplicate in the BHoM_UI, which is never called in the project, is that legacy?
It would be good to let the UI call the Serializer_Engine, since the methods are now public, in case we need them.

Test file: 
https://burohappold.sharepoint.com/:u:/s/BHoM/EW-D-XdtLKlEueSYDNlxYMMBkjojTw_KtlNd08Bm3omeTw?e=9XuROr#
Definition of done, <u>**_all_**</u>:
1. Copy and paste `DeepClone` and `ShallowClone` methods
2. Save and reopen the grasshopper file
3. Create `DeepClone` component from the global menu (CTRL + SHIFT + B)
4. Create `DeepClone` component from a magic mutable component


epignatelli:
Thanks @adecler, sorry I didn't notice that.

I fixed it, but I think may require changes to the UI as well. 
The BHoM_UI calling this other method `GetConstructedType`. Am I missing something in the flow?

https://github.com/BHoM/BHoM_UI/blob/612676c78d67f400b9b1fc12c04082088a5bf15f/BHoM_UI/Templates/MethodCaller.cs#L98-L102

I switched to the engine method and it succeeds in (de)serialise `Create.PriorityQueue<T>`.
I'll push the code I wrote for the Engine so you can test it, for the BHoM_UI is just about switching to it (project's locked now).

PR #781, opened 2019-01-28
Reflection engine issue780 find generic type

### LINKED ISSUE #780: Reflection_Engine: Provide a way to find the lowest common shared type between a list of types
Given a `BH.oM.Geometry.Point` and a `BH.oM.Geometry.Polyline`, I expect to get `BH.oM.Geometry.IGeometry`

epignatelli:

Closes #780 

Accepts a list of types and returns the closest parent type to them all.

### Test files

https://burohappold.sharepoint.com/:u:/s/BHoM/EVYkpCYXqI5GkqiUoq8AqfAB_NN2AHR9mVa5ghMMwZAZQg?e=7moUiu

### Additional comments

I am not sure this code is required. We may find other solutions within the `System.Reflection` namespace that are shorter and do not require packing in a method. Monitoring https://github.com/BHoM/BHoM_UI/pull/69 will solve this.


Reflection_Engine/Query/ClassHierarchy.cs
```diff
@@ -0,0 +1,49 @@
...
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        public static List<Type> ClassHierarchy(this Type type)
+        {
+            if (type == null)
+                return new List<Type>();
+
+            List<Type> classHierarchy = new List<Type>();
+            for (Type current = type; current != null; current = current.GetInterfaces().FirstOrDefault() ?? current.BaseType)
```
adecler: The `FirstOrDefault` is incorrect here. You need to collect all of them (plus base type) and then apply `ClassHierarchy` recursively on those for it to give the correct result.

epignatelli: I found `InheritedTypes` doing exactly what I need, with a small fix. I will delete `ClassHierarchy`

Reflection_Engine/Query/CommonBaseType.cs
```diff
@@ -0,0 +1,58 @@
...
+                return typeof(object);
+
+            types = types.Where(t => t != null).ToList();
+            if (types.Count == 0)
+                return typeof(object);
+
+            List<List<Type>> inheritanceChain = new List<List<Type>>();
+
+            foreach (Type type in types.Distinct())
+                inheritanceChain.Add(type.InheritedTypes());
```
IsakNaslundBh: `type.InheritedTypes()` does not cover the actual type itself, so you probably want to add that to the list. 

Then, for the inherited types, not sure if it is getting out all the base types. If you for example are working on type `C` where `C : B` and `B : A`, I am not sure if  `A` will be added to the list.

epignatelli: Yes, agreed. If the BHoM_UI pr is ok with the latest changes (does not rely on this anymore) I think I am going to close this without merging - this was a bad rush.

epignatelli:
Closing this, since it is not critical and can be solved later on. 
Also referencing https://github.com/BHoM/BHoM_UI/issues/72 (thanks @IsakNaslundBh ) as direct application of #780

Issue #804, opened 2019-02-17
DataStructure_Engine: remove reference to Geometry_Engine


pawelbaran:
Currently Geometry_Engine is referenced to DataStructure_Engine - this should be the other way around as the geometry should be able to use data structures without causing circular dependency.

Change necessary to finalize  #594.

pawelbaran:
Closed with #805

Issue #803, opened 2019-02-17
DataStructure_Engine: Typo in DiscretePoint


pawelbaran:
Following [#415](https://github.com/BHoM/BHoM/issues/415)

pawelbaran:
Closed with #805

Issue #416, opened 2018-07-17
Spatial_Engine: Merge FabricatedBox and GeneralisedFabricatedBox (Profiles and SectionCurves)


DanielJonssonBH:
### This issue has been archived 

Historic issue archive can be found [here](https://burohappold.sharepoint.com/sites/BHoM/01_Master%20File/Forms/AllItems.aspx?id=%2Fsites%2FBHoM%2F01_Master%20File%2F999_BHoM_GitHub_History_Archive)

FraserGreenroyd:
Having reviewed this historic issue as part of the 2024 kick-off, and reviewing the code in todays format, it appears that the context of the original issue is no longer valid and that time has superseded the original requirements of this issue. As such, I am closing this issue as stale, however, if anyone feels differently, please feel free to reopen :smile:

Issue #562, opened 2018-08-22
Global: Do we need a Maths_Engine?


alexradne:
### This issue has been archived 

Historic issue archive can be found [here](https://burohappold.sharepoint.com/sites/BHoM/01_Master%20File/Forms/AllItems.aspx?id=%2Fsites%2FBHoM%2F01_Master%20File%2F999_BHoM_GitHub_History_Archive)

pawelbaran:
Coming back to the subject, I believe we do, same as Math_oM. I would be glad to implement `Matrix` class (it is on my list for a few months already), a few methods could also migrate (e.g. `RowEchelonForm` etc.). What do you guys think @al-fisher @epignatelli @IsakNaslundBh?

IsakNaslundBh:
I would agree

FraserGreenroyd:
What would a `Math_oM` and `Math_Engine` provide us that `System.Math` does not already offer?

rwemay:
@pawelbaran , @IsakNaslundBh , @alexradne , agree, I do think we need a math engine to carry things like matrix definitions and methods. I know @alexradne raised this some time ago, needing to store engine methods for the structural Equilibriate wrapper methods. 

@FraserGreenroyd , the system.math library only provides some relatively basic methods, there are a number of engineering applications that would make use of a core math library.

FraserGreenroyd:
> @FraserGreenroyd , the system.math library only provides some relatively basic methods, there are a number of engineering applications that would make use of a core math library.

In that case have we considered `Math.NET` or similar that might provide that functionality without us having to replicate anything? Unless we wanted full ownership on the replication/don't wish to handle additional conversions (thought a `Math.Net_Toolkit` could resolve that?)? I'm just keen to make sure no one reinvents a wheel when I'm sure we can't be the first people to have encountered these problems? :smile:

rwemay:
@FraserGreenroyd , yes could be. As usual, the preference will be to wrap rather than implement directly, so we can change dependencies later/manage multiple dependencies without these needing to be loaded multiple times.

epignatelli:
I do agree we need a `Math` namespace. I can't see any drawback in using an external library as well. There are plenty of them. Depending on how good we want them to be, they could not be .NET, so we may need a wrapper.

Using an existing one doesn't automatically mean that its methods are exposed to UI, though. So, we may still need a wrapper around that, or a rule to expose its methods (which I don't think we ever done before @adecler).

[...]

PR #805, opened 2019-02-17
DataStructure&Geometry_Engine-Issues594,803,804-TypoInDiscretePoint&MigrateDBSCAN

pawelbaran:
**To be merged together with [#416](https://github.com/BHoM/BHoM/pull/416)!**

Closing #594 
Closing #803 
Closing #804 

A generic DBSCAN algorithm capable of taking any object type and any method as metric function has been implemented - no drop in performance was observed on PointClusters method. Test file located [here](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?id=%2Fsites%2FBHoM%2F02_Current%2F12_Scripts%2F01_Test%20Scripts%2FBHoM_Engine%2FGeometry_Engine%2FBHoM_Engine-Issue594-ExposeGenericDBSCAN).

In order to remove Geometry_Engine reference from DataStructure_Engine, a new `PMSquareDistance` method has been implemented. It is a bit lame as it is exactly same method as `Point.SquareDistance`.

As far as I understand, the DataStructure_oM and Engine should not be limited to 3 dimensions (as they might work on abstract data, not only geometry), therefore I would recommend:
- replacing the geometrical `Point` in `LocalData.Position` with a generic, n-dimensional `DataPoint` with coordinates stored in an array of doubles
- making `DiscretePoint` n-dimensional with coordinates stored in an array of integers
- declaring the number of dimensions when creating an instance of `PointMatrix`
- adjusting all methods to work on n dimensions instead of 3 (X, Y, Z)

This would be an elegant way to keep the PointMatrix generic and completely independent of Geometry_oM & Engine. I could raise an issue in the BHoM and implement it in the next PR. What do you guys think? @epignatelli @al-fisher @IsakNaslundBh


pawelbaran:
Good points, @epignatelli. I thought that DataStructure_Engine is, by definition, a generic, data-specific engine - the one in which one would keep general matrix/sorting/clustering-related stuff. Am I wrong? If so, where should all that go?

Concerning the code structure, I have seen the private helper classes in other places (even the header is copied), so assumed that in certain cases it is allowed. Exposing DBSCAN in chunks seems to be quite pointless to me (it works as one entity). An alternative could be rewriting the code to have everything clustered in one method - not sure if it makes any sense either.

epignatelli:
I think the DataStructure refers specifically to computational constructs. The limit with math is quite thin.
I see this pr more related to a Mathematics Engine, it is actually pure math. 
It is great to have a generic clustering method!!

For the code itself, yes there are other places where we have private classes. They exist either because there was no other way to achieve the goal, or they're incorrect, and should be targeted by a compliance issue. The general guideline is something like "always public, if you can't, go private", but not the opposite. In this case I think we can avoid doing that. It does work as one entity, but it's also true that the smaller function may serve other purposes: mean-shift, hierarchical, and so on. I'd keep the modular structure we're used to.

pawelbaran:
@epignatelli maybe I am ignorant, but do we have Mathematics Engine? And which part of this PR would you relate to it? I would say DBSCAN in its generic form is not pure mathematics, but more algorithm/data structure stuff.

Concerning the code structure, would you recommend making the classes public and moving them to the BHoM? I am good with that, what do you think @al-fisher @adecler?

epignatelli:
No, sorry - there is no Math namespace at the moment, I meant to create it. I would related just #594 to that.

My proposal for the code is:
- Move `DBScanObject<T>` to the oM, make it public
- Move it's constructor the the `Create` class of the engine
- Make `ComputeClusters` public static in the `Compute` class
- Make `ExpandCluster` public static in the `Compute` class
- Make `RegionQuery` public static in the `Query` class
- Move `PointClusters` to the Geometry_Engine and maybe call it `DBPointClusters`?
- We might remove the redundancy in the names as well and make them more specific to DBScan operations

Now, in what namespace/project, I think it's another issue and I am up for discussion about it!

If we isolate #594 from the other issues, you should be able to solve the others in this pr quite quickly.

In general, as you suggested I'd wait for @adecler and @al-fisher 's inputs before doing anything.

pawelbaran:
Oke, let's wait for the others to comment. In general I agree with everything you proposed.

pawelbaran:
Further changes made with regard to this issue:
- `PointClusters` had been deprecated, `PointClustersDBSCAN` method added
- `DBSCANAlgorithm` and `DBSCANObject` classes have been added in the [BHoM](https://github.com/BHoM/BHoM/pull/416) together with Create methods in BHoM_Engine
- DBSCAN-related method `ComputeClusters` has been refactored to comply with BHoM

I would say let's merge this PR together with [#416](https://github.com/BHoM/BHoM/pull/416) and raise an issue concerning a duplicated `SquareDistance` method in DataStructure_Engine.

Would be nice if @al-fisher @rwemay @adecler gave their opinion - this one is blocking the Geometry_Engine, which is starting to be an obstacle. Thanks!

epignatelli:
Nice, thanks @pawelbaran for addressing those! Having a generic `Func<T, T, bool>` potentially allows also to specify a custom function, and compile it on the fly. Looks great!

Still need to check the reference issue, but happy for the changes to the DBScan!

al-fisher:
Previously raised Maths engine issue: https://github.com/BHoM/BHoM_Engine/issues/562

pawelbaran:
@al-fisher thanks for merging. I let myself reopen the discussion concerning the Maths engine & oM - please see #562. Concerning public/private, so far it was not an issue for me at all.

PR #837, opened 2019-02-27
Environment_Engine/Architecture_Engine - add names to create methods

### LINKED ISSUE #800: Architecture Engine - Create-methods with Name (Grid, Level)
Create methods for Levels and Grids should have a Name parameter. Currently they do not. 
```c#
public static Level Level(double elevation, string name)
        {
            return new Level
            {
                Elevation = elevation,
                Name = name
            };
        }
```

### LINKED ISSUE #836: Environment_Engine: Name parameter for material and construction
To aid creation of Materials and Constructions in Rhino -> BHoM workflow we need names set for materials and constructions - similar to #800

FraserGreenroyd:
Fixes #800 
Fixes #836 

Adds name input parameters to creation methods of:
 - Grid (Architecture)
 - Level (Architecture)
 - Construction (Environment)
 - Material (Environment)

#### Added
 - Added Create Method with name parameter for `Grid` and `Level` in the `Architecture_Engine`
 - Updated Create Method to include name parameter for `Construction` and `Material` in the `Environment_Engine`

### DIFF
--- a/Architecture_Engine/Create/Grid.cs
+++ b/Architecture_Engine/Create/Grid.cs
@@ -50,5 +50,22 @@ namespace BH.Engine.Architecture.Elements
         }
 
         /***************************************************/
+
+        public static Grid Grid(ICurve curve, string name)
+        {
+            return new Grid
+            {
+                Curve = Geometry.Modify.IProject(curve, Plane.XY),
+                Name = name,
+            };
+        }
+
+        /***************************************************/
+
+        public static Grid Grid(Point origin, Vector direction, string name, double length = 20)
+        {
+            Line line = new Line { Start = new Point { X = origin.X, Y = origin.Y, Z = 0 }, End = origin + new Vector { X = direction.X, Y = direction.Y, Z = 0 }.Normalise() * length };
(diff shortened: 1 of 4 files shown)


Architecture_Engine/Create/Grid.cs
```diff
@@ -50,5 +50,22 @@ public static Grid Grid(Point origin, Vector direction, double length = 20)
...
+            return new Grid
+            {
+                Curve = Geometry.Modify.IProject(curve, Plane.XY),
+                Name = name,
+            };
+        }
+
+        /***************************************************/
+
+        public static Grid Grid(Point origin, Vector direction, string name, double length = 20)
```
alexradne: I would suggest having the name parameter being optional instead of having two create methods. 
I also think name should be the last parameter. I think this is the case for most create methods.

FraserGreenroyd: As discussed - we cannot change the method signatures of existing methods because it would cause problems with existing scripts so we need to be careful. We could add `deprecated` tag but that can be done at any time later if that makes sense :smile:

PR #851, opened 2019-03-04
Environment engine deprecated properties

### LINKED ISSUE #844: Environment_Engine: Remove deprecated properties
Where `ElementID` and `BuildingElementProperties` are  being deprecated from the Environment_oM, update the engine to remove dependencies on them to facilitate their removal from the Environment_oM.

The `BuildingElementProperties` class has been replaced by the `ElementProperties` class in the Environment_oM, so any methods related to `BuildingElementProperties` should also be removed.

### LINKED ISSUE #850: Environment_Engine: Fix ShadingElements query
`ShadingElements` query no longer needs to rely on the custom data that it currently is, and indeed, produces wrong results when used. Fix it.

FraserGreenroyd:
Fixes #844 
Fixes #850 

Fixes the shading query method and removes references to now deprecated properties

### Test files
N/A

#### Removed
 - Removed references to deprecated properties from Environment_oM


IsakNaslundBh:
Checked builds to master

PR #867, opened 2019-03-13
Structure_Engine: Adding Geometry() method To PanelPlanar

### LINKED ISSUE #859: Structure_Engine: Geometry() method for PanelPlanar
With the introduction of the PlanarSurface, https://github.com/BHoM/BHoM/pull/427, we can now return this from the Geometry() method for PanelPlanar. This should allow for automatic visualisation of the panel objects.

IsakNaslundBh:

Closes #859 

Adding Geometry() method to the PanelPlanar returning a PlanarSurface

### Test files

All panels should now automatically display their centre plane geometry, so any GH script using PanelPlanar should work for testing this.

Also, see simple script here:

https://burohappold.sharepoint.com/:f:/r/sites/BHoM/02_Current/12_Scripts/01_Test%20Scripts/BHoM_Engine/Structure_Engine/Structure_Engine-Issue859-GeometryForPanelPlanar?csf=1&e=2LqibL

#### Added:
- Geometry() method added for PanelPlanar automatically displaying its geometry in GH

### Additional comments

image

### DIFF
--- a/Structure_Engine/Query/Geometry.cs
+++ b/Structure_Engine/Query/Geometry.cs
@@ -23,6 +23,7 @@
 using BH.oM.Geometry;
 using BH.oM.Structure.Elements;
 using BH.oM.Structure.Properties.Section;
+using BH.Engine.Common;
 using System.Collections.Generic;
 using System.Linq;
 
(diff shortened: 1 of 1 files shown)


IsakNaslundBh:
Hmm, that is indeed strange, and yes, they both should be doing exactly the same thing, at least from the side of the engine. Not sure if there is any difference in how the methods are called on the side of Rhino/GH, but would assume not.

@epignatelli any ideas?

epignatelli:
Yes, I noticed that when converting. I didn't dig too deeply.
For what I check, Rhino always tries to make a surface with the normal facing upwards, whereas we instead take into account the sense of the curve.

Have you experienced the same?
Regardless, if there is a problem, it would be somewhere else than this pr.

IsakNaslundBh:
Thanks @peterjamesnugent . Yes, quite sure that this is a GH issue, not Engine. Think Eduardo is looking in to it.

@epignatelli 

> or this pr, I would not use Linq expressions as they may be slow if we have a lot of curves. Since this is used for visualisation, I'd go for a bit more performance and resort to for loops.

Is there a significant slowdown using linq? Do you have any link to any more information on that? If it is a noticeable performance hit, then should we use linq at all?

epignatelli:
In can't find the experiment anymore, but a significant hit happens after a thousands objects.

I am fine with the code as it is, it was more about flagging that up because to display PanelPlanar is already a little slow. But I think the main improvement will come from storing the rhino geometries in the grasshopper object anyway. If it is still slow, we can come back here and optimise it, I think.

IsakNaslundBh:
Ok, thanks, take that as you are fine with this as is then!

PR #879, opened 2019-03-26
Solves #878

### LINKED ISSUE #878: Reflection_Engine: Show list of compatible types in input descriptions
From @alelom 's https://github.com/BHoM/Grasshopper_Toolkit/issues/339:

> For some components it's not easy to guess what is the right type of input required.
> 
> For example, consider the creation of a structural `Bar`. The input `property` description says:
> 
> > property is a ISectionProperty
> 
> but for a newcomer this is not helpful in finding out what to type in the BHoM search bar next.
> A novice won't find a type compatible with `ISectionProperty` straight away and might struggle looking for terms like "crossSection" or similar.
> 
> A workaround could be to have a **list of types types compatible with the input type in the input description**. This way the user would easily find e.g. the `SteelSection` as a valid input for the `property` input of `ISectionProperty`.
> 
> image
>

epignatelli:

Closes #878

This PR changes the way `interface`s are described in the UIs, by appending all the types that implement the interface to the string.
For example:
If you have `ICurve`, the tooltip in the UI displays only "ICurve".
It now displays: "ICurve: Arc, Circle, Ellipse, Line, Polyline, Polycurve" and all the others I forgot to add.
If the number of types is more than 15, it stops and ends with "...and more".

NB. It does not work yet for lists where the generic type is an interface.

### Test files

Hover on the parameters to get the description.
https://burohappold.sharepoint.com/:f:/s/BHoM/EpZu2DXIL1VGlms6O3XwBz4BZ6usNVH1NeF0FILz3bVy_A?e=cFUgFK

### DIFF
--- a/Reflection_Engine/Query/Description.cs
+++ b/Reflection_Engine/Query/Description.cs
@@ -54,11 +54,38 @@ namespace BH.Engine.Reflection
         {
             IEnumerable<InputAttribute> inputDesc = parameter.Member.GetCustomAttributes<InputAttribute>().Where(x => x.Name == parameter.Name);
             if (inputDesc.Count() > 0)
+            {
                 return inputDesc.First().Description;
+            }
             else if (parameter.ParameterType != null)
-                return parameter.Name + " is a " + parameter.ParameterType.ToText();
+            {
+                if (parameter.ParameterType.IsInterface)
+                {
+                    Type type = parameter.ParameterType;
+                    string echo = $"This is a {type.ToText()}: ";
+
+                    List<Type> t = type.ImplementingTypes();
+                    int m = Math.Min(15, t.Count);
+
+                    for (int i = 0; i < m; i++)
+                        echo += $"{t[i].ToText()}, ";
+
+                    if (t.Count > m)
+                        echo += "and more...";
+                    else
+                        echo.Remove(echo.Length - 2, 2);
+
+                    return echo;
+                }
+                else
+                {
+                    return parameter.Name + " is a " + parameter.ParameterType.ToText();
+                }
+            }
             else
+            {
                 return "";
+            }
         }
 
         /***************************************************/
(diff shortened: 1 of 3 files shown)


alelom:
Great job @epignatelli! I really think this is immensely useful for newcomers - it makes it so much easier to see what is and isn't already implemented, and generally makes the workflow and learning process much faster.

Only comment I have is that there's always a trailing comma `,` at the end of the description.

epignatelli:
Thanks @alelom! I pushed the fix, that should do!

Reflection_Engine/Query/Description.cs
```diff
@@ -54,23 +54,50 @@ public static string Description(this ParameterInfo parameter)
...
-                return "This is a " + type.ToText();
+            }
+            else if (type.IsInterface)
+            {
+                string echo = $"This is a {type.ToText()}: ";
+
+                List<Type> t = type.ImplementingTypes();
+                int m = Math.Min(15, t.Count);
+
+                for (int i = 0; i < m; i++)
```
adecler: Little trick here, you can use Linq to save yourself the trouble of dealing with the final ',':

`t.Select(x => x.ToText()).Aggregate((a,b) => a + ", " + b)`

On a side note while I am here, try to avoid using one-letter variable names like 't' for other things than loops (or Linq  methods)

Reflection_Engine/Query/ImplementedTypes.cs
```diff
@@ -0,0 +1,55 @@
...
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Returns all the loaded types that implement the specified interface.")]
+        public static List<Type> ImplementingTypes(this Type @interface)
+        {
+            if (!@interface.IsInterface)
+                return new List<Type>();
+
+            List<Type> implemented = new List<Type>();
```
adecler: Since I am promoting Linq in this review already, you can use it here too to make things lighter:

`return BHoMTypeList().Where(t => @interface.IsAssignableFrom(t)).ToList()`

PR #872, opened 2019-03-15
Geometry engine issues 343, 637 centroid and normal of a curve

LMarkowski:
### Issues addressed by this PR

Closes #343 
Closes #637

Adding methods centroid(polyline) calculating a center of mass of planar polyline and centroid(polycurve) for planar polycurve (consisting of arcs and lines only).

Adding methods for polylines and polycurves to Normal() calculating normal vector; always according to the right-hand rule.

### Test files
[Here](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?id=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FTest%20Scripts%2FBHoM%5FEngine%2FGeometry%5FEngine%2FGeometry%5FEngine%2DIssues343%2C637%2DCentroidAndNormalOfACurve) is the GH test file for both methods.

### Changelog
**Added:**

- Centroid(Polyline), Centroid(PolyCurve)
- Normal(Polyline), Normal(PolyCurve)

### Additional comments
There are some errors (less than 0,5% of tested samples) in Normal(PolyCurve) method. In my understanig of normal vector those are caused by GH. The code works fine.

PR #892, opened 2019-04-05
Common_Engine-Issues886,889-RightHandNormalAndDistributeOutlinesFix

### LINKED ISSUE #886: Common_Engine: Implement right hand rule based normal for IElement2D
Following #872, implement right hand rule based normal for `IElement2D`

### LINKED ISSUE #889: Common_Engine: make DistributeOutlines method ignore openings when cannot be cut
Some IElements2D can have openings cut in them (e.g. PanelPlanar in the Structure_oM) and some not since they are at the bottom of the hierarchy (Opening in the Structure_oM). For the latter, placing of the openings needs to be blocked. To be resolved by an extra bool argument in the method header.

pawelbaran:

Closes #886 
Closes #889 

### Test files

No test file for the Normal method (I believe #872 should be enough), a DistributeOutlines test file is [here](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?id=%2Fsites%2FBHoM%2F02_Current%2F12_Scripts%2F01_Test%20Scripts%2FBHoM_Engine%2FCommon_Engine%2FCommon_Engine-Issues886%2C889-RightHandNormalAndDistributeOutlinesFix) - this one should also be safe.

### Changelog

Added Normal method for an IElement2D + an extra argument that makes the `DistributeOutlines` method work correctly on the IElement2D that do not allow cutting holes.

### DIFF
--- a/Common_Engine/Common_Engine.csproj
+++ b/Common_Engine/Common_Engine.csproj
@@ -72,6 +72,7 @@
     <Compile Include="Query\Geometry.cs" />
     <Compile Include="Query\InternalElements2D.cs" />
     <Compile Include="Query\IsSelfIntersecting.cs" />
+    <Compile Include="Query\Normal.cs" />
     <Compile Include="Query\OutlineElements1D.cs" />
     <Compile Include="Query\OutlineCurve.cs" />
     <Compile Include="Query\ShearModulus.cs" />
--- a/Common_Engine/Common_Engine.csproj
+++ b/Common_Engine/Common_Engine.csproj
@@ -42,6 +42,10 @@
       <SpecificVersion>False</SpecificVersion>
       <HintPath>..\..\BHoM\Build\Geometry_oM.dll</HintPath>
     </Reference>
+    <Reference Include="Reflection_oM">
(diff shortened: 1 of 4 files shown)


pawelbaran:
@IsakNaslundBh I needed to step into the Structure_Engine to make the PanelPlanar constructor work - hope this will not be an issue.

FraserGreenroyd:
@pawelbaran was the change to `PanelPlanar` truly necessary? It looks to me as if you've added a `true` input to the method, but the method it is calling already has `true` as the default bool value for that parameter? In my mind, this change to `PanelPlanar` isn't needed, other than providing clarification to someone scanning it that it `canCutOpenings`?

pawelbaran:
@FraserGreenroyd yes it was: without a boolean input, the method call was ambiguous and the code did not compile. It surprised me too 😸

FraserGreenroyd:
@pawelbaran ah is there another method with a similar name that wasn't changed for the files changed that I looked at? That makes more sense then :smile:

pawelbaran:
Yeah, there is the original, deprecated one without a Boolean and a new one with a default `True` value.

Issue #820, opened 2019-02-22
Structure_Engine: Standard name for properties


IsakNaslundBh:
Could be good to add some standard naming conventions for our different property types to easy the handling of pushing objects through the different adapters.

Would basically be a string that could be generated based on the properties of the objects.

For example a 200mm ConstantThickness property with C30/37 could be called `0.2 THK - C30/37` or similar. If no name is set, this could then be used as a fallback plan.

Any comments?

rwemay:
Looks good. Probably better to use mm in the name (and write 200mm so its clear).

peterjamesnugent:
I like this, in the Lusas_Toolkit we use the name to store the AdapterID because you cannot force IDs for properties. So it would be something like: 

G1/0.2 THK - C30/37

We might have to change how we read the AdapterID (currently just searches for a "'/" but that can be remedied fairly easily.

al-fisher:
Good idea!

JosefTaylor:
We're going to get into some units issues here... if you want to convert to mm then I want to convert to inches!

rwemay:
I think most materials that are imperial are US specific, so I think it follows that those descriptions would be in inches. Note that this is only the name tag though, the units themselves always default to metric.

JosefTaylor:
Agreed on that. Almost never do we use UK/European materials and call out dimensions in inches, or the other way around. If you want to do that, you can set the name yourself.

FredericBourgeon:
Chiming in on a side case, but there are some cases where the usual name for a section is different from the actual size (thinking timber, mostly on the smaller sections). Which name would be preferred in this case, usual or actual? In timber, section size varies with moisture content, should it be specified as well?

IsakNaslundBh:
Some good input here, thanks! :)

I am fine with mm or m, but maybe should stick to standard SI in m, to not confuse it for people used to imperial?

Regarding name of sections vs the properties:
I think whatever we implemented for this would just be generic string based on the actual magnitude of the properties it consists of. For example, an un-named UB section, like `UB 356x171x53` would be called something like `I - 363.4x173.2x9.1x15.7` where the numbers describe the actual heights, widths and thicknesses. For the name to be used, the name need to be provided.

Think we still should promote naming of sections as the standard workflow and see this as a _fallback_ if that has not been done. Name should ofc also be part of any sections we serialise, which should help with keeping the standard sections consistent! :)

[...]

PR #951, opened 2019-05-10
Structure geometry engine issue922 align with changes in b ho m

### LINKED ISSUE #862: Structure_Engine: Deprecate MeshFace class
Deprecate any method targeting the `MeshFace` class, as class itself should be deprecated by `FEMesh`

Related to:
https://github.com/BHoM/BHoM/issues/432

### LINKED ISSUE #922: Structure_Engine: Update engine to follow changes in BHoM
Update to cover:

- https://github.com/BHoM/BHoM/issues/399
- https://github.com/BHoM/BHoM/issues/448
- https://github.com/BHoM/BHoM/issues/459
- https://github.com/BHoM/BHoM/issues/461
- https://github.com/BHoM/BHoM/issues/468

### LINKED ISSUE #940: Geometry_Engine: Add Create method for Basis

IsakNaslundBh:
### NOTE: Depends on 

https://github.com/BHoM/BHoM/pull/481
https://github.com/BHoM/BHoM_Datasets/pull/3
   
Closes #922 Aligning with changes in BHoM
Closes #940 
Closes #862 

##### Aligning with changes med in Structures and Geometry_oM ( see https://github.com/BHoM/BHoM/pull/481)

- Namespace updates for usings
- Name of Elements updated
- Create methods pointing to old name deprecated
- Create methods for `ShapeProfiles` moved to Geometry_Engine. Old ones in Structure_Engine deprecated

###### Materials:

**EDIT**
Materials on structural objects are now always a structural 'IMaterialFragment' Tip for use in toolkits is to check `if(materialFragment is IIsotopic)` and `if(materialFragment is IOrthotropic)`. With a cast after, this will give access to the analytical proeprties of the material

**EDIT2** Make sure to rebuild the BHoM_Datasets to ensure the "Default" method for materials returns sane values! This is used by the SteelSEctionFromProfile and ConcreteSectionFromProfile methods

~~As the material has been tweaked from the old one in common to the new one in physical, several support methods have been added:~~

~~Checks of material type:~~
- ~~`IsStructural()`  - checks if the material has a a set structural material fragment~~
- ~~`IsIsotropic()` - checks if the material is of isotropic nature - does it have a Isotropic structural material fragment assigned~~
- ~~`IsOrthoTropic()`- checks if the material is of orthotropic nature - does it have a Orthotropic structural material fragment assigned~~
- ~~`IsAluminium()`~~
- ~~`IsConcrete()`~~
- ~~`IsSteel()`~~
- ~~`IsTimber()`~~

~~Query method added to simplify getting out specific material properties:~~

- ~~`YoungsModulus()`~~
- ~~`ShearModulus()`~~
- ~~`PoissonsRatio()`~~
- ~~`ThermalExpansionCoeff()`~~

~~All of the above returns an `object` as it could be either a `Vector` or a `double` depending on the material type. As addition all of them have methods that return a `Vector` or `double` explicitly. For example:~~

- ~~`YoungsModulusIsotropic()` return a `double` value (only works for Isotropic materials)~~
- ~~`YoungsModulusOrthotropic() returns a `Vector` value (only works for Orthotropic materials)~~

~~Several create methods added for materials, creating a specific material fragment, then creates a new material and assigns the fragment to the material:~~

- ~~`Create.AluminiumMaterial()`~~
- ~~`Create.ConcreteMaterial()`~~
- ~~`Create.SteelMaterial()`~~
-~~ `Create.TimberMaterial()`~~

~~Also added various set methods that will create a new structural material fragments and assign/overwrite any old this fragment to a `Material`~~

- ~~`SetAluminium()`~~
- ~~`SetSteel()`~~
- ~~`SetConcrete()`~~
- ~~`SetTimber()`~~

#### Library_Engine

All Structures related DataSets have been removed from the Library_Engine and moved into BHoM_Datasets. See separate PR for this https://github.com/BHoM/BHoM_Datasets/pull/3


rwemay:
Is is possible to add default names to constructors for materials, constraints etc?

IsakNaslundBh:
@rwemay we could. Reason I have not done it is that the structural adapters relies on unique names for them, so could lead to a situation where you construct like 3 different concrete materials that all have the name "Concrete" and try to push them to Etabs/GSA/Robot, where they all will be assumed to be the same thing. Same could happen for Constraints etc.

Basically worried that helping on this level might be worse for users in general and lead to some, from the users point of view, unexpected behaviours. Any thoughts?

rwemay:
Right, that is a good reason - bit of a chicken and egg! Most softwares require there to be a name so there's an overhead with needing to create it (and tracking the issue in a users script) when test-pushing in particular, but at the same time they need to be unique. Perhaps there's an adapter method that deals with uniqueness/naming etc.? (i.e. having our cake and eating it). For the constructors that create a specific material with defaults (which is great) I actually created a concrete material with all steel properties by accident (in the naming) which is also not great!

IsakNaslundBh:
Assume you mean something like this:

https://github.com/BHoM/BHoM_Engine/issues/820 ?

Can think of ideas of how to handle this for Sections, CostantThickness and constraints etc. A bit more unsure of the materials though.
"Concrete -CubeStrength 37000000, CylinderStrength 30000000, E 33000000000" does not look so good... Unsure how to make something generic enough to not be wrong/conflicting while still keeping in valid over different codes/regions. might be able to find some kind of variable for each material to point to. E for Concrete, fy for steel etc.

Also a questions if it should be used in the `Create` methods or just the adapters. The latter was my first idea, but maybe would be worth just setting the name to some default on construction as well.

epignatelli:
From https://github.com/BHoM/BHoM_Datasets/pull/3#pullrequestreview-238958106 the BHoM_Datasets repo works fine now. Details in the linked pull request.

PR #920, opened 2019-04-26
Geometry-Engine: Issue 874-BooleanDifference/Intersection/UnionForPolyCurves

LMarkowski:
### Issues addressed by this PR

Closes #874 

### Test files
Regular tests for Boolean operations are [here](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?id=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FTest%20Scripts%2FBHoM%5FEngine%2FGeometry%5FEngine%2FCompute) marked with ~~'V2'~~ 'V3' (the 2 recorded errors are the result of different tolerance handling by BHoM and Rhino, ours is more precise 🙊), `IsContaining` for `BoundingBoxes` is [here](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?id=%2Fsites%2FBHoM%2F02_Current%2F12_Scripts%2F01_Test%20Scripts%2FBHoM_Engine%2FGeometry_Engine%2FQuery)
~~Test for performance of rewritten Polylines methods is [here](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?id=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FTest%20Scripts%2FBHoM%5FEngine%2FGeometry%5FEngine%2FGeometry%2DEngine%2DIssue874%2DBooleanDifferenceIntersectionUnionForPolyCurvesPerformance)~~

#### Changed (with no change in input/output type or logics)
- `List<Polyline> BooleanIntersection(this Polyline region, Polyline refRegion, double tolerance = Tolerance.Distance)`
- `List<Polyline> BooleanIntersection(this List<Polyline> regions, double tolerance = Tolerance.Distance)`
- `List<Polyline> BooleanUnion(this List<Polyline> regions, double tolerance = Tolerance.Distance)`
- `IsContaining` methods for `BoundingBoxes`
 ~~booleanIntersection(PolyCurve,PolyCurve)
booleanIntersection(List<PolyCurve>)
booleanUnion(List<PolyCurve>)
booleanDifference(PolyCurve, List<PolyCurve>)   //returns List<PolyCurve>
booleanDifference(List<PolyCurve>, List<PolyCurve>)   //returns List<List<PolyCurve>>~~

#### Added
[...]

PR #961, opened 2019-05-20
Geometry_Engine: Issue918 Update IsContaining for BoundingBoxes

### LINKED ISSUE #918: Geometry_Engine: update IsContainig method for BoundingBox
Methods IsContaing(this box, box) and IsContaing(this box, point) do not contain tolerance nor acceptOnEgdge boolean. Method for list of points is missing.

LMarkowski:

Closes #918 

As mentioned in comment in PR #920 updating and properly testing new methods.
Old methods did not include tolerance and AcceptOnEdge bool.
I also changed them a bit after merge in PR #920 .

### Test files
Test file is [here](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?newTargetListUrl=%2Fsites%2FBHoM%2F02_Current&viewpath=%2Fsites%2FBHoM%2F02_Current%2FForms%2FAllItems%2Easpx&id=%2Fsites%2FBHoM%2F02_Current%2F12_Scripts%2F01_Test%20Scripts%2FBHoM_Engine%2FGeometry_Engine%2FQuery)

### Changelog
Updated:
- IsContaining(this BoundingBox box1, BoundingBox box2)
- IsContaining(this BoundingBox box, Point pt)
- IsContaining(this BoundingBox box, IGeometry geometry)

Added:
-  IsContaining(this BoundingBox box, List<Point> pts)

### Additional comments
- We discussed with @pawelbaran the matter of BBoxes which have one (or more) of dimensions equal zero. In such cases, if AcceptOnEgde bool=false, the methods would always return false (as the boxes would consist of egdes only). We agreed that it is correct solution as we have different isContaining methods for planar objects and isOnLine for linear objects.
- I wasn't sure if IsContainig(BBox, IGeometry) should have additional 'I' like an interface. It is not exactly an interface for this method, but refers to different types of input. I've added the 'I' but it may be discussed.

### DIFF
--- a/Geometry_Engine/Query/IsContaining.cs
+++ b/Geometry_Engine/Query/IsContaining.cs
@@ -59,7 +59,7 @@ namespace BH.Engine.Geometry
         public static bool IsContaining(this BoundingBox box, IGeometry geometry)
         {
             //return box.IsContaining(geometry.IBounds());
-            return box.IIsContaining(geometry, true, Tolerance.Distance);
+            return box.IsContaining(geometry, true, Tolerance.Distance);
         }
 
         /***************************************************/
@@ -125,7 +125,7 @@ namespace BH.Engine.Geometry
 
         /***************************************************/
 
-        public static bool IIsContaining(this BoundingBox box, IGeometry geometry, bool acceptOnEdge = true, double tolerance = Tolerance.Distance)
+        public static bool IsContaining(this BoundingBox box, IGeometry geometry, bool acceptOnEdge = true, double tolerance = Tolerance.Distance)
         {
             return box.IsContaining(geometry.IBounds(), acceptOnEdge, tolerance);
         }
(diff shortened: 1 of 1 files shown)


IsakNaslundBh:
> I wasn't sure if IsContainig(BBox, IGeometry) should have additional 'I' like an interface. It is not exactly an interface for this method, but refers to different types of input. I've added the 'I' but it may be discussed.

I would get rid of the extra I for this method. I would say the extra I should be added only for methods that just simply dispatch the code to other methods. For this case the logic is the method (although short and simple). 1. Get the bounding box of the geometry. 2. check if the boundingbox is inside.
It does not cast and call various methods depending on different geometry types (other methods called by you method does, but that does not really impact this decision) 

TL;DR: No extra I.

@al-fisher @adecler feel free to comment if you disagree.

al-fisher:
> TL;DR: No extra I.

Agreed. 😄

LMarkowski:
Ok, so I deleted the 'I'. Then reverted the commit. Then reverted the reversion. Sorry for the mess 😅 
Updated the test file also.

PR #962, opened 2019-05-20
Reflection engine issue923 956 is deprecated and description updates

### LINKED ISSUE #923: Reflection_Engine: Remove version check for deprecation
If a deprecated tag has been added to a method it should be safe to assume that the method should be deprecated and hidden for the UI, and the version could be seen more as information as to from when it happen.

Should remove the Version check from the `IsDeprecated` method.

### LINKED ISSUE #956: Reflection_Engine: Always point to method used for description
The Description method should be updated to always provide the default text generated, stating where the method is from etc, even if a Description attribute has been added. 

@michaldengusiak FYI

IsakNaslundBh:

Closes #923 
Closes #956 

- Always deprecate a method with deprecated attribute attached
- Always add default descriptions for methods, inputs and outputs
- Fixed getting available types for list inputs

### Test files

Just changing the UI. Please test how it looks on a few components

### Additional comments

@al-fisher @michaldengusiak @FraserGreenroyd @rwemay FYI.


FraserGreenroyd:
> Maybe the only thing is to prepend the default description before the custom one, but I am being tedious: :D

That was the subject of a good half hour debate with @IsakNaslundBh and @al-fisher when we first prototyped this last week!

epignatelli:
I feel less alone in the tediousness world now :D

adecler:
**On the code side alone**, this looks much cleaner with the last commit. As long as other are happy with the way it feels and work in the UIs, I'm all good for the merge.

PS: I disagree with @epignatelli and prefer the default description to be after the custom one. You want the most explicit and clear possible thing to be the first one people read, not the last.

al-fisher:
> You want the most explicit and clear possible thing to be the first one people read, not the last.

That was the reasoning @FraserGreenroyd @IsakNaslundBh and I came to too!
Looks good to me!

IsakNaslundBh:
Thanks all for comments :) Merging this now!

PR #904, opened 2019-04-12
Geometry_engine-issues-880,896-SplitAtPoints-SortAlongCurve-Distance-etc

LMarkowski:
Closes #880 
Closes #896 

### Test files
Test files are [here](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?id=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FTest%20Scripts%2FBHoM%5FEngine%2FGeometry%5FEngine%2FGeometry%5Fengine%2Dissues%2D880%2C896%2DSplitAtPoints%2DSortAlongCurve%2DDistance%2Detc)

### Changelog
New methods added:
- SortAlongCurve: Arc, Circle, PolyCurve
- SplitAtPoints: Arc, Circle PolyCurve
- Normal: Arc

Rewrited code for new inputs:
- IsCoplanar: PoluCurve (rewrited from polylines)
- IsOnCurve: everything but nurbs and ellipse. (Refer to Distance())
- Join: List PolyCurve (rewrited from Join(List ICurve))
- CurveIntersections: PolyCurve, PolyCurve (rewrited from CurveIntersections(ICurve, ICurve))

Fixed:
- ClosestPoint(Arc,Point) now checks if the point is arc.centre and then always returns StartPoint. It caused bug in Distance() -> #896  . Previous method is deprecated and functionality is preserved.

### Additional comments
These are all methods which were necessery to create booleanMethods for PolyCurves (#874 ). They are coming in another PR right after this one get merged.

PR #994, opened 2019-06-04
Geometry_Engine: issues 992,993, Normal(Arc), PointAtParameter(Arc)

### LINKED ISSUE #992: Geometry_Engine: PointAtParameter(Arc) does not always return correct value
As in header

EDIT:
Same issue with StartPoint(Arc) and EnPoint(Arc)

### LINKED ISSUE #993: Geometry_Engine: Normal(Arc) should be get from arc.CoordinateSystem
As in header it will faster and more reliable way to calculate normal of an arc, than by using cross-products.

LMarkowski:

Closes #992
Closes #993 

<s>There was a one-line bug in `PointAtParameter(Arc)`. If vectors determining CoordinateSystem of an arc were not normalised, the method returned wrong values. (EDIT: Found and fixed same issue in `StartPoint(Arc)` and `EndPoint(Arc)`).</s>

EDIT2: I've checked a bit deeper and found a bug in `SplitAtPoint(Circle)` which caused returning arcs with wrong coordinate systems. Now it's fixed and there is no need to change other 3 methods. Sorry for a confusion.
I'm linking a regular test of SplitAtPoints, the method is still 100% accurate.

I think it'll be much faster and more reliable to get normal of an arc from it's CoordinateSystem than from cross-products, as it is now. Current method was actually pushed by me in #904. _Mea culpa_ 😉 

### Test files
PointAtParameter (EDIT: + StartPoint + EndPoint) - made issue specific [test](https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/01_Test%20Scripts/BHoM_Engine/Geometry_Engine/Geometry_Engine-issue992-PointAtParameter(Arc)Bug/Geometry_Engine-issue992-PointAtParameter(Arc)Bug.gh?csf=1&e=H50d84)
Normal - updated regular [test file](https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/01_Test%20Scripts/BHoM_Engine/Geometry_Engine/Query/Normal.gh?csf=1&e=JkEQ9S)
EDIT: SplitAtPoints(Circle) - regular test [file](https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/01_Test%20Scripts/BHoM_Engine/Geometry_Engine/Modify/SplitAtPoints.gh?csf=1&e=b0iHAN)

### Changelog

Updated:
- <s>`PointAtParameter(Arc)`</s>
- `Normal(Arc)`
EDIT:
Also updated:
- <s>`StartPoint(Arc)`</s>
- <s>`EndPoint(Arc)`</s>
EDIT2:
Updated:
- `SplitAtPoint(Circle)`

### DIFF
--- a/Geometry_Engine/Modify/SplitAtPoints.cs
+++ b/Geometry_Engine/Modify/SplitAtPoints.cs
@@ -104,7 +104,7 @@ namespace BH.Engine.Geometry
             Vector normal = circle.Normal;
             Double rotAng;
 
-            Cartesian system = new Cartesian(circle.Centre, startVector, circle.PointAtParameter(0.25) - cirCen, normal);
+            Cartesian system = new Cartesian(circle.Centre, startVector.Normalise(), (circle.PointAtParameter(0.25) - cirCen).Normalise(), normal);
             
             Arc mainArc = new Arc();
             Arc tmpArc = new Arc();
--- a/Geometry_Engine/Query/EndPoint.cs
+++ b/Geometry_Engine/Query/EndPoint.cs
@@ -1,6 +1,6 @@
 /*
  * This file is part of the Buildings and Habitats object Model (BHoM)
- * Copyright (c) 2015 - 2018, the respective contributors. All rights reserved.
+ * Copyright (c) 2015 - 2019, the respective contributors. All rights reserved.
  *
  * Each contributor holds copyright over their respective contributions.
  * The project versioning (Git) records all such contribution source information.
--- a/Geometry_Engine/Query/EndPoint.cs
+++ b/Geometry_Engine/Query/EndPoint.cs
@@ -1,6 +1,6 @@
 /*
  * This file is part of the Buildings and Habitats object Model (BHoM)
- * Copyright (c) 2015 - 2019, the respective contributors. All rights reserved.
+ * Copyright (c) 2015 - 2018, the respective contributors. All rights reserved.
  *
  * Each contributor holds copyright over their respective contributions.
  * The project versioning (Git) records all such contribution source information.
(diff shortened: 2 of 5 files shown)


LMarkowski:
I've found same issue in `StartPoint(Arc)` and `EndPoint(Arc)`. @FraserGreenroyd, could you take another look and merge if it's okay?

IsakNaslundBh:
Just to comment on the changes here,

We have assumed that the coordinateSystem is "correct", that is with orthogonal unit vectors, for all the methods to save some computation time. All the vector normalisations etc where intentionally left out for this reasons. The coordinateSystem is immutable and the create methods for it should be doing all the checks necessary to enforce it to be correct.

I understand why you have done the changes, just not sure if they are needed or just a waste of computation without actually improving what the methods are doing. 

Have you come across any cases where the methods have been giving wrong results based on erroneous coordinate systems?

LMarkowski:
@IsakNaslundBh Yes, actually in the test file which I provided is the case with a coordinate system that look like this:
image

In general I agree that we should assume that coordinateSystem is correct. I needed this methods to work on boolean operations so I made that kind of quick fix. 
I could possibly dig in to find what is the cause of erroneous vectors in coordinateSystem. But postpone boolean operations a bit. It depends on how urgent are booleans.

FraserGreenroyd:
Given this is likely to now stay open until tomorrow's BHoM call, the reason I haven't merged is due to waiting for confirmation from code owner @pawelbaran on the suitability. So if it gets to BHoM call and anyone is wondering why I didn't merge when I approved - that's why :smile:

LMarkowski:
@IsakNaslundBh @FraserGreenroyd Following BHoM talk: I've checked a bit deeper and found a bug in SplitAtPoint(Circle) which caused returning arcs with wrong coordinate systems. Now it's fixed and there is no need to change other 3 methods. Sorry for a confusion 😉

PR #1007, opened 2019-06-07
Reflection engine issue982 get set nested properties

### LINKED ISSUE #982: Reflection Engine: Access Object information within an Object
Enable get property to access properties of objects within objects, thus preventing the need for example below.

image

Suggestion syntax could be using "." hence in this example "SectionProperty.Name" could be used to access the section name for the list of BHoM bars.

epignatelli:

Closes #982

Get method for nested properties allows to use the form `PropOne.PropTwo` as input in the GetProperty method.
The `SetPropertyValue`, at the moment, cannot work that way. The BHoM_Engine project, where the `DeepClone` resides, references the Reflection_Engine. Trying to use any cloning functionality would result in a circular dependency.

### Test files

https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/01_Test%20Scripts/BHoM_Engine/Reflection_Engine/Reflection_Engine-Issue985-IsDeprecatedInterfaceMethod/ReflectionEngine-Issue982-GetSetNestedProperties.gh?csf=1&e=2RnGdv

### DIFF
--- a/Reflection_Engine/Query/PropertyValue.cs
+++ b/Reflection_Engine/Query/PropertyValue.cs
@@ -39,23 +39,20 @@ namespace BH.Engine.Reflection
         [Output("value", "value of the property")]
         public static object PropertyValue(this object obj, string propName)
         {
-            string[] props = propName.Split('.');
-            foreach (string prop in props)
-            {
-                obj = obj.GetPropertyValue(prop);
-                if (obj == null)
-                    break;
-            }
-
-            return obj;
-        }
-
-        /***************************************************/
-
-        public static object GetPropertyValue(this object obj, string propName)
-        {
             if (obj == null || propName == null)
                 return null;
+
+            if (propName.Contains("."))
+            {
+                string[] props = propName.Split('.');
+                foreach (string innerProp in props)
+                {
+                    obj = obj.PropertyValue(innerProp);
+                    if (obj == null)
+                        break;
+                }
+                return obj;
+            }
 
             System.Reflection.PropertyInfo prop = obj.GetType().GetProperty(propName);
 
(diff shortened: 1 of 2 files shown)


epignatelli:
> The `SetPropertyValue`, at the moment, cannot work that way. The BHoM_Engine project, where the `DeepClone` resides, references the Reflection_Engine. Trying to use any cloning functionality would result in a circular dependency.

:)

jdarcyBHE:
Ok I can live with just GetProperty working 👍

epignatelli:
Thanks @jdarcyBHE, in the set-property the issue is a bit more widespread, so it will need a more coordinated action to happen.

Reflection_Engine/Query/PropertyValue.cs
```diff
@@ -39,31 +39,57 @@ public static partial class Query
...
+                if (obj == null)
+                    break;
+            }
+
+            return obj;
+        }
+
+        /***************************************************/
+
+        public static object GetPropertyValue(this object obj, string propName)
```
adecler: The name of this method is not compliant, especially since it is public.  

The way you need to do this is having `PropertyValue` calling itself on the parts of the split if `propName` is containing a dot. That way you have a single method with a valid name.

epignatelli: Thanks @adecler, that'd be solved.

jdarcyBHE:
Re tested and works as intended able to access multiple levels below the initial object.

epignatelli:
@adecler if all is ok, would you mind merging this?

al-fisher:
thanks @jdarcyBHE / @epignatelli 
I can merge

PR #1042, opened 2019-06-19
Environment_Engine: Floor geometry method fix

### LINKED ISSUE #871: Environment_Engine: Fix Floor Geometry method
The `FloorGeometry` method does not return the full floor geometry in the event where multiple `BuildingElement`s make up the space.

May be rewritten as part of ongoing oM/Engine changes with @IsakNaslundBh later in the month - capturing here to be considered.

Related to https://github.com/BHoM/XML_Toolkit/issues/103

FraserGreenroyd:
Fixes #871 

### Test files
Existing test files

#### Fixed
 - Fixed Floor Geometry method

### DIFF
--- a/Environment_Engine/Query/FloorGeometry.cs
+++ b/Environment_Engine/Query/FloorGeometry.cs
@@ -43,7 +43,7 @@ namespace BH.Engine.Environment
         [Description("Returns the floor geometry of a space represented by Environment Panels as a BHoM Geometry Polyline")]
         [Input("panelsAsSpace", "A collection of Environment Panels that represent a closed space")]
         [Output("polyline", "BHoM Geometry Polyline representing the floor of the space")]
-        public static Polyline FloorGeometry(this List<Panel> panelsAsSpace)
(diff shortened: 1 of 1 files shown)


rwemay:
Is it a good idea to hardcode this tolerance?

michaldengusiak:
it is exposed, just default


image

rwemay:
Ok cool. For some reason I thought we also had a global default tolerance setting/system which methods can use - a bit like the default tolerance in Rhino/Dynamo etc. Maybe one for later if now. 

@IsakNaslundBh

al-fisher:
Hey @rwemay - you are right - we do. https://github.com/BHoM/BHoM/blob/master/Geometry_oM/Misc/Tolerance.cs
@michaldengusiak good to reference tolerance from here to keep consistent. Although, still should also only be a default and still exposed as you have implemented.

Consider ``` double tolerance = BH.oM.Geometry.Tolerance.Distance ```

PR #1047, opened 2019-06-20
Geometry_Engine-Issue1004-IBoundsMethodBreaksIfObjectIsNull

### LINKED ISSUE #1004: Geometry_Engine: IBounds method breaks if object is null
The UIs are breaking most of the time when the `Bounds` method is called with a null argument

Proposed:
```CS
        /**** Public Methods - Interfaces               ****/

        public static BoundingBox IBounds(this IGeometry geometry)
        {
            if (geometry == null)
                return null;
            return Bounds(geometry as dynamic);
        }
```

I would also add a fallback case, since the second main reason the method breaks is that it cannot find a suitable overload.
```CS
        /**** Private Methods - Fallback                ****/

        private static BoundingBox Bounds(object value)
        {
            return null;
        }
```

pawelbaran:

Closes #1004 

### Test files

Dedicated test file located [here](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?id=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FTest%20Scripts%2FBHoM%5FEngine%2FGeometry%5FEngine%2FGeometry%5FEngine%2DIssue1004%2DIBoundsMethodBreaksIfObjectIsNull).

#### Added:
- Added null checks that prevent the code from breaking when no geometrical bounds are found for an object

### Additional comments
Should not interfere with #1022 as modifying different files.

### DIFF
--- a/Geometry_Engine/Query/Bounds.cs
+++ b/Geometry_Engine/Query/Bounds.cs
@@ -427,7 +427,7 @@ namespace BH.Engine.Geometry
         /**** Private Methods - Fallback                ****/
         /***************************************************/
 
-        private static BoundingBox Bounds(object value)
+        private static BoundingBox Bounds(IGeometry geometry)
         {
             return null;
         }
--- a/Geometry_Engine/Query/Bounds.cs
+++ b/Geometry_Engine/Query/Bounds.cs
@@ -416,7 +416,20 @@ namespace BH.Engine.Geometry
 
         public static BoundingBox IBounds(this IGeometry geometry)
         {
+            if (geometry == null)
+                return null;
+
             return Bounds(geometry as dynamic);
+        }
+
+
+        /***************************************************/
+        /**** Private Methods - Fallback                ****/
+        /***************************************************/
+
+        private static BoundingBox Bounds(object value)
+        {
+            return null;
         }
 
         /***************************************************/


Geometry_Engine/Query/Bounds.cs
```diff
@@ -416,9 +416,22 @@ public static BoundingBox Bounds(this CompositeGeometry group)
...
+
             return Bounds(geometry as dynamic);
         }
 
+
+        /***************************************************/
+        /**** Private Methods - Fallback                ****/
+        /***************************************************/
+
+        private static BoundingBox Bounds(object value)
```
IsakNaslundBh: Would probably switch this method to only accept `IGeometry` rather than `object`. Should not really make any real difference, but will look more sensible code-wise.

pawelbaran: @epignatelli could you please reflect on that?

epignatelli: Yeah, I agree with @IsakNaslundBh - it doesn't change anything, it may be just more sensible.
I am pretty flexible about it - happy with both anyway.

pawelbaran: Updated.

PR #1043, opened 2019-06-19
Data+structure issue1041 refactor queries

epignatelli:
Depends on https://github.com/BHoM/BHoM/pull/509

   
### Issues addressed by this PR

Closes #1041

This pr is to adapt the code to the changes in the dependent pr listed above.
It affects more projects:
- DataStructure_Engine
- DataManipulation_Engine
- Geometry_Engine: the project is pretty busy. Live coordination with @pawelbaran is keeping the modification free from conflicts.
- Libraries_Engine
-Structure_Engine

Changes involve:
- Adapt to the namespace changes in the BHoM
- Merge the DataManipulation_Engine and the DataStructure_Engine into a Data_Engine project

If you look at cumulative changes, it's difficult to catch what happened. I structured the commits in a way that the you can identify the namespace and reference changes, and a final one that merges the project.

There is no need to review all the code, I added more reviewers to split the review in parts:
- Libraries_Engine and Structure_Engine @IsakNaslundBh 
- Geometry_Engine @pawelbaran 
- Data_Engine and general structure @adecler 

Feel free to review what you want though, that's just an indication.

PR #1050, opened 2019-06-20
Reflection engine issue1002 extension method to return interface methods

### LINKED ISSUE #1002: Reflection_Engine: ExtensionMethods to return all available methods
Currently the `ExtentionMethods()` method only returns methods with exact type matching for the first argument. This can be a problem when wanting to write methods for interface types that you want to be accessed by the `RunExtentionMethod()`.

As an example, the following method

```
        public static ICurve Geometry(this IFramingElement framingElement)
        {
            return framingElement.Location;
        }
```

is never hit, as the actual type will be `Beam`/`Column` etc.

One solution would be to change the type check on line 45 from `==` to `IsAssignableFrom`.

https://github.com/BHoM/BHoM_Engine/blob/18fd2606dd551b4b3e4785e5a30fbbfebfd1dfd1/Reflection_Engine/Query/ExtentionMethods.cs#L37-L49

A problem with this though is that it would mean you could end up with multiple returns, for cases where someone had implemented a specific method for for example the `Beam` class.

This could be handled by sorting the methods to the "closests" as c# does automatically. Thas is, if you provide a beam, you will get two methods returned, sort them out by "closeness" and call the method for the beam. If you provide a column you will only get the method for the `IFramingElement` returned, hence call that.

Please let me know if you have any comments on how/if this change should be made.

IsakNaslundBh:

Closes #1002 

- `ExtensionMethods()` now returns all methods callable by the type. Change from type == to type.IsAsignableFrom()
- Simple sorting of the methods to ensure that exact type matching is rated higher than interface/baseclass types.
- Storing of methods having multiple input parameters

This now makes in possible to reach methods written for an interface type via the `RunExtensionMethod()` schema.

Further improvements to be done:
- Make the sort method handle the closeness sorting in a more extensive way. Currently only sorts exact type higher than non-exact type. Should be done in a more hierarchical way based on inheritance chain.

### Test files

Testfile trying to get out geometry for IFramingElement types
https://burohappold.sharepoint.com/:f:/r/sites/BHoM/02_Current/12_Scripts/01_Test%20Scripts/BHoM_Engine/Reflection_Engine/Rection_Engine-Issue1002-RunExtensionInterfaceTypes?csf=1&e=fvlJKz

@pawelbaran as a lot of the work you have been doing is making use of these methods, good if you can check that this have not had any unforeseen side effects your end. Have tested some of your testscripts and things seem to be running fine. SHould give a boost in performance for Set methods.

### Additional comments

Have checked this PR compared to https://github.com/BHoM/BHoM_Engine/pull/1043 and did not see any modifications to the Reclection_Engine in there so should not be any clashes. If I am mistaken, would wait with this PR until #1043 is merged.

@al-fisher @rwemay @epignatelli @FraserGreenroyd FYI

### DIFF
--- a/Reflection_Engine/Compute/RunExtentionMethod.cs
+++ b/Reflection_Engine/Compute/RunExtentionMethod.cs
@@ -46,13 +46,12 @@ namespace BH.Engine.Reflection
                 return m_PreviousInvokedMethods[key].Invoke(null, new object[] { target });
 
             // Otherwise, search for the method and call it if found
-            foreach (MethodInfo method in type.ExtentionMethods(methodName))
+            MethodInfo method = type.ExtentionMethods(methodName).Where(x => x.GetParameters().Length == 1).SortExtensionMethods(type).FirstOrDefault();
+
+            if (method != null)
             {
-                if (method.GetParameters().Length == 1)
-                {
-                    m_PreviousInvokedMethods[key] = method;
-                    return method.Invoke(null, new object[] { target });
-                }
+                m_PreviousInvokedMethods[key] = method;
+                return method.Invoke(null, new object[] { target });
             }
 
             // Return null if nothing found
@@ -64,13 +63,19 @@ namespace BH.Engine.Reflection
         public static object RunExtentionMethod(object target, string methodName, object[] parameters)
         {
             object result = null;
+            Type type = target.GetType();
 
-            foreach (MethodInfo method in target.GetType().ExtentionMethods(methodName))
+            // If the method has been called before, just use that
+            string name = methodName + parameters.Select(x => x.GetType().ToString()).Aggregate((a,b) => a+b);
+            Tuple<Type, string> key = new Tuple<Type, string>(type, name);
+            if (m_PreviousInvokedMethods.ContainsKey(key))
+                return m_PreviousInvokedMethods[key].Invoke(null, new object[] { target }.Concat(parameters).ToArray());
+
+
+            foreach (MethodInfo method in target.GetType().ExtentionMethods(methodName).Where(x => x.GetParameters().Length == parameters.Length +1).SortExtensionMethods(type))
             {
-                // Make sure the number of parameters is matching
+
                 ParameterInfo[] paramInfo = method.GetParameters();
-                if (paramInfo.Length != parameters.Length + 1)
-                    continue;
 
                 // Make sure the type of parameters is matching
                 bool matchingTypes = true;
@@ -85,6 +90,7 @@ namespace BH.Engine.Reflection
                 if (!matchingTypes)
                     continue;
 
+                m_PreviousInvokedMethods[key] = method;
                 return method.Invoke(null, new object[] { target }.Concat(parameters).ToArray());
             }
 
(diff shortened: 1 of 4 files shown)


IsakNaslundBh:
@pawelbaran Would rename in separate PR, given that this is a bit behind various things at this point.

Will add in Eduardo as a reviewer as well then, unless anyone of the two of you are comfortable approving now that both have been ok with it? :)

Reflection_Engine/Compute/RunExtentionMethod.cs
```diff
@@ -46,13 +46,12 @@ public static object RunExtentionMethod(object target, string methodName)
...
-                if (method.GetParameters().Length == 1)
-                {
-                    m_PreviousInvokedMethods[key] = method;
-                    return method.Invoke(null, new object[] { target });
-                }
+                m_PreviousInvokedMethods[key] = method;
+                return method.Invoke(null, new object[] { target });
             }
 
             // Return null if nothing found
```
epignatelli: Is `result` initialised as `null` and never assigned for a reason or is it just a leftover? 
It happens in the next overload as well.

Reflection_Engine/Compute/RunExtentionMethod.cs
```diff
@@ -46,13 +46,12 @@ public static object RunExtentionMethod(object target, string methodName)
                 return m_PreviousInvokedMethods[key].Invoke(null, new object[] { target });
 
             // Otherwise, search for the method and call it if found
-            foreach (MethodInfo method in type.ExtentionMethods(methodName))
```
epignatelli: So, let me check I I get what is happening. Since we removed the loop, why are we sorting first?
Can't we just take the correct overload?

IsakNaslundBh: so, this line contains quite a bit of information, maybe a bit to much, but to break it down for clarity:

- `ExtentionMethods(type, methodName)` Returns all methods with the given name that has the first argument of a type matching (being assignable from) the provided type
- `Where(x => x.GetParameters().Length == 1)` filters out so we only take the methods that have exactly one argument, as we know by the context that that is the only methods we want for this.
- `SortExtensionMethods(type)` sorts the methods based on "closeness" to the provided type. (assume this is the one you are questioning the most). This is currently quite simplistic for in that it ranks exact type match high and non-exact type match low. We will want to expand upon this going forward.
- `FirstOrDefault()` then simply returns the best match

As an example of what the sorting is there for:

Say we have a method `Foo(IFramingElement)` and a method `Foo(Beam)`. 
If one calls `RunExtensionMethod` providing a `Column` it will only match and call the first method,  `Foo(IFramingElement)`. 
But, if one calls `RunExtensionMethod` providing a `Beam` it will match with both methods, and this is where we need the sorting, to make sure we rank the method `Foo(Beam)` higher than `Foo(IFramingElement)`. What we are trying to do is to match the method overloading that c# does automatically for you normally.. 
Both methods would be "a correct" overload, but we need to figure out which of them that is most correct, if that makes sense?

Reflection_Engine/Compute/RunExtentionMethod.cs
IsakNaslundBh: Just a leftover, no real reason.

Could change if you think it is important?

Reflection_Engine/Compute/RunExtentionMethod.cs
epignatelli: As per our last chat, I see the picture - I'll get this approved!

IsakNaslundBh:
Thanks @epignatelli

PR #1089, opened 2019-07-08
BHoM_Engine-#1088-DeleteBHoM_Engine.nuspec

### LINKED ISSUE #1088: BHoM_Engine: BH.Engine.Base not found
### Description:
Methods in BH.Engine.Base are not found anymore:
image

same goes for [the CI that can't find BH.Engine.Base](https://github.com/BHoM/BHoM_Adapter/pull/120#issuecomment-509238379):
image

### Steps to reproduce:
As in above description; try to add `using BH.Engine.Base` in any other solution that references BHoM_Engine. It should compile fine, but if you try to raise a PR you should get an error from CI.

Equally, find should fail for any method included in the `BH.Engine.Base` namespace in any UI.

alelom:

Closes #1088

As discovered by @epignatelli, seems like there was a Nuget package reference that should't have been around anymore.

Removing that solves the issue:
- The CI is able to find `BH.Engine.Base` references successfully: https://github.com/BHoM/BHoM_Adapter/pull/120
- The methods included in `BH.Engine.Base` are again available in the UI: 
   

I'm sure this issue wasn't around until very recently -- one theory is that it might have been triggered by the renumbering of the Assembly versions.


epignatelli:
@al-fisher I am not sure what is the CI exactly checking. I was thinking that it should also check the repos downstream the one interested. E.g. if @alelom modified the BHoM_Engine repo, in this case, it would be helpful if the CI was checking if the BHoM_Adapter was compiling successfully - i.e. it doesn't break other repos. We might add it only in the core bhom repos, maybe include the UIs as well?

al-fisher:
>  I am not sure what is the CI exactly checking.

Thanks @epignatelli - I am thinking to split the CI Compilation checks into two specific categories
1) Compilation of the Repo and its dependencies
2) Compilation of Dependants (downstream)

The appveyor CI that is being activated currently is focused on 1 only.
I propose to activate check 2 as an additional discrete test or check - with potential to decide and control if the downstream dependants check 2 is only carried out once PR is raised (rather than with every branch commit. Although with move to leaning more on draft PRs - last point is a little moot - however very useful to be a discrete and separate check

PR #1095, opened 2019-07-09
Geometry_Engine: re-implement UVCount() and Degrees()

### LINKED ISSUE #1094: Geometry_Engine: Implement UVCount, Degrees() and Bounds() for NurbsSurface
#### Description:

`Geometry.Query.UVCount(NurbsSurface)` and `Geometry.Query.Degrees(NurbsSurface)` are required to re-enable the conversion of `NurbsSurface`s from and to Rhinoceros.

@al-fisher @rwemay @alelom FYI this is one of the fix to include nurbs into the beta

epignatelli:

Closes #1094 

In order to reinstate the ability to convert `NurbsSurface` to and from Rhinoceros.
See the issue for more.

### Test files

I simply reinstated the code that was in these commits, which, for the sake of celerity, I assume it is correct. The test file will come with the associated Grasshopper pull request.

### Changelog

- Add `BH.Engine.Geometry.Query.UVCount()`
- Add `BH.Engine.Geometry.Query.Degrees()`
- Add `BH.Engine.Geometry.Query.Bounds(NurbsSurface)`

### DIFF
--- a/Geometry_Engine/Query/Bounds.cs
+++ b/Geometry_Engine/Query/Bounds.cs
@@ -255,10 +255,9 @@ namespace BH.Engine.Geometry
 
         /***************************************************/
 
-        [NotImplemented]
         public static BoundingBox Bounds(this NurbsCurve curve)
         {
-            throw new NotImplementedException();
+            return curve.ControlPoints.Bounds(); // TODO: Bounds(NurbsCurve) incorrect
         }
 
         /***************************************************/
@@ -313,10 +312,9 @@ namespace BH.Engine.Geometry
 
         /***************************************************/
 
-        [NotImplemented]
         public static BoundingBox Bounds(this NurbsSurface surface)
         {
-            throw new NotImplementedException();
+            return surface.ControlPoints.Bounds(); // TODO: Bounds(NurbsSurface) incorrect
         }
 
         /***************************************************/
--- a/Geometry_Engine/Query/Bounds.cs
+++ b/Geometry_Engine/Query/Bounds.cs
@@ -257,7 +257,7 @@ namespace BH.Engine.Geometry
 
         public static BoundingBox Bounds(this NurbsCurve curve)
         {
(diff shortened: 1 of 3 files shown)


Geometry_Engine/Query/Bounds.cs
```diff
@@ -255,10 +255,9 @@ public static BoundingBox Bounds(this Line line)
 
         /***************************************************/
 
-        [NotImplemented]
         public static BoundingBox Bounds(this NurbsCurve curve)
         {
-            throw new NotImplementedException();
+            return curve.ControlPoints.Bounds(); // TODO: Bounds(NurbsCurve) incorrect
```
pawelbaran: Why is the comment left in this case? If there is a bug, please raise an issue and either refer to it or delete the comment altogether.

epignatelli:
@pawelbaran they are indeed incorrect, that's exactly what the code comment is saying.
The reason for all of this is to re-enable the conversion from and to Rhino, which has to be included in the beta. Specifically, `Bounds` must be implemented, because the preview in the UIs rely on those.

I have no idea how to calculate the `BoundingBox` of a `NurbsSurface`, and I don't think it is a task we can do anyway in the next 4 hours (beta release).
`ControlPoints.Bounds();` is the best trade-off I found between the time we have and the correctness of the method.

pawelbaran:
If the method is not correct, could you please raise an issue instead of leaving a comment? This will be much easier to manage in the future - it took me a longer moment to understand what did you mean and what you are trying to achieve with this PR.

I will delete the comments and approve the PR to speed up the process, but please raise a proper issue as you are enforcing this bug on purpose.

epignatelli:
Thanks @pawelbaran, I appreciate that! 
This pr now opens https://github.com/BHoM/BHoM_Engine/issues/1097

epignatelli:
@alelom because of the above, would you mind reviewing this?

PR #1132, opened 2019-08-08
BHoM_Engine: Prototyping reflection of external libraries

### LINKED ISSUE #1131: BHoM_Engine: Prototyping reflection of external libraries
#### Description:

The aim of this feature is to be able to call methods reflected from libraries that do not belong to the BHoM framework, i.e. Numpy

epignatelli:
There is no actual code-wise dependency. 
Although, to test the prototype, you need to switch to the correspondent (identically named) branches in the BHoM_UI and the Grasshopper_Toolkit, and this pr from the Numpy_Toolkit: https://github.com/BHoM/Numpy_Toolkit/pull/2
   
Closes #1131
This pr is adding a way to reflect assemblies that are not BHoM-compliant, but can be useful in the BHoM workflow.
I have put together a quick wiki for a more exhaustive explanation of the principles I followed, and how to implement another in the future: https://github.com/BHoM/documentation/wiki/The-%60External%60-class

The other main change is this pr is the migration of some of the methods from the UI, that was not really in the best place.
I will put specific comments on the diffs to explain them, I think it is easier to explain and review without missing something.

### Test files

The test comes with the PR on the Numpy_Toolkit: https://github.com/BHoM/Numpy_Toolkit/pull/2

### Changelog

- Migrated `GroupMethodsByName` from the UI to the `Data_Engine` and renamed it to `GroupByName`, since it is a generic method
- Refactored the UI method `CompileFunction` to `Engine.Reflection.Convert.ToFunc` in order to transform any delegate to a `Func<object[], object>`
- Provide a way to expose instance methods, and any method that returns `void`
- Migrate `Engine.Serialiser.Create.MethodBase` to `Engine.Reflection.Create.MethodBase`. The `List<string> paramTypes` now is the list of name of the types, rather than their json representation
- Added a way to create `MethodBase`s via a `List<Type>` as well, for robustness and performance
- Migrate `Engine.Serialiser.Create.MethodInfo` to `Engine.Reflection.Create.MethodInfo`
- Add the `IsExposed` method to centralise the check of which methods to expose to the UIs.
- Add `Engine.Reflection.Query.ExternalMethodList` as a way to collect methods from external libraries
- Add ability to collect methods declared in nested classes


Data_Engine/Modify/GroupByName.cs
```diff
@@ -0,0 +1,80 @@
...
+
+namespace BH.Engine.Data
+{
+    public static partial class Modify
+    {
+        /*************************************/
+        /**** Public Methods              ****/
+        /*************************************/
+
+        public static Tree<T> GroupByName<T>(this Tree<T> tree)
```
epignatelli: This method has been migrated from the UI: https://github.com/BHoM/BHoM_UI/blob/2ad2b8a3a7c1a27596a5c88097da34d28636e8f8/UI_Engine/Modify/GroupMethodsByName.cs#L39
Old:
```
public static Tree<MethodBase> GroupMethodsByName<MethodBase>(this Tree<MethodBase> tree)
```
New:
```
public static Tree<T> GroupByName<T>(this Tree<T> tree)
```

Note that the name `MethodBase` was actually used as the name of a generic parameter, and was not a real specialisation of a generic method.

Reflection_Engine/Convert/ToFunc.cs
```diff
@@ -0,0 +1,118 @@
...
+namespace BH.Engine.Reflection
+{
+    public static partial class Convert
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+
+        public static Func<object[], object> ToFunc(this MethodBase method)
```
epignatelli: This set of methods are the port of `CompileFunction` in the BHoM_UI: https://github.com/BHoM/BHoM_UI/blob/d065743b720aa171b442b91de4b07da27974509b/BHoM_UI/Templates/MethodCaller.cs#L144

It has now been refactored into a `Convert` method that returns a ` Func<object[], object>`.
The convergence of any method to a `Delegate` that takes `object[]` as input and returns `object` is necessary to generalise the `MethodCaller` in the UI.

Reflection_Engine/Convert/ToFunc.cs
```diff
@@ -0,0 +1,118 @@
...
+        {
+            ParameterExpression lambdaInput = Expression.Parameter(typeof(object[]), "x");
+            Expression[] inputs = ctor.GetParameters().Select((x, i) => Expression.Convert(Expression.ArrayIndex(lambdaInput, Expression.Constant(i)), x.ParameterType)).ToArray();
+            NewExpression constructorExpression = Expression.New(ctor as ConstructorInfo, inputs);
+            return Expression.Lambda<Func<object[], object>>(Expression.Convert(constructorExpression, typeof(object)), lambdaInput).Compile();
+        }
+
+        /***************************************************/
+
+        public static Func<object[], object> ToFunc(this Action<object[]> act)
```
epignatelli: This one solves https://github.com/BHoM/BHoM_UI/issues/118

Reflection_Engine/Convert/ToFunc.cs
```diff
@@ -0,0 +1,118 @@
...
+        /***************************************************/
+
+        public static Func<object[], object> ToFunc(this Func<object, object[], object> func)
+        {
+            return inputs => { return func(inputs[0], inputs); };
+        }
+
+        /***************************************************/
+
+        public static Func<object[], object> ToFunc(this Action<object, object[]> act)
```
epignatelli: This one solves https://github.com/BHoM/BHoM_UI/issues/118 for instance methods

Reflection_Engine/Convert/ToFunc.cs
```diff
@@ -0,0 +1,118 @@
...
+        /***************************************************/
+
+        public static Func<object[], object> ToFunc(this Action<object[]> act)
+        {
+            return inputs => { act(inputs); return true; };
+        }
+
+        /***************************************************/
+
+        public static Func<object[], object> ToFunc(this Func<object, object[], object> func)
```
epignatelli: This one solves the convergence to the `delegate` above for instance methods

Reflection_Engine/Create/MethodBase.cs
```diff
@@ -33,7 +33,7 @@ public static partial class Create
         /**** Public Methods                    ****/
         /*******************************************/
 
-        public static MethodBase MethodBase(Type type, string methodName, List<string> paramTypes)
+        public static MethodBase MethodBase(Type type, string methodName, List<string> paramTypeNames)
```
epignatelli: This change is necessary to solve the dependency between the `Serialiser_Engine` and the `Reflection_Engine`.

Reflection_Engine/Create/MethodBase.cs
```diff
@@ -66,5 +66,25 @@ public static MethodBase MethodBase(Type type, string methodName, List<string> p
         }
 
         /*******************************************/
+
+        public static MethodBase MethodBase(Type type, string methodName, List<Type> paramTypes)
```
epignatelli: This method improves the retrieval of `MethodBase`s by using the `Type.GetMehod` method rather than using `Type.GetMethods` and iterate on them.

Although, it fails on generic types, since we compile the methods on the constraints of the generic types, rather than on the generic types themselves.

It also raises  https://github.com/BHoM/BHoM_Engine/issues/1138

Reflection_Engine/Create/MethodBase.cs
```diff
@@ -66,5 +66,25 @@ public static MethodBase MethodBase(Type type, string methodName, List<string> p
...
+                method = type.GetMethod(methodName, paramTypes.ToArray());
+
+            // the above will return null if the type is a generic type
+            // that is because we serialise a generic method with its constraints
+            // rather than serialising the generics
+            if (method != null)
+                return method;
+
+            // So, let's try the other overload
+            return MethodBase(type, methodName, paramTypes.Select(x => x.Name).ToList());
```
epignatelli: Uses the old overload in case the new fails when dealing with generic types (I believe around 10% of the cases, maybe less)

Reflection_Engine/Create/MethodInfo.cs
```diff
@@ -0,0 +1,78 @@
...
+
+namespace BH.Engine.Reflection
+{
+    public static partial class Create
+    {
+        /*************************************/
+        /**** Public Methods              ****/
+        /*************************************/
+
+        public static MethodInfo MethodInfo(this Type declaringType, string methodName, List<Type> paramTypes)
```
epignatelli: Migration from the Serialiser_Engine

Reflection_Engine/Modify/MakeStatic.cs
```diff
@@ -0,0 +1,70 @@
...
+    public static partial class Modify
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Returns the specific type from a type that contains generic parameters")]
+        [Input("method", "The generic type")]
+        [Output("type", "The specific type constructed from the generic one")]
+        public static MethodInfo MakeStatic(this MethodInfo method)
```
epignatelli: This method is not correct, because I was not really able to convert an instance method into a static method.. Thus, it is not really used. We may consider removing it altogether.

Reflection_Engine/Query/IsExposed.cs
```diff
@@ -0,0 +1,78 @@
...
+
+namespace BH.Engine.Reflection
+{
+    public static partial class Query
+    {
+        /***************************************************/
+        /**** Public Methods - Interfaces               ****/
+        /***************************************************/
+
+        public static bool IIsExposed(this object obj)
```
epignatelli: This one centralises the check of what to reflect to the UIs. 
The current approach spreads around a combination of `IsNotImplemented` and `IsDeprecated`.

Reflection_Engine/Query/MethodList.cs
```diff
@@ -59,6 +59,18 @@ public static List<MethodBase> AllMethodList()
             return m_AllMethodList;
         }
 
+        /***************************************************/
+
+        public static List<MethodBase> ExternalMethodList()
```
epignatelli: This one collects the methods that come from external libraries.

I have chosen to store the methods into a standalone field to avoid contaminating the other ones. 
The reason is that the namespace of the methods the field stores do not start with `BH`.

Reflection_Engine/Query/NestedMethods.cs
```diff
@@ -0,0 +1,52 @@
...
+
+namespace BH.Engine.Reflection
+{
+    public static partial class Query
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        public static List<MethodInfo> NestedMethods(this Type type)
```
epignatelli: This add the feature to retrieve methods declared in nested classes

Serialiser_Engine/Objects/BsonSerializers/MethodBaseSerializer.cs
```diff
@@ -88,18 +88,20 @@ public override MethodBase Deserialize(BsonDeserializationContext context, BsonD
             bsonReader.ReadName();
             string methodName = bsonReader.ReadString();
 
-            List<string> paramTypes = new List<string>();
+            List<string> paramTypesJson = new List<string>();
```
epignatelli: this improves the clarity of the variable: it's not the name of the type, it is its json representation

Serialiser_Engine/Objects/BsonSerializers/MethodBaseSerializer.cs
```diff
@@ -88,18 +88,20 @@ public override MethodBase Deserialize(BsonDeserializationContext context, BsonD
...
+                paramTypesJson.Add(bsonReader.ReadString());
             bsonReader.ReadEndArray();
 
             context.Reader.ReadEndDocument();
 
             try
             {
-                MethodBase method = Create.MethodBase((Type)Convert.FromJson(typeName), methodName, paramTypes);
+                List<Type> types = paramTypesJson.Select(x => Convert.FromJson(x)).Cast<Type>().ToList();
+                MethodBase method = Create.MethodBase((Type)Convert.FromJson(typeName), methodName, types); // type overload
```
epignatelli: Passing the actual types to the reflection engine. In the current master, `Create.MethodBase` lies in the Serialisation_Engine exactly because `paramTypes` is a json string, and not the name of the method.

By deserialising the types before the method call allows us to move the `Create.MethodBase` in its correct namespase: `Engine.Reflection`.

Reflection_Engine/Modify/MakeStatic.cs
adecler: Yes, I suppose you now have figured out why it cannot work after your work on the UI side.
I would suggest to remove this then. No point on leaving public methods that don't work in the code.

epignatelli:
Thanks @adecler - yeah, that method doesn't have reason to exist.
Removed in the last commit.

PR #1134, opened 2019-08-14
Geometry_Engine: Added Functions for Distance

### LINKED ISSUE #1133: Geometry_Engine: Add Distance methods for Line, Arc, Circle, PolyLine, PolyCurve
#### Description:

As a support issue to other toolkits, I add an issue here to add Distance methods for Line, Arc, Circle, Polyline and Polycurve (as in title).

### LINKED ISSUE #1140: Geometry_Engine: Implement CurveProximity methods and move them to Compute
#### Description:
There is a CurveProximity.cs file, but:
1. It does not contain relevant methods, only one `NotImplemented` and one private helper
2. It sits in the Geometry_Engine/Query, while more appropriate location would be Geometry_Engine/Compute

Actions to be taken:
- [x] move the .cs file to Geometry_Engine/Compute
- [x] change the class from `Query` to `Compute`
- [x] fix the broken references to the `SkewLineProximity` method
- [x] implement the `CurveProximity` methods for Line, Arc, Circle, Polyline and PolyCurve
- [ ] add the relevant interface method

KonradStolarski:
... for Lines, Arcs, Circles, Polylines, Polycurves

Fixes #1133 
Fixes #1140 

Adds required functionality to further work on other toolkits.

### Test files

https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/01_Test%20Scripts/BHoM_Engine/Geometry_Engine/Compute/CurveProximity.gh?csf=1&e=NgNavl

### Changelog

Added functions related to Distance(ICurve, ICurve).
Added Testfile. The testfile uses "Distance()" method, which uses the distance between our points calculated by "CurveProximity()" method.


KonradStolarski:
I just noticed a huge problem with the code, so i'm adding WIP for now.

pawelbaran:
@KonradStolarski please change the name of the PR to include a relevant suffix (in this case it is _Geometry_Engine: _)

KonradStolarski:
By the way, @epignatelli @adecler , should I leave CurveProximity output as BH.oM.Reflection.Output<T1,T2>, or change it back to Point[]?

epignatelli:
I think we should avoid returning arrays altogether, unless there is no work around. and then strictly necessary.

As per the choice of `Output<>`, I am fine with that. I think it is up to you whether to return a `Tuple<>` or an `Output<>`. The difference is in the UI:
- `Tuple<>` will render in a component with one output
- `Output<>` will render in a multi-output component

image

Geometry_Engine/Compute/CurveProximity.cs
```diff
@@ -0,0 +1,549 @@
...
+        public static BH.oM.Reflection.Output<Point,Point> ICurveProximity(this ICurve curve1, ICurve curve2, double tolerance = Tolerance.Distance)
+        {
+            return CurveProximity(curve1 as dynamic, curve2 as dynamic);
+        }
+
+        /***************************************************/
+        /**** Private Methods                           ****/
+        /***************************************************/
+
+        internal static double[] SkewLineProximity(this Line line1, Line line2, double angleTolerance = Tolerance.Angle)
```
epignatelli: Is there a specific reason why this is `internal`? Why can't it fall into either `public` or `private`?

pawelbaran: That was my suggestion. I do not see the `public` use of the method (and do not want to flood the UI with useless stuff), but it cannot be `private` as it is used in 2 classes, therefore `internal`. What do you think?

pawelbaran:
@epignatelli concerning the `Output<T,T>` vs `Tuple<T,T>` I believe the former would be more suitable here: we have 2 separate inputs for curves, will be nice to have 2 separate outputs too.

Geometry_Engine/Compute/CurveProximity.cs
epignatelli: I remember we have had a pretty bad experience with `internal` methods.  Usually when you don't see a use case for a method is just because you haven't found it yet (it may be whatever). I vouch for the `public` one in here.

pawelbaran: What do you mean by pretty bad experience?

I would prefer not to change the return type of the method, which would result in a public method returning an array. Are you fine with that?

epignatelli: Mainly about wanting to access the method from other assemblies, and we couldn't. @adecler and @IsakNaslundBh can tell more than me about that.

For the return type, it shouldn't return an array even if it is `private` or `internal` preferably.

As for not exposing the method, take a look at the Reflection_Engine. There are a lot of methods that have very few use cases, some of them has none, but they're `public` anyway.

pawelbaran: Yeah, the point of using `internal` is to expose the method to other classes inside the assembly  while preventing it from being used by other assemblies, right? Concerning the return type, `SkewLineProximity` has been there for some time, not sure if there is any point in changing it now.

Just to be clear, I generally agree with your points, but in this particular case my idea was to implement as few changes as possible, i.e. not to expose the hidden method and not to change its return type. That's why `internal` and array.

epignatelli: I get your point and I still don't see a reason for not exposing it. Exposing even a hundred methods will not flood the UI, since we have more than 5000 methods there already.

In any case, we should not use access modifiers as a tool to hide/expose items from the UI, because it's not their purpose. They are usually used because exposing it publicly harms the logic of the class it is in, making it invalid - which is (almost) never our case with `static` classes.

Few changes possible is an empty diff - you mean to change as little as possible to implement the functionality? If so, compliance is part of that. The method is still not compliant, so at one point it will be changed anyway.

At least this is my opinion, but it's up to @adecler, I can just advice on what I see.

pawelbaran: The reason for not exposing is that it is a helper method that was sitting there for a year and would still be kept `private` if it was not used in 2 different classes due to the changes made in this PR.

Following your logic, does it mean that any helper method, whatever it does, if used by two classes need to go `public`?

epignatelli: No, because there is no such thing as a helper method the way the framework is configured.
Every method should be `public`, unless there is a very valid reason for not doing so, or it sits in a particular part of the diamond (e.g. Adapter, UI).

Again, I haven't require changes to the pr, feel free to do as you please.

pawelbaran: > No, because there is no such thing as a helper method the way the framework is configured.
> Every method should be `public`, unless there is a very valid reason for not doing so, or it sits in a particular part of the diamond (e.g. Adapter, UI).

Is that documented somewhere? Personally I have seen more than a few `private` helper methods, also created a few myself, never heard any complaints. Just trying to get my head around this.

KonradStolarski:
This is my first testfile prepared, could You let me know if it is correct or should I do something with it?

Geometry_Engine/Compute/CurveProximity.cs
pawelbaran: After another thought and a background conversation with @epignatelli I would opt for making `SkewLineProximity` public with `Output<double, double>` return type. However, it remains unclear to me whether the `private` helper methods are not compliant and will be removed/refactored or just not preferable - @al-fisher @rwemay @adecler could you please reflect on that? Thanks!

Geometry_Engine/Compute/CurveProximity.cs
```diff
@@ -0,0 +1,1004 @@
...
+        }
+
+        /***************************************************/
+
+        [NotImplemented]
+        public static BH.oM.Reflection.Output<Point, Point> CurveProximity(this ICurve curve1, NurbsCurve curve2, double tolerance = Tolerance.Distance)
+        {
+            throw new NotImplementedException();
+        }
+
```
pawelbaran: Headers are preceded by 2 blank lines

Geometry_Engine/Compute/CurveProximity.cs
```diff
@@ -0,0 +1,1004 @@
...
+                //    oldresult2.Item2 = result2.Item2.Clone();
+                //    result2.Item1 = curve2.ClosestPoint(result2.Item2);
+                //    result2.Item2 = curve1.ClosestPoint(result2.Item1);
+                //} while (oldresult2.Item2.Distance(result2.Item2) > tolerance * tolerance && oldresult2.Item1.Distance(result2.Item1) > tolerance * tolerance);
+                //if (result.Item1.Distance(result.Item2) > result2.Item1.Distance(result2.Item2))
+                //    result = result2;
+            }
+            else
+            {
+                Line intersect = Create.Line(curve1.Centre(), curve2.Centre);
```
pawelbaran: Please use `Line intersect = new Line {Start = curve1.Centre(), End = curve2.Centre};`

Geometry_Engine/Compute/CurveProximity.cs
```diff
@@ -0,0 +1,1004 @@
...
+
+        public static BH.oM.Reflection.Output<Point, Point> CurveProximity(this Arc curve1, Circle curve2, double tolerance = Tolerance.Distance)
+        {
+            if (curve1.CurveIntersections(curve2).Count > 0)
+                return new BH.oM.Reflection.Output<Point, Point> { Item1 = curve1.CurveIntersections(curve2)[0], Item2 = curve1.CurveIntersections(curve2)[1] };
+            BH.oM.Reflection.Output<Point, Point> result = new BH.oM.Reflection.Output<Point, Point>();
+            Plane ftPln1 = curve1.FitPlane();
+            Plane ftPln2 = curve2.FitPlane();
+            List<Point> intPts = new List<Point>();
+            Point tmp = new Point();
```
pawelbaran: Why is this Point instantiated here?

Geometry_Engine/Compute/CurveProximity.cs
```diff
@@ -0,0 +1,1004 @@
...
+                            result.Item1 = curve1.ClosestPoint(fitPoints[1]);
+                            check = true;
+                    }
+                }
+            }
+            if (check)
+            {
+                do
+                {
+                    oldresult.Item1 = result.Item1.Clone();
```
pawelbaran: Are all `Clone` calls necessary?

Geometry_Engine/Compute/CurveProximity.cs
```diff
@@ -0,0 +1,1004 @@
...
+            BH.oM.Reflection.Output<Point, Point> result2 = new BH.oM.Reflection.Output<Point, Point>();
+            BH.oM.Reflection.Output<Point, Point> result3 = new BH.oM.Reflection.Output<Point, Point>();
+            BH.oM.Reflection.Output<Point, Point> result4 = new BH.oM.Reflection.Output<Point, Point>();
+            BH.oM.Reflection.Output<Point, Point> result5 = new BH.oM.Reflection.Output<Point, Point>();
+            BH.oM.Reflection.Output<Point, Point> result6 = new BH.oM.Reflection.Output<Point, Point>();
+            Plane ftPln1 = curve1.FitPlane();
+            Plane ftPln2 = curve2.FitPlane();
+            List<Point> fitPoints = new List<Point>();
+            bool check = false;
+            bool isSet1, isSet2, isSet3;
```
pawelbaran: This Booleans stay unused

Geometry_Engine/Compute/CurveProximity.cs
KonradStolarski: I actually think that yes, I want to pass an object, not exactly a reference to an object. If I passed there result.item1, and then change result.item1, this would also modify oldresult.item1, as oldresult was only having a reference to an object, not a "physical" (funny word in programming) object.

KonradStolarski: But I will check if they are necessary, and if not, then will delete them.

KonradStolarski:
About SkewLineProximity, needed to change LineIntersection as well, as it was using this method.

Geometry_Engine/Compute/CurveProximity.cs
KonradStolarski: in 2 places it needs to be instantiated because of later comparisons/checks. In other places though, i removed the allocation.

Geometry_Engine/Compute/CurveProximity.cs
KonradStolarski: Update 1.
They aren't.

Geometry_Engine/Compute/CurveProximity.cs
pawelbaran: In this case it should be null, instantiating (0,0,0) is risky as it might meet some `if` statement (edge case). Wouldn't that be possible?

FraserGreenroyd:
As discussed with @pawelbaran - I will review this this morning/afternoon as well just to provide a secondary eye over things :smile:

Geometry_Engine/Query/Distance.cs
```diff
@@ -142,6 +142,215 @@ public static double Distance(this Point point, NurbsCurve curve)
         }
 
 
+        /***************************************************/
+        /****       Public Methods - Curve/Curve        ****/
+        /***************************************************/
+
+        public static double Distance(this Line curve1, Line curve2, double tolerance = Tolerance.Distance)
```
IsakNaslundBh: Question regarding all of those methods. Should we not be able to get away with just one, calling the interface methods for the Curve proximity, as :

```
 public static double Distance(this ICurve curve1, ICurve curve2, double tolerance=Tolerance.Distance)
        {
            BH.oM.Reflection.Output<Point, Point> results = curve1.ICurveProximity(curve2,tolerance);
            return results.Item1.Distance(results.Item2);
        }
```

KonradStolarski: Was thinking about this, because all of them is basically the same, but wasn't sure if this shouldn't be split like the rest. I can of course change this to the form accepting (ICurve, ICurve) if this is preferred way. Will have to just exclude NurbsCurve and Elipse.

pawelbaran: I believe @IsakNaslundBh is right - probably even NurbsCurve and Ellipse could be called dynamically -> it will hit against the NotImplemented `CurveProximity` method.

IsakNaslundBh: Can understand why you added them in, but the reasons they are split in general is due to necessity rather than that we want to. Always better to keep as few as we can (while still keeping all functionality) to keep both the code more maintainable as well as avoiding to flood the UI. I would go for updating to this.

Regarding the ellipse and NurbsCurve, good to get them into the curve proximity methods and just returning NotImplemented exception as well as `[NotImplmented]` attribute (this might already have been done). Then this will return a correct error, and will start working straigh away when/if they get added.

KonradStolarski: Okay, will correct in a second then.

KonradStolarski: Done.

Issue #1119, opened 2019-07-25
Diffing_Engine: profiling


alelom:
Create profiling tests and record results.

### ProfilingTest_01

For small to medium number of objects (10 to 10000) the algorithm has an efficiency of `O(n log n)`
![image](https://user-images.githubusercontent.com/6352844/61910084-90d76300-af2b-11e9-8420-c2369f14304a.png)

For larger numbers it seems that the efficency drops to `n²` or worse.

A big, probably unnecessary performance hit is caused by how we are forced to retrieve a specific fragment:
![image](https://user-images.githubusercontent.com/6352844/61909886-245c6400-af2b-11e9-8ec6-7f2aa9616722.png)

Largest percentage of performance is however lost in this loop:
![image](https://user-images.githubusercontent.com/6352844/61909997-5f5e9780-af2b-11e9-83c4-66ede2c0de42.png)

If deemed necessary there is room for improvement here.

alelom:
After removing the call to `GetHashFragment()`, storing the hashes and objects in a dictionary instead, the performance has improved: bottleneck becomes (as it should) the "property level diffing", that is implemented through the Kellerman Library in `BH.Engine.Testing.DifferentProperties`:
![image](https://user-images.githubusercontent.com/6352844/63603088-c6ba4680-c5c0-11e9-8402-fea8d58f1ea1.png)

The picture shows the result for 1000 elements compared by property.

For 100 elements, `DifferentProperties()` was closer to 80%, with the other 20% occupied by the "collection-level diffing".

This demonstrates that the more elements, the more the computation is taken by the "property-level diffing", as expected.

alelom:
After https://github.com/BHoM/BHoM_Engine/pull/1248

![image](https://user-images.githubusercontent.com/6352844/66659806-0ce46b80-ec3c-11e9-90f8-0b94a09d7b4f.png)

alelom:
With https://github.com/BHoM/BHoM_Engine/pull/1952, performance is about x100 better, especially on large collections.

![image](https://user-images.githubusercontent.com/6352844/91320420-63f02200-e7b5-11ea-9490-4cb18b064d4c.png)

![image](https://user-images.githubusercontent.com/6352844/91320450-6f434d80-e7b5-11ea-88d8-ed1d71ab7f0e.png)

alelom:
[...]

Issue #1113, opened 2019-07-24
Diffing_Engine: create initial examples (=unit tests) for different scenarios


alelom:
Create examples of different usage cases.

Should be used both as examples and as Unit Tests.

I will keep this post updated with new examples.

## Test_01
I've tried to keep this test clean and with a lot of comments, so it can be followed as a guide to understand how I envisioned the diffing_engine
https://github.com/BHoM/BHoM_Engine/blob/754b70841535a2596c61bc2c323c269e5c60cf49/Engine_Test/Diffing_Engine/Test01.cs#L21-L89

PR #1150, opened 2019-08-28
Diffing_Engine : Initial implementation

### LINKED ISSUE #1100: Diffing_Engine: Implement hashing
After some thoughts and discussions, we concluded that the diffing (at least its first step, which is the  "collection-level" diffing) should rather work with hashes than with custom comparers.

This is for several reasons:
* Comparing two objects is not enough to keep track of **who** has changed from an iteration to the next. When you modify the objects, you can't tell who originally was different. The old version of the hash can be saved in the object itself, making possible to discern which objects have been modified in different iterations.
* The base scenario of comparing two sets of object is expecting a deep (complete) comparison, so by default the diffing should look at all their properties. In some cases it could be required _not_ to consider some of their properties though; this is where the hashing will have to be computed with some _exceptions_.
* Hashing has the potential to make the "collection-level" diffing fast and data transmission lighter. 

Downside:
* Using hashing you lose the possibility of having two exactly identical objects. For example, you can't have two identical beams in the same position. 

I don't think the downside should be seen as a problem. I can't find any good reason why we should support duplicates. If a case like that exists, I think we should rather rethink why/how that case is allowed to exist. 

## 1. Hashing requirements
1. Generate a hash that acts as a "Fingerprint" for our objects.
2. The fingerprint must **exclude** a certain set of properties:
    - all properties that are change every time the user opens/closes the solution, notably `BH.oM.Base.BHoMObject GUID`
    - If the hash of the object is saved in the objects (e.g. in the Fragments), then that must be excluded as well.
    - Any other property that we decide is appropriate to exclude (`BH.oM.Base.BHoMObject CustomData`?)

## 2. Algorithm to generate the hash
The standard .NET GetHash() returns different results based on the environment. In other words, for example, [for the same identical string two different users could get two different hashes](https://stackoverflow.com/questions/4466132/string-gethashcode-returns-different-values).

**So we can't just implement the standard  .NET GetHash()**.

Since we need a platform-independent algorithm to generate hashes, we could use:
* [MD5](https://stackoverflow.com/a/2435932/3873799), however that's generally regarded as non-safe against clashes.
* SHA or variants (they are more safe against clashes)

## 3. Serialization to generate hashing
Hashing algorithms require the object to be **Serialised in Byte[]**.

The serialisation requires all classes to be marked as `[Serializable]`. Which we don't want.
We [can **not** add the Serializable attribute at runtime, either](https://stackoverflow.com/a/26957174/3873799).
So we need another serializer.

### 3.1 <del>Protobuf-net</del> (**deprecated proposal**) 
<del>A [workaround could be to use Protocol Buffers to serialise to byte[]](https://stackoverflow.com/a/7923451/3873799), selecting the properties we want to serialise. This would also allow to solve point (2): it seems easy to exclude properties we don't want to be part of the serialization (therefore not in the fingerprint).
</del>

### 3.2 Use of MongoDB
As Eduardo suggested, we already have a good serializer in Mongo_Toolkit and we can leverage it.
We just need to find a way to avoid serializing some types. These  _exceptions_ will work as custom comparer for objects.

### LINKED ISSUE #1110: Diffing_Engine: implement property-level diffing
### Description:
#1100 will identify which objects are different to each other, and at the same time track down which objects were changed between revisions (=_to be updated_).

This issue aims the implementation of a mechanism to find the differences between two objects.
This will intervene to compare objects _to be updated_ (a "new version" of the same object) and get out the properties that have changed, so we can include them in the delta.

After a chat with Arnaud, it came out that we already have something we can use:
https://github.com/BHoM/BHoM_Engine/blob/5fd2bf1871c9015c1bd138f975669471a7ee6acf/Testing_Engine/Query/IsEqual.cs#L44-L84
that implements `KellermanSoftware.CompareNetObjects`, a library we already reference.

alelom:
### NOTE: Depends on 
https://github.com/BHoM/BHoM/pull/541  

This is for a prototype of the Diffing_Engine.

Ported https://github.com/BHoM/BHoM_Engine/pull/1103 and adapted to the newly agreed concept as discussed with @al-fisher    

* Collection-level diffing: closes #1100.

* Property-level diffing: closes #1110 
   
* Tests and examples: #1113

* Profiling: #1119 
  
### Test files
Code examples: see #1113 and #1119.
Grasshopper script: https://burohappold.sharepoint.com/:f:/s/BHoM/EtVg-MNP33dHhB_XtRYojXQBXg4OXTMic7c-EQUoY-4ktw?e=XfXyCJ

### Changelog
* First prototype of the Diffing_Engine that performs Collection-level and property-level diffing for objects.


alelom:
@al-fisher have a go with the Grasshopper example file above and let me know!

al-fisher:
Great stuff - thanks @alelom will do!

FraserGreenroyd:
Hi @alelom per updated guidelines can you remove `BHoM_Engine` from the PR title (but keep `Diffing_Engine`) :smile:

epignatelli:
According to Sprint Closure call - @al-fisher , @alelom is asking for merging this to solve the milestone. To be taken offline.

al-fisher:
Adding the "Depends on:" to the PR description. Indeed obvious - and the robots :robot: :robot: found matching PR as branch names correspond - but useful for clarity to ensure referenced! 😄

PR #1161, opened 2019-09-04
Reflection_Engine: added methods' inputs parameters names in menu description

alelom:

Closes https://github.com/BHoM/BHoM_UI/issues/41

Adds parameter input names into the menu when browsing methods:
image

### Changelog

* Added parameter input names into the menu when browsing methods.


adecler:
Feels much better!
One thing that will have to be addressed in the BHoM UI though. Some method text are now insanely long:

image

I would be fine with it, if it was not for the fact that a part of the menu jumps to the next screen when it is too long. If there is no screen to the right, you just have no sub-menu.

image

Best thing would be to simply truncate the string to a given length, I have raised an issue [here](https://github.com/BHoM/BHoM_UI/issues/127)

PR #1163, opened 2019-09-05
Reflection_Engine: Added bool in ToText() method to choose follow return type instead of namespace

alelom:

Related to https://github.com/BHoM/BHoM_UI/issues/121

### Changelog

* Added bool in ToText() method to choose follow return type instead of namespace


alelom:
Moved to another Branch for CI

PR #1170, opened 2019-09-06
Geometry_Engine: Expose tolerances

### LINKED ISSUE #1156: Geometry_Engine: Expose tolerance in Query Centroid
#### Description:

Currently, we use default Distance.

image

I need to be able to adj tolerance to 0.001 but unable to do it,
 therefore, we need to expose tolerance input

### LINKED ISSUE #1157: Geometry_Engine: Expose tolerance in Query Normal
#### Description:

Currently, we use  default tolerance
image

I need the flexibility to adjust and decrease tolerance,
 therefore, we need to expose tolerance input
image

FraserGreenroyd:
Fixes #1156 
Fixes #1157 

### Test files
N/A

### Changelog
 - Expose tolerance parameter on `Normal` query for `Curve` inputs
 - Expose tolerance parameter on `Centroid` query for `Curve` inputs
 - Deprecate existing `Normal` and `Centroid` queries as a result of the above

### Additional comments
This exposes tolerances for methods which can utilise tolerances to provide greater control when utilising these geometry methods within code and visual programming.

### DIFF
--- a/Geometry_Engine/Query/Centroid.cs
+++ b/Geometry_Engine/Query/Centroid.cs
@@ -35,20 +35,31 @@ namespace BH.Engine.Geometry
         /***************************************************/
         /**** Public Methods - Curves                   ****/
         /***************************************************/
-
+        [Deprecated("2.4", "Deprecated to expose tolerance as optional parameter for greater control", null, "Centroid(this Polyline curve, double tolerance = Tolerance.Distance)")]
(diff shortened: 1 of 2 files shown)


pawelbaran:
I quickly went through the code and not sure I agree with the changes. As far as I can see, after the changes `Centroid` for some types takes `tolerance` argument, while for some it does not (incl. the interface method). 🤔

PR #1208, opened 2019-09-16
BHoM_Engine: fix fragment adding bug

### LINKED ISSUE #1206: BHoM_Engine: Bug in adding fragments
#### Description:
Adding two fragments of the same type is possible despite the attempts at checks.

#### How to replicate:
Add two fragments (different instances) of the same type to one `IBHoMObject` - the second fragment is incorrectly added.

#### Expected behaviour:
The second fragment should not be added

#### Test file(s):
[Here](https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/01_Test%20Scripts/BHoM_Engine/BHoM_Engine/FragmentAdding.gh?csf=1&e=jh2UMC)

FYI: @IsakNaslundBh

FraserGreenroyd:
Fixes #1206 

### Test files
[Here](https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/01_Test%20Scripts/BHoM_Engine/BHoM_Engine/FragmentAdding.gh?csf=1&e=jh2UMC)

### Changelog
 - Ensure that only one instance of a fragment can be added to an object at this time

### DIFF
--- a/BHoM_Engine/Modify/AddFragment.cs
+++ b/BHoM_Engine/Modify/AddFragment.cs
@@ -45,13 +45,12 @@ namespace BH.Engine.Base
             IBHoMObject o = iBHoMObject.DeepClone();
             o.Fragments = new List<IBHoMFragment>(iBHoMObject.Fragments);
 
-            if(o.Fragments.Where(x => x.GetType() == fragment.GetType()).FirstOrDefault() != null)
+            int index = o.Fragments.FindIndex(x => x.GetType() == fragment.GetType());
+
+            if(index >= 0)
             {
-                if(replace)
-                {
-                    o.Fragments.RemoveAt(o.Fragments.IndexOf(o.Fragments.Where(x => x.GetType() == fragment.GetType()).FirstOrDefault()));
-                    o.Fragments.Add(fragment);
-                }
+                if (replace)
+                    o.Fragments[index] = fragment;
                 else
                     Reflection.Compute.RecordError("That fragment already exists on this object. If you would like to replace the existing fragment set the 'replace' input to 'true'");
             }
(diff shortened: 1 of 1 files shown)


BHoM_Engine/Modify/AddFragment.cs
```diff
@@ -44,10 +44,14 @@ public static IBHoMObject AddFragment(this IBHoMObject iBHoMObject, IBHoMFragmen
             if (iBHoMObject == null) return null;
             IBHoMObject o = iBHoMObject.DeepClone();
             o.Fragments = new List<IBHoMFragment>(iBHoMObject.Fragments);
-            if(o.Fragments.Contains(fragment))
+
+            if(o.Fragments.Where(x => x.GetType() == fragment.GetType()).FirstOrDefault() != null)
```
adecler: Not the biggest fan of this. As a minimum, I would rather use:

```
o.Fragments.Find(x => x.GetType() == fragment.GetType()) != null
```

But, given what is coming below, I would prefer

```
int index = o.Fragments.FindIndex(x => x.GetType() == fragment.GetType());
if (index >= 0)
{
    if(replace)
        o.Fragments[index] = fragment;
    else
        ...
}
```

al-fisher:
>  vague memory of us discussing the possibility to use a Dictionary<Type, Fragment> instead. Is that still on the table?

I would say this is. 
And therefore we should plan for potential implementation. Although agree - this week may not be the moment to execute

BHoM_Engine/Modify/AddFragment.cs
FraserGreenroyd: Changes made :smile:

FraserGreenroyd:
Have made the requested change @adecler ready to re-review.

> But I have a vague memory of us discussing the possibility to use a `Dictionary<Type, Fragment>` instead. Is that still on the table?

Also agreed, this is still on the table and should be discussed as one of the earliest discussions at the start of the next milestone :smile:

PR #1209, opened 2019-09-16
Environment_Engine: expose multiple tolerance to prevent incorrect opening additions

### LINKED ISSUE #1195: Environment_Engine: IsContaining tolerance causing inverse downstream behaviour
#### Description:

After exposing tolerance in https://github.com/BHoM/BHoM_Engine/issues/1188 some odd behaviour was experienced resulting in low tolerances returning results whereas high tolerances (less stringent) didn't! 

#### How to replicate:

See test file.

#### Expected behaviour:

Lowering tolerance would result in no resultant downstream objects if passed more accurate geometries, while increasing tolerance would mean more *dodgy* geometry is accepted.

#### Test file(s):

[AddOpeningstest.zip](https://github.com/BHoM/BHoM_Engine/files/3605292/AddOpeningstest.zip)

FraserGreenroyd:
Fixes #1195 

### Test files
See issue raised

### Changelog
 - Expose `ContainingTolerance` on adding openings

### Additional comments
Test script may need updating to use the new method signature for this method.

### DIFF
--- a/Environment_Engine/Modify/AddOpening.cs
+++ b/Environment_Engine/Modify/AddOpening.cs
@@ -57,16 +57,17 @@ namespace BH.Engine.Environment
         [Description("Returns a list of Environment Panel with the provided openings added. Openings are added to the panels which contain them geometrically.")]
         [Input("panels", "A collection of Environment Panels to add the opening to")]
         [Input("openings", "A collection of Environment Openings to add to the panels")]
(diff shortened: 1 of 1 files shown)


FraserGreenroyd:
image

The test script may need to update so the `AddOpenings` component looks like the above @tg359

PR #1210, opened 2019-09-16
Geometry_Engine: Methods now work for open polylines

### LINKED ISSUE #1207: Geometry_Engine: RemoveLeastSignificantVertices and RemoveShortSegments does not work for open polylines
#### Description:

The methods "RemoveLeastSignificantVertices" and "RemoveShortSegments" only works for closed polylines at the moment. Make them work for open polylines as well.

RasmusCanfjorden:

Closes #1207

### Test files

https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/01_Test%20Scripts/BHoM_Engine/Geometry_Engine/Modify/RemoveLeastSignificantVertices_2019-10-14.gh?csf=1&e=qIGlV4
https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/01_Test%20Scripts/BHoM_Engine/Geometry_Engine/Modify/RemoveShortSegments_2019-10-14.gh?csf=1&e=ohKfnB

### Changelog

* Enabled RemoveShortSegments to handle open polylines.
* Enabler RemoveLeastSignificantVertices to handle open polylines.

### DIFF
--- a/Geometry_Engine/Modify/RemoveLeastSignificantVertices.cs
+++ b/Geometry_Engine/Modify/RemoveLeastSignificantVertices.cs
@@ -48,35 +48,17 @@ namespace BH.Engine.Geometry
             if (pnts.Count < 3)
                 return polyline; //If there's only two points here then this method isn't necessary
 
-            int startIndex = 0;
-
-            if (polyline.IsClosed())
+            int startIndex = 0;            
+            while (startIndex < pnts.Count)
             {
-                while (startIndex < pnts.Count)
-                {
-                    Point first = pnts[startIndex];
-                    Point second = pnts[(startIndex + 1) % pnts.Count];
-                    Point third = pnts[(startIndex + 2) % pnts.Count];
+                Point first = pnts[startIndex];
+                Point second = pnts[(startIndex + 1) % pnts.Count];
+                Point third = pnts[(startIndex + 2) % pnts.Count];
 
-                    if (first.Angle(second, third) <= angleTolerance)
-                        pnts.RemoveAt((startIndex + 1) % pnts.Count);  //Delete the second point from the list, it's not necessary
-                    else
-                        startIndex++; //Move onto the next point
-                }
-            }
-            else 
-            {
-                while ((startIndex) < (pnts.Count - 1))
-                {
-                    Point first = pnts[startIndex];
-                    Point second = pnts[(startIndex + 1) % pnts.Count];
-                    Point third = pnts[(startIndex + 2) % pnts.Count];
-
-                    if (first.Angle(second, third) <= angleTolerance)
-                        pnts.RemoveAt((startIndex + 1) % pnts.Count);  //Delete the second point from the list, it's not necessary
-                    else
-                        startIndex++; //Move onto the next point
-                }
+                if (first.Angle(second, third) <= angleTolerance)
+                    pnts.RemoveAt((startIndex + 1) % pnts.Count);  //Delete the second point from the list, it's not necessary
+                else
+                    startIndex++; //Move onto the next point
             }
 
             if (!polyline.IsClosed())
@@ -88,15 +70,11 @@ namespace BH.Engine.Geometry
                 }
             }
 
-            if (polyline.IsClosed()) //Only re-close if original polyline is closed
+            else if (pnts.First() != pnts.Last()) //Only re-close if original polyline is closed
             {
-                if (pnts.First() != pnts.Last())
-                {
-                    pnts.Add(pnts.First());
-                }
-                    
+                pnts.Add(pnts.First());
             }
-            
+
             Polyline pLine = new Polyline()
             {
                 ControlPoints = pnts,
@@ -107,3 +85,5 @@ namespace BH.Engine.Geometry
         /***************************************************/
     }    
 }
+
+
--- a/Geometry_Engine/Modify/RemoveLeastSignificantVertices.cs
+++ b/Geometry_Engine/Modify/RemoveLeastSignificantVertices.cs
@@ -41,35 +41,62 @@ namespace BH.Engine.Geometry
         [Input("angleTolerance", "The tolerance of the angle that defines a straight line. Default is set to the value defined by BH.oM.Geometry.Tolerance.Angle")]
         [Output("polyline", "The cleaned polyline")]
         public static Polyline RemoveLeastSignificantVertices(this Polyline polyline, double angleTolerance = Tolerance.Angle)
-        {
-            //This method is for closed polylines only at the moment
-            if (!polyline.IsClosed())
-            {
-                BH.Engine.Reflection.Compute.RecordError("The RemoveLeastSignificantVertices method is only for closed polylines and the input polyline is not closed.");
-                return polyline;
-            }
-
+        {            
             List<Point> pnts = polyline.DiscontinuityPoints();
+            List<Point> originalPnts = polyline.DiscontinuityPoints();
 
             if (pnts.Count < 3)
                 return polyline; //If there's only two points here then this method isn't necessary
 
             int startIndex = 0;
-            while (startIndex < pnts.Count)
-            {
-                Point first = pnts[startIndex];
-                Point second = pnts[(startIndex + 1) % pnts.Count];
-                Point third = pnts[(startIndex + 2) % pnts.Count];
 
-                if (first.Angle(second, third) <= angleTolerance)
-                    pnts.RemoveAt((startIndex + 1) % pnts.Count);  //Delete the second point from the list, it's not necessary
-                else
-                    startIndex++; //Move onto the next point
+            if (polyline.IsClosed())
+            {
+                while (startIndex < pnts.Count)
+                {
+                    Point first = pnts[startIndex];
+                    Point second = pnts[(startIndex + 1) % pnts.Count];
+                    Point third = pnts[(startIndex + 2) % pnts.Count];
+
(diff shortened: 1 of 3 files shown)


Geometry_Engine/Modify/RemoveLeastSignificantVertices.cs
```diff
@@ -41,43 +41,69 @@ public static partial class Modify
...
+                    else
+                        startIndex++; //Move onto the next point
+                }
+            }
+            else 
             {
-                Point first = pnts[startIndex];
-                Point second = pnts[(startIndex + 1) % pnts.Count];
-                Point third = pnts[(startIndex + 2) % pnts.Count];
+                while ((startIndex) < (pnts.Count - 1))
```
FraserGreenroyd: I do not believe that any of this `if` `else` is necessary - I believe the algorithm will work quite happily as it was for non-closed polylines as it did for closed polylines.

Alternatively, if the `if` `else` is necessary, then the `else` needs modifying to work with non-closed polylines, because currently that algorithm will loop to the beginning of the polyline unnecessarily (unnecessarily because the polyline is not closed so we don't need to loop to the beginning again).

Geometry_Engine/Modify/RemoveLeastSignificantVertices.cs
```diff
@@ -41,43 +41,69 @@ public static partial class Modify
...
+            if (!polyline.IsClosed())
+            {
+                if (pnts.Last() != originalPnts.Last())
+                {
+                    pnts.Remove(pnts.Last());
+                    pnts.Add(originalPnts.Last());
+                }
+            }
 
+            if (polyline.IsClosed()) //Only re-close if original polyline is closed
```
FraserGreenroyd: This can easily be an `else` rather than running a `IsClosed` check twice.

FraserGreenroyd: Use `else if(pnts.First() != pnts.Last())` to check the other bit in a single line.

Geometry_Engine/Modify/RemoveShortSegments.cs
```diff
@@ -41,20 +41,16 @@ public static partial class Modify
...
-                BH.Engine.Reflection.Compute.RecordError("The RemoveShortSegments method is only for closed polylines and the input polyline is not closed.");
-                return polyline;
-            }
-
+        {            
             List<Point> pnts = polyline.DiscontinuityPoints();
+            List<Point> originalPnts = polyline.DiscontinuityPoints();
 
             if (pnts.Count < 3)
+            {
```
FraserGreenroyd: These brackets aren't needed, remove them

Geometry_Engine/Modify/RemoveShortSegments.cs
```diff
@@ -67,17 +63,27 @@ public static Polyline RemoveShortSegments(this Polyline polyline, double minimu
...
+            if (!polyline.IsClosed())
+            {
+                if (pnts.Last() != originalPnts.Last())
+                {
+                    pnts.Remove(pnts.Last());
+                    pnts.Add(originalPnts.Last()); //Keep the original endpoint of the polyline                   
+                }
+            }
+
+            if (polyline.IsClosed())
```
FraserGreenroyd: See previous comment on the other method about using an `else` instead of two `IsClosed` checks

Geometry_Engine/Modify/RemoveLeastSignificantVertices.cs
```diff
@@ -41,20 +41,14 @@ public static partial class Modify
...
         public static Polyline RemoveLeastSignificantVertices(this Polyline polyline, double angleTolerance = Tolerance.Angle)
-        {
-            //This method is for closed polylines only at the moment
-            if (!polyline.IsClosed())
-            {
-                BH.Engine.Reflection.Compute.RecordError("The RemoveLeastSignificantVertices method is only for closed polylines and the input polyline is not closed.");
-                return polyline;
-            }
-
+        {
```
pawelbaran: I think `distanceTolerance` should also be an argument of this method:
`public static Polyline RemoveLeastSignificantVertices(this Polyline polyline,, double distanceTolerance = Tolerance.Distance, double angleTolerance = Tolerance.Angle)`

Geometry_Engine/Modify/RemoveLeastSignificantVertices.cs
```diff
@@ -41,20 +41,14 @@ public static partial class Modify
...
-            //This method is for closed polylines only at the moment
-            if (!polyline.IsClosed())
-            {
-                BH.Engine.Reflection.Compute.RecordError("The RemoveLeastSignificantVertices method is only for closed polylines and the input polyline is not closed.");
-                return polyline;
-            }
-
+        {            
             List<Point> pnts = polyline.DiscontinuityPoints();
+            List<Point> originalPnts = polyline.DiscontinuityPoints();
```
pawelbaran: The list is useless (plus duplicate of the above) - it could be `Point lastPoint = pnts.Last();`

Geometry_Engine/Modify/RemoveShortSegments.cs
```diff
@@ -41,43 +41,44 @@ public static partial class Modify
...
         public static Polyline RemoveShortSegments(this Polyline polyline, double minimumSegmentLength = Tolerance.Distance)
-        {
-            //This method is for closed polylines only at the moment
-            if (!polyline.IsClosed())
-            {
-                BH.Engine.Reflection.Compute.RecordError("The RemoveShortSegments method is only for closed polylines and the input polyline is not closed.");
-                return polyline;
-            }
-
+        {
```
pawelbaran: Same as above

Geometry_Engine/Modify/RemoveShortSegments.cs
```diff
@@ -41,43 +41,44 @@ public static partial class Modify
...
-            //This method is for closed polylines only at the moment
-            if (!polyline.IsClosed())
-            {
-                BH.Engine.Reflection.Compute.RecordError("The RemoveShortSegments method is only for closed polylines and the input polyline is not closed.");
-                return polyline;
-            }
-
+        {            
             List<Point> pnts = polyline.DiscontinuityPoints();
+            List<Point> originalPnts = polyline.DiscontinuityPoints();
```
pawelbaran: Same as above

Geometry_Engine/Modify/RemoveLeastSignificantVertices.cs
```diff
@@ -41,20 +41,14 @@ public static partial class Modify
...
-        {
-            //This method is for closed polylines only at the moment
-            if (!polyline.IsClosed())
-            {
-                BH.Engine.Reflection.Compute.RecordError("The RemoveLeastSignificantVertices method is only for closed polylines and the input polyline is not closed.");
-                return polyline;
-            }
-
+        {            
             List<Point> pnts = polyline.DiscontinuityPoints();
```
pawelbaran: `DiscontinuityPoints` should pass both distance and angle tolerances

Geometry_Engine/Modify/RemoveLeastSignificantVertices.cs
```diff
@@ -67,8 +61,18 @@ public static Polyline RemoveLeastSignificantVertices(this Polyline polyline, do
                     startIndex++; //Move onto the next point
             }
 
-            if (pnts.First() != pnts.Last())
-                pnts.Add(pnts.First()); //Reclose polyline
+            if (!polyline.IsClosed())
+            {
+                if (pnts.Last() != originalPnts.Last())
```
pawelbaran: This check is dangerous - the != operator does not take any tolerance into account, can imagine this being an issue. I would rather skip this line at all and replace the last point with the original last point `if (!polyline.IsClosed())`

Geometry_Engine/Modify/RemoveShortSegments.cs
```diff
@@ -41,43 +41,44 @@ public static partial class Modify
...
                 if (first.Distance(second) <= minimumSegmentLength)
                     pnts.RemoveAt((startIndex + 1) % pnts.Count); //Delete the second point as it is too close to the first to produce any meaningful change...
                 else
                     startIndex++; //Move onto the next point
             }
 
-            if (pnts.First() != pnts.Last())
+            if (!polyline.IsClosed())
+            {
+                if (pnts.Last() != originalPnts.Last())
```
pawelbaran: Same as above

Geometry_Engine/Modify/RemoveLeastSignificantVertices.cs
```diff
@@ -77,7 +81,8 @@ public static Polyline RemoveLeastSignificantVertices(this Polyline polyline, do
 
             return pLine;
         }
```
pawelbaran: Please leave the blank line here

FraserGreenroyd:
> I just realized: do these two methods not overlap with DiscontinuityPoints with tolerances set to values representing minimum angle and minimum segment length?

@pawelbaran we did discuss the subtle (and they are subtle) differences between this method and the `DiscontinuityPoints` when we first added `CleanPolyline` a few months ago - it was agreed this method gave us a cleaning ability that `DiscontinuityPoints` lacked so I think this point should be ok.

@RasmusCanfjorden please look into the other points @pawelbaran has raised though as, on reflection, I agree with a lot of them :smile:

pawelbaran:
Happy to tick off one of the points then!

PR #1222, opened 2019-09-25
Serialiser_Engine: Fix ToJson() method for when input is null

### LINKED ISSUE #1221: ToJson() throws exception when input is null
#### Description:

We found the error message of the BHoM object to Json component is confusing. The problem shown in the image is that the BHoM is empty, while the error message is saying not correct data. It will be helpful to change it to a warning saying empty input. 
image

Thanks.

adecler:
Closes #1221 

### Test files
See issue #1221

### DIFF
--- a/Serialiser_Engine/Convert/Json.cs
+++ b/Serialiser_Engine/Convert/Json.cs
@@ -40,6 +40,9 @@ namespace BH.Engine.Serialiser
         [Output("json", "String representing the object in json")]
         public static string ToJson(this object obj)
         {
+            if (obj == null)
+                return "";
+
             if (obj is string)
                 return "{ \"_t\": \"System.String\", \"_v\": " + BsonExtensionMethods.ToJson<string>(obj as string) + "}";
                 


adecler:
@BingWangUS, FYI.
I cannot add you as a reviewer for some reason.

BingWangUS:
Hi team,

Thanks for responding to this. However, it does not solved the bug for me, when using 2.4 alpha 66 installer, assuming this is included.

See the attach screenshot. I did not open Revit so clearly the pull function is giving empty data output. Therefore, the ToJson component is also taking empty inputs. Here is the error message:

"1. This component failed to run properly. Are you sure you have the correct type of inputs?
Check their description for more details. Here is the error provided by the method:
C# null values of type 'BsonDocument' cannot be serialized using a serializer of type 'BsonDocumentSerializer'."

The error message is misleading since it says not correct input at the beginning instead of no input. I was expecting a warning message as the second ToJson component says. See the test file here. 
[JsonError.zip](https://github.com/BHoM/BHoM_Engine/files/3658708/JsonError.zip)
 

image
 
I understand this is a small issue because the error message also says null values. So feel free to discard this.

michaldengusiak:
You need to connect type to you filter request ie Environment.Elements.Building

BingWangUS:
Thanks Michal. However, my point is, in any case the pull function is not working in this situation and not giving any data out, so the ToJson component should perform the same as the second ToJson, which I don't understand why not.

PR #1247, opened 2019-10-10
Geometry engine - Adding an Orient method and Translate, Scale and Rotate methods for Arcs, Circles and Ellipses

### LINKED ISSUE #1010: Geometry_Engine: Method to orient between two coordinatesystems
Want to be able to orient geometry from one Cartesian coordinate system to another.

### LINKED ISSUE #1236: Geometry_Engine: avoid converting arcs and circles to nurbs on Transform when not necessary
#### Description:
At the moment arcs and circles are converted to nurbs when any transformation is done on them, which leads to many `NotImplementExceptions` afterwards. Conversion could be avoided in most cases (as long as the curve is not scaled nonuniformly). A simple if statement that checks the transform matrix for uniform scaling factor should solve the problem.

FYI @IsakNaslundBh @al-fisher @epignatelli

### LINKED ISSUE #1237: Geometry_Engine: implement Scale, Rotate and Translate method for Ellipse
#### Description:
As in the header - atm the interface method crashes badly when the `ICurve` is `Ellipse`. If simple, then worth adding the actual functionality, if not, then `NotImplemented` should be enough.

LMarkowski:

Closes #1010 
Closes #1236 
Closes #1237 

- Added three new TransformMatrices:
  - `OrientationMatrixGlobalToLocal` - transforming geometry from global coordinate systems to given local coordinate system
  - `OrientationMatrixLocalToGlobal` - transforming geometry from given local coordinate systems to global coordinate system
  - `OrientationMatrix` - transform geometry between two given coordinate systems
- Added new method `Modify.Orient`, which orients geometry between two coordinate systems using `Transform(OrientationMatrix)`. **( IMPORTANT )** With @IsakNaslundBh we came up with an idea to cover all IGeometry classes with one Generics method. It works quite good for Orient so we'd like to hear your opinions and maybe discuss spreading this approach also over Rotate, Scale and Translate methods. **( /IMPORTANT )**
- Added methods for Ellipse in Rotate, Scale and Translate methods.
- Updated `Tranform(Arc)`, `Tranform(Circle)` and `Tranform(Ellipse)`. Now instead of instantly converting to Nurbscurve it checks if transformation is rigid ([by checking a determinant](https://en.wikipedia.org/wiki/Rigid_transformation#Formal_definition)) or is uniform scaling. If neither it converts to NurbsCurve with a proper note. For now the check won't pass a compounding of rigid transformation and uniform scaling despite it doesn't change the shape. (#toBeResolvedLater) But I believe it is still worth pushing as in most cases user would perform two transformations one after another rather than multiply TransformMatrices. Unless maybe you have any ideas?
- `Orient(Vector)` doesn't work as it should.. because I'm not sure how it should work. When GH orients Vectors it gives same results as when orienting Points. Should our method do the same or should the local vector point from csFrom.Origin.. or something else. Honestly I don't know.

### Test files
[Here](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?viewid=a5a93cba%2Dfcca%2D46cc%2Da31a%2D80b8d5cbfd7b&id=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FTest%20Scripts%2FBHoM%5FEngine%2FGeometry%5FEngine%2FModify) is general test file for Orient Method.
[Here](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4&viewid=a5a93cba%2Dfcca%2D46cc%2Da31a%2D80b8d5cbfd7b&id=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FTest%20Scripts%2FBHoM%5FEngine%2FGeometry%5FEngine%2FGeometry%5FEngine%231236%231237%2DAvoidConvertingArcsAndCirclesToNurbsAndAddEllipseTransforms) is test file for new Ellipse, Arc and Circle methods. I've created a separate test file because generic test files for Translate, Rotate and Scale are empty and outdated.

### Changelog

- `Create.OrientationMatrixGlobalToLocal()` public method added in the `Geometry_Engine` for `Cartesian` class
- `Create.OrientationMatrixLocalToGlobal()` public method added in the `Geometry_Engine` for `Cartesian` class
- `Create.OrientationMatrix()` public method added in the `Geometry_Engine` for `Cartesian` class
- `Modify.Orient()` generic public method added in the `Geometry_Engine` for `IGeometry` class
- `Modify.Rotate()` public method added in the `Geometry_Engine` for `Ellipse` class
- `Modify.Scale()` public method added in the `Geometry_Engine` for `Ellipse` class
- `Modify.Translate()` public method added in the `Geometry_Engine` for `Ellipse` class
- `Modify.Transform()` public methods updated in the `Geometry_Engine` for `Arc`, `Circle` and `Ellipse` class
- `Modify.IsUniformScaling()` private method added in the `Geometry_Engine` for `TransformMatrix` class

### DIFF
--- a/Geometry_Engine/Modify/Orient.cs
+++ b/Geometry_Engine/Modify/Orient.cs
@@ -1,0 +1,44 @@
+﻿/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2019, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 
+ * GNU Lesser General Public License for more details.                          
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using BH.oM.Geometry;
+using BH.oM.Geometry.CoordinateSystem;
+using BH.oM.Reflection.Attributes;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace BH.Engine.Geometry
+{
+    public static partial class Modify
+    {
+        /***************************************************/
+        /****         public Generic Interface          ****/
+        /***************************************************/
+
+        public static T Orient<T>(this T geometry, Cartesian csFrom, Cartesian csTo) where T: IGeometry
+        {
+            TransformMatrix orientationMatrix = Create.OrientationMatrix(csFrom, csTo);
+            return (T)ITransform(geometry, orientationMatrix);
+        }  
+    }
+}
--- a/Geometry_Engine/Modify/Orient.cs
+++ b/Geometry_Engine/Modify/Orient.cs
@@ -25,6 +25,7 @@ using BH.oM.Geometry.CoordinateSystem;
 using BH.oM.Reflection.Attributes;
 using System;
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.Linq;
 
 namespace BH.Engine.Geometry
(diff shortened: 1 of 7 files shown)


Geometry_Engine/Modify/Transform.cs
```diff
@@ -214,6 +249,26 @@ public static ISurface ITransform(this ISurface geometry, TransformMatrix transf
...
+                    if (i != j && Math.Abs(transform.Matrix[i, j]) > tol)
+                        return false;
+
+            if (Math.Abs(transform.Matrix[0, 0] - transform.Matrix[1, 1]) <= tol &&
+                Math.Abs(transform.Matrix[1, 1] - transform.Matrix[2, 2]) <= tol &&
+                Math.Abs(transform.Matrix[3, 3] - 1) <= tol)
+                return true;
+
+            return false;
+        }
```
IsakNaslundBh: /***************************************************/

Geometry_Engine/Modify/Orient.cs
```diff
@@ -0,0 +1,50 @@
...
+        [Description("Orients geometry from one coordinate sytem to another")]
+        [Input("geometry", "Geometry to be transformed")]
+        [Input("csFrom", "Coordinate system in which geometry is now")]
+        [Input("csTo", "Coordinate system in which we want geometry to be")]
+        [Output("G", "Geometry in new coordinate system")]
+        public static T Orient<T>(this T geometry, Cartesian csFrom, Cartesian csTo) where T: IGeometry
+        {
+            TransformMatrix orientationMatrix = Create.OrientationMatrix(csFrom, csTo);
+            return (T)ITransform(geometry, orientationMatrix);
+        }
```
IsakNaslundBh: /***************************************************/

LMarkowski:
@al-fisher I am aware that this way of entering toleration is not correct but in this case it was neither distance nor angle and it seems we do not have a policy of dealing with computational tolerance. Maybe it's worth raising an issue?

al-fisher:
Perfect! Yes - no problem - thanks @LMarkowski This is indeed a distinct type of tolerance - so agree - should raise a new issue to capture how best to store/control going forward! 
Thanks 👍

Issue #259, opened 2018-04-04
Geometry_Engine: Ensure all section profile constructors are correct


IsakNaslundBh:
Check that all the input values for all of the section profiles in create correspond to the correct values.

For example, the webthickness of a boxprofile needs to be smaller or equal to the width.

Test file:

https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/01_Test%20Scripts/BHoM_Engine/Structure_Engine/Create.gh?csf=1&e=jaliHb

IsakNaslundBh:
closed by https://github.com/BHoM/BHoM_Engine/pull/1252

PR #1252, opened 2019-10-14
Geometry_Engine: Ensures all section profile constructors are correct

JonathanNillius:
Closes Issue #259

### Changes
Added code to not make profile components not output inconceivable profiles.
With inconceivable i mean for example where the profile is inside out or it self intersects. With the changes you are still able to make unrealistic profiles structure wise, but not geometry wise. 

 ### Changelog

 - Adding checks for all ratios used to generate section profiles, raising error messages and returning null if the provided values does not generate a functioning profile.

### Test files

https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?csf=1&e=CJWjNz&cid=0b727f4a%2Daf39%2D4d1d%2D8942%2D0c26c33d7c20&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4&viewid=a5a93cba%2Dfcca%2D46cc%2Da31a%2D80b8d5cbfd7b&id=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FTest%20Scripts%2FBHoM%5FEngine%2FGeometry%5FEngine%2FGeometry%5FEngine%2DPull%20Request%201252%20section%20profiles

### How to test
Slide number sliders to try and make a profile that is inconceivable. If the component stops working and returns null and an error message the changes works as intended.

### DIFF
--- a/Geometry_Engine/Create/ShapeProfile.cs
+++ b/Geometry_Engine/Create/ShapeProfile.cs
@@ -58,6 +58,18 @@ namespace BH.Engine.Geometry
 
         public static BoxProfile BoxProfile(double height, double width, double thickness, double outerRadius, double innerRadius)
         {
+            if (thickness > height / 2 || thickness > width / 2 || outerRadius > height / 2 || outerRadius > width / 2 || innerRadius * 2 > width - thickness * 2 || innerRadius * 2 > height - thickness * 2 ||
+                Math.Sqrt(2) * thickness <= Math.Sqrt(2) *outerRadius - outerRadius - Math.Sqrt(2) * innerRadius + innerRadius)
+            {
+                Engine.Reflection.Compute.RecordError("The ratio between inputs makes section inconceivable");
+                return null;
+            }
+            if (height <= 0 || width <= 0 || thickness <= 0 || outerRadius < 0 || innerRadius < 0)
+            {
+                Engine.Reflection.Compute.RecordError("Input length less or equal to 0");
+                return null;
+            }
+
             List<ICurve> curves = BoxProfileCurves(width, height, thickness, thickness, innerRadius, outerRadius);
             return new BoxProfile(height, width, thickness, outerRadius, innerRadius, curves);
         }
--- a/Geometry_Engine/Create/ShapeProfile.cs
+++ b/Geometry_Engine/Create/ShapeProfile.cs
@@ -171,6 +171,16 @@ namespace BH.Engine.Geometry
 
         public static FabricatedISectionProfile FabricatedISectionProfile(double height, double topFlangeWidth, double botFlangeWidth, double webThickness, double topFlangeThickness, double botFlangeThickness, double weldSize)
         {
+            if (height <= topFlangeThickness + botFlangeThickness + 2 * Math.Sqrt(2) * weldSize || botFlangeWidth <= webThickness + 2 * Math.Sqrt(2) * weldSize || topFlangeWidth <= webThickness + 2 * Math.Sqrt(2) * weldSize)
+            {
+                Engine.Reflection.Compute.RecordError("The ratio between inputs makes section inconceivable");
+                return null;
+            }
+            if (height <= 0 || topFlangeWidth <= 0 || botFlangeWidth <= 0 || webThickness <= 0 || topFlangeThickness <= 0 || botFlangeThickness <= 0 || weldSize < 0)
+            {
+                Engine.Reflection.Compute.RecordError("Input length less or equal to 0");
+                return null;
+            }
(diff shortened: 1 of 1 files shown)


Geometry_Engine/Create/ShapeProfile.cs
```diff
@@ -148,6 +256,16 @@ public static TSectionProfile TSectionProfile(double height, double width, doubl
 
         public static GeneralisedTSectionProfile GeneralisedTSectionProfile(double height, double webThickness, double leftOutstandWidth, double leftOutstandThickness, double rightOutstandWidth, double rightOutstandThickness, bool mirrorAboutLocalY = false)
         {
+            if (height <= leftOutstandThickness || height <= rightOutstandThickness)
+            {
+                Engine.Reflection.Compute.RecordError("The ratio between inputs makes section inconceivable");
+                return null;
+            }
+            if (height <= 0 || webThickness <= 0 || leftOutstandThickness <= 0 || leftOutstandThickness <= 0 || rightOutstandThickness < 0 || rightOutstandWidth <= 0)
```
IsakNaslundBh: Some inconsistencies here for righthand vs lefthand side. 

Think the check needs to be that outstands widths can be 0 and that if outstandwidths are non-0, then the thickness need to be above 0. Basically, need to avoid being able to make these kinds of sections:

image

but this should still be allowed:

image

Geometry_Engine/Create/ShapeProfile.cs
```diff
@@ -39,6 +40,16 @@ public static partial class Create
 
         public static ISectionProfile ISectionProfile(double height, double width, double webthickness, double flangeThickness, double rootRadius, double toeRadius)
         {
+            if (height < flangeThickness * 2 + rootRadius * 2 || width < webthickness + rootRadius * 2 +toeRadius*2 || toeRadius > flangeThickness)
+            {
+                Engine.Reflection.Compute.RecordError("The ratio between inputs makes section inconceivable");
```
IsakNaslundBh: If possible, would be good to specify in the error message which parameters that were involved in causing the fail. for example "The ratio between height, flangethickness and rootradius makes section inconceivable" if the first of the check here is the one causing problems.

Will make it easier for users to know why the section was failing.

Could also set up a small private helper method for returning the error for this, so you only have to specify which paramters are failing for the input. Let me know if you want to have a quick chat about this.

pawelbaran:
Happy to lean on @IsakNaslundBh in terms of the functionality, I can only recommend adding blank lines (e.g. between `if` statements) to improve readability.

Geometry_Engine/Create/ShapeProfile.cs
```diff
@@ -194,5 +481,19 @@ private static List<ICurve> MirrorAboutLocalZ(this List<ICurve> curves)
...
+        private static void InvalidRatioError(string inputs)
+        {
+            Engine.Reflection.Compute.RecordError("The ratio between " + inputs + " makes section inconceivable");
+        }
+
+        /***************************************************/
+
+        private static void InvalidRatioError2(string first, string second)
+        {
+            Engine.Reflection.Compute.RecordError("The ratio between " + first + " in relation to " + second + " makes section inconceivable");
```
IsakNaslundBh: Know I kind of wrote this, but would change this to

"The ratio of the " + first + " in relation to the " + second + " makes section inconceivable"

Geometry_Engine/Create/ShapeProfile.cs
```diff
@@ -194,5 +481,19 @@ private static List<ICurve> MirrorAboutLocalZ(this List<ICurve> curves)
         }
 
         /***************************************************/
+
+        private static void InvalidRatioError(string inputs)
```
IsakNaslundBh: Would consider just using the second method instead to keep things consistent, but not a big one.

Geometry_Engine/Create/ShapeProfile.cs
```diff
@@ -194,5 +481,19 @@ private static List<ICurve> MirrorAboutLocalZ(this List<ICurve> curves)
...
+        {
+            Engine.Reflection.Compute.RecordError("The ratio between " + inputs + " makes section inconceivable");
+        }
+
+        /***************************************************/
+
+        private static void InvalidRatioError2(string first, string second)
```
IsakNaslundBh: This is kind of on me as well, but the 2 is not needed here. Method overloading will take care of it with different number of inputs.

IsakNaslundBh:
Have tested through the actual logics of all now, and it all seem to make sense. Able to generate up til the limit. If we just can fix the small tweaks to the error messages, I will be happy to approve.

IsakNaslundBh:
Looks fine to me now! 👍

PR #1255, opened 2019-10-16
Geometry_Structure_Engine: Green's Therom Integration of Regions

kThorsager:
Everything is done and should mostly need testing now
I have not yet exchanged `Integrate` for `IntegrateSection` where it is needed
and it still returns the `polyline` versions in its dictionary, which will make some testing easier but is unnecessary looking forward

Will keep this set as WIP during the reviewing process as that might take a while

#### Review testing checklist
Feel free to add tests

- [ ] Compare results with old `Integrate` methods for all sections in BHoM
- [ ] Compare results with tabular values in the BHoM
- [ ] Compare results with old `Integrate` methods for a lot of arbitrary shapes
- [ ] Test for discontinuous sections (Think it works for most values, but there are no errors as of now)
- [ ] Test for deeply "nested sections" (a Russian doll of Tex CHSs) (shouldn't give any errors or problems, which might be a problem)
- [ ] Do more extensive testing for chosen tolerances

### Integration of Regions
Creates a Method which utilises [Green's Therom](https://mathinsight.org/greens_theorem_find_area) to integrate regions by evaluating the regions perimeter. This removes the need for integrations slices which should speed things up some.

For now I added some defined solutions for `x^0`, `x^1`, & `x^2` which is area, "finds the centre(?)", and second moment of inertia respectively. And I also have an general solution for any* `x^n`.
Now also solutions for `x^-2`

*(have not tested non integer numbers but any real number should work(?), but as of now the method only takes `int` since it mostly seems like a bother)

The formulas are defined for Lines, Arcs, and Circles, (and hence PolyLine, etc)

Can add formulas for other region formulas (`f(x,y)`) maybe

### Shear Area
Capture
I have added a method which calculates the denominator of this term for special regions

The region needs to be defined counter-clockwise simple region with its entire upper side against the y-axis without overhangs. The right- and leftmost line must be vertical (this is because diagonal lines calculations uses `ln(a.Y/b.Y)`).
Region

### Plastic Modulus
Needs a region in the WetBlanket Format
But works and calculates the plastic modulus of a Region

### Wet Blanket Interpretation
Will convert any multiple `Polyline` region into a region that the Shear Area method can handle.
T.ex.
Capture

### CollapseToScaledPolyline
Will convert a `PolyCurve` to a `PolyLine` while taking into account both curvature and size of the curves. Because small arc fillets in a I-section are not as important as the circle(s) in a circular sections. This method will cause them to be broken down into reasonable pieces for computation.

### Test files
[Testfile for Integration of regions](https://burohappold.sharepoint.com/:u:/s/BHoM/EZg1Q_2HFK5FtDxAvcrhwagBUrXDxObnIsrkVVlbHETP2A?e=WRTW4s)
[Testfile for Creating SectionProperties](https://burohappold.sharepoint.com/:u:/s/BHoM/EWnx9uHVds5ApVCBl85FokMBcv66tPYSoiKnsTt3TJOXJQ?e=hcX4az)

### Changelog
- `IntegrateRegion` added to `Geometry_Engine`, preforms a region integration of a region in the x-direction, can not handle nurbs or ellipses.
- `CollapseToScaledPolyline` added to `Geometry_Engine`, Will convert a `PolyCurve` to a `PolyLine` while taking into account both curvature and size of the curves.
- `WetBlanketInterpretation` added to `Geometry_Engine`, Takes a List of `Polylines` and creates a `squished` version with the same area at every X-location.
- `ShearArea` added to `Structure_Engine`, calculates the shear area for a region passed trough the `WetBlanketInterpretation`.
- `PlasticModulus` added to `Structure_Engine`, calculates the plastic modulus for a region passed trough the `WetBlanketInterpretation`.
- `IntegrateSection` added to `Structure_Engine`, calculates the same thing as the previous `Integrate` but less numerically and quicker
- Cleaned up `IsContained` for `PolyCurve` and added special case for single `Circle`
- Consistent tolerances for `LineIntersection`, specifically for `Arc` and `Circle`
- Constrained the results for `ParameterAtPoint` to be in-between 0 and 1

### Additional comments
How much documentation is needed on the principles that the calculation works from?

Should probably justify my changes to `IsContained` & `LineIntersection`. 
`IsContained` assumes that a sub segments end point is the start point of another segment, and that `LineIntersection` returns a point if the line is within the tolerance from the curve. 
A `PolyCurve` can consist of a single `Circle` which for special cases breaks the method as it attaches to itself.
`LineIntersection` for `Arc` did not add its tolerance for the intersection checks around its endpoints.

Also did some further changes to the tolerance comparisons as their consistency with inputted value seems integral for `IsContaining`. It has a lot of comparisons of square values, but Abs(a-b) < tol does not equal Abs(a^2-b^2) < tol^2. I think I have made it consistent now.

@pawelbaran could you check that I understood this correctly? And that my in code comments are correct.


Geometry_Engine/Compute/CollapseToPolylineEq.cs
```diff
@@ -79,7 +79,7 @@ private static Polyline Divide(this PolyCurve line, double shortestLength)
         private static Polyline Divide(this Arc arc, double shortestLength)
         {
             Polyline result = new Polyline();
-            double res = Math.Floor(arc.Length() / shortestLength);
+            double res = Math.Floor((shortestLength * 10 + 0.15 * arc.Length()) / shortestLength);
```
FraserGreenroyd: Out of curiosity, where did these numbers come from? Are they part of the underlying theory?

kThorsager: They are part of the underlying WIP and may not be questioned

FraserGreenroyd: As long as they get documented properly by the time this is no longer WIP... :wink:

kThorsager:
Thanks, @IsakNaslundBh 
The plastic modulus error has been fixed and is now 0.07% from hand calculations for CHS with default tolerance.
And happy to see this back on the agenda again

al-fisher:
:rocket:

kThorsager:
Added a fix for `LineIntersections` as the tolerance was not carried through. Might want to put that in a different issue, but the fix is needed for this current configuration as some small sections fail with higher tolerances.
Can however also see that this used to be the case but was removed? Should be investigated further.

peterjamesnugent:
@peterjamesnugent @JosefTaylor to test various sections before end of the week as per Structure Sprint Kickoff call.

IsakNaslundBh:
Have checked through the full UK ans US steel libraries, comparing the old and new integration methods. Worst difference is for ShearAreas (still very low difference of worst case 0.1%) and almost tempted to think that the new might be slightly sharper than the old. Will try to check against some table values as well, even though they can be hard to find for shear areas.

Worst case difference for other properties seem to be more or less within 0.01% and for the spot checks I have made, the new integration beats the old in accuracy.

Will try to see if I can find any old bugs left (from geometry engine or elsewhere) for edge cases, if not I think this starts to look ready to merge. But good for @peterjamesnugent and/or @JosefTaylor to have a look as well.

Structure_Engine/Compute/IntegrateSection.cs
```diff
@@ -0,0 +1,186 @@
...
+
+namespace BH.Engine.Structure
+{
+    public static partial class Compute
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Calculates Section properties for a region")]
```
IsakNaslundBh: Maybe add some more comments about what this method requires. That it assumes the curves to be in the XY-plane for example, and how the constants outputted relates to this.

Geometry_Engine/Compute/IntegrateRegion.cs
```diff
@@ -0,0 +1,236 @@
...
+namespace BH.Engine.Geometry
+{
+    public static partial class Compute
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Integrates a closed region with x to the specified power on the XY-Plane")]
+        [Input("curve", "defined counter clockwise on the XY-Plane")]
```
IsakNaslundBh: good! (small tweak to start the sentences in the description with capital letter).

IsakNaslundBh:
Also found one more edge case bug for "InContaining", seem to have to do with some tolerance issues Test file here:

https://burohappold.sharepoint.com/:f:/s/BHoM/Egm1Ia-4R_NBifddFZW6WvwBb9OAn8_2Zm0ASZidzxCPKA?e=qLnrJu

`TangentAtPoint` returns null (called on line 335 in IsContaining). as the ParameterAtPoint > 1+ paramTol.

kThorsager:
Ready for review again,
And it ran fine on the master as the line-intersection for arcs does not incorporate tolerances there (but those tolerances are needed for `IsContaining` to work on other edge cases)

Structure_Engine/Compute/ShearArea.cs
```diff
@@ -0,0 +1,191 @@
...
+namespace BH.Engine.Structure
+{
+    public static partial class Compute
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Calculates the the exact shear area of an simplified boundery of an section.")]
+        [Input("pLine", "The Polyline should have the upper side along the x-axis and the rest of the lines should be defineble as /n" +
```
IsakNaslundBh: would get rid of the new line signs (\n) here. If they are to be used, I would but them after full sentences.

Geometry_Engine/Compute/IntegrateRegion.cs
```diff
@@ -0,0 +1,236 @@
...
+
+            for (int i = 0; i < pts.Count - 1; i++)
+                result += IntSurfLine(pts[i], pts[i + 1], powX, tol);
+
+            return result;
+        }
+
+        /***************************************************/
+
+        private static double IntegrateRegion(NurbsCurve pLine, int powX, double tol = Tolerance.Distance)
```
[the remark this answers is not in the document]
kThorsager: That would create a loop no? 

Ellipse enters `IntegrateRegion`
get sent back to `IntegrateRegion`

Will add another fallback method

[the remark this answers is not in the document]
kThorsager: Oh missread, sorry

Geometry_Engine/Compute/IntegrateRegion.cs
```diff
@@ -0,0 +1,236 @@
...
+        {
+            throw new NotImplementedException("NurbsCurve is not imlemented yet so this cannot be calculated");
```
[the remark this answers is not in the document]
kThorsager: if it's just a error then we would need to provide a result value, and if it's a PolyCurve with a nurbs part the value might seem right if one does not see the error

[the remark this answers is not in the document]
IsakNaslundBh: agreed, so lets keep the exception!

Geometry_Engine/Compute/WetBlanketInterpetation.cs
```diff
@@ -0,0 +1,287 @@
...
+            if (center.Y + tol < a.Y)
+                return 1;   // over
+            if (center.Y - tol > a.Y)
+                return 3;   // under
+            return -100;  // the same
+        }
+
+        /***************************************************/
+
+        public static Point PointAtX(Point a, Point b, double x)    //Public??
```
IsakNaslundBh: no, make it private

JosefTaylor:
Sorry for the merge commit; I told Isak I would do a spellcheck but I didn't tell Kalle. Looks like there are no conflicts. (and no more spelling errors in Kalle's work). Also fixed WetBlanketInterpretation.cs

PR #1259, opened 2019-10-17
Update assembly version

### LINKED ISSUE #1258: Update assembly version
`AssemblyFileVersion` is at `2.4.0.0` and should be `3.0.0.0` for the beta development cycle.
We should also check the `AssemblyVersion` value, which should be `3.0.0.0`

FraserGreenroyd:
Fixes #1258


FraserGreenroyd:
> PS: Interesting how we have 5 more projects in the Engine compared to the oM. Some make sense but others are a bit more surprising.

I agree @adecler - something to perhaps review this quarter/next to see what alignment might be needed? :smile:

PR #1270, opened 2019-10-22
Leverage the versioning engine for deserialisation of objects

### LINKED ISSUE #1267: Serialiser_Engine: Leverage the versioning engine for deserialisation of objects
#### Description:
Make sure deprecated object are automatically upgraded when deserialised by using the versioning engine

adecler:
### NOTE: Depends on 
https://github.com/BHoM/Versioning_Toolkit/pull/2

   
Closes #1267 

See the PR in the Versioning_Toolkit for more details

### Additional comments
Recreating with proper branch name

Edit: This doesn't really depends on the Version toolkit per se anymore (no direct reference) but you will need it for testing.

### DIFF
--- a/BHoM_Engine.sln
+++ b/BHoM_Engine.sln
@@ -57,6 +57,8 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Physical_Engine", "Physical
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Diffing_Engine", "Diffing_Engine\Diffing_Engine.csproj", "{073DFD36-0829-4792-8C32-67BF692A9413}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Versioning_Engine", "Versioning_Engine\Versioning_Engine.csproj", "{3A1179AA-695B-40DF-8055-ABA7ACB51D4C}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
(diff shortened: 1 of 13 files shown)


adecler:
Although we don't have circular dependencies between projects, we now have one between repos because
- The Versioning Engine depends on the Reflection-Engine for logging messages
- The Serialisation_Engine depends on the Versioning_Engine
This means that AppVeyor and Code_Assistant will stop working properly

The only thing I can think of to avoid this is to move the Versioning_Engine project into the BHoM_Engine (contains a single file). The Versioning_Toolkit would then contain only the upgrader dlls. Maybe we shouldn't call it a toolkit then as it doesn't cover any part of the diamond in that case.

Any thoughts?

IsakNaslundBh:
Was going to ask you about the dependencies when trying to compile this yesterday, and agree with the solution. To me it does not make sense for the core engine to depend on any toolkit at all, so agree with moving the Versioning_Engine in here.

When it comes to the Versioning_Toolkit only containing the upgrader dlls, I am personally fine with that solution, can not think of anything better of the top of my head.

al-fisher:
> The only thing I can think of to avoid this is to move the Versioning_Engine project into the BHoM_Engine

Good idea - agree best to avoid circular dependencies across repos also 👍

adecler:
See [the comment on the Versioning_Toolkit PR](https://github.com/BHoM/Versioning_Toolkit/pull/2#issuecomment-544871222) for moredetails about the exact scope covered here.

al-fisher:
With introduction of new `Compute partial class` in `BH.Engine.Serialiser` creates an error in Excel_Toolkit:

> `Compute` is an ambiguous reference between `BH.Engine.Reflection.Compute` and `BH.Engine.Serialiser.Compute` in Excel_UI	

Have raised an issue here: https://github.com/BHoM/Excel_Toolkit/issues/159

al-fisher:
Also - on building and running the Engine PR _with_ Versioning_Toolkit PR built etc. https://github.com/BHoM/Versioning_Toolkit/pull/2

Still get warnings regarding "Cannot find a definition" and a `CustomObject` is returned
image

al-fisher:
Although the simple "primitive" `Level` class works - consisting of only simple type Properties?
image

adecler:
oh boy, the one time I don't test with Bars and Nodes. Sorry about that. You're right, I had tested on Level but it doesn't contain other BHoM objects. I am still a bit confused why it was complaining for the bar type itself though and not just its children properties. 
Anyway, it should be fixed now.

adecler:
@al-fisher , the conversion now returns the old version when the upgraders are missing. I have also turned the messages into warnings instead of errors for that situation. Let me know if that works for you.

al-fisher:
Tested this again _without_ Versioning_Toolkit.  The above three scenarios:

> 1. non-deprecated objects will simply work
> 2. deprecated will return as "old" (as converts will naturally not be found)
> 3. deleted will be returned as CustomObjects (as also converts will not be found)

 are all behaving as expected. Thanks @adecler 

For case 2. Noting deprecated objects are deserialsed without warning they are Deprecated. That could be a nice to have? - but not critical for this PR perhaps.

al-fisher:
Tested _with_ Versioning_Toolkit and case 2. upgrades:

image

al-fisher:
Also important we address https://github.com/BHoM/Excel_Toolkit/issues/159 as noted above before merging.

Leaving **`status:do-not-merge`** label on this PR

al-fisher:
Thanks @alelom

PR #1276, opened 2019-10-23
Reflection_Engine: Add a ToText() method for MemberInfo

adecler:
This assists solving this issue: https://github.com/BHoM/BHoM_UI/issues/124

### Test files
I honestly don't think this needs testing. Inspection of the code and making it compiles should be enough here. 

Unless you see a reason for the fallback case to not be `member.ToString()` ?

### DIFF
--- a/Reflection_Engine/Convert/ToText.cs
+++ b/Reflection_Engine/Convert/ToText.cs
@@ -35,6 +35,19 @@ namespace BH.Engine.Reflection
         /**** Public Methods                            ****/
         /***************************************************/
 
+        public static string ToText(this MemberInfo member, bool includePath = false)
+        {
+            if (member is MethodBase)
+                return ToText(member as MethodBase, includePath);
+            else if (member is Type)
+                return ToText(member as Type, includePath);
+            else
+                return member.ToString();
+        }
+
+
+        /***************************************************/
+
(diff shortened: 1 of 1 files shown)


Reflection_Engine/Convert/ToText.cs
```diff
@@ -35,6 +35,19 @@ public static partial class Convert
         /**** Public Methods                            ****/
         /***************************************************/
 
+        public static string ToText(this MemberInfo member, bool includePath = false)
```
epignatelli: I get that `MemberInfo` is neither acting on an `interface`nor is receiving a `BHoMObject`, but I think the pattern is still the one of the interface method. Is there any reason why we are not using that pattern?

adecler: I considered it but 
- MemberInfo is not an interface. 
- We are not dealing with BHoMObjects as you said so I didn't feel like the pattern was relevant here as you would have to know the details of the MemberInfo class to even consider using the IToText method instead. So I felt more intuitive to no ask the user to choose between IToText and ToText 
- It is not actually a dynamic call to another method as it can also handle a default case on its own. Admittedly, we could reformat the code to change that but this keeps things nice and contained.

PR #1295, opened 2019-11-04
Structure_Engine: Description query

### LINKED ISSUE #820: Structure_Engine: Standard name for properties

JonathanNillius:
A general description to profiles, sections and surface properties.
WIP adding a NameOrDescription method that outputs the description name if there is no name.

Closes #820 

### Test files
https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?csf=1&e=CJWjNz&cid=0b727f4a%2Daf39%2D4d1d%2D8942%2D0c26c33d7c20&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4&viewid=a5a93cba%2Dfcca%2D46cc%2Da31a%2D80b8d5cbfd7b&id=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FTest%20Scripts%2FBHoM%5FEngine%2FStructure%5FEngine%2FStructure%5FEnigne%2DPR1295%2D%20DescriptionQuery

How to test:
Description: Look at the pannels are see if the names are appropriate.
DescriptionOrName: Look at the pannels and see if the component only renames into description if there isn't already a name.


IsakNaslundBh:
@JonathanNillius Nice! 👍 Generally looking good!

Some initial comments (know still WIP).

- Need some comment header and method separator lines, i.e.:
```
```
and

`        /***************************************************/`

Have a look here for example https://github.com/BHoM/BHoM_Engine/blob/master/Geometry_Engine/Modify/Translate.cs

- Need private fallback method for when this is being called for something not implemented yet (might as well add one accepting `IObject`)

- Think some Profiles might be missing (Could not find `ISectionProfile` for example)

- Would skip all corner/root/toe radii for the description strings in general and stick with the _height x width x web thickness x flange thickness_ approach in general (ofc to be modified when not applicable)

Hope this makes sense!

Structure_Engine/Query/Description.cs
```diff
@@ -0,0 +1,232 @@
...
+        public static string Description(this KiteProfile profile)
+        {
+            return "Kite " + Math.Round(profile.Angle1,2) + "x" + profile.Width1 + "x" + profile.Thickness;
+        }
+
+        /***************************************************/
+
+        public static string Description(this FabricatedISectionProfile profile)
+        {
+            return "FabI " + profile.Height + "x" + profile.WebThickness + "x" + profile.TopFlangeWidth + "x" + profile.BotFlangeWidth + "x" + profile.TopFlangeThickness + "x" + profile.BotFlangeThickness;
```
IsakNaslundBh: Would change this to:

`FabI " + profile.Height + "x" + profile.TopFlangeWidth + "x" + profile.BotFlangeWidth + "x" + profile.WebThickness + "x" + profile.TopFlangeThickness + "x" + profile.BotFlangeThickness;`

I.e. main parameters first (heights, and widths etc) then corresponding thicknesses.

Structure_Engine/Query/Description.cs
```diff
@@ -0,0 +1,246 @@
...
+            return "Tube " + profile.Diameter + "x" + profile.Thickness;
+        }        
+
+        /***************************************************/
+        /**** Public Methods - Sections                 ****/
+        /***************************************************/
+
+        public static string Description(this SteelSection section)
+        {
+            return "Steel " + section.SectionProfile.IDescription() + " - " + CheckGetMaterialName(section.Material);
```
IsakNaslundBh:
```suggestion
            return "Steel " + section.SectionProfile.IDescription() + CheckGetMaterialName(section.Material);
```
One final change, then good to go!

with the " - " included in the method, we now end up with two "-". So need to either remove for the method, or change to suggestion above everywhere

PR #1301, opened 2019-11-06
Serialiser_Engine: Handle the case of a class containing another class that was only changed internally

adecler:
### NOTE: Depends on 
https://github.com/BHoM/Versioning_Toolkit/pull/6

See https://github.com/BHoM/Versioning_Toolkit/pull/6 for more details

### Additional comments
When a class is modified only internally, there is no `Deprecated` tag used anywhere (since the class is still active, just modified). This makes it very difficult to detect when deserialising another class that has this modified class as one of the inner properties. The only solution I could find here is to deserialise as custom object so the text now only contains upgraded json/bson and then deserialise again. This works fine and doesn't break the case of deleted classes. Make sure you test any case you can think of though as this is probably the part that is most likely to fail in some weird cases.


FraserGreenroyd:
Just added which engine is affected to PR title for change log :smile: will try and review today unless someone else jumps in first :smile:

adecler:
> Just added which engine is affected to PR title for change log 😄 will try and review today unless someone else jumps in first 😄

I will never learn 😿

PR #1289, opened 2019-10-29
Data_Engine: Adding create methods and first query methods for Table class

IsakNaslundBh:
### NOTE: Depends on 

https://github.com/BHoM/BHoM/pull/595
   
### Issues addressed by this PR

Closes #1285 

Adding create methods and first Query methods for Tables.

Please note that the DataTable can not be serialised to Bson by default. I have code ready for this, but will need to be aligned with the work done by @adecler in https://github.com/BHoM/BHoM_Engine/pull/1270

### Test files

https://burohappold.sharepoint.com/:f:/r/sites/BHoM/02_Current/12_Scripts/01_Test%20Scripts/BHoM_Engine/Data_Engine/Issue_1289_CreateTables?csf=1&e=KcLoUo

### Changelog

- Adding Create method for table class. One from `List<IBHoMObject>` and one from 2D table with header values.
- Query method for getting table rows matching exact values in table
- Query method for getting table rows matching values above set axis-values

### Additional comments

@Martian42

PR #1312, opened 2019-11-07
Data_Engine: Enable advanced sorting and filtering BHoM object lists, by wrapping 'Linq' methods

### LINKED ISSUE #1305: Data_Engine: Expose Linq methods
#### Description:

Add methods to expose Linq into the UI's by wrapping Linq. 

Capturing issues raised elsewhere: 
 - Issue #223 
 - Issue #1211 
 - Issue #772

rwemay:

Closes #1305 

### Test files

https://burohappold.sharepoint.com/:f:/r/sites/BHoM/02_Current/12_Scripts/01_Test%20Scripts/BHoM_Engine/Data_Engine/Issues?csf=1&e=fk4xir

**Additional Test script by @FraserGreenroyd **
[Here is a GH script and a JSON file of test data to pull from for the script to work with](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?csf=1&e=fk4xir&cid=4eedc797%2D06e0%2D4cc5%2Daec9%2D39a8d560956c&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4&viewid=a5a93cba%2Dfcca%2D46cc%2Da31a%2D80b8d5cbfd7b&id=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FTest%20Scripts%2FBHoM%5FEngine%2FData%5FEngine%2FIssues%2F1305%20%2D%20Linq)

### Changelog

 - Add OrderBy and FilterBy methods to wrap Linq OrderBy and Where respectively. 

### Additional comments

Linq

### DIFF
--- a/Data_Engine/Query/Linq/FilterBy.cs
+++ b/Data_Engine/Query/Linq/FilterBy.cs
@@ -21,7 +21,9 @@
  */
 
 
+using BH.oM.Base;
 using BH.oM.Reflection.Attributes;
+using System.Collections;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Linq;
@@ -39,37 +41,131 @@ namespace BH.Engine.Data
         [Input("objects", "List of objects to be filtered. All objects in the list should be of similar type")]
         [Input("propertyName", "The name of the property to filter the objects by")]
         [Input("value", "The value of the selected property to filter the objects by")]
-        public static List<T> FilterBy<T>(this List<T> objects, string propertyName = null, object value = null, bool ignoreStringCase = false, bool exactStringMatch = false, double rounding = 0.0)
+        [Input("ignoreStringCase", "Ignorse upper/lower case letters if the property filtered by is a text/string.")]
+        [Input("exactStringMatch", "If true checks if the property filtered by contains the text value (filtering must be a string/text)")]
+        public static List<T> FilterBy<T>(this List<T> objects, string propertyName = null, object value = null, bool ignoreStringCase = false, bool exactStringMatch = true)
         {
             if (objects == null || objects.Count == 0) return new List<T>();
             if (propertyName == null) return new List<T>();
             if (value == null) return new List<T>();
-            System.Reflection.PropertyInfo prop = objects.First(x => x != null).GetType().GetProperty(propertyName);
+            object firstObject = objects.First(x => x != null);
+            bool isNested = propertyName.Contains(".");
+            bool isCustomData = false;
+            System.Reflection.PropertyInfo prop = firstObject.GetType().GetProperty(propertyName);
+            if (prop == null && firstObject is IBHoMObject)
+                if ((firstObject as IBHoMObject).CustomData.ContainsKey(propertyName)) isCustomData = true;
+            System.Type type = PropertyType(firstObject, propertyName);
             int groupedObjects = objects.GroupBy(x => x.GetType()).Count();
             if (groupedObjects != 1)
             {
                 BHRE.Compute.RecordError("All objects in the list to be sorted should be of similiar type.");
                 return null;
             }
-            if (!propertyName.Contains("."))
+            if (type == typeof(System.String))
             {
-                if (prop != null)
-                {
-                    if (value.GetType() == typeof(System.String))
-                    {
+                string stringValue = value.ToString();
+                if (ignoreStringCase && exactStringMatch)
+                    if (!isNested)
+                        return objects.Where(x => string.Equals((string)prop.GetValue(x), stringValue, System.StringComparison.CurrentCultureIgnoreCase)).ToList();
+                    else
+                        return objects.Where(x => string.Equals((string)BHRE.Query.PropertyValue(x, propertyName), stringValue, System.StringComparison.CurrentCultureIgnoreCase)).ToList();
+                if (!exactStringMatch)
+                    if (!isNested)
                         if (ignoreStringCase)
-                            return objects.Where(x => string.Equals((string)prop.GetValue(x), (string)value, System.StringComparison.CurrentCultureIgnoreCase)).ToList();
-                        if (!exactStringMatch)
-                            return objects.Where(x => (prop.GetValue(x) as string).ToUpper().Contains((value as string).ToUpper())).ToList();
-                        return objects.Where(x => prop.GetValue(x).Equals(value)).ToList();
-                    }
-                return objects.Where(x => prop.GetValue(x) == value).ToList();
-                }
+                            return objects.Where(x => (prop.GetValue(x) as string).ToUpper().Contains(stringValue.ToUpper())).ToList();
+                        else
+                            return objects.Where(x => (prop.GetValue(x) as string).Contains(stringValue)).ToList();
+                    else
+                        if (ignoreStringCase)
+                        return objects.Where(x => (BHRE.Query.PropertyValue(x, propertyName) as string).ToUpper().Contains(stringValue.ToUpper())).ToList();
+                    else
+                        return objects.Where(x => (BHRE.Query.PropertyValue(x, propertyName) as string).Contains(stringValue)).ToList();
+                if (!isNested)
+                    return objects.Where(x => prop.GetValue(x).Equals(stringValue)).ToList();
+                else
+                    return objects.Where(x => BHRE.Query.PropertyValue(x, propertyName).Equals(stringValue)).ToList();
             }
-            BHRE.Compute.RecordNote("CustomData or nested property is used as the sorting property (using 'Object.Property.Property...') which is slower than a base property.");
-            return objects.Where(x => BHRE.Query.PropertyValue(x, propertyName).Equals(value)).ToList();
+            if (type == typeof(System.Double))
+            {
+                double doubleValue = 0.0;
+                string searchValue = value.ToString();
+                int rounding = searchValue.Contains('.')? searchValue.Split('.').Last().Length : 0;
+                try
+                { double.TryParse(searchValue, out doubleValue); }
+                catch
+                { BHRE.Compute.RecordError("Property value to search for cannot be converted to a number"); }
+                if (!isNested && !isCustomData)
+                    return objects.Where(x => System.Math.Round((double)prop.GetValue(x), rounding) == doubleValue).ToList();
+                else
+                    return objects.Where(x => System.Math.Round((double)BHRE.Query.PropertyValue(x, propertyName), rounding) == doubleValue).ToList();
+            }
+            if (prop != null && !isNested)
+                return objects.Where(x => prop.GetValue(x) == value).ToList();
+            return objects.Where(x => BHRE.Query.PropertyValue(x, propertyName) == value).ToList();
         }
         /***************************************************/
 
+        /***************************************************/
+        /**** Private Methods                           ****/
+        /***************************************************/
+
+        private static System.Type PropertyType(this object obj, string propName)
+        {
+            if (obj == null || propName == null)
+                return null;
+
+            if (propName.Contains("."))
+            {
+                string[] props = propName.Split('.');
+                for (int i = 0; i < props.Count() - 1; i++)
+                {
+                    obj = BHRE.Query.PropertyValue(obj, props[i]);
+                    if (obj == null)
+                        break;
+                }
+                return PropertyType(obj, props.Last());
+            }
+            
+            System.Reflection.PropertyInfo prop = obj.GetType().GetProperty(propName);
+
+            if (prop != null) return prop.PropertyType;
+
+            else if (obj is IBHoMObject)
+            {
+                IBHoMObject bhom = obj as IBHoMObject;
+                if (bhom.CustomData.ContainsKey(propName))
+                {
+                    if (!(bhom is CustomObject))
+                        BHRE.Compute.RecordNote($"{propName} is stored in CustomData");
+                    return bhom.CustomData[propName].GetType();
+                }
+                else
+                {
+                    BHRE.Compute.RecordWarning($"{bhom} does not contain a property: {propName}, or: CustomData[{propName}]");
+                    return null;
+                }
+
+            }
+            else if (obj is IDictionary)
+            {
+                IDictionary dic = obj as IDictionary;
+                if (dic.Contains(propName))
+                {
+                    return dic[propName].GetType();
+                }
+                else
+                {
+                    BHRE.Compute.RecordWarning($"{dic} does not contain the key: {propName}");
+                    return null;
+                }
+            }
+            else
+            {
+                BHRE.Compute.RecordWarning($"This instance of {obj.GetType()} does not contain the property: {propName}");
+                return null;
+            }
+        }
+
+        /***************************************************/
     }
 }
--- a/Data_Engine/Query/Linq/FilterBy.cs
+++ b/Data_Engine/Query/Linq/FilterBy.cs
@@ -27,7 +27,7 @@ using System.Collections;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Linq;
-using BHRE = BH.Engine.Reflection;
+using BH.Engine.Reflection;
 
 namespace BH.Engine.Data
 {
@@ -43,6 +43,7 @@ namespace BH.Engine.Data
         [Input("value", "The value of the selected property to filter the objects by")]
         [Input("ignoreStringCase", "Ignore upper/lower case letters if the property filtered by is a text/string.")]
         [Input("exactStringMatch", "If true checks if the property filtered by contains the text value (filtering must be a string/text)")]
+        [Output("filteredObjects", "The collection of objects that match the given filter conditions")]
         public static List<T> FilterBy<T>(this List<T> objects, string propertyName = null, object value = null, bool ignoreStringCase = false, bool exactStringMatch = true)
         {
             if (objects == null || objects.Count == 0)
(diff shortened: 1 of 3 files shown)


adecler:
I see this is marked as `do-not-merge` already but putting it here as a reminder:
This is clashing with a pre-existing PR:  https://github.com/BHoM/BHoM_Engine/pull/1289
So, please, do not merge.

adecler:
@FraserGreenroyd , you might want to look at the title of this PR

Data_Engine/Query/Linq/OrderBy.cs
```diff
@@ -0,0 +1,71 @@
...
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Orders a list of objects by a property of the object - e.g. input 'Name' to sort objects by their names. Note that not all object properties have sortable values")]
+        [Input("objects", "List of objects to be sorted. All objects in the list should be of similar type")]
+        [Input("propertyName", "The name of the property to sort the list of objects by. Note that the property must be able to be sorted - e.g. numbers/text")]
+        public static List<T> OrderBy<T>(this List<T> objects, string propertyName)
+        {
+            if (objects == null || objects.Count == 0) return null;
```
adecler: it is probably better to return an empty list instead of null

Data_Engine/Query/Linq/OrderBy.cs
```diff
@@ -0,0 +1,71 @@
...
+            int groupedObjects = objects.GroupBy(x => x.GetType()).Count();
+            System.Type objectType = objects[0].GetType();
```
adecler: Two things here:
- `objects[0]` can potentially be null. So this would crash
- Am I blind or is it not used anywhere ?

Data_Engine/Query/Linq/OrderBy.cs
```diff
@@ -0,0 +1,71 @@
...
+            if (groupedObjects != 1)
+            {
+                BHRE.Compute.RecordError("All objects in the list to be sorted should be of similiar type.");
+                return null;
+            }
+            try
+            {
+                if (!propertyName.Contains("."))
+                {
+                    System.Reflection.PropertyInfo prop = objects[0].GetType().GetProperty(propertyName);
```
adecler: Same thing here: make sure to pick the first non-null object to get the type from.

Data_Engine/Query/Linq/FilterBy.cs
```diff
@@ -0,0 +1,63 @@
...
+
+        [Description("Filters a list of objects to output only those with a property value matching that input")]
+        [Input("objects", "List of objects to be filtered. All objects in the list should be of similar type")]
+        [Input("propertyName", "The name of the property to filter the objects by")]
+        [Input("value", "The value of the selected property to filter the objects by")]
+        public static List<T> FilterBy<T>(this List<T> objects, string propertyName, object value)
+        {
+            if (objects == null || objects.Count == 0) return null;
+            int groupedObjects = objects.GroupBy(x => x.GetType()).Count();
+            System.Type objectType = objects[0].GetType();
```
adecler: objectType variable is not used

Data_Engine/Query/Linq/FilterBy.cs
```diff
@@ -0,0 +1,63 @@
...
+            if (groupedObjects != 1)
+            {
+                BHRE.Compute.RecordError("All objects in the list to be sorted should be of similiar type.");
+                return null;
+            }
+            if (!propertyName.Contains("."))
+            {
+                System.Reflection.PropertyInfo prop = objects[0].GetType().GetProperty(propertyName);
```
adecler: Make sure to pick the first non-null object to get the type from

Data_Engine/Query/Linq/FilterBy.cs
```diff
@@ -0,0 +1,63 @@
...
+                if (prop != null)
+                    return objects.Where(x => prop.GetValue(x).Equals(value)).ToList();
```
adecler: `Equal` will not work if `prop.GetValue(x)` is null. Better to use `==`

Data_Engine/Query/Linq/OrderBy.cs
rwemay: What are the consequences of returning null? One issue with empty lists is, it's sometimes difficult to know something went 'wrong' even if that's the user inputting something that they know the list doesn't contain. Have found myself having to do ICount on lists in Excel quite a bit, to check that an error isn't because a list is empty. Maybe that's a different problem - need an 'empty list' warning in Excel.

Data_Engine/Query/Linq/OrderBy.cs
rwemay: No not blind, is a WIP hangover. The problem I was trying to solve was to try and cast/be explicit in object types as at the moment, the specific IComparable methods on objects are not found. The example is a BHoM Point - it has an IComparer<Point>, but as we're using List<T> it doesn't work. Not sure yet how to get around this - @awakeman  and I did a quick hack to explicitly cast the OrderBy Linq method to (Point) and the objects inside, and the IComparer worked, but don't know how to cast to a variable... I'm sure this doesn't make sense, sorry! 

image

Data_Engine/Query/Linq/OrderBy.cs
rwemay: Makes sense. Wasn't sure if this mattered as won't the method fail if any objects are null - through any of the Linq methods? In any case does the below look good? ` System.Reflection.PropertyInfo prop = objects.First(x => x != null).GetType().GetProperty(propertyName);`

Data_Engine/Query/Linq/FilterBy.cs
rwemay: Had a problem with `==` is why I changed this to `.Equals.` I tested on a simple string and would not say the value (to search for) and the propertyValue were the same unless I used `.Equals`. I think related to the List<T> issue above? Not sure how to get around it...

Data_Engine/Query/Linq/OrderBy.cs
adecler: I has more chances to cause problems upstream as it is not always covered.
If anything though, an empty list as input should return an empty list, not null.

Data_Engine/Query/Linq/OrderBy.cs
adecler: nearly, you might a case where all the objects inside objects are null. your new code would still crash for that. One thing to consider is to first filter the objects to remove all the nulls and work only on that. The question of whether to append the null objects at the end of the sorted list is up to you.

Data_Engine/Query/Linq/FilterBy.cs
adecler: You might want to do your own comparison method that you call within the Where then. That way, you can return false, when null, use Contains for string, and == for the rest.

al-fisher:
> @FraserGreenroyd , you might want to look at the title of this PR

Thanks @adecler. 
@rwemay don’t forget PR titles are reflected automatically as change log items - so care must be taken to make human readable - serving different purpose from the branch naming convention.
I have just tweaked PR title to be more suitable. ^^
The change log list giving more specifics in the PR body looks good though FYI!
Cheers

rwemay:
Sorry guys - was tired. I might go a bit further with title to make it easier for non-coder. Saying all that was also conscious that this probably wont get merged - will have to re PR.

rwemay:
Prototype of the FilterBy wrapper below. Some string/double parsing to do etc, but looking promising. 

Linq2

rwemay:
Just going to post this here quietly. 

asdf

rwemay:
Pushed latest FilterBy method just now - see below example of how it's working. One thing I can't get working is to be able to filter by an actual BHoM property - e.g. Level object. It works if I select a level directly from one of the exploded object - but only returns that object. There are only 6 levels attached to 999 bars, so plugging in one level should return multiple filtered results. Something to do with IEquatable interface, I think... @awakeman  are you able to help on that? Is line 102 in the FilterBy.cs. @adecler good to get your view on the current hack(!!) I think could do with being a bit more elegant/refined. Note that I put in a method for PropertyType(object, propertyName) as I couldn't find a better way/method - thoughts much appreciated? One thing @jdarcyBHE mentioned is it would be great to a) return both the included and excluded filter items @adecler , can we do multiple outputs? b) allow lists with different objects in to be filtered - bit of a can of worms, but have some ideas so will raise a separate issue for this. 

Linq3

adecler:
@rwemay , yes, you can do multiple outputs. You can find an example of how it is done here: https://github.com/BHoM/BHoM_Engine/blob/977443201f405a70635299ab61c40f8a51827ebe/Diffing_Engine/Compute/Diffing.cs#L129

rwemay:
Thanks @adecler . Thinking about it, it might be tricky for some UI's like Excel.

FraserGreenroyd:
@adecler @rwemay is there any movement on this PR from November? Just doing some workshops with @enarhi and @michaelhoehn and this Linq exposed stuff would really help with some of the object querying that is being attempted without trying to cast it to Data Tables.

So long story short, what's the current status on this PR and what resources are needed to try and maybe look to include this in 3.1? :smile:

rwemay:
No progress since last comment, sorry! Could do with a bit of help on the BHoM object comparer question. From memory I felt like it was a cast I was missing. Don't supposed you'd have a few minutes to look? Otherwise I'll try and get back around to it.

FraserGreenroyd:
I can try and take a look this morning and see what I can find out :smile:

FraserGreenroyd:
@rwemay @adecler I have taken a look at the OrderBy method and fixed the casting issue @rwemay was referring to.

I've added a test script to the tests folder and linked in the original comment up top ([but also here just for you](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?csf=1&e=fk4xir&cid=4eedc797%2D06e0%2D4cc5%2Daec9%2D39a8d560956c&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4&viewid=a5a93cba%2Dfcca%2D46cc%2Da31a%2D80b8d5cbfd7b&id=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FTest%20Scripts%2FBHoM%5FEngine%2FData%5FEngine%2FIssues%2F1305%20%2D%20Linq)) to test the change I've made.

@rwemay I've tidied the file up a little bit to help me read it and added some user errors/warnings to the initial checks to help the user understand why they might get an empty list back initially, hope you don't mind?

@adecler I've done what made it work on my test set and what I think will work for @michaelhoehn and @enarhi so your critical eye is definitely needed please :smile:

I've added @enarhi and @michaelhoehn as reviewers as well to get their thoughts/input.

I have removed the 'do not merge' label because the PR which was blocking this one (highlighted by @adecler ) has since been merged so I do not believe there is anything that would prevent this being merged. I've retained the 'WIP' label for the time being, but happy for that being removed if we're happy with the functionality being added here to get the ball rolling on this.

Hope that makes sense! :smile:

Data_Engine/Query/Linq/FilterBy.cs
```diff
@@ -0,0 +1,214 @@
...
+                            return objects.Where(x => (BHRE.Query.PropertyValue(x, propertyName) as string).Contains(stringValue)).ToList();
+                    }
+                }
+                if (!isNested)
+                    return objects.Where(x => prop.GetValue(x).Equals(stringValue)).ToList();
+                else
+                    return objects.Where(x => BHRE.Query.PropertyValue(x, propertyName).Equals(stringValue)).ToList();
+            }
+
+            if (type == typeof(System.Double))
```
adecler: Should it not also cover `int`, `float`, and `decimal` ?

Data_Engine/Query/Linq/FilterBy.cs
```diff
@@ -0,0 +1,214 @@
...
+            System.Reflection.PropertyInfo prop = firstObject.GetType().GetProperty(propertyName);
+            if (prop == null && firstObject is IBHoMObject)
+            {
+                if ((firstObject as IBHoMObject).CustomData.ContainsKey(propertyName))
+                    isCustomData = true;
+            }
+
+            System.Type type = PropertyType(firstObject, propertyName);
+
+            int groupedObjects = objects.GroupBy(x => x.GetType()).Count();
```
adecler: Ouch! If objects contains null elements, this will throw an exception. 
As I said before, the best is to first do `objects = objects.Where(x => x != null).ToList()`

Data_Engine/Query/Linq/FilterBy.cs
```diff
@@ -0,0 +1,214 @@
...
+                if (!isNested && !isCustomData)
+                    return objects.Where(x => System.Math.Round((double)prop.GetValue(x), rounding) == doubleValue).ToList();
+                else
+                    return objects.Where(x => System.Math.Round((double)BHRE.Query.PropertyValue(x, propertyName), rounding) == doubleValue).ToList();
+            }
+
+            if (prop != null && !isNested)
+            {
+                dynamic val = System.Convert.ChangeType(value, prop.PropertyType);
+                return objects.Where(x => (System.Convert.ChangeType(prop.GetValue(x), type) as dynamic) == val).ToList();
```
adecler: Not sure I am clear why you need to do this.
What about this?
```c#
if (value != null) 
    return objects.Where(x => value.Equals(prop.GetValue(x))).ToList();
else 
    return objects.Where(x =>prop.GetValue(x) == null).ToList();
```

Data_Engine/Query/Linq/OrderBy.cs
```diff
@@ -0,0 +1,70 @@
...
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Orders a list of objects by a property of the object - e.g. input 'Name' to sort objects by their names. Note that not all object properties have sortable values")]
+        [Input("objects", "List of objects to be sorted. All objects in the list should be of similar type")]
+        [Input("propertyName", "The name of the property to sort the list of objects by. Note that the property must be able to be sorted - e.g. numbers/text")]
+        public static List<T> OrderBy<T>(this List<T> objects, string propertyName)
+        {
+            if (objects == null || objects.Count == 0) return new List<T>();
+            int groupedObjects = objects.GroupBy(x => x.GetType()).Count();
```
adecler: Same thing here, you first need to remove the null items from the object list.

Data_Engine/Query/Linq/OrderBy.cs
```diff
@@ -0,0 +1,70 @@
...
+            if (groupedObjects != 1)
+            {
+                BHRE.Compute.RecordError("All objects in the list to be sorted should be of similiar type.");
+                return null;
+            }
+            try
+            {
+                if (!propertyName.Contains("."))
+                {
+                    System.Reflection.PropertyInfo prop = objects.First(x => x != null).GetType().GetProperty(propertyName);
```
adecler: `objects.First(x => x != null)` might throw an exception if there isn't any non-null elements in the list

Data_Engine/Query/Linq/FilterBy.cs
```diff
@@ -0,0 +1,214 @@
...
+                return new List<T>();
+            }
+
+            if (value == null)
+            {
+                BH.Engine.Reflection.Compute.RecordError("value cannot be null to filter the objects");
+                return new List<T>();
+            }
+
+            object firstObject = objects.First(x => x != null);
```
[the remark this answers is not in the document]
FraserGreenroyd: Have changed to a `where` with `FirstOrDefault` and then a `null` check.

FraserGreenroyd:
Ok so I started looking at your comments @adecler and managed to resolve the first one (maybe) but then saw the others need a bit more thinking than I can do right this moment so I'll try and continue this tomorrow for you to take another look ASAP to avoid it slipping too much longer :smile:

Data_Engine/Query/Linq/FilterBy.cs
rwemay: This was me early on and needing an equality for different cases (strings and objects from memory?) as one did not cover the other.

FraserGreenroyd:
I have undertaken a refactor of the `FilterBy` method to try and make it easier to follow, and to make use of the existing `PropertyValue` method in the Reflection_Engine. Still using dynamic at the moment because that's what is working and the original `prop` variable has gone in the refactor, but open to suggestions on improvements on that part of it (FAO: @adecler )

I have not changed the `OrderBy` much but I have added some `null` checking per @adecler comments.

I might have refactored too much, I'm not sure. This is working for my test scripts that I could generate, but @rwemay it would probably be good if you could rerun your scripts and see if it's also still working for what you were looking at? I'm happy to disregard the commit which does the refactoring if it doesn't help and try again from what we had before if necessary.

Anyhow, review needed please @adecler @rwemay @awakeman @al-fisher @enarhi @michaelhoehn :smile:

rwemay:
Thanks guys - much neater! The methods in the test script I saved in the folder work as previous with the exception of FilterBy. The filter by BHoM object (Level in this case) works well, but I had previously set up so you could put in a partial double and filter by that. So in the example below, all nodes with an x value that rounds to 75 would be returned. I was probably trying to be a bit too helpful/guessing most typical use-cases, in that + including the string case/match. 

I did try and avoid the PropertyValue method in the first commit, where I could, with a bit of gymnastics - mainly due to the performance (was _much_ slower in earlier prototype. I think that's resolved now, with update to the method. 

I wonder if we should actually pull ignoreStringCase, and exactStringMatch into a separate predicate (or a more user-friendly name - FilterCondition?!) class. The predicate could be optional, with ignoreStringCase and exactStringMatch set to false by default. 

Is everyone happy with the name of the method (FilterBy)? It's wrapping Linq 'where', but I didn't think `BH.Engine.Data.Query.Where` would be intuitive. `Match`, `Find`, `Search` were others I thought of (there is a `Match` in library engine query). 

image

Data_Engine/Query/Linq/FilterBy.cs
```diff
@@ -0,0 +1,162 @@
...
+                return new List<T>();
+            }
+
+            if (value == null)
+            {
+                BH.Engine.Reflection.Compute.RecordError("value cannot be null to filter the objects");
+                return new List<T>();
+            }
+
+            object firstObject = objects.Where(x => x != null).FirstOrDefault();
```
adecler: You should do the same as for `OrderBy`: 
`objects = objects.Where(x => x != null).ToList();`

The added benefit is that you don't need to check for null when getting for object or anywhere after.

Data_Engine/Query/Linq/FilterBy.cs
```diff
@@ -0,0 +1,162 @@
...
+            if(firstObject == null)
+            {
+                BH.Engine.Reflection.Compute.RecordError("All objects in the list to filter are null, please try with valid objects");
+                return new List<T>();
+            }
+
+            System.Type type = PropertyType(firstObject, propertyName);
```
adecler: what happens if `type` is null ?

Data_Engine/Query/Linq/FilterBy.cs
```diff
@@ -0,0 +1,162 @@
...
+
+            List<T> filteredObjects = new List<T>();
+
+            if (type == typeof(System.String))
+            {
+                if (ignoreStringCase)
+                    filteredObjects.AddRange(objects.Where(x => x.PropertyValue(propertyName).ToString().Equals(value.ToString(), System.StringComparison.CurrentCultureIgnoreCase)).ToList());
```
adecler: Assuming that you have removed the null objects already, there is still a risk that `x.PropertyValue(propertyName)` might be null

Data_Engine/Query/Linq/FilterBy.cs
```diff
@@ -0,0 +1,162 @@
...
+                else
+                    filteredObjects.AddRange(objects.Where(x => x.PropertyValue(propertyName).ToString().Equals(value.ToString())).ToList());
+
+                if (!exactStringMatch)
+                    filteredObjects.AddRange(objects.Where(x => x.PropertyValue(propertyName).ToString().Contains(value.ToString())).ToList());
```
adecler: Doesn't that mean that values matching exactly are added twice ?

Data_Engine/Query/Linq/FilterBy.cs
```diff
@@ -0,0 +1,162 @@
...
+            }
+            else
+            {
+                dynamic val = System.Convert.ChangeType(value, type);
+                filteredObjects = objects.Where(x => (System.Convert.ChangeType(x.PropertyValue(propertyName), type) as dynamic) == val).ToList();
```
adecler: I feel we had a discussion already about `ChangeType()` vs `Equals()` but I honestly don't remember so just putting this is as a note to check.

Data_Engine/Query/Linq/FilterBy.cs
```diff
@@ -0,0 +1,162 @@
...
+            }
+
+            return filteredObjects.Distinct().ToList();
```
adecler: Ha I see. Well I guess I should have read it all before commenting then 😋 .

I get that you want to have the exact matches first in your list but I wonder if that is really a feature you want. The user doesn't really have a way to know where in the list objects stop being exact matches so having them first is only partially helpful. On the other hand, it messes up the order of the input objects (say they where order alphabetically or something) which might be annoying for the user.

Data_Engine/Query/Linq/OrderBy.cs
```diff
@@ -0,0 +1,84 @@
...
+            if(objects.Count == 0)
+            {
+                BH.Engine.Reflection.Compute.RecordError("All objects in the list to order are null, please try with valid objects");
+                return new List<T>();
+            }
+
+            int groupedObjects = objects.GroupBy(x => x.GetType()).Count();
+            if (groupedObjects != 1)
+            {
+                BHRE.Compute.RecordError("All objects in the list to be sorted should be of similiar type.");
```
adecler: Just a note but should it not say this instead:
All objects in the list to be sorted should be of the _**exact same**_ type

Data_Engine/Query/Linq/OrderBy.cs
```diff
@@ -0,0 +1,84 @@
...
+                return null;
+            }
+            try
+            {
+                if (!propertyName.Contains("."))
+                {
+                    System.Reflection.PropertyInfo prop = objects.First(x => x != null).GetType().GetProperty(propertyName);
```
adecler: The null check is not needed

Data_Engine/Query/Linq/FilterBy.cs
FraserGreenroyd: Am clicking resolve only to take it off the to-do list of changes I need to work on but this comment can be taken to assume I have not done anything with this

Data_Engine/Query/Linq/FilterBy.cs
FraserGreenroyd: I've refactored the statements slightly to try and avoid adding entries twice so this might be resolved by that. It might not be also, but I'm personally not too concerned about the order changing as we're adding the `OrderBy` as well/lists can be reordered, and nothing in our filter description suggests we will preserve current order - but maybe we can make that more explicit in the description if needed?

FraserGreenroyd:
I've made some of those changes suggested @adecler as best as I could think of for a Friday morning 😋  and put a couple of comments on bits I haven't resolved (or resolved in a way that 100% matches your comment).

For your comment @rwemay I'm not convinced by the attempt to allow numbers 'close to' 75 to be matched on a filter by 75 just because, to me, 74.9 is not 75 - but we could maybe look to add a filter by range as an additional component to handle some of those options?

rwemay:
I was holding back a bit on this as was keen to make it more generic and have a Comparer or Equalities input, but I think there's a bit of work to do to consolidate a few concepts bubbling up elsewhere (logical and/or etc), and have a family resemblance to Requests. 

Now that understand there's a real use-case/need asap in the US (thanks @FraserGreenroyd ) suggest we merge this one. At worst we'll deprecate but maybe just another method alongside, since a huge number of filters are going to be by strings/exact matches. 

I can't approve this as I think I created it, @al-fisher might you be able to take a look? The test scripts are pretty quick/small.

adecler:
@rwemay , I'll let you press the button 😄

PR #1324, opened 2019-11-14
Library_Engine: Update libraries to work with Datasets

### LINKED ISSUE #1286: Library_Engine: Update Library method to work with Dataset object rather than flat list of obejcts
#### Description:

Going forward, the default for adding datasets should be through the use of the Dataset object. The library engine needs to be updated to work under this assumption. To avoid disruption, the current functionality of the Library engine will be kept intact until all current Datasets have been given a chance to get migrated to the new format.

IsakNaslundBh:

Closes #1286 

### Test files

To Test, the following can be placed in the Dataset folder on Appdata. These are not finalised datasets yet, but just contructed for testing:
[SectionProperties_DataSets.zip](https://github.com/BHoM/BHoM_Engine/files/3845503/SectionProperties_DataSets.zip)

Simple GH testfile
https://burohappold.sharepoint.com/:f:/s/BHoM/ErnA5fNcgJ9Dohk_eyWUZ7EBR_zd4Lp16Azw_-al3BepCA?e=6FE2Bv

Would also encourage to test that current libraries work as expected and to generate a new one using Dataset to test the functionality.

### Changelog

- Library_Engine now assumes data stored in 1 Dataset per json file
- Still supporting previous format with a list of BHoMObjects
- Added Source method to extract Source from Library in Dataset


Library_Engine/Query/Libraries.cs
```diff
@@ -74,47 +92,94 @@ public static void RefreshLibraries()
...
                         {
-                            obj = (IBHoMObject)BH.Engine.Serialiser.Convert.FromJson(entry);
+                            IBHoMObject obj;
+                            try
+                            {
+                                obj = (IBHoMObject)BH.Engine.Serialiser.Convert.FromJson(entry);
+                            }
+                            catch
+                            {
+                                AddDeserialisationEvent(name, oM.Reflection.Debugging.EventType.Warning, "Failed to deserialise at least one item in the dataset named " + name);
```
FraserGreenroyd: Would it be helpful to state which `entry` has failed to deserialise?

IsakNaslundBh: Presume it could be useful, but two points:
1. As it failed to deserialise, I will not really have anything to look at for a generic case to state what it was. As in, failing to deserialise means I dont have a BHoMObject to take name or similar from to display. Could paste the whole string out or something like that, but not sure how useful that would be?
1. In general when we raise multiples of the same error, it generally feels as it is more clogging the UI than helping out + at least GH has a maximum error count of 15.

Please let me know if you are ok with this, or want me to update!

FraserGreenroyd:
@IsakNaslundBh will let you decide to merge based on my review or wait for others or addressing my comment as appropriate, but all checks are happy :smile:

Library_Engine/Query/Libraries.cs
FraserGreenroyd: I'm happy with that, it makes sense to me, happy to go ahead with merging without changes :smile:

PR #1326, opened 2019-11-14
Reflection_Engine: Add EnumValues method

### LINKED ISSUE #1325: Reflection_Engine - add EnumValues method
#### Description:
It would be good to be able to get out all values of an enum with one component in GH - it is way more handy than dropping multiple dropdowns on a canvas.

pawelbaran:

Closes #1325

### Test files

Uploaded to [SharePoint](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?viewid=a5a93cba%2Dfcca%2D46cc%2Da31a%2D80b8d5cbfd7b&id=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FTest%20Scripts%2FBHoM%5FEngine%2FReflection%5FEngine%2FReflection%5FEngine%2DIssue1325%2DAddEnumValuesMethod).

### Changelog
- `EnumValues()` method has been added in `BH.Engine.Reflection.Query`

### DIFF
--- a/Reflection_Engine/Query/EnumValues.cs
+++ b/Reflection_Engine/Query/EnumValues.cs
@@ -1,0 +1,45 @@
+﻿/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2019, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 
+ * GNU Lesser General Public License for more details.                          
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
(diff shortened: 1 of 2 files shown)


Reflection_Engine/Query/EnumValues.cs
```diff
@@ -0,0 +1,45 @@
...
+{
+    public static partial class Query
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        public static IEnumerable<T> EnumValues<T>(this T enumeration)
+        {
+            return Enum.GetValues(enumeration.GetType()).Cast<T>();
```
FraserGreenroyd: Some comments:

 - I personally would like to see some description/input/output attributes but won't hold up the PR over them.
 - I tried this with plugging in some enum components in GH and it worked as expected, however, when I plugged in the `type` of the enum (using the type component), it throws an error because `type` is not an `enum`. When I plug in the enum component I can still change the drop down of that enum, which doesn't make a lot of sense when getting just the full list of enum options. So I wonder if it's worth attempting to get all the values from the type rather than the enum itself? For a better UX? Thoughts @al-fisher @adecler ?

pawelbaran: Yeah, the motivation to create this method is purely practical: I was asked by the end users to investigate how to 'explode' the dropdown into a list. Happy to refactor, as long as the functionality is in place.

pawelbaran: @FraserGreenroyd I've added a method that takes the type - now we have two for both scenarios.

adecler:
I am just a bit unclear why this takes an enum instance instead of an enum type as input.

adecler:
Hmm OK, I am still not convinced that it wouldn't be better to have to use the type component in the UI. But I don't see this as a big enough problem than I would hold up merging the PR.

PR #1328, opened 2019-11-14
Geometry_Engine CurveProximity, return single point twice for intersection events

### LINKED ISSUE #1327: Geometry_Engine: CurveProximity has index out of bounds errors
#### Description:

image
Seems like there are some issues regarding assumptions of intersections

kThorsager:

Closes #1327 
CurveProximity always tried to return both intersection events for some curves, which caused issues when there only was one.

Now it returns the first point of the intersection twice if there is an intersection event.
The method is really not structured to be able to return all closest points

### Test files

[CurveProximity](https://burohappold.sharepoint.com/:u:/s/BHoM/EW3dPy2DKiBEuYL3nxchY6sBKpYfVyxIj032G682HsQSOg?e=saKZxY)

### Changelog

- `BH.Engine.Geometry.Compute.CurveProximity(Line, Circle)` coplanar curve edge case bug fixed

### Additional comments
also @pawelbaran  @FraserGreenroyd  @KonradStolarski  why did it try to return separate intersections (i.e both of two planar circles intersection points)? I thought that this was used in to find distance between curves (and this would cause some intersecting curves to have a distance between them)

The points it retuned:
image

### DIFF
--- a/Geometry_Engine/Compute/CurveProximity.cs
+++ b/Geometry_Engine/Compute/CurveProximity.cs
@@ -59,7 +59,8 @@ namespace BH.Engine.Geometry
             if (curve1.IsCoplanar(curve2))
                 return new Output<Point, Point> { Item1 = min1, Item2 = min2 };
 
-            double[] t = new double[] { curve1.SkewLineProximity(curve2).Item1, curve1.SkewLineProximity(curve2).Item2 };
+            Output<double,double> skewLineProximity = curve1.SkewLineProximity(curve2);
+            double[] t = new double[] { skewLineProximity.Item1, skewLineProximity.Item2 };
             double t1 = Math.Max(Math.Min(t[0], 1), 0);
             double t2 = Math.Max(Math.Min(t[1], 1), 0);
             Vector e1 = curve1.End - curve1.Start;
--- a/Geometry_Engine/Compute/CurveProximity.cs
+++ b/Geometry_Engine/Compute/CurveProximity.cs
@@ -176,6 +176,25 @@ namespace BH.Engine.Geometry
             plnPts.Add(curve2.Centre);
             Point tmp;
             Plane ftPln = plnPts.FitPlane();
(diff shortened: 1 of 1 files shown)


pawelbaran:
Yep the method is just wrong I think. It should work similar to Rhino: find a pair of points on curves that are closest to each other. If the curves intersect, intersection point should be returned twice in a form of `Output<Point, Point>{Item1 = intersectionPoint, Item2 = intersectionPoint}`.

pawelbaran:
@kThorsager:
- please update the test file so that it exposes the issue
- please share its location
- please fill in the changelog following the guidelines
- please add the PR to SCRUM board

It is important to understand that **programming is not a race and testing/documentation is more important than the actual implementation** because:
- without following guildelines you create more mess than added value
- without a clear test case the code is unverified
- it takes time to understand the code
- it takes time to find/create the test cases
- it takes time to understand and evaluate the provided test files

The more streamlined you make the review and merging process the better.

kThorsager:
Sorry, didn't know there was a test file but I'll get on updating it

And I thought the PR title was the change log for smaller PRs, but I'll clarify it a little

And sorry about the SCRUM board, I'd forgotten about it

kThorsager:
I have updated the test file to include forced intersection events (random curves in 3d doesn't touch very often) for the affected curve combinations, and am on my way to add it for the others. 

But as it turns out, testing is good, because `CurveProximity` for two co-planar circles is completely broken as well, so I'll keep this in WIP until I've fixed that and made sure there are no such surprises in the rest of them

kThorsager:
Test file updated and every `Circle` - `ICurve` had some problems when they were both placed on a plane

pawelbaran:
Yeah, just looked into the test file - looks like there is an issue with `Circle`-`Line` proximity method.

kThorsager:
The Circle - Line problem has been fixed by adding a special case for 2d on top of the existing functionality.

No errors in the test file any more, should be good to go @pawelbaran

PR #1335, opened 2019-11-20
Reflection_Engine: Add possibility to include full path for parameters in method.ToText()

### LINKED ISSUE #1334: Reflection_Engine: Add possibility to include full path for parameters in method.ToText()
#### Description:
This is necessary to guarantee uniqueness of the method string representation. 
This change is in anticipation of the changes in the versioning engine where old method description is relying on the format provided by the `ToText()` method

adecler:

Closes #1334

### Test files
You should get something like this when testing the ToText component on method (i.e. full path for method parameters):
image

### Additional comments
Nothing depends on this. This is purely to help users generate the correct string when adding a method upgrader to the versioning engine.

### DIFF
--- a/Reflection_Engine/Convert/ToText.cs
+++ b/Reflection_Engine/Convert/ToText.cs
@@ -48,7 +48,7 @@ namespace BH.Engine.Reflection
 
         /***************************************************/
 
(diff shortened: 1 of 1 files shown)


FraserGreenroyd:
Have updated title for Change Log :smile:

PR #1338, opened 2019-11-20
Serialiser_Engine: Remove attempt to re-deserialise a custom object

### LINKED ISSUE #1337: Serialiser_Engine: Stack overflow in deserialising fragments
#### Description:
Trying to deserialise fragments, which changes from inheriting `BHoMObject` to `IObject` cause old scripts to crash without attempting to load

#### How to replicate:
Load an oldish script (month or so old) containing fragments (possibly Environment Fragments to be specific)

#### Expected behaviour:
Script loads but with errors where appropriate

#### Test file(s):
[Here](https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/01_Test%20Scripts/BHoM_Engine/Serializer_Engine/ROI_Rhino%20to%20IES_RoomReference(1).gh?csf=1&e=pitETz)

FraserGreenroyd:
Fixes #1337 

### Test files
[Here](https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/01_Test%20Scripts/BHoM_Engine/Serializer_Engine/ROI_Rhino%20to%20IES_RoomReference(1).gh?csf=1&e=pitETz)

### Additional comments
During an investigation into Excel problems with @awakeman and @IsakNaslundBh today, it was discovered that some old scripts weren't loading following a change to fragment structure done by myself and @alelom . Namely, the PR to Serialiser_Engine of the 9th November suddenly caused old scripts not to deserialise.

The line which appeared to be causing it was `object result = Convert.FromBson(customObject.ToBson());` it appeared to be getting recursively stuck and never moving onto the next line.

Commenting out this line (and then removing it) appeared to fix the problem, in so much as the attached script could open, and Excel could load again.

I opted to remove the line, because I am unsure what it was adding to the function, because it appeared to be serialising the `customObject` to BSon, and then immediately deserialising it back to an `object`.

Which is why I'm only asking for a review from @adecler explicitly at this stage, as I am unsure what potential consequences this will have on the versioning stuff, and hopefully you'll be able to explain the above to me :smile:

### DIFF
--- a/Serialiser_Engine/Objects/BsonSerializers/ObjectSerializer.cs
+++ b/Serialiser_Engine/Objects/BsonSerializers/ObjectSerializer.cs
@@ -31,6 +31,7 @@ using BH.Engine.Versioning;
 using System;
 using System.Reflection;
 using BH.Engine.Serialiser.Objects;
+using System.Collections.Generic;
 
 namespace BH.Engine.Serialiser.BsonSerializers
 {
@@ -294,12 +295,30 @@ namespace BH.Engine.Serialiser.BsonSerializers
                 {
                     // A child of the object is causing problems. Try to recover from custom object
                     IBsonSerializer customSerializer = BsonSerializer.LookupSerializer(typeof(CustomObject));
-                    object customObject = customSerializer.Deserialize(context, args);
+                    object result = customSerializer.Deserialize(context, args);
+                    Guid objectId = ((CustomObject)result).BHoM_Guid;
 
-                    if (customObject is CustomObject)
+                    if (!Config.TypesWithoutUpgrade.Contains(actualType))
+                    {
+                        if (m_StackCounter.ContainsKey(objectId))
+                            m_StackCounter[objectId] += 1;
+                        else
+                            m_StackCounter[objectId] = 1;
+
+                        if (m_StackCounter[objectId] < 10)
+                        {
+                            result = Convert.FromBson(result.ToBson());
+                            m_StackCounter.Remove(objectId);
+                        }
+                    }
+                    
+                    if (result is CustomObject)
+                    {
                         Engine.Reflection.Compute.RecordWarning("The type " + actualType.FullName + " is unknown -> data returned as custom objects.");
-
-                    return customObject;
+                        Config.TypesWithoutUpgrade.Add(actualType);
+                    }
+                        
+                    return result;
 
                 }
                 else if (actualType != typeof(IDeprecated))
@@ -324,6 +343,8 @@ namespace BH.Engine.Serialiser.BsonSerializers
         /***************************************************/
 
         private readonly IDiscriminatorConvention _discriminatorConvention;
+
+        private static Dictionary<Guid, int> m_StackCounter = new Dictionary<Guid, int>();
 
 
         /*******************************************/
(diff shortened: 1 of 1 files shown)


adecler:
Well, the problem is that you have modified objects but nothing in the versioning engine to upgrade them. The obvious solution would be to add the missing upgrade in the versioning engine but we can all agree that a stack overflow is never a good thing.

"_But, wait, why the infinite loop ? Aren't we just returning Custom objects when we cannot upgrade?_" you might ask. The answer to that is slightly complicated but let me give it a shot: All objects to deserialise first arrive as `object` and are handled by us through the `BH_ObjectSerializer`. The children of those objects, on the other hand, already have an expected type since the parent knows what to expect. those children are therefore going through an automatically generated deserialiser that we have no control over. When a children has been modified through its type or namespace, we can catch that and fix it since Mongo Bson doesn't even know what serialiser to use for it. When the child class was only modified internally, Bson just go ahead with the corresponding deserialiser but then throws an exception when the internals don't match. The problem is that we only catch that exception at the level of the deserialisation of the parent (this is why I am looking for internal exception in the code). If you turn the parent object into a custom object, all the children are suddenly treated as 'object' and we can go back to being in control. This is why we have this weird `Convert.FromBson(customObject.ToBson())` so that the Bson for the problematic child is fixed and the parent is then able to deserialise without problem...... I hope that makes at least a bit of sense 😋.

I'll push a new version of that code to make sure we don't go infinite while still handling the case detailed above.

adecler:
Alright, this should fix it. I have used an arbitrary number of 10 for the level difference between the parent and the modified (grand-)child. I couldn't really come up with a solution on the spot for how to avoid that. I am sure this solution can be improved but it should do the trick for now.

FraserGreenroyd:
> "_But, wait, why the infinite loop ? Aren't we just returning Custom objects when we cannot upgrade?_" you might ask. The answer to that is slightly complicated but let me give it a shot: All objects to deserialise first arrive as `object` and are handled by us through the `BH_ObjectSerializer`. The children of those objects, on the other hand, already have an expected type since the parent knows what to expect. those children are therefore going through an automatically generated deserialiser that we have no control over. When a children has been modified through its type or namespace, we can catch that and fix it since Mongo Bson doesn't even know what serialiser to use for it. When the child class was only modified internally, Bson just go ahead with the corresponding deserialiser but then throws an exception when the internals don't match. The problem is that we only catch that exception at the level of the deserialisation of the parent (this is why I am looking for internal exception in the code). If you turn the parent object into a custom object, all the children are suddenly treated as 'object' and we can go back to being in control. This is why we have this weird `Convert.FromBson(customObject.ToBson())` so that the Bson for the problematic child is fixed and the parent is then able to deserialise without problem...... I hope that makes at least a bit of sense 😋.

That makes sense to me, and it appears the fix pushed handles the issue this PR aims to address so I'm happy I've learnt something :smile:

rwemay:
FWIW I've tested this as I needed to test something else in Excel. All worked fine. Will be good to get this merged as Excel_Toolkit currently won't function without as it relies on serialisation on load.

FraserGreenroyd:
Thanks @rwemay - I'm hopeful that @awakeman @IsakNaslundBh or @alelom will be here very soon to expedite this :smile:

FraserGreenroyd:
Am awaiting approval from @IsakNaslundBh who is looking at a specific case before hitting the green button but I'm hopeful @IsakNaslundBh will be happy :smile:

alelom:
Worked for @kThorsager as well 😄

kThorsager:
Yes, before using this branch the enable BHoM button caused Excel to crash and restart, to crash again until you stopped it. Now it works as I would expect (no crash)

FraserGreenroyd:
Thanks everyone, great team effort - I do enjoy the connectivity first thing in the morning when we all come together round something like this :smile:

PR #1352, opened 2019-11-26
Geometry_Engine: Add constructor for TaperedProfile

### LINKED ISSUE #1351: Geometry_Engine: Constructor associated with TaperedProfile
#### Description:

Constructor and other methods associated with `TaperedProfile `objects:

https://github.com/BHoM/BHoM/issues/597

peterjamesnugent:
### NOTE: Depends on 

https://github.com/BHoM/BHoM/pull/629
   
Closes #1351 

Added consturctor for `TaperedProfile`

### Test files

https://burohappold.sharepoint.com/:f:/s/BHoM/EnUcHH0s8oBEjpv_0llfrNYBaqnCW7KtgEVismCaC4kmow?e=uPvp2p

### Changelog

Added consturctor for `TaperedProfile`


IsakNaslundBh:
Agree with Josef that having one taking start and end would be good.

For the second one, not 100% sure. could assume for the bar to be prismatic from the lowest value to 0 as well as between highest value and 1 potentially? 
But should probably add in a check that all positional values are withing the 0-1 range as well at least.

peterjamesnugent:
Added the following error messages:
- Check whether both lists are the same length
- Check whether a Start and EndProfile are provided (i.e. 0 and 1 in positions)
- Check whether all positions lie in the following range 0<= t <= 1

I have added the second constructor too, all of this can be tested in the test file.

@IsakNaslundBh I think we should require the StartProfile and EndProfile, that way it's clearer when creating the object that it is prismatic from the position to the start/end.

IsakNaslundBh:
@peterjamesnugent Yes, thinking a bit more about it I think you and @JosefTaylor are correct. A lot more clear to force a start and end instead of assuming extrapolation from the first/last.

peterjamesnugent:
Can this be reviewed soon (to allow time for changes), I am on leave tomorrow afternoon until Tuesday, so I will not be available for the sprint closure.

JosefTaylor:
I'm able to create the profile, but not the section. A new issue?
image

peterjamesnugent:
> I'm able to create the profile, but not the section. A new issue?

I think this would require rethinking the `ISectionProperty` class, because every property would have to be a List<double> instead of just double - maybe this is right as it allows varying profiles and uniform ones too all within the same class. It does have a wider impact on every structural toolkit that interacts with a `SectionProperty` object when pushing/pulling as well as all the create, query, compute and modify methods.

It would be a significant refactor I believe.

The current workaround (as I see it), is to extract the profiles on a toolkit level and create the section there when you want to push it to a structural adapter.

```
namespace BH.oM.Structure.SectionProperties
{
    public interface ISectionProperty : IBHoMObject, IObject
    {
        IMaterialFragment Material { get; set; }
        double Vpy { get; }
        double Vy { get; }
        double Vpz { get; }
        double Vz { get; }
        double CentreY { get; }
        double CentreZ { get; }
        double Wplz { get; }
        double Wply { get; }
        double Welz { get; }
        double Wely { get; }
        double Iw { get; }
        double Iz { get; }
        double Iy { get; }
        double J { get; }
        double Rgz { get; }
        double Rgy { get; }
        double Area { get; }
        double Asy { get; }
        double Asz { get; }
    }
}
```

IsakNaslundBh:
Parking this to 3.1. No Real point forcing this through without any support method for it in the engine and no support in the adapters. Lets get it in early in 3.1.

PR #1368, opened 2019-12-03
Geometry_Engine: Add Triangulate method

### LINKED ISSUE #1367: Geometry_Engine: Add capacity to triangulate polylines
Add ability to triangulate polylines to create cuts from internal polylines.

FraserGreenroyd:
Fixes #1367 

### Test files
[Here](https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/01_Test%20Scripts/BHoM_Engine/Geometry_Engine/Compute/Triangulate.gh?csf=1&e=GlcKf4)

### DIFF
--- a/Geometry_Engine/Compute/Triangulate.cs
+++ b/Geometry_Engine/Compute/Triangulate.cs
@@ -1,0 +1,126 @@
+﻿/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2019, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 
+ * GNU Lesser General Public License for more details.                          
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using BH.oM.Geometry;
+using BH.oM.Reflection.Attributes;
+using System;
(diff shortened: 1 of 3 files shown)


pawelbaran:
@FraserGreenroyd could you please fill the changelog?

As discussed on the call, not sure if I like the idea of referencing an external library for the sole reason of implementing one method (so far we managed not to use such shortcuts). If that were to happen, shouldn't we be more deliberate on which library to choose? There is plenty of them out there and we definitely do not want to create a patchwork of references in the Engine. @adecler @al-fisher

As an interim solution, couldn't we implement this method in the specific repo that needs it?

FraserGreenroyd:
As discussed with @IsakNaslundBh and @pawelbaran there is an agreement among all of us that we should be careful on what external libraries we include and ensure we do not sleepwalk into a chaos of numerous libraries unnecessarily. However, there is no functionality currently provided within BHoM to allow for triangulation of geometry in the way which resolves the issues faced at least within BEnv at this time. I am not against us writing out the need for the library if we can write our own triangulation method, but I am not convinced that we have the resources to do that this week when we have, readily available, this method which has been prototyped ahead of this PR.

I am certain there will not be one catch-all library available which will handle every extra/advanced geometry operation we want, because generally the libraries specialise and refine their working thing without trying to be too blanket cover-all and spread itself too thin. Therefore we should not be against using external libraries that do that work for us, but, it needs to be on a case-by-case basis. This is the first case, where triangulation of geometry is not available in the Geometry_Engine at this time, but it is needed, and we do not have the resource to achieve it ourselves at this time, so we're relying on an accepted library at least in the interim. In the same way as we have chosen to use the Mongo library for our BSon requirements in other engines, we too we need to evaluate the needs of this library rather than reject it outright.

 > As an interim solution, couldn't we implement this method in the specific repo that needs it?

It has already been prototyped on several repos that needed it, and it is becoming unwieldy to continue duplicating functionality unnecessarily as it is needed to unblock a bug on XML Toolkit as well. Hence, moving it here, to the central Geometry Engine, so that it may be more accessible and open for discussion than continuing to hide it away in other toolkits.

al-fisher:
Agreed with above - in addition to simple management of dependencies - is also super important to help with _communication/documentation_ of libraries the (in particular) **core BHoM** depends on.
I would not rule out Triangle.NET_Toolkit ? @pawelbaran @FraserGreenroyd

pawelbaran:
Another repo wrapping external libraries could be the way to go, I like the idea!

FraserGreenroyd:
Following discussion with @al-fisher it is agreed to wrap libraries like this within their own toolkit repo. This gives us flexibility and modularity as appropriate for this library, and additional libraries when we deem necessary.

This PR will close in favour of a new TriangleNet_Toolkit :smile:

PR #1385, opened 2019-12-12
Serialiser_Engine: Enable serialisation of immutable objects with immutable base class

### LINKED ISSUE #1384: Serialiser_Engine: IImutable objects with abstract base class fail to serialise
#### Description:

All `MeshElementResults` currently fail to serialise. This seem to stem from them being immutable 
and having an abstract base class which also is immutable.

After some investigation it seems as the the class map is failing when trying to run the AutoMap on row 181 in the RegisterClassMap method. https://github.com/BHoM/BHoM_Engine/blob/ec3a2a04a589b1925d975a92b8d875682f36f151/Serialiser_Engine/Convert/Bson.cs#L180-L182

From checking the stack trace and by trying to debug placing breakpoints in our methods it looks as if the standard [ImmutableTypeClassMapConvention](https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Bson/Serialization/Conventions/ImmutableTypeClassMapConvention.cs) is being used rather than our own.

This then fails as it tries to register memberinfo from the abstract base class to the superclass here:

https://github.com/mongodb/mongo-csharp-driver/blob/4fc6a30bd8b62ed1acbec01bd6b8df435f01a99d/src/MongoDB.Bson/Serialization/BsonClassMap.cs#L1252-L1262

I tried adding a dummy member to the MeshElementResult with a setter and that made the class registration hit [our method](https://github.com/BHoM/BHoM_Engine/blob/master/Serialiser_Engine/Objects/MemberMapConventions/ImmutableBHoMClassMapConvention.cs), but then all got registred without any of the members from the baseclass which made it fail to serialise the object. (plus we ofc do not want to enforce having at least one public setable property on all immutable obejcts)

I also tried making the mesh class mutable (removing interface and adding setters to all properties) which made it work fine.

have also tried various things like setting the abstract base class to a root class but no luck.

Found this after googling:

https://jira.mongodb.org/browse/CSHARP-1559

Seems like that guy has the same issue as we do and have solved it by forking and tweaking the mongo drivers. Something I guess we would like to avoid if at all possible.

Would be nice if this can be done before the beta, as we have put a bit of focus on making working with results smoother in this quarter, but if not possible, so be it.

Will check if @epignatelli knows anything more, if not would be grateful if you could have a look at this @adecler .

#### How to replicate:

try ToJson on any mesh result

#### Expected behaviour:

For the object to serialise correctly

#### Test file(s):

https://burohappold.sharepoint.com/:f:/r/sites/BHoM/02_Current/12_Scripts/01_Test%20Scripts/BHoM_Engine/Serializer_Engine/Serialiser_Engine-Issue1384-FailingToDeserialiseSomeImmutableObjects?csf=1&e=O7JV7F

Note test files requires robot. Can not just store out a mesh result for obvious reasons.

IsakNaslundBh:

Closes #1384 

For background of problem, please see issue.

Tried fixing the problem by adding our own version of BsonClassMap and adding another method for registering members without checking if the member is declared in the type of the class map.

This feels like sort of a hacked solution, and not sure if it is the right way forward.

Could get this to work if I added one extra public get-set property to the `MeshElementResult` :

```
    public abstract class MeshElementResult : IResult, IImmutable
    {
        /**** Properties                                ****/

        public string TestProeprty { get; set; } = "";

        public IComparable ObjectId { get; } = "";

....
```

Without the extra property, it still fails as the [default immutable mapper](https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Bson/Serialization/Conventions/ImmutableTypeClassMapConvention.cs) is being used, and is crashing out on the [EnsureMemberInfoIsForThisClass](https://github.com/mongodb/mongo-csharp-driver/blob/4fc6a30bd8b62ed1acbec01bd6b8df435f01a99d/src/MongoDB.Bson/Serialization/BsonClassMap.cs#L1252-L1262) is being called on the BsonClassMap.

Hence this is not a full solution, but raising draft PR anyway in case any of it is useful.

Again, would be nice if you had a chance to look at this @adecler as you know the inner workings of the Bson stuff a lot better than me.

### Test files

https://burohappold.sharepoint.com/:f:/r/sites/BHoM/02_Current/12_Scripts/01_Test%20Scripts/BHoM_Engine/Serializer_Engine/Serialiser_Engine-Issue1384-FailingToDeserialiseSomeImmutableObjects?csf=1&e=tZxYxA


adecler:
@IsakNaslundBh , as promised I have fixed the issue with ImmutableTypeClassMapConvention by replacing it with our own version in the ConventionRegistry. The `RegisterTypes()` method was becoming a bit too big for my liking so I reformatted it but the actual changes to the code are fairly small (only `RegisterPacks()` contains new code.

I also provided a fallback when people just try to be mean in their argument naming for the constructor since I noticed that MeshResult was still failing to deserialise:

image


@FraserGreenroyd , I suggest we add a check in CI to make sure that names are matching between properties and arguments in immutable objects as my fallback solution only works when a single property is of the type required. Name matching is something I use for the versioning engine as well so a good habit to get people into.

Finally, @IsakNaslundBh , you'll obviously need another reviewer now that I have committed code too.

IsakNaslundBh:
Thanks @adecler ! Will test it through myself and find another reviewer as well!

Regarding the `MeshResult`, I know, PR raised for it yday here: https://github.com/BHoM/BHoM/pull/640

Assume this still is a bit too dangerous to merge before beta?

FraserGreenroyd:
Hi @adecler , I think [this issue](https://github.com/BHoM/Test_Toolkit/issues/103) captures what you describe above? I'd not I'll raise one for 3.1 :smile:

adecler:
ha, yes, perfect! Thanks @FraserGreenroyd !

al-fisher:
Great minds! 😆

adecler:
@IsakNaslundBh , the change feels pretty safe but I would still hold until the beta is released...
Unless we can safely guarantee that nothing is broken. I had written some code that tests serialisation of all BHoM objects in the old test engine so @FraserGreenroyd might have something available to do just that.

al-fisher:
> the change feels pretty safe but I would still hold until the beta is released...

Yes - I would probably agree on this too!
Great to have get this resolved! Thanks @adecler @IsakNaslundBh And worth ploughing on with testing etc. 
But don’t think any benefit of actually merging until after release?

Perhaps pop a `do-not-merge` If all happy!
Cheers

al-fisher:
Haha! 😆 
Great minds again!!

IsakNaslundBh:
All sounds good to me! 😄

FraserGreenroyd:
Has anyone seen @adecler and @al-fisher in the same room together recently? Starting to wonder if they've become the same person after reading all this 🤔

IsakNaslundBh:
To and From Json works as expected for the MeshResults and internalize data works fine as well. Have also tried to (de)serialise some various other things and pulling from Revit which all seem fine.

One thing to note is that when pulling results from Etabs the panel turns to blood when trying to check the Json. Aslo, when hovering over the ouput you end up with a red box with a cross over it. In Rhino 5 the hovering over leads to a full RHino crash, in Rhino 6 they seem to have caught that error, still showing the box, but no crash. Had a chat with @epignatelli who told me this has been happening before for some various objects turned to JSON, so not really a problem with this PR.

For Robot this is not a problem, even if the result is containing more information which is a bit strange. 

image


Mostly putting here to note!

PR #1386, opened 2019-12-12
Structure_Engine: the fallback `public static IGeometry Geometry(this object section)` is now private 

### LINKED ISSUE #1383: Structure_Engine: make the fallback `public static IGeometry Geometry(this object section)` private
#### Broken rules:

The scope of the method is too large because it's public and targets `System.Object`.

This causes problem when dynamically collecting all `Geometry()` methods.

As per title.

alelom:

Closes #1383


alelom:
Thank you @IsakNaslundBh !

PR #1387, opened 2019-12-12
Environment_Engine deprecate create methods 

### LINKED ISSUE #1382: Environment_Engine: Deprecate create methods which are just default
Following the inclusion of automatic create methods into UIs, there is no need for the create methods in the engine as they are duplicate options. They should be deprecated ready for removal.

MajaLindroth:

Closes #1382 

Deprecate create methods which are now created by default

### DIFF
--- a/Environment_Engine/Create/Location.cs
+++ b/Environment_Engine/Create/Location.cs
@@ -47,6 +47,7 @@ namespace BH.Engine.Environment
         [Input("utcOffset", "The offset from UTC for the location (positive or negative), default 0")]
         [Input("name", "The name of the location, default empty string")]
         [Output("location", "An Environment location object - used for defining locations in space for climate analysis")]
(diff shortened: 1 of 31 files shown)


Environment_Engine/Create/Location.cs
```diff
@@ -47,6 +47,7 @@ public static partial class Create
         [Input("utcOffset", "The offset from UTC for the location (positive or negative), default 0")]
         [Input("name", "The name of the location, default empty string")]
         [Output("location", "An Environment location object - used for defining locations in space for climate analysis")]
+        [Deprecated("3.0", "Deprecated in favour of default create components produced by BHoM")]
```
FraserGreenroyd: Hmm on reflection, could we add a new query method which checks if a `Location` object is valid which runs the `if` checks that were included in this create method?

PR #1390, opened 2019-12-17
Environment_Engine Change tolerance default for FloorGeometry

### LINKED ISSUE #1389: Environment_Engine: Use BH.oM.Geometry.Tolerance default in FloorGeometry method
See comment on this PR: https://github.com/BHoM/BHoM_Engine/pull/1042#issuecomment-503773822

This tolerance has never been updated from `0.00001` to a default provided by `BH.oM.Geometry.Tolerance`. This should be updated.

Use `BH.oM.Geometry.Tolerance.Distance` as the default.

MajaLindroth:

Closes #1389 

Change tolerance default to be BH.oM.Geometry.Tolerance.Distance for FloorGeometry query method. 
Add input description.

### DIFF
--- a/Environment_Engine/Query/FloorGeometry.cs
+++ b/Environment_Engine/Query/FloorGeometry.cs
@@ -42,7 +42,7 @@ namespace BH.Engine.Environment
 
         [Description("Returns the floor geometry of a space represented by Environment Panels as a BHoM Geometry Polyline")]
         [Input("panelsAsSpace", "A collection of Environment Panels that represent a closed space")]
-        [Input("tolerance", "Distance tolerance for planar checks, default set to BH.oM.Geometry.Tolerance.Distance")]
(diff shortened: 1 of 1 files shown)


Environment_Engine/Query/FloorGeometry.cs
```diff
@@ -42,8 +42,9 @@ public static partial class Query
 
         [Description("Returns the floor geometry of a space represented by Environment Panels as a BHoM Geometry Polyline")]
         [Input("panelsAsSpace", "A collection of Environment Panels that represent a closed space")]
+        [Input("tolerance", "Distance tolerance for planar checks, default set to BH.oM.Geometry.Tolerance.Distance")]
```
FraserGreenroyd: The tolerance in this instance is actually for the `BooleanUnion` of the floor lines, rather than planar checks :smile:

PR #1402, opened 2020-01-07
Physical_Geometry_Engine: Add `Normal` method for IFramingElement

### LINKED ISSUE #1401: Physical_Engine: Add `Normal` method for IFramingElement
#### Description:

IFramingElement need a method to get the normal. 

- For a Linear element the same principle as for a bar should be used
- For a planar curved centrecurve the Normal should be taken as the plane that the curve can be fitted through. 
- For nonplanar curved it gets a bit more tricky. Can leave this case as not implemented for now until a convention has been decided. @rwemay

kThorsager:
Closes #1401 
Adds a method for getting the Normal (Z-axis) of a IFramingElement and a central method in the Geometry_Engine which mirrors the existing one in the Structure_Engine.

#### Current snags:
Named the file in Geometry_Engine `ElementNormal` since it already existed a `Normal.cs` with other kinds of normal, but it still doesn't feel like the right place for it.

The `orientationAngle` should rotate the default normal around an axis, and for a curved element, is that the initial direction, start/end point or something else? Almost feels like one would like to be able to query the normal at every point of the element.

A very low amount of change to a curved element can flip the default normal:
image
@IsakNaslundBh That feels very unsafe, it is however technically not a problem yet because... (turns out I forgot `PolyCurves` which will also have this problem)

For some reason `FitPlane()` works fine for `NurbCurve`, but `IsLinear()` and `IsPlanar()` it is'nt implemented. (It is also not implemented for `IStartDir()` which is more understandable, so that would also be an hold-up if we go with that route), so this method will not run with `NurbCurve`. Is this a case of not sure of how to solve it right or haven't done it yet @pawelbaran 

### Test files

https://burohappold.sharepoint.com/:u:/s/BHoM/EWDDw9vvDOBPmkDCdsU_ZEUBlqbWmfcoJymc3IVqkTBehg?e=JysInw

### Changelog

- Adds `Normal(ICurve curve, double orientationAngle)` in the Geometry_Engine

### DIFF
--- a/Geometry_Engine/Query/ElementNormal.cs
+++ b/Geometry_Engine/Query/ElementNormal.cs
@@ -60,9 +60,11 @@ namespace BH.Engine.Geometry
             }
             else if (curve.IIsPlanar())
             {
-                Vector tan = curve.IStartDir();   // Is this how we should define it?
+                Engine.Reflection.Compute.RecordError("The normal for non-linear elements is not implemented");
+                return null;
+                //Vector tan = curve.IStartDir();   // Is this how we should define it?
                 //Vector tan = framingElement.Location.IEndPoint() - framingElement.Location.IStartPoint();
-                return curve.IFitPlane().Normal.Rotate(orientationAngle, tan);    // The normal could potentially flip by moving some control points
+                //return curve.IFitPlane().Normal.Rotate(orientationAngle, tan);    // Warning: The normal can flip by moving some control points
             }
             else
             {
--- a/Geometry_Engine/Query/ElementNormal.cs
+++ b/Geometry_Engine/Query/ElementNormal.cs
@@ -24,6 +24,7 @@ using BH.oM.Geometry;
 using BH.oM.Reflection.Attributes;
 using System;
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.Linq;
 
 namespace BH.Engine.Geometry
@@ -34,7 +35,10 @@ namespace BH.Engine.Geometry
         /*** Public Methods -One dimentional Element Normal*/
         /***************************************************/
 
-        public static Vector Normal(this ICurve curve, double orientationAngle)
+        [Description("Gets the Normal/local Z-axis from a element")]
+        [Input("curve","The curve to evaluate the Normal from")]
+        [Input("orientationAngle", "How much the normal is rotated about the curves axis in radians")]
+        public static Vector ElementNormal(this ICurve curve, double orientationAngle)
         {
             if (curve.IIsLinear())
             {
--- a/Geometry_Engine/Query/ElementNormal.cs
+++ b/Geometry_Engine/Query/ElementNormal.cs
@@ -1,0 +1,91 @@
+/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2020, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 
+ * GNU Lesser General Public License for more details.                          
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using BH.oM.Geometry;
+using BH.oM.Reflection.Attributes;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace BH.Engine.Geometry
+{
+    public static partial class Query
+    {
+        /***************************************************/
+        /*** Public Methods -One dimentional Element Normal*/
+        /***************************************************/
+
+        public static Vector Normal(this ICurve curve, double orientationAngle)
+        {
+            if (curve.IIsLinear())
+            {
+                Point p1 = curve.IStartPoint();
+                Point p2 = curve.IEndPoint();
+
+                Vector tan = (p2 - p1).Normalise();
+                Vector normal;
+
+                if (!IsVertical(p1, p2))
+                {
+                    normal = Vector.ZAxis;
+                    normal = (normal - tan.DotProduct(normal) * tan).Normalise();
+                }
+                else
+                {
+                    Vector locY = Vector.YAxis;
+                    locY = (locY - tan.DotProduct(locY) * tan).Normalise();
+                    normal = tan.CrossProduct(locY);
+                }
+
+                return normal.Rotate(orientationAngle, tan);
+            }
+            else if (curve.IIsPlanar())
+            {
+                Vector tan = curve.IStartDir();   // Is this how we should define it?
+                //Vector tan = framingElement.Location.IEndPoint() - framingElement.Location.IStartPoint();
+                return curve.IFitPlane().Normal.Rotate(orientationAngle, tan);    // The normal could potentially flip by moving some control points
+            }
+            else
+            {
+                Engine.Reflection.Compute.RecordError("The normal for non-planar elements is not implemented");
+                return null;
+            }
+
+        }
+
+        /***************************************************/
+        /**** Private Methods                           ****/
+        /***************************************************/
+        
+        private static bool IsVertical(Point p1, Point p2)
+        {
+            double dx = p1.X - p2.X;
+            double dy = p1.Y - p2.Y;
+
+            return Math.Sqrt(dx * dx + dy * dy) < 0.0001;
+        }
+
+        /***************************************************/
+
+    }
+}
+
--- a/Geometry_Engine/Query/ElementNormal.cs
+++ b/Geometry_Engine/Query/ElementNormal.cs
@@ -67,9 +67,6 @@ namespace BH.Engine.Geometry
             {
                 Engine.Reflection.Compute.RecordError("The normal for non-linear elements is not implemented");
                 return null;
-                //Vector tan = curve.IStartDir();   // Is this how we should define it?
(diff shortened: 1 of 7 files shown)


IsakNaslundBh:
Thanks @kThorsager . Some initial comments:

> Adds a method for getting the Normal (Y-axis) of a IFramingElement and a central method in the Geometry_Engine which mirrors the existing one in the Structure_Engine.

The normal should relate to the local Z-axis, not the local Y-axis, just as an initial comment.

> The orientationAngle should rotate the default normal around an axis, and for a curved element, is that the initial direction, start/end point or something else? Almost feels like one would like to be able to query the normal at every point of the element.

I think you are right in this actually, thinking about it. Could limit this to line based elements for now and have a check for the normal in the method for the `IFramingElement` in the Physical_Engine. I know this is just kicking the can down the road, but to unblock this for what you need it for, it could be a "for now" solution.

> For some reason FitPlane() works fine for NurbCurve, but IsLinear() and IsPlanar() it is'nt implemented. (It is also not implemented for IStartDir() which is more understandable, so that would also be an hold-up if we go with that route), so this method will not run with NurbCurve. Is this a case of not sure of how to solve it right or haven't done it yet @pawelbaran

Would not worry to much about this not working for Nurbs yet, as they are generally not that well supported in the engine.

kThorsager:
Can build my repos now, and didn't remove, only deprecated methods, so I feel fairly confident that I haven't broken anything

Geometry_Engine/Query/ElementNormal.cs
```diff
@@ -0,0 +1,85 @@
...
+namespace BH.Engine.Geometry
+{
+    public static partial class Query
+    {
+        /***************************************************/
+        /*** Public Methods -One dimentional Element Normal*/
+        /***************************************************/
+
+        [Description("Gets the Normal/local Z-axis from a element")]
+        [Input("curve","The curve to evaluate the Normal from")]
```
FraserGreenroyd: Need as space after the comma here between the two elements

Geometry_Engine/Query/ElementNormal.cs
```diff
@@ -0,0 +1,85 @@
...
+        [Input("orientationAngle", "How much the normal is rotated about the curves axis in radians")]
+        public static Vector ElementNormal(this ICurve curve, double orientationAngle)
```
FraserGreenroyd: Need an `output` attribute

Geometry_Engine/Query/ElementNormal.cs
```diff
@@ -0,0 +1,85 @@
...
+                    normal = tan.CrossProduct(locY);
+                }
+
+                return normal.Rotate(orientationAngle, tan);
+            }
+            else if (curve.IIsPlanar())
+            {
+                Engine.Reflection.Compute.RecordError("The normal for non-linear elements is not implemented");
+                return null;
+                //Vector tan = curve.IStartDir();   // Is this how we should define it?
```
FraserGreenroyd: If this code isn't needed, delete it.

Geometry_Engine/Query/IsVertical.cs
```diff
@@ -0,0 +1,54 @@
...
+
+namespace BH.Engine.Geometry
+{
+    public static partial class Query
+    {
+        /***************************************************/
+        /**** Private Methods                           ****/
+        /***************************************************/
+
+        public static bool IsVertical(this Line line)
```
FraserGreenroyd: Needs `Description`, `Input`, `Output` documentation

Geometry_Engine/Query/IsVertical.cs
```diff
@@ -0,0 +1,54 @@
...
+        /***************************************************/
+        /**** Private Methods                           ****/
+        /***************************************************/
+
+        private static bool IsVertical(Point p1, Point p2)
+        {
+            double dx = p1.X - p2.X;
+            double dy = p1.Y - p2.Y;
+
+            return Math.Sqrt(dx * dx + dy * dy) < 0.0001;
```
FraserGreenroyd: Is this a tolerance `0.0001` or part of a pre-determined calculation? If it's a tolerance, I'd advise we expose it as an input and give it the default from `BH.oM.Geometry.Tolerance`

Physical_Engine/Query/Normal.cs
```diff
@@ -0,0 +1,63 @@
...
+namespace BH.Engine.Physical
+{
+    public static partial class Query
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Returns the normal of a FramingElement, which would be the Z-axis in the local coordinate syetem")]
+        [Input("framingElement", "A FramingElement")]
```
FraserGreenroyd: This input description could use some work, it won't provide anything greater than the default in its current state.

Physical_Engine/Query/Normal.cs
```diff
@@ -0,0 +1,63 @@
...
+        [Output("normal", "The FramingElements normal vector")]
+        public static Vector Normal(this IFramingElement framingElement)
+        {
+            double orientationAngle;
```
FraserGreenroyd: Give the explicit default here rather than the implicit.

Physical_Engine/Query/Normal.cs
```diff
@@ -0,0 +1,63 @@
...
+            if (!(framingElement.Property is ConstantFramingProperty))
+            {
+                Reflection.Compute.RecordWarning("No ConstantFramingProperty found, OrientationAngle set as 0");
+                orientationAngle = 0;
```
FraserGreenroyd: Giving the explicit default above means you can delete this line.

Structure_Engine/Query/Normal.cs
```diff
@@ -34,29 +36,12 @@ public static partial class Query
         /**** Public Methods                            ****/
         /***************************************************/
 
+        [Description("Returns the bars local Z-axis")]
+        [Input("bar","The relevant Bar")]
```
FraserGreenroyd: Space needed after comma.
What is the 'relevant' bar? Maybe a better description for the input is needed?

Structure_Engine/Query/Normal.cs
```diff
@@ -34,29 +36,12 @@ public static partial class Query
...
+        [Output("normal","Vector representing the bars local Z-axis")]
```
FraserGreenroyd: Space needed after the comma.

Geometry_Engine/Query/IsVertical.cs
kThorsager: This is copied code (from the BHoM), and it is defined in the [documentation](https://github.com/BHoM/documentation/wiki/BHoM-Structural-Conventions). Or do we want to change it @IsakNaslundBh ?

FraserGreenroyd: The documentation says:
 > A bar is vertical if it's projected length to the horizontal plane is less than 0.001

Whereas your code says `< 0.0001` (an extra 0) - if it's to match the documentation then one of them is wrong (or I'm looking at the wrong bit of documentation?) so might need to change one of them.

If this is a structural convention then that would be ok for when it was in the Structure_Engine, but if this is now coming into the more common Geometry_Engine I think exposing that parameter would be more suitable? @IsakNaslundBh - thoughts?

kThorsager: Would guess that it's a typo in the documentation since it also says 0.1mm directly contradicting it self, but does match the code. But would also like some thoughts @IsakNaslundBh

IsakNaslundBh: > Is this a tolerance 0.0001 or part of a pre-determined calculation? If it's a tolerance, I'd advise we expose it as an input and give it the default from BH.oM.Geometry.Tolerance

Documentation is wrong. Will fix. Good spot, thanks! 👍 

> If this is a structural convention then that would be ok for when it was in the Structure_Engine, but if this is now coming into the more common Geometry_Engine I think exposing that parameter would be more suitable? @IsakNaslundBh - thoughts?

There are multiple conventions for this, and it varies across software packages etc. The important thing is to have one singly defined convention somewhere. I am not particularly attached to the current convention we have and be fine to change it to something else if something is suggested, but as I said, we need it defined somewhere, and only once.

The important part is that this _needs_ to be consistent, _always_. That is the reason for the hard coded value, rather than having it exposed. We could potentially solve this by exposing the value in geometry engine and hard coding it in the physical and structure engine, but that would mean typing it out twice, and hence have the potential of them getting out of sync, which is something I really want to avoid.

IDK if this is enough for you @FraserGreenroyd ? If not we might have to think of another solution to this whole PR.

FraserGreenroyd: > IDK if this is enough for you @FraserGreenroyd ? If not we might have to think of another solution to this whole PR.

I'm think this PR solves the problem best for now so I don't think we should spend more time on another solution in itself.

I am happy with a 1mm tolerance being the standard and hard-coding it (at least for now) if we're happy it meets the criteria required given its new location (Geometry_Engine rather than Structure_Engine). To avoid hard-coding in Physical/Structure engines though, we could do a tolerance input which has a default input of `0.001` but this of course hides that tolerance from the code side on Physical/Structural (though this is no different to the current implementation).

Ultimately, I'm not too concerned as I've raised my query and I feel this satisfactorily answers it for now, so if you're happy with this implementation for now @IsakNaslundBh then I'm happy to support it :smile:

Geometry_Engine/Query/IsVertical.cs
```diff
@@ -0,0 +1,59 @@
...
+{
+    public static partial class Query
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Evaluates whether a line is vertical from its start- and endpoints projected 2d-distance")]
+        [Input("line", "The line to determine the verticality of")]
+        [Output("isVertical", "If it is vertical")]
```
FraserGreenroyd: Can this be improved? Not sure this is the most beneficial of output descriptions? @IsakNaslundBh thoughts?

al-fisher: Good to link to our agreed definition of what vertical means - i.e. tolerances etc. on wiki perhaps

IsakNaslundBh: If this can be fixed, Linking over to the Wiki, then I am Happy to approve and merge this PR

kThorsager: Sure, but it is a bit strange with a link to the structure wiki in the geometry engine

IsakNaslundBh: > Sure, but it is a bit strange with a link to the structure wiki in the geometry engine

Think that is fine, at least for now. BEnv will need to have those kinds of things defined for things like ducts etc at one point as well, and once that happens, we could move the wiki out of the structures convention page.

PR #1423, opened 2020-01-21
NotExposedAttribute

kThorsager:
### NOTE: Depends on 

https://github.com/BHoM/BHoM/pull/662
   
Closes https://github.com/BHoM/BHoM/issues/661

### Test files

Test any file to make sure it doesn't break anything.
But as this PR is not for setting anything as `NotExposed` no proper test file will be provided for that, instead set a current method as `NotExposed` and test that it disappeared from the UI


alelom:
@al-fisher would you mind having a look and a test go too?

al-fisher:
See conversation here https://github.com/BHoM/BHoM/pull/662

PR #1429, opened 2020-01-23
Geometry_Engine: Implemented Normal, IsPlanar and FitPlane for PlanarSurface

### LINKED ISSUE #1428: Geometry_Engine: Implement missing methods for PlanarSurface
#### Description:
A few obvious methods missing for `PlanarSurface`:
- `IsPlanar`
- `FitPlane`
- `Normal`

pawelbaran:

Closes #1428

### Test files

No test files as the methods are mere wrappers of others. One thing that came across my mind: should we implement explicit `IsPlanar` check for `PlanarSurface`, or is it OK to _assume_ that it has been constructed correctly (current approach, returning `true` without any computations)?

### Changelog
- `IsPlanar`, `FitPlane` and `Normal` methods have been implemented for `BH.oM.Geometry.PlanarSurface`

### DIFF
--- a/Geometry_Engine/Compute/FitPlane.cs
+++ b/Geometry_Engine/Compute/FitPlane.cs
@@ -149,6 +149,16 @@ namespace BH.Engine.Geometry
 
 
         /***************************************************/
+        /**** Public Methods - Surfaces                 ****/
+        /***************************************************/
+
+        public static Plane FitPlane(this PlanarSurface surface, double tolerance = Tolerance.Distance)
+        {
+            return IFitPlane(surface.ExternalBoundary, tolerance);
+        }
+
+
+        /***************************************************/
         /**** Public Methods - Interfaces               ****/
         /***************************************************/
 
--- a/Geometry_Engine/Query/IsPlanar.cs
+++ b/Geometry_Engine/Query/IsPlanar.cs
@@ -107,7 +107,14 @@ namespace BH.Engine.Geometry
 
 
         /***************************************************/
-        /**** Public Methods - Surfaces                ****/
+        /**** Public Methods - Surfaces                 ****/
+        /***************************************************/
+
+        public static bool IsPlanar(this PlanarSurface surface, double tolerance = Tolerance.Distance)
+        {
+            return true;
+        }
+
         /***************************************************/
 
         public static bool IsPlanar(this Extrusion surface, double tolerance = Tolerance.Distance)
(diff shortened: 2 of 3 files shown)


pawelbaran:
@IsakNaslundBh @epignatelli @al-fisher do you have thoughts on the planarity check case?

al-fisher:
Thanks for the nudge.
Looking at this now I actually this that is a very good question.
At the moment you can easily construct an "invalid" `PlanarSurface` 

So I think there is an argument to change `PlanarSurface` to `Immutable` (bit like the `BoundaryRepresentaion`

We should do either the idea above (and keep your simple check) or make the IsPlanar a formal validation.

I currently am considering the case for making `PlanarSurface` Immutable )with a create method that validate planarity at time of creation) 
@pawelbaran what do you think?

pawelbaran:
I like the idea @al-fisher, did not think of that! Just check it once at the beginning and never bother any more :)

al-fisher:
Yep! Good way of putting it! 😄

rwemay:
Yes, like it, believe we should make immutable and check on creation - fits with self conflicting objects rule. Also much better to know upstream/at creation than downstream. One thing, though... tolerance compatibility could be an issue where the default or constructed tolerance is less onerous than a downstream workflow.

al-fisher:
> could be an issue where the default or constructed tolerance is less onerous than a downstream workflow.

I think it is fair to say - all checks (and therefor definitions of objects as appropriate) are assumed to be to standard BHoM tolerances.
If downstream you have more onerous requirements you should recheck with those more stringent tolerances at that point in the workflow.
How about that? 
Hopefully that logic can apply more generally

rwemay:
Yep. Was going to say the answer is ‘both’ so the downstream check is only required if you’re not matching BHoM tolerance. Is similar to the isvertical discussion.

pawelbaran:
@al-fisher @rwemay I implemented the tolerance value check in `IsPlanar` - once we get this one merged, I will tackle [#676](https://github.com/BHoM/BHoM/issues/676) both on oM and Engine side.

pawelbaran:
Your point is valid @kThorsager - even though we do not support modification of nurbs, we do aim for allowing conversions, which require planarity checks sometimes. Therefore, I believe it is worth changing the method following your suggestion.

pawelbaran:
Or, as an afterthought, after discussion with @kThorsager: deriving normal from a `NurbsCurve` might be more tricky than just a cross product when we take weights into account, so I would rather not bypass that even for planar curves.

Especially taken into account that for converts `IsPlanar` query should be enough, normal is not needed. Ergo, I would stick to the current form of this PR.

al-fisher:
Thanks @pawelbaran @kThorsager all makes sense

also on 

> once we get this one merged, I will tackle #676 both on oM and Engine side.

Perfect. 👍

pawelbaran:
Cool, thanks! Shall we merge then? @al-fisher @FraserGreenroyd

PR #1436, opened 2020-01-28
Serialiser_Engine: Nested objects with headers supported using iterative approach

### LINKED ISSUE #1435: Serialiser_Engine: Serialization fails if data is nested as per World Bank response data
#### Description:
-If the json being serialized contains separate header object data it throws a casting error

#### How to replicate:
-This can be replicated when serializing a nested json response, ie when using HTTP toolkit to get a response from the World Bank Open Data and serialize in the test script.

#### Expected behaviour:
-Serialization should work, returning separate objects for the header info and the data objects.

#### Test file(s):
Test file attached 
[200127_HTTPTest-Adapter.zip](https://github.com/BHoM/BHoM_Engine/files/4124559/200127_HTTPTest-Adapter.zip)

enarhi:

Closes #1435 

-Supports json serialization where json is nested with header or other data

### Test files
[200127_HTTPTest-Adapter.zip](https://github.com/BHoM/BHoM_Engine/files/4124570/200127_HTTPTest-Adapter.zip)

### DIFF
--- a/Serialiser_Engine/Convert/Json.cs
+++ b/Serialiser_Engine/Convert/Json.cs
@@ -91,7 +91,7 @@ namespace BH.Engine.Serialiser
                 {
                     CustomData = new Dictionary<string, object>()
                     {
-                        {"Objects", objects}
+                        { "Objects", array.Select(b => b.IsBsonDocument ? Convert.FromBson(b.AsBsonDocument) : Convert.FromJson(b.ToString())).ToList() }
                     }
                 };
             }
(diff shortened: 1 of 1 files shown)


Serialiser_Engine/Convert/Json.cs
```diff
@@ -77,16 +77,25 @@ public static object FromJson(string json)
...
+                {
+                    object obj = FromJson(array[i].ToString());
+                    objects.Add(obj);
+                }
+                return new BH.oM.Base.CustomObject()
                 {
                     CustomData = new Dictionary<string, object>()
                     {
-                        { "Objects", array.Select(b => Convert.FromBson(b.AsBsonDocument)).ToList() }
+                        {"Objects", objects}
```
epignatelli: Thanks @enarhi for spotting it.
I suppose we can just change this line if that's the issue, no?
```suggestion
                        { "Objects", array.Select(b => Convert.FromJson(b.ToString())).ToList() }
```

enarhi:
Oops, correct, fixed now.

PR #1463, opened 2020-02-06
Reflection_Engine: Extracting descriptions of properties for Create methods

### LINKED ISSUE #1462: Reflection_Engine: Expose property attributes in create methods
#### Description:

We have some create methods that have some logic in them, but where the rest is simple property assignments. An example is the Bar create that accepts a Line. here the only difference between a straight property assignment is that the line is used to grab out the start and end node positions, the rest goes straight through.

For these types of methods is would be good to be able to pick up descriptions from the base class, and not have to add new input attributes to describe them.

IsakNaslundBh:
### Depends on:
https://github.com/BHoM/BHoM/pull/695

Closes #1462 

Makes use of new InputFromProperty attribute to grab descriptions from the oM. Lets you do for example:
```
        [InputFromProperty("property","SectionProperty")]
        [InputFromProperty("orientationAngle")]
        [InputFromProperty("release")]
        [InputFromProperty("feaType","FEAType")]
        public static Bar Bar(Line line, ISectionProperty property = null, double orientationAngle = 0, BarRelease release = null, BarFEAType feaType = BarFEAType.Flexural, string name = "")
        {          
            return Bar(Node(line.Start), Node(line.End), property, orientationAngle, release, feaType, name);
        }
```
### Test files

Add a Create.Bar component accepting a Line as its first input. OrientationAngle and Release should have descriptions grabbed from the bar class.

### DIFF
--- a/Reflection_Engine/Query/Description.cs
+++ b/Reflection_Engine/Query/Description.cs
@@ -73,23 +73,18 @@ namespace BH.Engine.Reflection
             }
             else
             {
-                //If no inputs are provided, try to grab description from object for Create methods
+                //If no input descs are found, check if inputFromProperty descs can be found
+                IEnumerable<InputFromProperty> inputFromPropDesc = parameter.Member.GetCustomAttributes<InputFromProperty>().Where(x => x.InputName == parameter.Name);
+
+                //Only valid for engine type methods
                 MethodInfo methodInfo = parameter.Member as MethodInfo;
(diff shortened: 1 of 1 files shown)


adecler:
See comment here: https://github.com/BHoM/BHoM_Engine/issues/1462#issuecomment-583225451
Only my preference but that's one thing for the majority to decide.

FraserGreenroyd:
Was just giving this a quick test as @adecler merged - but all good from me, played with a couple of Environment objects and worked as expected :smile:

PR #1465, opened 2020-02-06
Physical_Engine: Volume Query Method for Columns

### LINKED ISSUE #1464: Physical_Engine: Add Method for Calculating Volume of Columns
#### Description:
Add method for calculating volume of columns using the length and the Constant Framing Property Profile curves

image

kayleighhoude:
Method for calculating the volume based on the length and the section profile area

Closes #1464

### Test files
image

https://burohappold.sharepoint.com/:u:/r/sites/Direction/designtechnology/Script%20Library/00312_BHoM%20Life%20Cycle%20Assessment%20Methods/Calculate%20GWP_Columns.gh?csf=1&e=TfAP8z

### DIFF
--- a/Physical_Engine/Query/Volume.cs
+++ b/Physical_Engine/Query/Volume.cs
@@ -1,0 +1,67 @@
+﻿/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2020, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 
+ * GNU Lesser General Public License for more details.                          
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+using BH.oM.Reflection.Attributes;
+using System.ComponentModel;
+using BH.oM.Physical.Elements;
+using BH.oM.Geometry.ShapeProfiles;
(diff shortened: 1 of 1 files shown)


Physical_Engine/Query/Volume.cs
```diff
@@ -0,0 +1,67 @@
...
+using BH.Engine.Reflection;
+
+namespace BH.Engine.Physical
+{
+    public static partial class Query
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+        [Description("Returns the volume of a column")]
```
IsakNaslundBh: Would expand the description here a bit to say what the method is doing (and thereby not doing). What I mean by this is stating that this is simply taking the length of the centre line times the area of the profile, nothing more, nothing less. This is so that it is clear to everyone that this will not account for cutouts, connections, plates, bolts etc etc.

Physical_Engine/Query/Volume.cs
```diff
@@ -0,0 +1,67 @@
...
+        [Input("column", "A physical oM column object")]
+        [Output("volume", "The total volume of the column object")]
+        public static double Volume (this Column column)
```
IsakNaslundBh: Could potentially change the method to accept `IFramingElement` instead of just column. Then this method could be used on Beams as well!

Physical_Engine/Query/Volume.cs
```diff
@@ -0,0 +1,67 @@
...
+            List<ICurve> edges = prop.Profile.Edges.ToList();
+            List<PolyCurve> perimeters = BH.Engine.Geometry.Compute.IJoin(edges);
+
+            List<double> areas = perimeters.Select(x => BH.Engine.Geometry.Query.Area(x)).ToList();
+
+            double maxArea = areas.Max();
+            areas.Remove(maxArea);
+
+            if(areas.Count > 0)
+                maxArea -= areas.Select(x => maxArea - x).ToList().Sum();
```
IsakNaslundBh: Think this area calculation will be fine for 99% of cases (if not more).

Can see a few cases, such as discontinuous sections (double angles/double channels etc) where this will not work, but could be fine to get this merged as is to unblock and in wait for the updated section integration to get merged.

Physical_Engine/Query/Volume.cs
al-fisher: Yep - agreed on this @IsakNaslundBh 
Might we go as far as raising a friendly warning?

al-fisher: Or even call the method - `LinearExtrudedVolume` or something similar?

Physical_Engine/Query/Volume.cs
rwemay: I would go a step further than this... I think we want volume methods that operate on IElements, that way we’ll be writing a very small amount of code to get volumes.

IsakNaslundBh:
Think this can be closed without merging now, as https://github.com/BHoM/BHoM_Engine/pull/1591 has been merged which adds the capabilities and some more, with method `SolidVolume`.

@kayleighhoude leave it for you to check.

PR #1474, opened 2020-02-10
Structure_Engine: Update Embodied Carbon

### LINKED ISSUE #1473: Structure_oM: correct embodied carbon numbers for structural materials
Bring the embodied carbon numbers from the quartz database into the default creates

#### Description:
Currently, create methods default to:
Aluminum: 7.9
Concrete: 0.12
Steel: 1.3
Wood: 0.4

#### Expected behaviour:

These should be:
Aluminum: 6.57
Concrete: 0.242
Steel: 0.938
Wood: -1.15

JosefTaylor:

Closes #1473

### DIFF
--- a/Structure_Engine/Create/MaterialFragment.cs
+++ b/Structure_Engine/Create/MaterialFragment.cs
@@ -45,7 +45,7 @@ namespace BH.Engine.Structure
         [Input("dampingRatio", "Dynamic damping ratio of the material. Will be stored on the material fragment")]
         [Input("embodiedCarbon", "Embodied carbon for the material. Will be stored on the material fragment")]
         [Output("Material", "The created material with a aluminium fragment")]
-        public static Aluminium Aluminium(string name, double E = 70000000000, double v = 0.34, double tC = 0.000023, double density = 2710, double dampingRatio = 0, double embodiedCarbon = 7.9)
+        public static Aluminium Aluminium(string name, double E = 70000000000, double v = 0.34, double tC = 0.000023, double density = 2710, double dampingRatio = 0, double embodiedCarbon = 6.57)
         {
             return new Aluminium()
             {
@@ -71,7 +71,7 @@ namespace BH.Engine.Structure
         [Input("cylinderStrength", "Cylinder strength of the concrete material. Given in [Pa] or [N/m].Will be stored on the material fragment")]
         [Input("embodiedCarbon", "Embodied carbon for the material. Will be stored on the material fragment")]
         [Output("Material", "The created material with a concrete fragment")]
-        public static Concrete Concrete(string name, double E = 33000000000, double v = 0.2, double tC = 0.00001, double density = 2550, double dampingRatio = 0, double cubeStrength = 0, double cylinderStrength = 0, double embodiedCarbon = 0.12)
+        public static Concrete Concrete(string name, double E = 33000000000, double v = 0.2, double tC = 0.00001, double density = 2550, double dampingRatio = 0, double cubeStrength = 0, double cylinderStrength = 0, double embodiedCarbon = 0.242)
         {
             return new Concrete()
             {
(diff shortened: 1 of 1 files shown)


FraserGreenroyd:
@JosefTaylor I have updated the PR title and added the 'bug' label to match the issue for change log purposes :smile:

IsakNaslundBh:
>  would prefer to see the embodied carbon units in the description themselves if we're not ready to use the Quantities engine (I think it needs the attribute adding anyway @al-fisher ?).

This has been added to the oM. See this https://github.com/BHoM/BHoM/pull/686
Once the oM is done and the current Greens theorem PR has been merged, I will comb through the full engine and add inputs and quantities where missing.

> Also still a little bit nervous about hardcoding these defaults. As a silly question.... is it possible/practical to have the method call default databases? (Multistorey structure constructor works this way). Would need to be careful about what we're reporting back to the human. Also should we return a warning where any of the inputs are defaulted for materials, IMHO. So component is always orange unless you've been explicit or loaded from a database.

We could be doing this, if we have a dataset for embodied carbon. We are currently doing it for materials for all section properties (not currently with a warning, but could easily add that).

IsakNaslundBh:
> The compliance reports has shown that the create methods in the file `MaterialFragment.cs` do not match the return type for `MaterialFragment`. To assist in matching what users see in the UIs and they see on the code, Create methods should be housed in a file which is named after the object type they create. In this instance, this would require the fragments to be split into 4 files, named `Aluminium.cs`, `Concrete.cs`, `Steel.cs`, and `Timber.cs` with each create method housed in them. To avoid this needing to be done as a broader issue in the future, could we make this change on this PR? 😄

@JosefTaylor please hold off doing this now, until we have merged the GreenTheorem (might do that today)
Want to avoid merge conflicts, and I am planning to run through the full Structure_Engine anyway and fix all compliance issues as well as adding descriptions.

When doing this these methods might even get removed as they are just simple property initialisers. The only thing I have against that is loosing the default values, as I am a lot more hesitant to populate those in the oM in general (steel might be fine to do this for, as analytical properties are constant, but would not want to have default values for concrete/timber in the oM). 

Kind of touches on @rwemay s point though, if we want defaults at all..

al-fisher:
> I would prefer to see the embodied carbon units in the description themselves if we're not ready to use the Quantities engine (I think it needs the attribute adding anyway @al-fisher ?).

@rwemay @JosefTaylor 
`MassFraction` Dimensionless with units kg/kg has been added already to the Quantities_oM

See https://github.com/BHoM/BHoM/blob/master/Quantities_oM/Attributes/MassFraction.cs

JosefTaylor:
Agreed on (most) counts;

Currently because of the location of these defaults, they're not always set when various toolkits make new materials- Revit for example, comes out with EmbodiedCarbon = 0 for all materials, as it doesn't use this create method. So for that reason I'd definitely support moving them

We now have Quartz database serialized, thanks to Michael Hoehn. This is where my default numbers are now coming from. It would be neat use that database, but it would create a dependency which might not be desired.

Alternately, spending a bit of time on the engine methods @michaelhoehn created could make the defaults superfluous, by auto-selecting an EPD for standard materials.

Also, now that we have default create methods, these ones might be superfluous, as all they do is set every property. Could we encode the defaults in the oM instead of the engine?

As for quantity descriptions, I'm certainly happy to add them.

The important thing for me to remember is that I stuck these in, on a bit of a whim, to get this conversation started- I must not be wedded to any particular implementation!

In summary, I'll hold out till @IsakNaslundBh is finished with Green's Theorem and then we can make the decisions around splitting these methods up, deleting them entirely, migrating the defaults somewhere else, or developing the LCA toolkit with @michaelhoehn.

IsakNaslundBh:
@JosefTaylor will close this one out now, as conflicting with master. Also, given all the comments above, think we should generally revisit the approach to embodied carbon given the work on the LCA toolki. Might mean no longer storing it as a parameter on the structural materials.

PR #1538, opened 2020-03-03
Reflection_Engine: Compliance Update

### LINKED ISSUE #1519: Reflection_Engine - Compliance issues
This report has been generated as part of a close out for the 3.1 milestone compliance and outlines all of the compliance issues currently highlighted by the CI team. I will be undertaking this issue to resolve all compliance issues which do not impact scripts or versioning, with the remainder being moved to the 3.2 milestone for further discussion and resolution.

The following compliance issues have been found for this project. Please see the checklist below.

##### Method name must start with or end with the name of the file
 - [x] in BHoM_Engine\Reflection_Engine\Compute\RecordEvent.cs at line 49, column 27 
 - [x] in BHoM_Engine\Reflection_Engine\Compute\RecordEvent.cs at line 56, column 27 
 - [x] in BHoM_Engine\Reflection_Engine\Compute\RecordEvent.cs at line 63, column 27 
 - [x] in BHoM_Engine\Reflection_Engine\Compute\UnqualifiedName.cs at line 83, column 35 
 - [x] in BHoM_Engine\Reflection_Engine\Modify\SetPropertyValue.cs at line 44, column 33 
 - [x] in BHoM_Engine\Reflection_Engine\Query\Count.cs at line 47, column 26 
 - [x] in BHoM_Engine\Reflection_Engine\Query\Count.cs at line 54, column 26 
 - [x] in BHoM_Engine\Reflection_Engine\Query\ImplementedTypes.cs at line 39, column 33 
 - [x] in BHoM_Engine\Reflection_Engine\Query\TypeList.cs at line 35, column 33 

##### Modify methods must return the same type as their first parameter
 - [ ] in BHoM_Engine\Reflection_Engine\Modify\Cast.cs at line 38, column 22 
 - [ ] in BHoM_Engine\Reflection_Engine\Modify\SetPropertyValue.cs at line 53, column 22 
 - [ ] in BHoM_Engine\Reflection_Engine\Modify\SetPropertyValue.cs at line 102, column 22 
 - [ ] in BHoM_Engine\Reflection_Engine\Modify\SortExtensionMethods.cs at line 43, column 22 

##### Invalid Engine class: Engine classes must be public
 - [x] in BHoM_Engine\Reflection_Engine\Query\ExtensionMethods.cs at line 31, column 4

FraserGreenroyd:
Fixes #1519 

### Test files
N/A

### Additional comments
The issue will be resolved by the changes made. Outstanding issues will be raised in 3.2 compliance reports as discussed.

There are some compliance issues surrounding Modify methods in this engine, which will need addressing as part of a 3.2 discussion as there may be script breakages as a result. Therefore #1519 can close as this completes the 3.1 requirements, and a new compliance report will be generated for 3.2.

There is an incompliance on `BHoMInterfaceList` sitting within `TypeList.cs` - however this causes breaking changes to BHoM_UI and I agree with @adecler that moving to a new file by itself doesn't make sense - so this will be left to be fixed in 3.2


FraserGreenroyd:
I am happy that compliance reports only the issues which we are accepting on this PR so I am happy for this to be merged with that compliance report following review.

Reflection_Engine/Query/BHoMAssemblyList.cs
```diff
@@ -41,16 +41,6 @@ public static List<Assembly> BHoMAssemblyList()
             return m_BHoMAssemblies.Values.ToList();
         }
 
-        /***************************************************/
```
adecler: I think this one is similar to my comment on the BHoM_UI PR regarding partial name matching. IT makes al lot of sense to have both `BHoMAssemblyList` and `AllAssemblyList` under the same file as they are basically dispatching to the `ExtractAllAssemblies` method below. Since the file was named `AssemblyList`, it actualy matches my recommendation of having the method names ending (or starting) with the file name instead of matching it exactly.

Reflection_Engine/Query/Count.cs
```diff
@@ -39,24 +39,6 @@ public static int ICount<T>(this List<T> list)
         {
             return Item(list as dynamic);
         }
-
-        /***************************************************/
-        /**** Public Methods                            ****/
-        /***************************************************/
-
-        public static int Item<T>(this object obj)
```
adecler: I think in this case, the name of the method should have been corrected instead of moving the methods to a new file

Reflection_Engine/Modify/SetPropertyValue.cs
```diff
@@ -36,20 +36,6 @@ public static partial class Modify
         /**** Public Methods                            ****/
         /***************************************************/
 
-        [Description("Set the value of a property with a given name for an object")]
-        [Input("obj", "object to set the value for")]
-        [Input("propName", "name of the property to set the value of")]
-        [Input("value", "new value of the property")]
-        [Output("result", "New object with its property changed to the new value")]
-        public static BHoMObject PropertyValue(this BHoMObject obj, string propName, object value)
```
adecler: In this case, the method should be renamed instead of moved

Reflection_Engine/Query/AllEvents.cs
```diff
@@ -38,18 +38,6 @@ public static List<Event> AllEvents()
             Log log = Query.DebugLog();
             return log.AllEvents;
         }
-
-
-        /***************************************************/
```
adecler: Same thing for here. I do prefer to keep the two methods in a single file called `Events`. It makes the code more readable to have them together on a file and it would be a bit silly to have to use an enum as argument just so we can have a matching method name.

Reflection_Engine/Query/BHoMMethodList.cs
```diff
@@ -45,33 +45,6 @@ public static List<MethodInfo> BHoMMethodList()
             return m_BHoMMethodList;
         }
 
-        /***************************************************/
-
-        public static List<MethodBase> AllMethodList()
```
adecler: Same thing here. clarity of the code is reduced by moving methods in separate files. The file name was allowing for those methods to be together.

Reflection_Engine/Query/BHoMInterfaceList.cs
```diff
@@ -44,49 +44,6 @@ public static List<Type> BHoMInterfaceList()
             return m_InterfaceList;
         }
 
-        /***************************************************/
```
adecler: Same again. Better in one file

FraserGreenroyd:
> Except for the few things that need fixing, there is the topic of partial name matching that we need to discuss. I think the examples here are a strong example of why this should be allowed (especially since we are doing it for create already)

Hi @adecler , I feel that will be something we need to discuss at the end-of-milestone workshops, because the discussion we had in December agreed that only Create methods should have partial naming matching, to aid the findability of code from a UI -> Code basis. With partial naming of, say `Query` methods, we run the risk of a situation where on a UI canvas I can have the following components:

 - `UniquePanels`
 - `SplitPanels`
 - `PanelsByType`
 - `PanelsByPoint`
 - `PanelsByHoppingOnOneLeg`
 - `HoppingPanels`

All contains within a Panels.cs file - but the functionality of each method might be drastically different. We could, and I believe did, argue that perhaps that should be down to the engine owner/toolkit lead to ensure functionality is grouped appropriately into CS files, but our final decision was that the component name seen in the UI (`UniquePanels`) should be matching the file name (`UniquePanels.cs`) so that the 1:1 matching is upheld.

I do agree, as I did with @IsakNaslundBh yesterday, that this can result in some situations where, from a coding perspective, things start to get odd, or reduces readability from the coding viewpoint. However, current BHoM philosophy remains to reduce the barrier for entry for end-users wishing to lift the lid off the 'black box' of the code.

Ultimately, we could argue that some engines are too 'core' or 'framework specific' that users wouldn't want to inspect them, but this starts us on a path of 'one rule for them, and another rule for us' with the obvious question being how do we decide which engines would be exempt and which would not.

Thus, to me, with my coding hat on, it is a pain to split the methods into the corresponding files to have that 1:1 matching. But with my CI hat on, I can see why it makes it easier for the new people to find their way around the code base. This is a very long way of saying I'm sitting on the fence I guess because I can see the arguments for both sides, but am currently learning towards the CI side of my fence with this round of compliance.

Coming back to this PR, this PR resolves the issues of compliance found to the guidelines we agreed upon in December at the start of the 3.1 milestone. But it is your engine, so I see that we have the following options for this PR:

 - Accept it as is to resolve compliance to current guidelines, but risk making more changes if we end up changing guidelines in the upcoming workshops
 - Make some changes (such as method renaming rather than moving) and then accept as above bullet point
 - Close the PR and the 3.1 compliance report, accept this engine is incompliant to the tune of the notes in #1519 , and await a 3.2 report after the workshop

I am happy for you to choose which option you'd like to take forward for this PR @adecler - but I would prefer that the PR does not stay open past the sprint unless it is being worked on (under point 2) :smile:

FraserGreenroyd:
@adecler I have rerun the report based on our call this morning and subsequent agreed changes which were resolved in https://github.com/BHoM/Test_Toolkit/issues/176

The new report in #1519 is the latest state. I have continued with not making changes to the modify methods as per the original PR notes. Let me know what you think :smile:

Reflection_Engine/Query/Count.cs
adecler: Same as before, this should not be moved but renamed instead

FraserGreenroyd:
@adecler I have reversed those changes for this PR then as both are sets of public methods which may (though I suspect the odds are low but they're not 0) be being used in scripts and so would require a broader deprecation/versioning change. On that basis I am happy to leave `PropertyValue` and `Item` (that should be `Count`) for this PR/compliance run and we can address in 3.2 if you're happy with that :smile:

adecler:
@FraserGreenroyd , works for me. Although I am not sure why you removed the separator comments on the Query.Count file.
FYI, `SetPropertyValue` is only used in Excel and SharePoint toolkit (do I see a pattern ?).
`Item` is a bit more difficult to track down due to GitHub not supporting '(' as a valid character in searches.

FraserGreenroyd:
> @FraserGreenroyd , works for me. Although I am not sure why you removed the separator comments on the Query.Count file.

Woops - I'll add them back now, 2 seconds!

FraserGreenroyd:
> @FraserGreenroyd , works for me. Although I am not sure why you removed the separator comments on the Query.Count file.

Should be fixed now :smile:

PR #1543, opened 2020-03-04
Physical_Engine: Add IElementXD Extension methods to physical objects

kThorsager:
#### NOTE: Depends on 
https://github.com/BHoM/BHoM/pull/730
   
https://github.com/BHoM/BHoM/issues/605

### Test files
https://burohappold.sharepoint.com/:u:/s/BHoM/Ef4UbFLZmdhIuyBg0NTDu9sBM3wxwcMkojtBIn8gAWLtWA?e=Wy3eXq

### Changelog
- Fixes a bug in Create Wall where resulting wall was self-intersecting
- Adds `Geometry()` for Physical ISurface which allows Rhino to automatically display the geometry


Physical_Engine/Query/InternalElements2D.cs
```diff
@@ -0,0 +1,64 @@
...
+        [Output("openings", "The IOpenings of the ISurface.")]
+        public static List<IElement2D> InternalElements2D(this oM.Physical.Elements.ISurface surface)
+        {
+            return new List<IElement2D>(surface.Openings);
+        }
+
+        /***************************************************/
+
+        [Description("Gets the internal 2D elements of an IOpening.")]
+        [Input("opening", "The IOpening to get the internal elements from.")]
```
FraserGreenroyd: This should be surface per compliance report

kThorsager: almost agree

kThorsager:
The extension of the folder structure is due to compliance complaints and is inspired by the folder structure in Structure_Engine.

The test file won't have any errors on this branch, but will formalise the error reporting in the file.

kThorsager:
Testfile is updated,

Then on the file structure, think the only other alternative is changing the filename from `NewElement1D.cs` to `NewIElement1D.cs` or `IElement1D.cs` and I kind of don't like any of the alternatives or the current one, but this is how it was done elsewhere... 
Perhaps some salvation from @FraserGreenroyd ?

FraserGreenroyd:
> Then on the file structure, think the only other alternative is changing the filename from `NewElement1D.cs` to `NewIElement1D.cs` or `IElement1D.cs` and I kind of don't like any of the alternatives or the current one, but this is how it was done elsewhere...

The other engines were moved the way they were to avoid conflicts in versioning as we made things compliant. If this is a new set of files, that don't exist previously, I would be happy with us doing a bit of something different that keeps compliance in a different (more accurate) way. Something to discuss perhaps.

IsakNaslundBh:
> Testfile is updated,
> 
> Then on the file structure, think the only other alternative is changing the filename from `NewElement1D.cs` to `NewIElement1D.cs` or `IElement1D.cs` and I kind of don't like any of the alternatives or the current one, but this is how it was done elsewhere...
> Perhaps some salvation from @FraserGreenroyd ?

Agreed on this. I used the same approach in structures, inspired by @FraserGreenroyd .

I think this is all quite awful, but might highlight that the methods themselfs should be moved, rather than the create compliance being to strict in this case. Could def argue that all of these "New" methods actually should be Queries, not Creates. If we see Create methods as constructors, then these should most definitely not be Creates at least, and seeing it as "I get a new compatible type of a lower level object" rather than "I create a new lower level object, based on this higher level obejct" I think actually even semantically makes more sense.

This would ofc need to be a coordinated code change, so if everyone else is happy, I will raise an issue for it, and would probably do it after this PR has been merged, as changes required in other repos that currently have PRs open.

pawelbaran:
@kThorsager the test is more readable than before, but still it is _messy_. Think of someone else thinking of tweaking or checking it in, say a year from now - this may be really hard taken into account the wires flying backwards or Explodes being left here and there. A few more comments/groups and slight reshuffling of the components will help a lot.

The time I am spending on complaining about your test files and PR descriptions takes more time than it would be for you to just do it properly from the beginning. **Please** be more neat in your work for the good of us all.

Concerning the code itself, after rebuilding BHoM, BHoM_Engine and BHoM_UI, I am getting the following error:

image

Am I missing something?

kThorsager:
> > Testfile is updated,
> > Then on the file structure, think the only other alternative is changing the filename from `NewElement1D.cs` to `NewIElement1D.cs` or `IElement1D.cs` and I kind of don't like any of the alternatives or the current one, but this is how it was done elsewhere...
> > Perhaps some salvation from @FraserGreenroyd ?
> 
> Agreed on this. I used the same approach in structures, inspired by @FraserGreenroyd .
> 
> I think this is all quite awful, but might highlight that the methods themselfs should be moved, rather than the create compliance being to strict in this case. Could def argue that all of these "New" methods actually should be Queries, not Creates. If we see Create methods as constructors, then these should most definitely not be Creates at least, and seeing it as "I get a new compatible type of a lower level object" rather than "I create a new lower level object, based on this higher level obejct" I think actually even semantically makes more sense.
> 
> This would ofc need to be a coordinated code change, so if everyone else is happy, I will raise an issue for it, and would probably do it after this PR has been merged, as changes required in other repos that currently have PRs open.

The methods are only to be called by Reflection which does not care which namespace it sits in, which means that if we would like the change does not need to happen at once for every project.
The existing methods would of course need deprecation and such but there's not code limitation to not place them in `Query` now. Could create confusion when trying to understand the code however when they are placed in different location depending on Engine

kThorsager:
> @kThorsager the test is more readable than before, but still it is _messy_. Think of someone else thinking of tweaking or checking it in, say a year from now - this may be really hard taken into account the wires flying backwards or Explodes being left here and there. A few more comments/groups and slight reshuffling of the components will help a lot.
> 
> The time I am spending on complaining about your test files and PR descriptions takes more time than it would be for you to just do it properly from the beginning. **Please** be more neat in your work for the good of us all.
> 
> Concerning the code itself, after rebuilding BHoM, BHoM_Engine and BHoM_UI, I am getting the following error:
> 
> image
> 
> Am I missing something?

Sorry @pawelbaran , my bad due to working out the fail conditions based on being on the old branch without thinking about that saving breaks the components.
But I have replaced the entire test file with something which is structured and should be fairly easily modified to be usable for similar PRs

al-fisher:
> Could def argue that all of these "New" methods actually should be Queries, not Creates.


Yes! Afraid agree! Creates should not be extension methods.
But also agree - one to raise issue on and resolve _after_ this PR!

IsakNaslundBh:
merged master into this, as CI installer check complained on https://github.com/BHoM/BHoM/pull/730 . Hope this will resolve the problem.

al-fisher:
I have raised an issue [here](https://github.com/BHoM/BHoM_Engine/issues/1572) to capture the notes above around reclassifying the NewElement Spatial Interface methods as Queries as well as tiding up the folder structures.
To be coordinated across all required namespaces.

PR #1551, opened 2020-03-06
Geometry_Engine: remove the ClipPolylines method

### LINKED ISSUE #1536: Geometry_Engine: ClipPolyline is not compliant and overlapping
#### Broken rules:

The method setup is itself not fully compliant, as it contains the extra Edge class. Also, reading the method description of what the method should be doing, it is overlapping with the BooleanIntersection method for Polylines. 

I think the method itself can be removed in full, and all usages switched to the BooleanIntersection method.

rolyhudson:

Closes #1536 
Removed the ClipPolylines method and updated all calls in AvalueAnalysis to use BooleanIntersection.

Tested changes on AvalueAnalysis where view cones are used to trim projected pitches. Seen below trimming a projected football pitch (black) with a spectators view cone (red) giving the trimmed results (purple)


rolyhudson:
I would defer to your view on version or deprecation - probably should be yes to conform? 
We do lose the opportunity to use the real name of the algorithm Sutherland–Hodgman passing due credit its creators Mr Sutherland and Mr Hodgman 🤣 🤣 
It was implemented in ignorance (by me) of the better more general BooleanIntersection alternative.
So its use is most likely limited to two calls in the Theatron AvalueAnalysis class.

FraserGreenroyd:
@IsakNaslundBh be good to get your thoughts on the above :smile:

IsakNaslundBh:
Could maybe add a Versioning method upgrade for it, for UI purposes, and still jsut delete it from the code?

I supose something like

```
            {
                "BH.Engine.Geometry.Compute.ClipPolylines(Polyline subject, Polyline clip)",
                typeof(BH.Engine.Geometry.Compute).GetMethod("BooleanIntersection", new Type[] { typeof(BH.oM.Geometry.Polyline), typeof(BH.oM.Geometry.Polyline) })
            },
```

Added to the ToNewMethod dictionary in the MethodConverter should do the trick in upgrading anu component using the old CLipPolyline on the canvas (but have not tested).

Would require a dependency on Geometry_oM and Geometry_Engine in the Versioning, but I think that is absolutely fine, and what it is there for?

rolyhudson:
Where are ToNewMethod dictionary in the MethodConverter? I will add this @FraserGreenroyd @IsakNaslundBh both think its a good plan.

IsakNaslundBh:
> Where are ToNewMethod dictionary in the MethodConverter?

All in the Versioning_Toolkit, under BHoMUpgrader31

rolyhudson:
I can't seem to get that too work. Made a gh file with ClipPolylines. Removed the method from BHoM_Engine. Rebuilt the BHoM_Engine Versioning_Toolkit and BHoM_UI
When I run the gh file I get 1. The component is not linked to a method.
Did I miss something?

IsakNaslundBh:
might be that the method extraction I added did not work( just added it untested) could be worth digging into if the GetMethod i suggested works as expected, or if the method needs to be retrieved some other way...

IsakNaslundBh:
Ah, silly mistake on my end. Missed the tolerance input! Should be:

```
            {
                "BH.Engine.Geometry.Compute.ClipPolylines(BH.oM.Geometry.Polyline subject, BH.oM.Geometry.Polyline clip)",
                typeof(BH.Engine.Geometry.Compute).GetMethod("BooleanIntersection", new Type[] { typeof(BH.oM.Geometry.Polyline), typeof(BH.oM.Geometry.Polyline), typeof(double) })
            },
```

Might be the case the wires get unplugged as inputs have different names, but not sure how else to get around it.

rolyhudson:
Thanks @IsakNaslundBh I added the tolerance but still the component is not linking to the method. Other than the Versioning_Toolkit and the BHoM_UI do I need to (re)build anything else?

rolyhudson:
Following the call with @IsakNaslundBh this is now dealt with. In Versioning_Toolkit. See this PR https://github.com/BHoM/Versioning_Toolkit/pull/36

PR #1558, opened 2020-03-10
Architecture_Engine: Add extension methods for Room

FraserGreenroyd:
### NOTE: Depends on 
https://github.com/BHoM/BHoM/pull/742
   
N/A


Architecture_Engine/Modify/SetInternalElements2D.cs
```diff
@@ -0,0 +1,52 @@
...
+        [Description("Assign a new collection of internal 2D elements to an Architecture Room")]
+        [Input("room", "An Architecture Room to update")]
+        [Input("internalElements2D", "A collection of internal 2D elements to assign to the panel")]
+        [Output("room", "The updated Architecture Room")]
+        public static Room SetInternalElements2D(this Room room, List<IElement2D> internalElements2D)
+        {
+            Room rr = room.GetShallowClone() as Room;
+            rr.Perimeter = internalElements2D.Cast<ICurve>().ToList()[0];
+            return rr;
+        }
```
kThorsager: This looks like it's replacing the the rooms perimeter with a new one, which is not the point of the method. 

I'm unsure of how the location point ties into this but as a Room is defined by a perimeter it can never have any internal elements and should simply return the room here

Architecture_Engine/Modify/SetOutlineElements1D.cs
```diff
@@ -0,0 +1,52 @@
...
+        [Description("Assign a new collection of external 1D boundaries to an Architecture Room")]
+        [Input("room", "An Architecture Room to update")]
+        [Input("outlineElements1D", "A collection of outline 1D elements to assign to the Room")]
+        [Output("room", "The updated Architecture Room")]
+        public static Room SetOutlineElements1D(this Room room, List<IElement1D> outlineElements1D)
+        {
+            Room r = room.GetShallowClone() as Room;
+            r.Perimeter = outlineElements1D.Cast<ICurve>().ToList()[0];
+            return r;
+        }
```
kThorsager: at least try to join the incoming IElement1ds and if the Room definition requires it, check for closed-ness

Architecture_Engine/Query/InternalElements2D.cs
```diff
@@ -0,0 +1,55 @@
...
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Returns the internal 2D elements of an Architecture Room")]
+        [Input("room", "An Architecture Room")]
+        [Output("element2D", "A collection of internal 2D elements")]
+        public static List<IElement2D> InternalElements2D(this Room room)
+        {
+            return room.Perimeter.ISubParts().Cast<IElement2D>().ToList();
+        }
```
kThorsager: Still don't think there's any internal elements, return a empty list

A few methods use recursion to work through the element 2d and this will cause stack overflows

Architecture_Engine/Query/NewInternalElement2D.cs
```diff
@@ -0,0 +1,49 @@
...
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Returns an instance of an IElement2D from the given object")]
+        [Input("room", "An Architecture Room object")]
+        [Output("element", "A 2D element object")]
+        public static IElement2D NewInternalElement2D(this Room room)
+        {
+            return new Room();
+        }
```
kThorsager: This method is not needed as there are no internal 2d elements

I should update the documentation

FraserGreenroyd: > I should update the documentation

Yes you should, I used that to do this 😕 but that's why I added you as a reviewer, I'll make some changes asap.

Architecture_Engine/Modify/SetInternalElements2D.cs
kThorsager: Should clarify that attempting to set internal elements to a IElement2d which does not have any should return a error.
i.e. if `internalElements2D` Count isn't zero, there should be an error

FraserGreenroyd:
Made the changes suggested @kThorsager if you can do another comb over :smile:

Architecture_Engine/Modify/SetInternalElements2D.cs
```diff
@@ -0,0 +1,54 @@
...
+{
+    public static partial class Modify
+    {
+        /***************************************************/
+        /****               Public Methods              ****/
+        /***************************************************/
+
+        [Description("Assign a new collection of internal 2D elements to an Architecture Room")]
+        [Input("room", "An Architecture Room to update")]
+        [Input("internalElements2D", "A collection of internal 2D elements to assign to the panel")]
```
kThorsager: this isn't a panel

Architecture_Engine/Modify/SetOutlineElements1D.cs
```diff
@@ -0,0 +1,52 @@
...
+        /***************************************************/
+
+        [Description("Assign a new collection of external 1D boundaries to an Architecture Room")]
+        [Input("room", "An Architecture Room to update")]
+        [Input("outlineElements1D", "A collection of outline 1D elements to assign to the Room")]
+        [Output("room", "The updated Architecture Room")]
+        public static Room SetOutlineElements1D(this Room room, List<IElement1D> outlineElements1D)
+        {
+            Room r = room.GetShallowClone() as Room;
+            r.Perimeter = BH.Engine.Geometry.Compute.Join(outlineElements1D.Cast<Polyline>().ToList())[0];
```
kThorsager: Any specific reason this isn't `ICurve`?

FraserGreenroyd: Because `List<ICurve>` does not have an implementation for `Join` that I could find 😞

FraserGreenroyd: Because it's `IJoin` 🤦‍♂  I'll change that now...

Architecture_Engine/Modify/SetInternalElements2D.cs
FraserGreenroyd: Good spot, thanks!

FraserGreenroyd:
> Works fine now, do however still not know what the property `Point location` is and if it needs to be considered.
> 
> Would like some response on that before I'm properly happy

The point location is an alternative option for the `room` but does not need to be considered for this scope of work.

PR #1565, opened 2020-03-11
Reflection_Engine: Improve support to generic types

adecler:
Provide the Engine side of the corresponding PR in the BHoM_UI.

On top of the couple of suggestions from @IsakNaslundBh  in https://github.com/BHoM/BHoM_UI/issues/45, this contain extra check to prevent stach overflow when making types from generic and a gigantic messy method `MakeGenericFromInputs(MethodInfo method, List<Type> inputTypes)` that creates the proper `MethodInfo` given a generic method definition and a set of input types. That method will certainly need a bit of TLC but should do the trick for now at least.

### Test files
See related PR in BHoM_UI

### Additional comments
- @FraserGreenroyd I blame you for requesting a meeting when I am trying to push this so I will let you fix what needs fixing for the changelog to be correct
- This should be safe enough to merge but if you have any doubts, I would rather merge next quarter to be safe.


Reflection_Engine/Compute/MakeGenericFromInputs.cs
```diff
@@ -0,0 +1,121 @@
...
+
+        public static MethodInfo MakeGenericFromInputs(MethodInfo method, List<Type> inputTypes)
+        {
+            if (!method.IsGenericMethod)
+                return method;
+
+            Type[] genericArguments = method.GetGenericArguments();
+            List<Type> paramTypes = method.GetParameters().Select(x => x.ParameterType).ToList();
+
+            Type t = typeof(List<>).MakeGenericType(new Type[] { typeof(List<>) });
```
IsakNaslundBh: Do not think this parameter is used anywhere?

adecler: indeed

adecler:
Note: New code was merged on the master for the oM since yesterday, preventing this PR to compile successfully and therefore to pass the CI test. I simply created a branch on the oM with the same name as this one and that did the trick to keep CI happy. In this case, you don't want to merge the master into this branch as I see some people doing. It might be worth keeping this in mind when asked for guidance.

@al-fisher, @FraserGreenroyd , @IsakNaslundBh

adecler:
PerimeterLength method removed. I'll be on holiday next week so I'll let you guys merge if you now think this is ready

PR #1571, opened 2020-03-12
Serialiser_Reflection_Engine: Store all engine types and extract them based on type string

### LINKED ISSUE #1569: Serialiser_Engine: Get Engine types based on string without assembly version
#### Description:

To be able to handle versioning of engine methods, need to be able to retrieve types without assembly version.

### LINKED ISSUE #1570: Reflection_Engine: Add method to get all Engine types
#### Description:

Add method to extract all engine types to the reflection Engine.

In support of #1569

IsakNaslundBh:

Closes #1569 
Closes #1570 

### Test files

All done in support of versioning https://github.com/BHoM/Versioning_Toolkit/pull/40

### Changelog

- Extract and store all Engine types
- Add method for creating Engine types from string without assembly version
- Make Serialiser_Engine use the new method to find appropriate type.


IsakNaslundBh:
Putting on do-not-merge tag, as I realised this might clash badly with https://github.com/BHoM/BHoM_Engine/pull/1565

Type serialiser is changed, just a few lines away between the two PRs. This was pushed as a separate commit, so if it is causing issues, can try to roll that back, merge master into this, and then re-apply that change again if this does not causes complaints.

IsakNaslundBh:
> Look ok to me but I would make one minor comment:
> The method is called `MethodTypeList` but the private list is called `m_EngineTypeList`. It would be better to have them ore consistent with each other.

Thanks @adecler . Updated all names to "EngineType"

IsakNaslundBh:
Removing do-not-merge tag now that the generics PR has been merged

PR #1579, opened 2020-03-13
Reflection_Engine: Added set property functionality for custom data

### LINKED ISSUE #1578: Reflection_Engine: Set Custom Data Property
Inability to set property for custom data properties

customdataerror

enarhi:

Closes #1578 

Adeed dictionary functionality to set property so that custom data properties can be set.

### Test files
[SetCustomDataProperty.zip](https://github.com/BHoM/BHoM_Engine/files/4331627/SetCustomDataProperty.zip)

### DIFF
--- a/Reflection_Engine/Modify/SetPropertyValue.cs
+++ b/Reflection_Engine/Modify/SetPropertyValue.cs
@@ -87,7 +87,7 @@ namespace BH.Engine.Reflection
                 IBHoMObject bhomObj = obj as IBHoMObject;
                 if (bhomObj == null) return false;
 
-                if (!(bhomObj is CustomObject))
+                if (!bhomObj.CustomData.ContainsKey(propName))
                     Compute.RecordWarning("The objects does not contain any property with the name " + propName + ". The value is being set as custom data");
 
                 bhomObj.CustomData[propName] = value;
--- a/Reflection_Engine/Modify/SetPropertyValue.cs
+++ b/Reflection_Engine/Modify/SetPropertyValue.cs
@@ -93,6 +93,12 @@ namespace BH.Engine.Reflection
                 bhomObj.CustomData[propName] = value;
                 return true;
             }
+            else if (obj is IDictionary)
+            {
+                IDictionary dic = obj as IDictionary;
+                dic[propName] = value;
+                return true;
+            }
 
             return false;
         }


Reflection_Engine/Modify/SetPropertyValue.cs
```diff
@@ -93,6 +93,12 @@ public static bool SetPropertyValue(this object obj, string propName, object val
                 bhomObj.CustomData[propName] = value;
                 return true;
             }
+            else if (obj is IDictionary)
+            {
+                IDictionary dic = obj as IDictionary;
+                dic[propName] = value;
```
FraserGreenroyd: What happens if the dictionary does not contain a key for `propName`? I'd like a check on it as a safety precaution please :smile:

enarhi: Key and value are added to CustomData accordingly.

image

enarhi:
Name and labels updated, thanks @FraserGreenroyd !

Reflection_Engine/Modify/SetPropertyValue.cs
adecler: @FraserGreenroyd , long day ? 😉

adecler:
Works for me. I like the fact that we are now getting to those issues where we can afford to worry about the different ways people understand and approach a component instead of being content with having one way that works (i.e. in this case `Comment` was already working but `CustomData.Comment` was not). That's a good sign the BHoM is getting more mature 👍 .

I would just take the opportunity of this PR to fix a little oddity:

image

The warning message there feels wrong as there is a matching property in CustomData already when we call the second `SetProperty`. It is even more clear when you put it next to the long form `CustomData.Comment` since that one doesn't give a warning.

So, what about this ?
image

@enarhi , I'll let you do the change if you agree with this since this is your PR after all.

Reflection_Engine/Modify/SetPropertyValue.cs
FraserGreenroyd: @adecler long week 😿 - marking as resolved cause this isn't relevant 🤦‍♂

IsakNaslundBh:
Could also mention in relation to this: https://github.com/BHoM/BHoM_Engine/issues/1590

Problem with this inner setting of properties making GH components mutable, i.e. changing input data. Would need a level cloning on each object down to the one being modified, and then have them re-applied back to the main object.

Alternatively switch the main method to always preform a deep clone, but could lead to other side effects if we are relying on same reference elsewhere.

IsakNaslundBh:
> Could also mention in relation to this: #1590
> 
> Problem with this inner setting of properties making GH components mutable, i.e. changing input data. Would need a level cloning on each object down to the one being modified, and then have them re-applied back to the main object.
> 
> Alternatively switch the main method to always preform a deep clone, but could lead to other side effects if we are relying on same reference elsewhere.

Answering myself here. Raised a fix for this in BHoM_UI instead, as it was easier to do it from there due to reference issues.

https://github.com/BHoM/BHoM_UI/pull/227

enarhi:
Took a minute on my end but updated as per @adecler .

PR #1586, opened 2020-03-17
Reflection_Engine: Extract descriptions of enum options

### LINKED ISSUE #1584: Reflection_Engine: Get Enum item options and Descriptions
#### What is missing/incorrect?

Add method to get all enum options and descriptions (if any) for enum descriptions.

IsakNaslundBh:

Closes #1584 

Adding the options of the enums, and their inner descriptions, to the description text of an enum. Enum values and text below added:

image

### DIFF
--- a/Reflection_Engine/Query/Description.cs
+++ b/Reflection_Engine/Query/Description.cs
@@ -108,6 +108,7 @@ namespace BH.Engine.Reflection
         }
 
         /***************************************************/
+
         [Description("Return the custom description of a C# class")]
         public static string Description(this Type type, QuantityAttribute quantityAttribute)
         {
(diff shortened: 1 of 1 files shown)


IsakNaslundBh:
@FraserGreenroyd Thought about that, but thought it never would get tooo bad. But could put a limit of say 10 items, and state something like "...and more" after.

Do you have an example of a really long enum, so I can check how it looks?

FraserGreenroyd:
@IsakNaslundBh I would support an `and more` option - Environment Space Type is the first one that springs to mind as being a bit on the longer side.

IsakNaslundBh:
BHoM_Engine.CheckCompliance

PR #1588, opened 2020-03-17
Humans_Engine: fix c value calculation

### LINKED ISSUE #1585: Humans_Engine: c value results not correct for first row of spectators
A focal point for the first row is not created. This means the horizontal and absolute distances for the front row are measure to the default point at 0,0,0:

rolyhudson:
Following on from conversation with @jdarcyBHE PR title and intial comment edited to reflect scope of fixes addressed.

This PR now addresses and fixes items related to incorrect calculation of c values.
Specifically issues #1585 and #1589.
- Method for c value calculation has been updated to function in the the plane defined by the vector from the eye point to the focal point.
- Vectors are now compared using Tolerance.Angle
- Focal point for all rows even the front row
- Perpendicular focal points now generated on edge of pitch / focal curve
- Query method added to easily access ref points for audience / spectator eyes

@jdarcyBHE to review and make further comments or remove the do-not-merge when satisfied.

21/05/2020: Updated test file after changes and merge of Venue_oM and Engine
Can be tested with 
[cValueTest.zip](https://github.com/BHoM/BHoM_Engine/files/4663525/cValueTest.zip)


rolyhudson:
Further issues with c value analysis discovered. Relating to tolerance and method of finding focal points.

Humans_Engine/Query/CvalueAnalysis.cs
```diff
@@ -84,18 +84,18 @@ private static List<Cvalue> EvaluateCvalue(Audience audience, CvalueSettings set
                 else
                 {
                     //check the infront and current are on parallel rows
-                    if (infront.Head.PairOfEyes.ViewDirection != s.Head.PairOfEyes.ViewDirection)
+                    if (infront.Head.PairOfEyes.ViewDirection.Angle(s.Head.PairOfEyes.ViewDirection)> 0.00872665)
```
rolyhudson: Odd to compare and angle to a hard coded half degree

Humans_Engine/Query/CvalueAnalysis.cs
```diff
@@ -114,6 +114,22 @@ private static double GetRowWidth(Spectator current, Spectator nearest, Vector r
         return width;
         }
         /***************************************************/
+        private static double GetRowWidth(Spectator current, Spectator nearest, Point focal, Vector rowV)
```
rolyhudson: this method appears unnecessary as the previous calculation of row width has been test and works

rolyhudson:
This PR is expanding to incorporate discussion of methodology + fixes / checks in the analysis of c value - not just including adding all focal points (as per the original issue + pr).
A change in description is required and perhaps further issues related to the notes below.
@jdarcyBHE would be good to have your file here show the issues you note for replication and discussion.

Other issues identified.
- Comparison of view vectors appears to need a tolerance
- The method for defining focal points perpendicular is incorrect
- current methods of finding the shortest distance between pitch (seatfrontplan) and front row do not (but should) take into account the different methods of defining a focal point (offset through corner, perpendicular, closest)

To allow easier testing I added a query for easy access of eye reference points from the audience. This means the c values can be linked to a point in space more easily. 
A couple of points of note from my tests so far:
- it is possible to have negative c values in situations where the shortest distance between front row and pitch is not used for setting out the profile
- the idea of a uniform bowl in terms of c value is hard to define due to the corner conditions typically having differences between pitch and front row

rolyhudson:
@jdarcyBHE did you get a chance to have a look. Would be good to confirm and get if merged or fixed. 👍

jdarcyBHE:
I will make sure I look at it tomorrow before starting on projects.

rolyhudson:
@jdarcyBHE did you get a chance to take a look?

rolyhudson:
@jdarcyBHE I think you need to update the branch as AudienceEyeReference is missing on the build you have there.

rolyhudson:
# Further notes on the expected behaviour when running the test file in the first comment above.
## Test section is located at the corner of the pitch.
The model is set up to provide the specified c value (0.09) at this location when the `CvalueFocalMethodEnum` is set to: OffsetThroughCorners


The distance between the corner and the front of the first row of seats is 5m. This distance is specified in the `ProfileParameters` and `BowlParameters`:

`CvalueFocalMethodEnum.OffsetThroughCorners` locates focal points on a line parallel to the row that passes through the corner:

Sight lines can be created from the spectators to each of their respective focal point:

Changing to another of the `CvalueFocalMethodEnum` will give different c-values for all spectators.
`CvalueFocalMethodEnum.Closest` locates all focal points on the corner:

`CvalueFocalMethodEnum.Perpendicular` locates focal points on the pitch edge directly in front of spectators:

`CvalueFocalMethodEnum.OffsetThroughCorners` is often used as it gives the worst case: the focal points are closer to the spectators than the other methods.
## Expected behaviour when changing the target c-value specified in `ProfileParameters`:
Cvalues
The front row uses a default value as a c-value cannot be calculated without a spectator in front.
As long as the distance between the front row and the focal point remains as specified in the `ProfileParameters` all spectators have the target c-value.

Adjusting the parameters to give varying distances between spectators and focal points through the corner varying c-values are expected:

rolyhudson:
Closing this following recent attempts to apply the c value and view quality analysis to use case for arenas.  
The methodology requires generalisation to handle view quality c value calculations for standing crowds and spectators / audiences seated on a horizontal surface.
Further experimentation and wider testing is needed to clearly define the use cases and best methodology.

Issue #1582, opened 2020-03-16
Environment_Engine: MapToLevels creating unexpected warning messages


BingWangUS:
#### Description:

The MapToLevels method is creating warning messages that are not expected. After talking to Fraser, very likely it is caused by decimal point.

BingWangUS:
This issue is closed by PR #1598

PR #1598, opened 2020-03-20
Environment_Engine: Sort Outputs based on Level in MapToLevel

### LINKED ISSUE #1593: Environment_Engine: MapToLevel to sort geometry by Z-Height
#### Description:

Adjust output so that all groups of geometry are structured according to their Z-Height. Branches of geometry is currently structured randomly (Floor 1 is path (0,0,4) in this case). And should start with the lowest Z Value (negative in some cases) and move forward from there. 

image

BingWangUS:

Closes #1593 

### Changelog

Sort Room in outputs
Sort LevelInUsed in outputs
Remove unused variable

### DIFF
--- a/Environment_Engine/Compute/MapToLevel.cs
+++ b/Environment_Engine/Compute/MapToLevel.cs
@@ -51,7 +51,6 @@ namespace BH.Engine.Environment
         {
             List<List<Room>> roomsByLevel = new List<List<Room>>();
             List<oM.Geometry.SettingOut.Level> levelsInUse = new List<oM.Geometry.SettingOut.Level>();
-
             List<oM.Geometry.SettingOut.Level> roundedLevels = new List<oM.Geometry.SettingOut.Level>();
 
             for (int x = 0; x < levels.Count; x++)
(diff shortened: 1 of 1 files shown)


BingWangUS:
If I could use some help here:
1. The OrderBy Key does not work for Room when the Level is negative. I suspect this is a issue of the LevelIndex part.
2. Looks like searchLevels was never used that we can remove it.

BingWangUS:
Questions above solved by latest commit

BingWangUS:
@FraserGreenroyd I changed back the dictionary for discussion. I would like to have a simple solution if possible. 
The issue is, during foreach loop of //map everything, the levelindex is referring a index from original level list thus cannot do the sort correctly, while levelinused in not complete until finish this for loop, thus cannot be sorted either. 
So I think the sort function need to be in adding roomsbylevels, to do that we need some elements being able to be sorted. So instead of using LevelIndex as key, I will take elevation as Key. 
Let me know your thought. Thanks.

michaelhoehn:
@BingWangUS @FraserGreenroyd I haven't been asked to review this PR but wanted to report that I'm happy with this functionality. Seems to work effectively and outputs are as expected in collections of geometry by level :D Nice one!

BingWangUS:
This PR also closed issue #1582

Issue #1572, opened 2020-03-12
Spatial_Engine: Create NewElement1D etc. should be Query methods


al-fisher:
As discussed here: https://github.com/BHoM/BHoM_Engine/pull/1543#issuecomment-595757027

> [...] all of these "New" methods actually should be Queries, not Creates. If we see Create methods as constructors, then these should most definitely not be Creates at least, and seeing it as "I get a new compatible type of a lower level object" rather than "I create a new lower level object, based on this higher level object" I think actually even semantically makes more sense.

The `NewElement0D` `NewElement1D`, `NewInternalElement2D` need to be migrated to Query namespace as above.

This will need to be coordinated with Physical_Engine, Structure_Engine, Environment_Engine and any other IElementXD implementations of those interfaces.

As part of this I would suggest we flatten the folder structures completely. Compliance for for Query should not be concerned with this in any case I do not believe. 

FYI @adecler @IsakNaslundBh @FraserGreenroyd @pawelbaran

al-fisher:
Thinking of these as Query methods should also eschew the prefix "New". 
For example being simply:

``` C#
public static IElement2D InternalElement2D(this Panel Panel);
public static IElement1D Element1D(this Opening opening, ICurve curve);
public static IElement0D Element0D(this Bar bar, Point point);
```

IsakNaslundBh:
> Thinking of these as Query methods should also eschew the prefix "New".
> For example being simply:
> 
> ```cs
> public static IElement2D InternalElement2D(this Panel Panel);
> public static IElement1D Element1D(this Opening opening, ICurve curve);
> public static IElement0D Element0D(this Bar bar, Point point);
> ```

I would keep the "New" for this case actually, or some other way to distinguish this. (think this is the reason it ended up in the create in the first palce).

Important to highlight that this is not grabbing anything out from the elements themself, but it is a completely new element being returned, matching the type appropriate for the bar/panels etc. Just calling it `Element0D` would at least make me think that it would be the endnodes of the bar.

al-fisher:
> I would keep the "New" for this case actually, or some other way to distinguish this. (think this is the reason it ended up in the create in the first place).

Ah I see! I wonder if a word like "Extract" might be better than "New"?
Saying that how is this any different from say `Structure.Query.CentreLine(this Bar bar)` ? This returns a "New" generated object.

@IsakNaslundBh @adecler

IsakNaslundBh:
[...]

PR #1601, opened 2020-03-23
Spatial_Engine:  Descriptions for Methods 

### LINKED ISSUE #1595: Spatial_Engine: Descriptions for methods
#### Broken rules:

There are no descriptions

Add descriptions

kThorsager:

Closes #1595
Adds descriptions for all Spatial_Engine methods

### Test files
Not needed
### Additional comments

### DIFF
--- a/Spatial_Engine/Query/Centroid.cs
+++ b/Spatial_Engine/Query/Centroid.cs
@@ -53,7 +53,7 @@ namespace BH.Engine.Spatial
 
         [Description("Queries the centre of area for a IElement2Ds surface representation. For an IElement2D with homogeneous material and thickness this will also be the centre of weight.")]
         [Input("element2D", "The IElement2D with the geometry to get the centre of area of.")]
-        [Output("centroid", "The Point at the centre of weight for the homogeneous geometrical representation of the IElement2D.")]
+        [Output("centroid", "The Point at the centre of area for the homogeneous geometrical representation of the IElement2D.")]
         public static Point Centroid(this IElement2D element2D)
         {
             Point tmp = Geometry.Query.Centroid(element2D.OutlineCurve());
--- a/Spatial_Engine/Query/Centroid.cs
+++ b/Spatial_Engine/Query/Centroid.cs
@@ -50,8 +50,8 @@ namespace BH.Engine.Spatial
         /****            IElement2D            ****/
         /******************************************/
 
-        [Description("Queries the centre of weight for a IElement2Ds surface representation.")]
-        [Input("element2D", "The IElement2D with the geometry to get the centre of weight of. The IElement2D will be considered homogeneous.")]
+        [Description("Queries the centre of area for a IElement2Ds surface representation. For an IElement2D with homogeneous material and thickness this will also be the centre of weight.")]
+        [Input("element2D", "The IElement2D with the geometry to get the centre of area of.")]
         [Output("centroid", "The Point at the centre of weight for the homogeneous geometrical representation of the IElement2D.")]
         public static Point Centroid(this IElement2D element2D)
         {
@@ -81,4 +81,3 @@ namespace BH.Engine.Spatial
         /******************************************/
     }
 }
-
--- a/Spatial_Engine/Query/Centroid.cs
+++ b/Spatial_Engine/Query/Centroid.cs
@@ -36,6 +36,7 @@ namespace BH.Engine.Spatial
         /****            IElement1D            ****/
         /******************************************/
 
+        [NotImplemented]
         [Description("Queries the centre of weight for a IElement1Ds ICurve representation.")]
         [Input("element1D", "The IElement1D with the geometry to get the centre of weight of. The IElement1D will be considered homogeneous.")]
         [Output("centroid", "The Point at the centre of weight for the homogeneous geometrical representation of the IElement1D.")]
--- a/Spatial_Engine/Query/Centroid.cs
+++ b/Spatial_Engine/Query/Centroid.cs
@@ -23,8 +23,10 @@
 using BH.Engine.Geometry;
 using BH.oM.Dimensional;
 using BH.oM.Geometry;
+using BH.oM.Reflection.Attributes;
 using System;
 using System.Collections.Generic;
+using System.ComponentModel;
 
 namespace BH.Engine.Spatial
 {
@@ -34,6 +36,9 @@ namespace BH.Engine.Spatial
         /****            IElement1D            ****/
         /******************************************/
 
+        [Description("Queries the centre of weight for a IElement1Ds ICurve representation.")]
+        [Input("element1D", "The IElement1D with the geometry to get the centre of weight of. The IElement1D will be considered homogeneous.")]
+        [Output("centroid", "The Point at the centre of weight for the homogeneous geometrical representation of the IElement1D.")]
         public static Point Centroid(this IElement1D element1D)
         {
             //TODO: find a proper centre of weight of a curve (not an average of control points)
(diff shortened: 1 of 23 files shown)


Spatial_Engine/Query/ElementVertices.cs
```diff
@@ -52,6 +56,9 @@ public static List<Point> ElementVertices(this IElement1D element1D)
         /****            IElement2D            ****/
         /******************************************/
 
+        [Description("Queries the vertecies from a IElement2D. They are the defining element curves discontinuity points.")]
```
IsakNaslundBh:
```suggestion
        [Description("Returns the discontinuity points from the defining ICurves of the IElement2D.")]
```

Spatial_Engine/Query/ElementVertices.cs
```diff
@@ -66,6 +73,9 @@ public static List<Point> ElementVertices(this IElement2D element2D)
 
         /******************************************/
 
+        [Description("Queries the vertecies from a collection of IElements. They are the defining element curves discontinuity points.")]
```
IsakNaslundBh:
```suggestion
        [Description("Returns the discontinuity points from the defining ICurves of the IElements.")]
```

Spatial_Engine/Query/ElementVertices.cs
```diff
@@ -81,6 +91,9 @@ public static List<Point> ElementVertices(this IEnumerable<IElement> elements)
         /**** Public Methods - Interfaces      ****/
         /******************************************/
 
+        [Description("Queries the vertecies from a IElement. They are the defining element curves discontinuity points.")]
```
IsakNaslundBh:
```suggestion
        [Description("Returns the discontinuity points from the defining ICurves of the IElement.")]
```

Spatial_Engine/Query/IsSelfIntersecting.cs
```diff
@@ -34,6 +37,10 @@ public static partial class Query
         /****            IElement1D            ****/
         /******************************************/
 
+        [Description("Returns if the one dimensional representation of the IElement1D is closer to itself than the tolerance at any two points.")]
+        [Input("element1D", "The IElement1D to evaluate self intersections from.")]
+        [Input("tolerance", "Minimum distance to be considered intersecting.", typeof(Length))]
+        [Output("o", "A boolean which is true if the IElement1Ds curve is self intersecting.")]
```
IsakNaslundBh: not sure I get the "o"?

Spatial_Engine/Query/IsSelfIntersecting.cs
```diff
@@ -44,6 +51,10 @@ public static bool IsSelfIntersecting(this IElement1D element1D, double toleranc
         /****            IElement2D            ****/
         /******************************************/
 
+        [Description("Returns if any of the element curves of the IElement2D is closer to itself than the tolerance at any two points. Does not check for intersections between different curves.")]
```
IsakNaslundBh:
```suggestion
        [Description("Returns if any of the element curves of the IElement2D is closer to itself than the tolerance at any two points. Does not check for intersections between external and internal curves, or between different internal curves.")]
```

Spatial_Engine/Query/IsSelfIntersecting.cs
```diff
@@ -44,6 +51,10 @@ public static bool IsSelfIntersecting(this IElement1D element1D, double toleranc
...
+        [Input("element2D", "The IElement2D which curves are to be evaluated for self intersection.")]
+        [Input("tolerance", "Minimum distance to be considered intersecting.", typeof(Length))]
+        [Output("o", "A boolean which is true if any of the IElement2Ds element curves are self intersecting.")]
```
IsakNaslundBh: "o" ?

Spatial_Engine/Query/Normal.cs
```diff
@@ -33,6 +35,9 @@ public static partial class Query
         /****            IElement2D            ****/
         /******************************************/
 
+        [Description("Returns the normal of the external outline curve.")]
```
IsakNaslundBh:
```suggestion
        [Description("Returns the normal based on the external outline curve.")]
```

Spatial_Engine/Query/Normal.cs
```diff
@@ -33,6 +35,9 @@ public static partial class Query
...
+        [Input("element2D", "The IElement2D to get the normal from.")]
+        [Output("n", "The normal to the IElement2D which is perpendicular to its plane and oriented according to the rigth hand rule in relation to the outline curve.")]
```
IsakNaslundBh: Really good description, but would maybe move some of this text from the output to the description of the method instead.

Spatial_Engine/Query/Centroid.cs
```diff
@@ -45,6 +50,9 @@ public static Point Centroid(this IElement1D element1D)
         /****            IElement2D            ****/
         /******************************************/
 
+        [Description("Queries the centre of weight for a IElement2Ds surface representation.")]
```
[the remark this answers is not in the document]
kThorsager: Said it like this mostly as IElement1Ds descriptions are about centre of weight and wanted to keep some consistency.
agree to this but would like some suggestions on some equivalent word for the IElement1D one

Spatial_Engine/Query/IsSelfIntersecting.cs
kThorsager: output, seemed silly to write `isSelfIntersecting` again

Spatial_Engine/Query/IsSelfIntersecting.cs
kThorsager: same ^

Spatial_Engine/Query/Centroid.cs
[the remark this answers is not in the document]
IsakNaslundBh: The one for `IElement1D` is not implemented from what I can see, so would rather consider stripping down that description all in all until we have a method in place for it, I think, and maybe add a `[NotImplemented]` tag.

I think the comment in that method is basically refering to centre of area. I am pretty sure it says "centre of weight" as opposed to "average control point", not as opposed to "centre of area".

Spatial_Engine/Query/Centroid.cs
```diff
@@ -34,6 +36,9 @@ public static partial class Query
         /****            IElement1D            ****/
         /******************************************/
 
+        [Description("Queries the centre of weight for a IElement1Ds ICurve representation.")]
```
IsakNaslundBh:
```suggestion
        [NotImplemented]
        [Description("Queries the centre of weight for a IElement1Ds ICurve representation.")]
```

Spatial_Engine/Query/IsSelfIntersecting.cs
al-fisher: I think for consistency of approach let's keep the full output name `isSelfIntersecting`.

Actually agree on the `OutputAttribute` there is duplication of information - forced from the format of the constructor requiring `name` & `description`  
``` C#
OutputAttribute(string name, string description, Type quantity = null)`
```

One to review more generally, but for now for clarity let us simply stick with full output name.
@IsakNaslundBh @kThorsager sound OK?

Spatial_Engine/Query/IsSelfIntersecting.cs
al-fisher: See above ^ 😄

Spatial_Engine/Query/IsSelfIntersecting.cs
kThorsager: Changed it and am OK, but should perhaps waited for happiness confirmation from @IsakNaslundBh first

al-fisher: Sorry - 🙈 just realised we have the constructor _already_ for OutputAttribute that takes a single string for description only. 👍 

``` C#

public OutputAttribute(string description)

```

@kThorsager  suggest you use this instead!

kThorsager: oh, when is the general case for not specifying a name?

IsakNaslundBh:
Have checked the compliance issues, and it is nothing that needs to be resolved in this PR.

Will be fixed in https://github.com/BHoM/BHoM_Engine/issues/1572

PR #1623, opened 2020-03-30
Spatial_Engine: Add default behaviour for IElement2D extension methods

### LINKED ISSUE #1561: Spatial_Engine: Add default behaviour for IElement2D extension methods
#### Description:
Specifically `InternalElements2D` and `SetInternalElements2D` 

`IElement2D` which do not have internal elements should not have to implement methods to return empty lists or itself.

Example of a method from Structure_Engine which should be defaulted to:
``` C#
public static List<IElement2D> InternalElements2D(this Opening opening)
{
    return new List<IElement2D>();
}
```

The change would be to return `new List<IElement2D>()` for `IInternalElements2D()` if no extension method could be found

Think that `ISetInternalElements2D()` can removed as well
Example from Structure_Engine:
``` C#
public static Opening SetInternalElements2D(this Opening opening, List<IElement2D> internalElements2D)
{
     if (internalElements2D.Count != 0)
        Reflection.Compute.RecordError("Cannot set internal 2D elements to an opening.");

     return opening.GetShallowClone() as Opening;
}
```

But will need to test if it's possible as it returns a specific type here, but otherwise the same principle.

This would alleviate the implementation of `IElement2D`

FYI @pawelbaran @IsakNaslundBh

kThorsager:
Closes #1561 
`IElement2D` enforced two methods with very default behaviour, this has been captured in the interface method instead.

### Test files

https://burohappold.sharepoint.com/:f:/s/BHoM/EpPN0YICITFLm92G8BBjy0QBJDpEv7r1bs98jeaESVBOIQ?e=cdsjKY

### Additional comments

does not overlap with #1630


IsakNaslundBh:
Nice one, good to get rid of the boilerplate methods. Had a hard time describing them in structure_engine, and motivate their existence, so nice to get rid of them!

PR #1630, opened 2020-03-31
Spatial_Engine: Add HasMergeablePropertiesWith method

### LINKED ISSUE #1629: Spatial_Engine: IsEqualProperties
#### Description:
The ability to evaluate whether two `IElement` can be merged will be useful as we could clean objects without removing important additions from users.

kThorsager:

Closes #1629 
This adds a method which compares the properties of `IElement`s, atm only does base comparison and returns false for objects of equal type which do not have a method specified.

### Test files
https://burohappold.sharepoint.com/:f:/s/BHoM/EgIee2Xt7DJFu--Isb6FWAkBZot7MPpF8YhyOAKcKy9r8Q?e=sIuvRw

### Additional comments
Feels like I'm using the Interface method in the exact opposite manner compared to how they're usually used, but I think it's still how we'd want it(?)

does not overlap with #1623


pawelbaran:
Initial comment: if CI/CD is scraping project names from PR titles, then there is a typo in there 🐒

Spatial_Engine/Query/IsEqualProperties.cs
```diff
@@ -0,0 +1,97 @@
...
+            // Geometrical objects don't have properties
+            if (typeof(IGeometry).IsAssignableFrom(element.GetType()) && typeof(IGeometry).IsAssignableFrom(other.GetType()))
+                return true;
+
+            // Different types are different
+            if (element.GetType() != other.GetType())
+                return false;
+
+            // look for a specific comparing method
+            object result = Reflection.Compute.RunExtensionMethod(element, "IsEqualProperties", new object[] { other });
```
IsakNaslundBh, line 83: Wont this lead to stack overflow given that the method is called the same thing?

kThorsager: Hmm, you'd think so, but the test file runs and passes that part, but I'll change it to be on the safe side 👍

Spatial_Engine/Query/IsEqualProperties.cs
```diff
@@ -0,0 +1,97 @@
...
+        public static bool IIsEqualProperties(this IElement2D element, IElement2D other)
+        {
+            return IIIsEqualProperties(element, other);
+        }
+
+        /******************************************/
+        /****        Private Methods           ****/
+        /******************************************/
+
+        private static bool IIIsEqualProperties(this IElement element, IElement other)
```
pawelbaran, line 72: III?

kThorsager: To avoid stack overflow as I have methods calling this method and this method should not call it self or the previous ones

See:
https://github.com/BHoM/BHoM_Engine/pull/1630#discussion_r401390723

kThorsager:
> Looks like there is a typo in one of the method names - see code comment. Besides, not sure if _Is_ would be my first choice for prefix here. What about `HasEqualProperties`?

I'm open for other names, other suggestion is `HasMergebleProperties`

pawelbaran:
> > Looks like there is a typo in one of the method names - see code comment. Besides, not sure if _Is_ would be my first choice for prefix here. What about `HasEqualProperties`?
> 
> I'm open for other names, other suggestion is `HasMergebleProperties`

Even better! Possibly we could have 2 methods, one checking whether the properties are equal and the other checking if they can be merged. Something for Diffing_Engine, @alelom?

Dropping one more idea: `CanMergeProperties`

Spatial_Engine/Query/IsEqualProperties.cs
pawelbaran: Hmm, makes sense, but looks weird. Maybe it could be done somehow neater (can't come up with anything nice myself though 🐒)?

kThorsager: Think a different name is the way to go without repeting the code or just going with input: (`IElement`, `IElement`)

some suggestions if it's the double I which is disturbing

`IsEqualPropertiesDispatcher` 
`IsEqualPropertiesProgram`
`IsEqualPropertiesMeat`

pawelbaran: `IsEqualPropertiesInterface`?

al-fisher:
I like `HaveMergebleProperties` as a naming convention. "Have" though rather than "Has" as is operating on a pair of objects.

Spatial_Engine/Query/IsEqualProperties.cs
```diff
@@ -0,0 +1,97 @@
...
+            if (element.GetType() != other.GetType())
+                return false;
+
+            // look for a specific comparing method
+            object result = Reflection.Compute.RunExtensionMethod(element, "IsEqualProperties", new object[] { other });
+
+            if (result == null)
+            {
+                Reflection.Compute.RecordWarning("No comparer found for comparing the properties, diffrent kind of objects may have been merged");
+                return true;
```
IsakNaslundBh, line 88: would switch default to false here.

Spatial_Engine/Query/IsEqualProperties.cs
```diff
@@ -0,0 +1,97 @@
...
+
+            // Different types are different
+            if (element.GetType() != other.GetType())
+                return false;
+
+            // look for a specific comparing method
+            object result = Reflection.Compute.RunExtensionMethod(element, "IsEqualProperties", new object[] { other });
+
+            if (result == null)
```
IsakNaslundBh, line 85: can be good to check if the result is of type bool here, to avoid potential crashes on the return line (should probably never ever happen, but still good to get it in).

kThorsager:
> * The file name is now incorrect.

Yes, fixed
> * Shouldn't it be `mergeable` instead of `mergeble`?

Yes, fixed
> * I am not convinced by the use of `have` here. The method is applied to the `this` argument only not to all the inputs together (i.e. `x.HasMergeableProperties(y)`, not `(x&y).HaveMergeableProperties()`). I would be fine with `x.HasMergeablePropertiesWith(y)` though.

Maybe, will hold off on action as it sounds like there should be some discussion and am unsure myself

al-fisher:
> I am not convinced by the use of have here. The method is applied to the this argument only not to all the inputs together (i.e. x.HasMergeableProperties(y), not (x&y).HaveMergeableProperties()). I would be fine with x.HasMergeablePropertiesWith(y) though.

Yes - I like `x.HasMergeablePropertiesWith(y)` more @adecler 
Great point. It was indeed the `this` keyword that was throwing me to start with. “Have” did not really scan perfectly. Appending “With” is neat and could be a pattern to reuse elsewhere for clarity.

PR #1632, opened 2020-04-01
Geometry_Engine: Clarify input naming in CollapseToScaledPolyline.cs

### LINKED ISSUE #1631: Geometry_Engine: weird hardcoded values in CollapseToScaledPolyline
#### Description:

Hardcoded values got sneaked in to `BH.Engine.Geometry.Convert.CollapseToScaledPolyline` while working on https://github.com/BHoM/BHoM_Engine/pull/1255:

https://github.com/BHoM/BHoM_Engine/blob/a6b00d12e00910d618ce1b8c233cd41e7c079fe9/Geometry_Engine/Compute/CollapseToScaledPolyline.cs#L44

Could we please change the values (particularly `tolerance` to restore compliance?

kThorsager:
Closes #1631

### Test files

Not needed, and only code implementation was in structures where the tolerance value was defined, doubt it has gotten mush use elsewhere.

### DIFF
--- a/Geometry_Engine/Compute/CollapseToScaledPolyline.cs
+++ b/Geometry_Engine/Compute/CollapseToScaledPolyline.cs
@@ -36,18 +36,18 @@ namespace BH.Engine.Geometry
 
         [Description("Collapses a PolyCurve to a polyline where curved segments are split into Polyline segments for computations. Where their partitioning is based on both their curvature and size, ergo; shorter segment less partitions, very curved segment more partitions")]
         [Input("curve", "The curve to collapse")]
-        [Input("tolerance", "tolerance is the angleTolerance for a unit circle, the tolerance value will vary by: toleranceGrowth over curvature")]
-        [Input("toleranceGrowth", "toleranceGrowth is the value that decides how much the angleTolerance varies due to curvature")]
+        [Input("relativeAngleTolerance", "relativeAngleTolerance is the angleTolerance for a unit circle, the tolerance value will vary by: toleranceGrowth over curvature")]
+        [Input("toleranceScaleFactor", "toleranceScaleFactor is the value that decides how much the angleTolerance varies due to curvature")]
         [Input("maxDivisionsPerSegment", "The maximum number of segment each sub-curve can be broken into")]
         [Input("scale", "the radius of a baseline circle for the angle tolerance, smaller circles will have less segments and bigger more. Default of 0 means the program will determine an appropriate radius for the PolyCurve")]
         [Output("C", "A polyline approximating the provided curve")]
-        public static Polyline CollapseToScaledPolyline(this PolyCurve curve, double tolerance, double toleranceGrowth = 0.001, int maxDivisionsPerSegment = 200, double scale = 0)
+        public static Polyline CollapseToScaledPolyline(this PolyCurve curve, double relativeAngleTolerance = 0.04, double toleranceScaleFactor = 0.001, int maxDivisionsPerSegment = 200, double scale = 0)
         {
-            if (tolerance <= 0)
+            if (relativeAngleTolerance <= 0)
                 return null;
-            if (toleranceGrowth < 0)
-                toleranceGrowth = 0;
-            else if (toleranceGrowth > 0.01)
+            if (toleranceScaleFactor < 0)
+                toleranceScaleFactor = 0;
+            else if (toleranceScaleFactor > 0.01)
                 Reflection.Compute.RecordWarning("High values for toleranceGrowth can be unstable and are not recommended");
             if (scale <= 0)
             {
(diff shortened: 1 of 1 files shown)


Geometry_Engine/Compute/CollapseToScaledPolyline.cs
```diff
@@ -41,7 +41,7 @@ public static partial class Compute
         [Input("maxDivisionsPerSegment", "The maximum number of segment each sub-curve can be broken into")]
         [Input("scale", "the radius of a baseline circle for the angle tolerance, smaller circles will have less segments and bigger more. Default of 0 means the program will determine an appropriate radius for the PolyCurve")]
         [Output("C", "A polyline approximating the provided curve")]
-        public static Polyline CollapseToScaledPolyline(this PolyCurve curve, double tolerance = 0.04, double toleranceGrowth = 0.001, int maxDivisionsPerSegment = 200, double scale = 0)
+        public static Polyline CollapseToScaledPolyline(this PolyCurve curve, double tolerance, double toleranceGrowth = 0.001, int maxDivisionsPerSegment = 200, double scale = 0)
```
FraserGreenroyd, line 44: The description for tolerance says that it is an angle tolerance, so is there a reason why it cannot default to `BH.oM.Geometry.Tolerance.Angle`?

Also, the `toleranceGrowth` is equal to `BH.oM.Geometry.Tolerance.MacroDistance` so it might be worthwhile using that one rather than a decimal here to be consistent with other methods/standards? :smile:

kThorsager: From memory I only have bad experiences with values under 0.035 or something like that, and would rater let the user make the mistake than have it as default.

`toleranceGrowth` is not a tolerance, but a growth of a tolerance and I figured I should not muddy the waters in this PR by defaulting non-tolerances as tolerances

FraserGreenroyd: Ok I'm happy with the defence of that then and happy to change my review :smile:

pawelbaran: I do not know the application of this value, but if `tolerance` argument in this method works differently from what we usually understand by geometrical tolerance in BHoM, maybe we could rename it to something else?

kThorsager: I'm going by memory and my old description here but it's an attempt for a tolerance which would equate to the angleTolerance for a circle if the `scale = 0` i.e. auto. otherwise maintain that level of fidelity for curves with more or less curvature without adding an obscene amount of control-points. 

The point of which was to get a good polyline representation of sections without a hundred control points for both, say a circular concrete column and its re-bar holes, as the small holes have no reason for that level of fidelity.

So it's kind of a angleTolerance but not really, as it will be a different angleTolerance based on the other parameters.

So unsure of what would be a good name, which is how I landed on `tolerance` in the first place

pawelbaran: Hmm not sure. Maybe we could rename the method so that it was pointing directly to integration (it seems to be very specific), e.g. `PrepareToIntegrate` or similar? Then the value could `fidelity`?

Or a step further: make the method `private` and put in integration file? Not sure if that brings any value to the user without the code base that sits on top of it? @IsakNaslundBh do you have opinion on that?

kThorsager:
After some discussion the option of clarifying the naming and keeping the value was decided on.

Geometry_Engine/Compute/CollapseToScaledPolyline.cs
```diff
@@ -36,18 +36,18 @@ public static partial class Compute
...
-            if (tolerance <= 0)
+            if (relativeAngleTolerance <= 0)
                 return null;
-            if (toleranceGrowth < 0)
-                toleranceGrowth = 0;
-            else if (toleranceGrowth > 0.01)
+            if (toleranceScaleFactor < 0)
+                toleranceScaleFactor = 0;
+            else if (toleranceScaleFactor > 0.01)
                 Reflection.Compute.RecordWarning("High values for toleranceGrowth can be unstable and are not recommended");
```
epignatelli, line 51: `toleranceGrowth` to `toleranceScaleFactor`?

Geometry_Engine/Compute/CollapseToScaledPolyline.cs
```diff
@@ -36,18 +36,18 @@ public static partial class Compute
 
         [Description("Collapses a PolyCurve to a polyline where curved segments are split into Polyline segments for computations. Where their partitioning is based on both their curvature and size, ergo; shorter segment less partitions, very curved segment more partitions")]
         [Input("curve", "The curve to collapse")]
-        [Input("tolerance", "tolerance is the angleTolerance for a unit circle, the tolerance value will vary by: toleranceGrowth over curvature")]
-        [Input("toleranceGrowth", "toleranceGrowth is the value that decides how much the angleTolerance varies due to curvature")]
+        [Input("relativeAngleTolerance", "relativeAngleTolerance is the angleTolerance for a unit circle, the tolerance value will vary by: toleranceGrowth over curvature")]
```
epignatelli, line 39: `toleranceGrowth` to `toleranceScaleFactor`?

Geometry_Engine/Compute/CollapseToScaledPolyline.cs
```diff
@@ -61,17 +61,17 @@ public static Polyline CollapseToScaledPolyline(this PolyCurve curve, double tol
                 scale = Math.Max(totalHeight, totalWidth) / 2;
             }
 
-            double scaledTolerance = scale * toleranceGrowth;
+            double scaledTolerance = scale * toleranceScaleFactor;
```
epignatelli, line 64: I didn't get why you are using two factors to control the refinement rate of the curves. Maybe it would be easier for the user to use just one of them?

kThorsager: probably but that would just hide an hard-coded value inside the method no?

And it factors the tolerance to its scale, so `toleranceScaleFactor`

al-fisher:
> Sorry, just noticed that the are some parts of the decription that have not been fixed yet, e.g. https://github.com/BHoM/BHoM_Engine/pull/1632/files#r405439808

Can't see what you are referring to there @epignatelli, sorry? 😄 
@kThorsager - can you recheck and confirm also tweaked all descriptions where necessary. Can then get this merged. 👍

kThorsager:
Missed to update a word, thanks @epignatelli 
Should be good to go again

PR #1637, opened 2020-04-02
Data_Engine: Add generic graph query methods

### LINKED ISSUE #1468: Data_Engine add methods for depth first graph traversal and adjacency dictionary
Edit 02/04/2020 to include 4 graph methods:

1. DepthDictionary
2. AdjacencyDictionary
3. GetSubGraph
4. GetLeafNodes

#### Description:
For scoring using the decision graph established in the Qualia_Toolkit: https://github.com/BuroHappoldEngineering/Qualia_Toolkit/blob/master/Qualia_Engine/Convert/ToDecisionGraph.cs

A depth first traversal method is required. 
This should return a ``Dictionary<GraphNode<T>, GraphNode<T>>`` which is in the form
``<Node @ nodeLevel_n-1,Node @ nodeLevel_n>``
and the root is stored as ``<Node @ root,Node @ root>``
This is a predecessor list which given any node allows fast traversal from that node back to the root.

This process would be helped by a single call to a method to determine a representation of the graph as an adjacency dictionary
in the form ``Dictionary<GraphNode<T>, List<GraphNode<T>>>`` or ``<node, neighbours>``

Also in relation to https://github.com/BuroHappoldEngineering/Qualia_Toolkit/issues/7
The methods 
1. DepthDictionary
2. AdjacencyDictionary
are also required

Any system that can be represented in graph form (space adjacency, road / service routing etc) could be a potential use case for these methods.

rolyhudson:
Generic graph methods added to `BH.Engine.Data.Query`:

###Depends on:
https://github.com/BHoM/BHoM/pull/788

1. `AdjacencyDictionary` (Could also be considered a convert as adjacency lists are also representations of graphs)
2. `DepthDictionary`
3. `LeafNodes`
4. `FindGraphNodes`

Closes #1468 
Added `AdjacencyDictionary` and `DepthDictionary` methods as per issue.
Also added related methods `SubGraph` and `LeafNodes`
And helper method `FindGraphNodes`
### Test files
Simple test in grasshopper
[TestFileGraphMethods2 (2).zip](https://github.com/BHoM/BHoM_Engine/files/4469410/TestFileGraphMethods2.2.zip)
### Changelog
Generic graph methods added to `BH.Engine.Data.Query`:

1. `AdjacencyDictionary` 
2. `DepthDictionary`
3. `LeafNodes`
4. `FindGraphNodes`


Data_Engine/Query/DepthDictionary.cs
```diff
@@ -0,0 +1,81 @@
...
+{
+    public static partial class Query
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+        [Description("Gets the depth dictionary of a graph using breadth first search, each key value pair in the resulting dictionary is in the form <graph node, depth>")]
+        [Input("adjacency", "The adjacency dictionary to extract the depth dictionary from")]
+        [Input("startNode", "The graph node from which the depth dictionary is created")]
+        public static Dictionary<GraphNode<T>, int> DepthDictionary<T>(Dictionary<GraphNode<T>, List<GraphNode<T>>> adjacency, GraphNode<T> startNode)
```
adecler, line 40: I suppose you are taking the adjacency dictionary as an input to avoid having to recompute it every time but it would be good to also have the method that takes the graph itself as an input. Otherwise, it might be a bit confusing/fiddly on the level of the UI.

Data_Engine/Query/DepthDictionary.cs
```diff
@@ -0,0 +1,81 @@
...
+            level[startNode] = 0;
+            // marked it as visited  
+            marked[startNode] = true;
+            // do until queue is empty  
+            while (que.Count > 0)
+            {
+                // dequeue element  
+                startNode = que.Dequeue();
+                // traverse neighbors of node x  
+                foreach (GraphNode<T> b in adjacency[startNode])
```
adecler, line 63: Since adjacency created outside of this method, it would be good to make sure that `startNode` is a key of `adjacency`.

Data_Engine/Query/FindGraphNodes.cs
```diff
@@ -0,0 +1,43 @@
...
+namespace BH.Engine.Data
+{
+    public static partial class Query
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+        [Description("Finds the nodes of a graph with value that matches the input value")]
+        [Input("value", "The value to match")]
+        public static List<GraphNode<T>> FindGraphNodes<T>(this Graph<T> graph, T value)
```
adecler, line 38: `FindGraphNodes` is not a valid name for a query method. It should simply be `GraphNodes`

Data_Engine/Query/FindGraphNodes.cs
```diff
@@ -0,0 +1,43 @@
...
+        {
+            return graph.Nodes.FindAll(x => x.Value.Equals(value));
```
adecler, line 40: Keep in mind that it is unlikely to work with `double`. As long as you're fine with that, I am ok too.

Data_Engine/Query/LeafNodes.cs
```diff
@@ -0,0 +1,51 @@
...
+using BH.oM.Reflection.Attributes;
+
+namespace BH.Engine.Data
+{
+    public static partial class Query
+    {
+        /***************************************************/
+        /****           Public Methods                  ****/
+        /***************************************************/
+        [Description("Extracts a list of leaf nodes from a graph, leaf nodes have no incoming links")]
```
adecler, line 36: That's interesting. In the first example of a tree that came to my mind, leafs would have no **outgoing** links. Take for example a tree representing a population split over multiple levels with the root being the total population. A link could be going from parent to children using the weight as the ratio. Same would go for a water/gas distribution network. 

If I want to be annoying, I could also argue that a leaf node could have both incoming and outgoing links to the same parent.

So it might be safer to return as leaves the nodes that are not connected to any node deeper than themselves.

Data_Engine/Query/SubGraph.cs
```diff
@@ -0,0 +1,63 @@
...
+        /***************************************************/
+        [Description("Extracts a sub graph from a graph")]
+        [Input("graph", "The graph to extract the sub graph from")]
+        [Input("rootNode", "The graph node that is the root of the sub graph")]
+        public static Graph<T> SubGraph<T>(this Graph<T> graph, GraphNode<T> rootNode)
+        {
+            Graph<T> subgraph = new Graph<T>();
+            subgraph.Nodes.Add(rootNode);
+            if (!graph.Nodes.Contains(rootNode))
+                throw new ArgumentException("rootnode provided cannot be found in the original graph. Ensure the node is from the original graph");
```
adecler, line 44: Same thing here: Don't throw an Exception, use RecordError instead

Data_Engine/Query/SubGraph.cs
```diff
@@ -0,0 +1,63 @@
...
+            List<GraphLink<T>> links = graph.Links.FindAll(x => x.EndNode.Value.Equals(rootNode.Value));
```
adecler, line 45: It would be safer to use the same technique you have used for `DepthDictionary` to avoid all the issues mentioned above about direction of the links.

Data_Engine/Query/DepthDictionary.cs
```diff
@@ -0,0 +1,81 @@
...
+        /**** Public Methods                            ****/
+        /***************************************************/
+        [Description("Gets the depth dictionary of a graph using breadth first search, each key value pair in the resulting dictionary is in the form <graph node, depth>")]
+        [Input("adjacency", "The adjacency dictionary to extract the depth dictionary from")]
+        [Input("startNode", "The graph node from which the depth dictionary is created")]
+        public static Dictionary<GraphNode<T>, int> DepthDictionary<T>(Dictionary<GraphNode<T>, List<GraphNode<T>>> adjacency, GraphNode<T> startNode)
+        {
+            //https://www.geeksforgeeks.org/level-node-tree-source-node-using-bfs/
+            if(!adjacency.ContainsKey(startNode))
+                throw new ArgumentException("startNode provided cannot be found in the adjacency dictionary. Ensure the node is from the original graph");
```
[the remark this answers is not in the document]
rolyhudson: No problem -changing to `Reflection.Compute.RecordError` and returning null. Why is this better?

Data_Engine/Query/AdjacencyDictionary.cs
```diff
@@ -0,0 +1,55 @@
...
+            foreach (GraphNode<T> node in graph.Nodes)
+            {
+                List<GraphNode<T>> neighbours = new List<GraphNode<T>>();
+                //neigbours in both directions
+                List<GraphNode<T>> neighboursOut = graph.Links.Where(x => x.StartNode == node).Select(x => x.EndNode).ToList();
+                List<GraphNode<T>> neighboursIn = graph.Links.Where(x => x.EndNode == node).Select(x => x.StartNode).ToList();
+
+                neighbours.AddRange(neighboursOut);
+                neighbours.AddRange(neighboursIn);
+                adjacency.Add(node, neighbours);
```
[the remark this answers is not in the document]
rolyhudson: Yes good point. 👍

Data_Engine/Query/DepthDictionary.cs
rolyhudson: yes already using `if(!adjacency.ContainsKey(startNode))` on line 43. throwing ArgumentException but changing to `Reflection.Compute.RecordError` in line with other comments

Data_Engine/Query/FindGraphNodes.cs
rolyhudson: Yeah think I'm happy here with that

Data_Engine/Query/DepthDictionary.cs
rolyhudson: Yes good point. Added the additional method.

Data_Engine/Query/LeafNodes.cs
rolyhudson: Good points. And useful to see this from another perspective. My use case for this is coming from the decision graph where leaves contribute to the next level up. 

Another classic case I have worked on where its useful to have leaves with only outgoing links is computing flow accumulation in water sheds in river networks. Here the graph can be defined between cells in a landscape linked by negative gradient in topology.  Rainfall from individual cells in a landscape contribute to others

Agreed also that incoming and outgoing links from a leaf are possible. You'll have to try harder than that to annoy 😄. We need to accommodate all these cases. Going to switch to a depth based method of determining a leaf.

Data_Engine/Query/AdjacencyDictionary.cs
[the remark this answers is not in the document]
rolyhudson: Actually on this should we add the option to the existing `Neighbours` query to retrieve neighbours using incoming and outgoing links? Currently `Neighbours` only captures neighbours at the `EndNode` of the nodes links. Also its not using `Distinct()`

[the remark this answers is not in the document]
adecler: Yes, definitely

Data_Engine/Query/DepthDictionary.cs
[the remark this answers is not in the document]
adecler: It just fits better with the way the BHoM architecture is working

Data_Engine/Query/DepthDictionary.cs
adecler: ha yes, true

Data_Engine/Query/SubGraph.cs
rolyhudson: I'm looking at some background here. https://mathworld.wolfram.com/InducedSubgraph.html
According to that source the naming I have used is too vague. Looks like induced subgraphs graphs are really either edge induced or link induced. Neither of these applies to what I have needed.
Here what I need for the particular use is all "downstream" nodes in a decision graph. So maybe this one not generic enough to be here. Suggesting this is removed and located in the specific toolkit. What do you think?

rolyhudson: Or rename to be more specific?

Data_Engine/Query/AdjacencyDictionary.cs
```diff
@@ -0,0 +1,49 @@
...
+namespace BH.Engine.Data
+{
+    public static partial class Query
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+        [Description("Gets the adjacency dictionary of a graph, each key value pair in the resulting dictionary is in the form <node, neighbours>")]
+        [Input("graph", "The graph to extract the adjacency dictionary from")]
+        public static Dictionary<GraphNode<T>, List<GraphNode<T>>> AdjacencyDictionary<T>(this Graph<T> graph)
```
epignatelli, line 38: Would it be more helpful to have an `AdjacencyMatrix` object?
https://en.wikipedia.org/wiki/Adjacency_matrix

Can be an `AdjacencyDictionary` in case, but I think the `AdjacencyMatrix` is one of the standard format to represent graphs, at least in causality, decisions, and particle models.

Data_Engine/Query/LeafNodes.cs
epignatelli: I think I agree with the first part of @adecler's comment. The most common definition of a `leaf` is a node with no children. If a most depth-wise node has outgoing links to the parent is a just a cyclic graph with no leaves. 

This includes cases where a node is parent of itself as well, [loops](https://en.wikipedia.org/wiki/Loop_(graph_theory)).

I think it would be good to clarify what is `Graph`. To me, the `Graph` class can represent non non-directed cyclic graphs and all its restrictions, either [directed](https://en.wikipedia.org/wiki/Directed_graph) and [acyclic](https://en.wikipedia.org/wiki/Directed_acyclic_graph).

Data_Engine/Query/SubGraph.cs
epignatelli: As per the other comment _downstream_ makes sense only for directed graphs. `SubGraph` makes sense to me if the recipe takes a list of nodes, or edges, as you say, and the methods figures out, respectively, either the connections, or the nodes.

I really think we should start splitting the two classes,

Data_Engine/Query/LeafNodes.cs
```diff
@@ -0,0 +1,56 @@
...
+                int nodeDepth = depthDictionary[node];
+                bool leaf = true;
+                foreach (GraphNode<T> n in neighbours)
+                    if (depthDictionary[n] > nodeDepth)
+                        leaf = false;
+                if(leaf) leafnodes.Add(node);
+            }
+            return leafnodes;
+        }
+    }
```
adecler, line 55: While technically correct, this is far more complex than it should be. 
Simply copy-paste the `DepthDictionary` method, remove the level variable, and determine that a node is a leaf when all its neighbours are marked already.

Data_Engine/Query/AdjacencyDictionary.cs
rolyhudson: Good point we should include other adjacency representations. Implemented as a dictionary just to support a specific case. Can we put this in the wish list to add later or when the need arises?

al-fisher: Yes - I really like the idea of the Adjacency Matrix too going forward. 
@rolyhudson - makes sense to me - I would be happy for this is be raised as issue for future development after this PR.

Data_Engine/Query/Neighbours.cs
```diff
@@ -32,9 +32,19 @@ public static partial class Query
         /**** Public Methods                            ****/
         /***************************************************/
 
-        public static List<GraphNode<T>> Neighbours<T>(this Graph<T> graph, GraphNode<T> node)
+        public static List<GraphNode<T>> Neighbours<T>(this Graph<T> graph, GraphNode<T> node,bool useBothDirections = false)
```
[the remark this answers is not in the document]
rolyhudson: Yes great 👍 . Going to add an enum in Data_oM and dependent PR.

Data_Engine/Query/LeafNodes.cs
rolyhudson: 👍 got it. Thanks

Data_Engine/Query/SubGraph.cs
rolyhudson: > I really think we should start splitting the two classes,

@epignatelli 
Is this a suggestion we should have classes for `DirectedGraph` and `UndirectedGraph`?

adecler: Not sure I like that idea too much if we can avoid it. The only difference between those two graphs is inside the links themselves. So we could just have one type of graph and two types of links (not necessarily two classes, it can just be a Boolean poverty of the link)

Data_Engine/Query/LeafNodes.cs
```diff
@@ -0,0 +1,86 @@
...
+                foreach (GraphNode<T> b in adjacency[startNode])
+                {
+                    // b is neighbor of node x  
+                    // if b is not marked already  
+                    if (!marked.ContainsKey(b))
+                    {
+                        // enqueue b in queue  
+                        que.Enqueue(b);
+                        // mark b  
+                        marked[b] = true;
```
adecler, line 74: Actually, my bad. This doesn't work if two nodes share a "child". So you were better off using depth.

Data_Engine/Query/SubGraph.cs
rolyhudson: Would it be simpler if there is a property like `IsDirected` on the `Graph` rather than on every link? 

Pleased the discussion is rolling a little 😃 now.

I think this is in need of raising a new issue and beyond the scope of the present PR - I have removed the SubGraph method from this PR so we can discuss further.

Data_Engine/Query/LeafNodes.cs
rolyhudson: Oh yes good spot. Reverting to previous version of LeafNodes

Data_Engine/Query/SubGraph.cs
adecler: > I think this is in need of raising a new issue and beyond the scope of the present PR

Agreed

epignatelli: Yeah, I am fine with having only one class, as long as any manipulation of `Graph` makes sense.

Data_Engine/Query/LeafNodes.cs
```diff
@@ -0,0 +1,60 @@
...
+            {
+                List<GraphNode<T>> neighbours = Neighbours(graph, node, GraphLinkDirection.Both);
+                //its a leaf if no neighbour is deeper in the graph
+                int nodeDepth = depthDictionary[node];
+                bool leaf = true;
+                foreach (GraphNode<T> n in neighbours)
+                    if (depthDictionary[n] > nodeDepth)
+                        leaf = false;
+                if (leaf) leafnodes.Add(node);
+            }
```
adecler, line 55: Alright, I am going to be annoying one last time. The initial reasons that made me asked to change this code was that neighbours are calculated twice for each node. Can you switch to getting the adjacency dictionary as a variable and then using it both for the `DepthDictionary` method and inside the loop (instead of `Neigbours`)

adecler:
@rolyhudson , FYI, CI is complaining about this PR because it depends on an oM PR but they don't have the same branch name. I will merge the oM now so it will remove the problem but something to keep in mind for next time.

FraserGreenroyd:
This PR also needs a suitable `feature` `bug` `documentation` or `compliance` label for change log purposes please :smile:

Data_Engine/Query/LeafNodes.cs
rolyhudson: Yes OK no problem. 👍

Data_Engine/Query/LeafNodes.cs
```diff
@@ -39,18 +39,30 @@ public static partial class Query
...
+                        {
+                            leaf = false;
+                            break;
+                        }
+                    }         
+                    if (leaf) leafnodes.Add(node);
+                }
+                catch
+                {
+                    int b = 0;
```
adecler, line 64: Looking better but it seems that you left some personal debugging code in there.

rolyhudson: Whoops. Sloppy. 😸 Cleaned it up. 👍

PR #1639, opened 2020-04-02
Diffing_Engine: Correct fieldsToIgnore not working for nested fields; added ability to diff Fragments

### LINKED ISSUE #1633: Diffing_Engine: changes needed to allow only specific Fragment types to be excluded
#### Description:

To be solved together with https://github.com/BHoM/BHoM/issues/775.

Currently the DiffConfig only has the PropertiesToIgnore list property that allows to set exceptions. This is translated in the Diffing_Engine as `fieldsToIgnore` in the Compute.DiffHash().

Will need to consider how to allow the user to easily input exceptions in a simple manner but allowing for specific Fragment types to be excluded.

### LINKED ISSUE #1638: Diffing_Engine: ToDiffingByteArray() does not remove nested properties from the "fieldsToIgnore"
#### Description:

`ToDiffingByteArray()` must remove all nested properties from the "fieldsToIgnore".

Currently, it only removes the first-level ones, e.g. for a Bar it removes the top `BHoM_Guid`, but not the StartNode/Endnode/etc `BHoM_Guid`.

This means that the hash of two identical objects come up different every time the hash is computed, which prevents a proper version control.

This eluded my attention until now for two reasons:
1. Initially this bug was not present, and was introduced by switching to Mongo's BSON bytearray serialisation. This was suggested by Arnaud as an alternative to my Serialization to Json string and consequent serialisation as bytearray using standard .NET BinaryFormatter.
2. This does not preclude workflows that rely on "relative" hash calculation to work. I.e. every time you calculate an hash it's on an entirely new object or an existing one. However, calculating the hash twice for two identical objects would expose the issue. This is why the Grasshopper script used until now will not show this. _**See steps to reproduce.**_

#### Steps to reproduce
Create a BHoMObject. Compute the Diffing Hash (e.g. by including it into a new Revision).
Compute again the Diffing Hash (e.g. by including it into another revision). 

Explode the object -> fragments -> HashFragment -> CurrentHash will be different for the two revisions. It shouldn't.

#### Proposed solution
The initial approach that did not cause the bug (see point 1 above) comes with two main drawbacks:
- [The standard .NET binaryFormatter does not exclude the assembly version](https://stackoverflow.com/questions/19289898/binaryformatter-ignore-assembly-version). While it may be acceptable in some sense, I'm sure in general we don't want this.
- The JSON serialization needed before the format to binary means that we will have to manually look for and exclude all fields that we want to exclude in the string, which is (probably?) slower than Mongo's `BsonDocument.Remove()` method.

This needs looking into.

alelom:
### NOTE: Depends on 

https://github.com/BHoM/BHoM/pull/776
   
Closes #1638 
Closes #1633

### DIFF
--- a/Diffing_Engine/Compute/Diffing.cs
+++ b/Diffing_Engine/Compute/Diffing.cs
@@ -55,7 +55,7 @@ namespace BH.Engine.Diffing
             List<IBHoMObject> readObjs = previousObjects.ToList();
 
             // Make dictionary with object hashes to speed up the next lookups
-            Dictionary<string, IBHoMObject> readObjs_dict = readObjs.ToDictionary(obj => obj.GetHashFragment().Hash, obj => obj);
+            Dictionary<string, IBHoMObject> readObjs_dict = readObjs.ToDictionary(obj => obj.GetHistoryFragment().Hash, obj => obj);
 
             // Dispatch the objects: new, modified or old
             List<IBHoMObject> newObjs = new List<IBHoMObject>();
@@ -67,7 +67,7 @@ namespace BH.Engine.Diffing
 
             foreach (var obj in currentObjs)
             {
-                var hashFragm = obj.GetHashFragment();
+                var hashFragm = obj.GetHistoryFragment();
 
                 if (hashFragm?.PreviousHash == null)
                 {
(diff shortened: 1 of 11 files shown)


alelom:
Replaced by https://github.com/BHoM/BHoM_Engine/pull/1640

PR #1640, opened 2020-04-06
Diffing_Engine: Correct fieldsToIgnore not working for nested fields; added ability to diff Fragments

### LINKED ISSUE #1633: Diffing_Engine: changes needed to allow only specific Fragment types to be excluded

### LINKED ISSUE #1638: Diffing_Engine: ToDiffingByteArray() does not remove nested properties from the "fieldsToIgnore"

alelom:
### NOTE: Depends on 

https://github.com/BHoM/BHoM/pull/776
   
Closes #1638 
Closes #1633

### Test files

https://burohappold.sharepoint.com/:f:/s/BHoM/EhZG8bcBYJJAvU00CCGa8AwBVxDOnuqe_vJiTp6v_e5W8g?e=Qeletj

### Changelog

- Bug fix: the Hash compute now correctly removes all the fields specified in the  `DiffConfig.PropertiesToIgnore` also from nested objects. This previously produced a bug where `Guid`s were not removed from nested objects; the resulting hash was different every time you instantiated BHoMObjects, even if they were identical in all properties (the `BHoM_Guid` is always generated new at every instantiation).
- Feature add: You can now diff the Fragments because `DiffConfig` does not have to mandatorily include `Fragments` in the `PropertiesToIgnore` anymore. `HashFragment` is now always ignored by the Hash compute function.
- Align the `Delta` Create methods to the corrections in the oM. Details here: <details> Since `Delta` implements `IImmutable`, the constructor must include all get-only parameters in order for Reflection and serialisation to work. Therefore, the constructor should be logic-free; it follows that also auto-initializers are to avoid. All default property values for the Delta had to be specified in the Create where the constructor is invoked from the UI viewpoint. </details>

### DIFF
--- a/Diffing_Engine/Compute/Diffing.cs
+++ b/Diffing_Engine/Compute/Diffing.cs
@@ -55,7 +55,7 @@ namespace BH.Engine.Diffing
             List<IBHoMObject> readObjs = previousObjects.ToList();
 
             // Make dictionary with object hashes to speed up the next lookups
-            Dictionary<string, IBHoMObject> readObjs_dict = readObjs.ToDictionary(obj => obj.GetHistoryFragment().CurrentHash, obj => obj);
+            Dictionary<string, IBHoMObject> readObjs_dict = readObjs.ToDictionary(obj => obj.GetHistoryFragment().Hash, obj => obj);
 
             // Dispatch the objects: new, modified or old
             List<IBHoMObject> newObjs = new List<IBHoMObject>();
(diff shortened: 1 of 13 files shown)


alelom:
@al-fisher I've applied the most recent corrections as discussed and tested thoroughly. 
This is ready for you to review. Thanks

PR #1650, opened 2020-04-09
Structure_Engine: Add comparer for name or description for IProperty

### LINKED ISSUE #1398: Structure_Engine: Add comparer for Name or Description
#### Description:

With Description method added for all structural properties, a new comparer can be added, comparing name and description. This will be useful for all structural properties when pushed with the adapter and make it possible to push over properties without pre-assigned names.

IsakNaslundBh:

Closes #1398 

Adding a new comparer making use of the DescriptionOrName method, that first checks for the name, and if empty, uses the autogenerated description.

To be used across most structural adapters for Property type obejcts.

Also added missing description for LinkConstraint

### Test files

See test case for method here https://github.com/BHoM/Robot_Toolkit/pull/353

### DIFF
--- a/Structure_Engine/Query/Description.cs
+++ b/Structure_Engine/Query/Description.cs
@@ -355,11 +355,38 @@ namespace BH.Engine.Structure
         /***************************************************/
 
         [Description("Generates a default description for the BarRelease as 'StartReleaseNameOrDesc - EndReleaseNameOrDesc'.")]
-        [Input("release", "The Constraint6DOF to get a default description for.")]
-        [Output("desc", "The generated description for the constraint.")]
+        [Input("release", "The BarRelease to get a default description for.")]
+        [Output("desc", "The generated description for the BarRelease.")]
         public static string Description(this BarRelease release)
         {
             return release.StartRelease.DescriptionOrName() + " - " + release.EndRelease.DescriptionOrName();
+        }
+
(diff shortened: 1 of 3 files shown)


JosefTaylor:
Why does it check if the property and `null` reference the same object, rather than checking if the property is `null`?

Structure_Engine/Query/Description.cs
```diff
@@ -355,13 +355,40 @@ public static string Description(this Constraint6DOF constraint)
...
+        [Output("desc", "The generated description for the BarRelease.")]
         public static string Description(this BarRelease release)
         {
             return release.StartRelease.DescriptionOrName() + " - " + release.EndRelease.DescriptionOrName();
         }
 
+        /***************************************************/
+
+        [Description("Generates a default description for the LinkConstraint based on its fixities where x denoted fixed and f denotes free.")]
+        [Input("release", "The LinkConstraint to get a default description for.")]
```
FraserGreenroyd, line 368: CheckCompliance will flag this up, but I'll comment here to highlight it for you @IsakNaslundBh - input attribute name doesn't match the method argument name.

IsakNaslundBh: 🤦

FraserGreenroyd: ❤️

PR #1655, opened 2020-04-14
Reflection_Engine: Migrate Dummy and IsEqual methods to the Test Toolkit

### LINKED ISSUE #1653: Reflection_Engine: Kellerman library causing load errors
#### Description:
Getting an error when opening Revit after building off the BHoM_Engine when it includes the Kellerman dll reference added in #1643 as per below. Have not been able to dig deeply into it but I would assume similar issues exist elsewhere, if it is specific to www.github.com/BHoM/Revit_Toolkit we can move this issue there.

image

#### Expected behaviour:
Revit opens as usual with BHoM Addin working.

#### Test file(s):
Build, open Revit 2020.

adecler:
Closes #1653 

Move those method to eliminate the dependency to the Kellerman Compare-Object library that is causing issues in code executed outside the BHoM assembly folder.

### Test files
Just start Revit and make sure the error highlighted in issues #1653 is not present anymore.


pawelbaran:
I got rid of that by cleaning up all Build folders and rebuilding all assemblies and works fine for me now, so cannot test myself. @enarhi @kayleighhoude maybe you can give it a shot?

PR #1659, opened 2020-04-14
Structure_Engine: Off-centre section issues

### LINKED ISSUE #1657: Structure_Engine: Off-centre section issues
#### Description:

https://github.com/BHoM/BHoM_Engine/blob/9ea9155f8dcd6ed0fdb020e1ae99e2f2f39790eb/Structure_Engine/Compute/IntegrateSection.cs#L97
This creates a singular `BoundingBox` at Origo, which will then be a part of the total `BoundingBox` for the section. This means that the placement of the section curves in the XY-plane affects the section properties. This shouldn't happen as the section after this has it's position adjusted, with no corresponding change in values.

IMG: four identical section edges with different section property values
image

FYI @IsakNaslundBh 

Would suggest adding a metod in Geometry_Engine which gets the total bounding box from a collection of `BoundingBoxes` to reduce the amount of people who make this mistake
#### Steps to reproduce:

Create a freefrom section away from the origo on the XY plane

#### Expected behaviour:

the edges on the section should have the same property values as the section

#### Test file(s):

https://burohappold.sharepoint.com/:f:/s/BHoM/EpXcV7sd78dOksKEPpfFg3oB4AgrgAKGQ18kYHcORuzmsA?e=5hG7pm

### LINKED ISSUE #1658: Geometry_Engine: Add Bounds(List<BoundingBox> boxes)
#### Description:
A method which gets the combined `BoundingBox` for a collection of boxes

would reduce incidents like these: #1657

kThorsager:
Closes #1657 
Closes #1658 

### Test files
https://burohappold.sharepoint.com/:f:/s/BHoM/EsCnlfKi1CVKl2RxG-SXCEUBuXEWwwikU8m6BwmSjoqaVQ?e=7tKiR3

### Changelog
- Adds Geometry.Query.Bounds(List<BoundingBox> boxes)

### DIFF
--- a/Geometry_Engine/Query/Bounds.cs
+++ b/Geometry_Engine/Query/Bounds.cs
@@ -25,6 +25,7 @@ using BH.oM.Geometry.CoordinateSystem;
 using BH.oM.Reflection.Attributes;
 using System;
 using System.Collections.Generic;
+using System.Linq;
 
 namespace BH.Engine.Geometry
 {
@@ -382,6 +383,27 @@ namespace BH.Engine.Geometry
             }
 
             return new BoundingBox { Min = new Point { X = minX, Y = minY, Z = minZ }, Max = new Point { X = maxX, Y = maxY, Z = maxZ } };
+        }
+
+        /***************************************************/
+
+        public static BoundingBox Bounds(this List<BoundingBox> boxes)
+        {
+            return new BoundingBox
+            {
+                Min = new Point
+                {
+                    X = boxes.Min(x => x.Min.X),
+                    Y = boxes.Min(x => x.Min.Y),
+                    Z = boxes.Min(x => x.Min.Z)
+                },
+                Max = new Point
+                {
+                    X = boxes.Max(x => x.Min.X),
+                    Y = boxes.Max(x => x.Min.Y),
+                    Z = boxes.Max(x => x.Min.Z)
+                }
+            };
         }
 
         /***************************************************/
--- a/Geometry_Engine/Query/Bounds.cs
+++ b/Geometry_Engine/Query/Bounds.cs
@@ -399,9 +399,9 @@ namespace BH.Engine.Geometry
                 },
                 Max = new Point
                 {
-                    X = boxes.Max(x => x.Min.X),
-                    Y = boxes.Max(x => x.Min.Y),
-                    Z = boxes.Max(x => x.Min.Z)
+                    X = boxes.Max(x => x.Max.X),
+                    Y = boxes.Max(x => x.Max.Y),
+                    Z = boxes.Max(x => x.Max.Z)
                 }
             };
         }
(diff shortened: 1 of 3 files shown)


Structure_Engine/Compute/Integrate.cs
```diff
@@ -51,12 +51,7 @@ public static partial class Compute
         {
             Dictionary<string, object> results = new Dictionary<string, object>();
 
-            BoundingBox box = new BoundingBox();
-
-            for (int i = 0; i < curves.Count; i++)
-            {
-                box += curves[i].IBounds();
-            }
+            BoundingBox box = Geometry.Query.Bounds(curves.Select(x => x.IBounds()).ToList());
```
epignatelli, line 54: Have you tried using `curvers.Sum(x => x.IBounds())`? I think it should do the same thing and can avoid adding a new method?

Structure_Engine/Compute/IntegrateSection.cs
```diff
@@ -94,9 +94,7 @@ public static partial class Compute
                 area += Math.Abs(areaTemp) * depth[i];
             }
 
-            BoundingBox box = new BoundingBox();
-            for (int i = 0; i < curvesZ.Count; i++)
-                box += curvesZ[i].IBounds();
+            BoundingBox box = Geometry.Query.Bounds(curvesZ.Select(x => x.IBounds()).ToList());
```
epignatelli, line 97: Ditto

Geometry_Engine/Query/Bounds.cs
```diff
@@ -386,6 +387,27 @@ public static BoundingBox Bounds(this List<Point> pts)
 
         /***************************************************/
 
+        public static BoundingBox Bounds(this List<BoundingBox> boxes)
```
epignatelli, line 390: If we want to keep this, we can just use the operator `+` between `BoundingBox`es maybe?

Structure_Engine/Compute/Integrate.cs
kThorsager: Yes, `Sum` only returns numbers, so no dice

Geometry_Engine/Query/Bounds.cs
kThorsager: If you'd like, I just went straight for the definition, (mostly for the aesthetics of avoiding assigning the first value and then looping over the rest of them, but maybe silly)

Structure_Engine/Compute/Integrate.cs
kThorsager: I did however now find this way:
`            return boxes.Aggregate((x, y) => x + y);
`
which ofc would work, but adding the method was also to avoid the mistake in #1657 by having a method for it, (and we have one for `List<Point>` which can also be solved in the same way, (know that two wrongs don't make it right, but still))

epignatelli:
@BHoM/bhom_engine can merge for me if checks succeed!

FraserGreenroyd:
I'd want to see a review from @IsakNaslundBh as there is also structure engine changes just to be double-safe :smile:

PR #1666, opened 2020-04-16
Diffing_Engine: allow to use Id stored in CustomData; fix Explode `modifiedPropertiesPerObj`

### LINKED ISSUE #1660: Diffing_Engine: change `ModifiedPropertiesPerObject` string keys to avoid Explode component issue
#### Description:

Connected to https://github.com/BHoM/BHoM/pull/776
and the following https://github.com/BHoM/BHoM_UI/issues/241

We decided to avoid the problem by changing `ModifiedPropertiesPerObject` string key to avoid including `.` "full stops" in order to maintain existing functionality.

The full stop is to be replaced by another character.

### LINKED ISSUE #1665: Diffing_Engine: allow to use an Id stored in CustomData instead of the Hash
#### Description:

Allow to use an Id stored in CustomData instead of the Hash.

##### Why we need it
To fully use the Diffing we'd need to ensure that we can store the object hashes (current and previous) into the external software. 

That may be possible for a variety of software: the ones where we can save the tag. We can rely on the tag mechanism to do that, assuming the field is long enough to store two hashes.

If that is not possible, this feature delegates the "object recognition" to the external software.  
This relies on the assumption that **if an object is modified, its CustomData id is retained**.

This assumption may well be too much of a simplification in many cases, but at least it enables more use workflows where that is acceptable for now.

alelom:

Closes #1665 
Closes #1660 

This PR primarily adds the feature where you can use an Id stored in CustomData instead of the object Hash, as explained in #1665.

Additionally, it also fixes the Explode not working properly for the `Diff` property called `modifiedPropertiesPerObj`, see #1660. It does so:
- by implementing a workaround to make the Explode work. However, going through the various exploded levels is not very user friendly. Therefore:
- it adds an Engine method `ListModifiedProperties` that "properly" queries and displays the data.

### Test files

@kayleighhoude this should enable the workflow in Revit. 

image

Workflow:
- Create your model in Revit
- Pull it in Grasshopper. Internalize the objects somewhere.
- Go back to Revit. Modify the model.
- Pull it again in Grasshopper.
- Attach the internalized objects and the newly pulled model to the `Diffing` Engine method as per image above. You need `customdataIdName` to be set to ~~`Revit_elementId`~~ `Revit_uniqueId`.
- Check the diff:
  - explode the Diff
  - check what properties have been modified using the `ListModifiedProperties` method as in the picture above. 

### Changelog

- Can now input a CustomData Id Name to the Diffing method so it will be using a certain software id to recognise objects instead of hashes.
- Added `ListModifiedProperties` method to better view the Diff property `modifiedPropertiesPerObj`
- Added workaround to make BHoM Explode work properly with `modifiedPropertiesPerObj`


FraserGreenroyd:
@alelom can you confirm the status of this PR please as I note it has been 10 days since any activity on it?

alelom:
@kayleighhoude any chance you could have a look if this solves your issue?

@michaelhoehn adding you as a reviewer, perhaps you'd like to have a look?

kayleighhoude:
@alelom Just tested and got the following error: 
image

alelom:
> @alelom Just tested and got the following error:
 Thanks! Could I get the test file and Revit file you tested with?

alelom:
@kayleighhoude Actually I might have suggested the wrong Revit Id to use. You might want to try using `Revit_UniqueId` instead of `Revit_ElementId`. I've been told that in Revit the id called `Revit_Unique ID` is more "stable" and, as the name implies, should ensure Uniqueness per each object. However it worked for me. 
If you keep having the error even with Revit_UniqueId, please leave a link with the script and Revit test file you are using, so I can debug.

al-fisher:
Hey @alelom - I am getting the same error as @kayleighhoude - and the `Revit_uniqueId` is not working as a key either. Worth a look together tomorrow

al-fisher:
@alelom - have raised an issue to capture our discussion - https://github.com/BHoM/BHoM_Engine/issues/1730

PR #1682, opened 2020-04-21
Spatial_Engine: Add PointLayout methods and Create methods for Layout objects

### LINKED ISSUE #1608: Spatial_Engine: Add method for getting out points in XY plane from a Layout object
#### Description:

Related to https://github.com/BHoM/BHoM/issues/758

All Layouts should have a method like

`public static List<Point> PointLayout(this ILayout2D layout, ICurve hostRegionCurve)`

generating all the base points for the particular layout object. name for the curve parameter to be agreed.

IsakNaslundBh:
### NOTE: Depends on 

https://github.com/BHoM/BHoM/pull/811
   
Closes #1608 

- Add constructors for Layout obejcts that requires it.
- Add PointLayout method for all Layout obejcts, generating a layout of points from a layout object and a region curve, with potential openings.

### Test files

https://burohappold.sharepoint.com/:f:/s/BHoM/Emvmzm2j1xBNhG1JBGXJhUwBfpkHzOg5NOSmEexzbWxDBw?e=dspVaa

### Additional comments

The PointLayout methods became a lot more complex than I first anticipated once I started to account for various edgecases, but seem to be working now. For most use cases a much simpler approach could have been used, but wanted it sort out as many cases as possible from the get-go, hence the quite long file..


kThorsager:
Annoying question this late but are things like discontinuous sections simply not the Layouts problem? (As the Host Curve is a single curve)

kThorsager:
This feels like an edge case which can be picked up later, but as the points are always added to the curves one can get points closer than the parallel distance from each other
image

IsakNaslundBh:
> This looks fantastic! I found a condition in which the multilinear layout doesn't pack as many in as it could:
> image

Thanks @JosefTaylor , could you give me the setup so I can try to debug?

IsakNaslundBh:
> This feels like an edge case which can be picked up later, but as the points are always added to the curves one can get points closer than the parallel distance from each other
> image

Well found... more edgecases. WIll have a look if I can do anything quick to fix it!

JosefTaylor:
> Thanks @JosefTaylor , could you give me the setup so I can try to debug?

@IsakNaslundBh I've put a .GH in the test script folder

Spatial_Engine/Query/PointLayout.cs
```diff
@@ -0,0 +1,389 @@
...
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Returns the points in the layout. Will not take the host geometry into account for ExplicitLayout")]
+        [Input("layout2D", "The layout object to query the points from.")]
+        [Input("hostRegionCurve", "The region curve of the objects associated with the layout. Unused for Explicit layouts.")]
+        [Input("openingCurves", "Optional opening curves in the region. Unused for ExplicitLayout.")]
+        [Output("points", "The Points stored in the ExplicitLayout object.")]
+        public static List<Point> PointLayout(this ExplicitLayout layout2D, ICurve hostRegionCurve, List<ICurve> openingCurves = null)
```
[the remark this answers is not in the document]
IsakNaslundBh: true, could add a "=null"

IsakNaslundBh:
Had to update from master, as this is now lagging behind BHoM and BHoM_UI on master will not build without the changes.

IsakNaslundBh:
> Annoying question this late but are things like discontinuous sections simply not the Layouts problem? (As the Host Curve is a single curve)

Good point. Wonder if that is worth tackling now, or better left for future iteration.

I have made a first fix for your findings of things to close together. Decided not to try to tackle every single edgecase of it, but at least now handles it as long as the distribution lines can be moved apart without collapsing completely.

Spatial_Engine/Query/PointLayout.cs
```diff
@@ -381,6 +385,32 @@ private static List<Line> IntersectionLines(this ICurve hostRegionCurve, List<IC
...
+                        if (result[i].Length() > diff / 2 && result[i + 1].Length() > diff / 2)
+                        {
+                            result[i].End -= tan * diff / 2;
+                            result[i + 1].Start += tan * diff / 2;
+                        }
+                        else
+                        {
+                            //For now, if the above does not work, leave a warning and continue.
+                            //TODO: make this work for these cases as well.
+                            Engine.Reflection.Compute.RecordWarning("The ends om some distribution lines are closer together than then minimumspacing, but could not be separated automatically. Please check the result of the Pointlayout distribution.");
```
JosefTaylor, line 408:
```suggestion
                            Engine.Reflection.Compute.RecordWarning("The ends of some distribution lines are closer together than the minimum spacing, but could not be separated automatically. Please check the result of the PointLayout distribution.");
```

Spatial_Engine/Query/PointLayout.cs
```diff
@@ -0,0 +1,419 @@
...
+                return new List<Point>();
+            }
+
+            return DivisionPoints(distributionLines, layout2D.NumberOfPoints);
+        }
+
+        /***************************************************/
+
+
+        [Description("Returns Points along straight lines throughout the region. If the line is segemented by openings or the region curve the points will be distributed to the different segments based on their length. \n" +
```
JosefTaylor, line 139:
```suggestion
        [Description("Returns Points along a straight line through the region curve. If the line is discontinuous because of openings or concave host region, Points will be distributed to the different segments based on their length.\n" +
```

IsakNaslundBh:
> > Annoying question this late but are things like discontinuous sections simply not the Layouts problem? (As the Host Curve is a single curve)
> 
> Good point. Wonder if that is worth tackling now, or better left for future iteration.

Changed my mind regarding this,. will tackle it straight away instead to avoid having to come back to this yet another time.

Spatial_Engine/Create/LinearLayout.cs
```diff
@@ -0,0 +1,65 @@
...
+        [Output("linLayout", "The created LinearLayout.")]
+        public static LinearLayout LinearLayout(int numberOfPoints, Vector direction = null, double offset = 0, ReferencePoint referencePoint = ReferencePoint.BottomCenter)
+        {
+            if (numberOfPoints <= 0)
+            {
+                Engine.Reflection.Compute.RecordError("LinearLayout requires numberOfPoints to be at least 1.");
+                return null;
+            }
+            Vector projDir = direction ?? Vector.XAxis;
+            if (direction.Z != 0)
```
kThorsager, line 54: Can't get direction.Z if it's `null`, what you get is an hard error

Spatial_Engine/Query/PointLayout.cs
```diff
@@ -164,7 +164,7 @@ public static List<Point> PointLayout(this LinearLayout layout2D, IEnumerable<IC
         /***************************************************/
 
 
-        [Description("Returns Points along straight lines throughout the region. If the line is segemented by openings or the region curve the points will be distributed to the different segments based on their length. \n" +
+        [Description("Returns Points along one or more straight lines through the region curves. If the line is discontinuous because of openings or concave host region, Points will be distributed to the different segments based on their length.\n" +
```
JosefTaylor, line 167:
```suggestion
        [Description("Returns Points along one or more straight lines through the region curves. If the line is discontinuous because of openings, a concave host region, or multiple host regions, Points will be distributed to the different segments based on their length.\n" +
```

kThorsager:
Getting duplicate Points on the Perimeter when I enforce Discontinuity Points
(The highly saturated one are two)
image
Updated the testfile

Spatial_Engine/Query/PointLayout.cs
```diff
@@ -0,0 +1,457 @@
...
+                }
+                else
+                {
+                    int divisions = curveDivs;
+                    bool closed = regionCurve.IIsClosed();
+
+                    if (!closed)
+                        divisions--;
+
+                    List<Point> pts = regionCurve.SamplePoints(divisions);
```
kThorsager, line 126: Perimeter with divisions = 1 (or less, no checks on that) send 0 or less to sample points, which throws an exception.

kThorsager: Realise extra important as, if the number of host region curves are < 2*curveDivisions this problem will appear (this part is called for each host curve)

Spatial_Engine/Query/PointLayout.cs
```diff
@@ -0,0 +1,457 @@
...
+        [Input("openingCurves", "Optional opening curves in the region. Unused for PerimeterLayout.")]
+        [Output("points", "Point layout generated by the layout objects and region curve.")]
+        public static List<Point> PointLayout(this PerimeterLayout layout2D, IEnumerable<ICurve> hostRegionCurves, IEnumerable<ICurve> openingCurves = null)
+        {
+            List<ICurve> regionCurveList = hostRegionCurves.ToList();
+            List<int> divs = new List<int>();
+            if (regionCurveList.Count == 1)
+                divs.Add(layout2D.NumberOfPoints);
+            else
+                divs = DistributeDivisions(regionCurveList, layout2D.NumberOfPoints);
```
kThorsager, line 77: The divisions per curve are calculated before the discontinuity points are considered, meaning all divisions can go to the bigger curve, which won't have enough divisions to make up the difference.

Here I asked for 10 points and got 17, (extreme case)
image

IsakNaslundBh: This was actually intentional. Considered whether to check fro number of points before/after and went with before to as it ensures long curves get more points. If checking after, a large circle in relation to a small polyline with multiple point would give kind of strange results.

Issue #1676, opened 2020-04-20
Spatial_Enigne: Add RoundCoordinates


kThorsager:
#### Description:

Extending the functionality of #1675 to `IElement` objects

PR #1684, opened 2020-04-21
Geometry_Engine: Add RoundCoordinates

### LINKED ISSUE #1675: Geometry_Enigne: Add RoundCoordinates
#### Description:

A method to round the coordinates of a geometrical object in such a way to make it suitable for various softwares

Separate issue raised about removing `RoundPoint` for compliance

kThorsager:

Closes #1675 
Adds `RoundCoordinates` which rounds the coordinates of a `IGeometry`, currently supports `ICurves`

#1676 will be addressed once Spatial_Engine clears up

### Test files

https://burohappold.sharepoint.com/:u:/s/BHoM/EeDbV5uQCIFEgOxEzGM3FuUBey_qVsrbuNa4JPjrtGpV0g?e=KPhnSe

### Changelog

- Deprecates `RoundPoint`

### DIFF
--- a/Geometry_Engine/Modify/RoundCoordinates.cs
+++ b/Geometry_Engine/Modify/RoundCoordinates.cs
@@ -36,7 +36,7 @@ namespace BH.Engine.Geometry
         /**** Interface Methods                         ****/
         /***************************************************/
 
-        [Description("Modifies a BHoM IGeometry's Points' coordinates to be rounded to the number of provided decimal places.")]
+        [Description("Modifies a BHoM IGeometry's coordinates to be rounded to the number of provided decimal places.")]
         [Input("geometry", "The BHoM IGeometry to modify.")]
         [Input("decimalPlaces", "The number of decimal places to round to, default 6.")]
         [Output("geometry", "The modified BHoM IGeometry.")]
@@ -47,7 +47,7 @@ namespace BH.Engine.Geometry
 
         /***************************************************/
 
-        [Description("Modifies a BHoM ICurve's Points coordinates to be rounded to the number of provided decimal places.")]
+        [Description("Modifies a BHoM ICurve's coordinates to be rounded to the number of provided decimal places.")]
         [Input("curve", "The BHoM ICurve to modify.")]
         [Input("decimalPlaces", "The number of decimal places to round to, default 6.")]
         [Output("curve", "The modified BHoM ICurve.")]
@@ -77,7 +77,7 @@ namespace BH.Engine.Geometry
 
         /***************************************************/
 
-        [Description("Modifies a BHoM Geometry Line's end points to be rounded to the number of provided decimal places.")]
+        [Description("Modifies a BHoM Geometry Line's points to be rounded to the number of provided decimal places.")]
         [Input("line", "The BHoM Geometry Line to modify.")]
         [Input("decimalPlaces", "The number of decimal places to round to, default 6.")]
         [Output("line", "The modified BHoM Geometry Line.")]
@@ -92,7 +92,7 @@ namespace BH.Engine.Geometry
 
         /***************************************************/
 
-        [Description("Modifies a BHoM Geometry Arc's end points to be rounded to the number of provided decimal places.")]
+        [Description("Modifies a BHoM Geometry Arc's start and end coordinates to be rounded to the number of provided decimal places.")]
         [Input("arc", "The BHoM Geometry Arc to modify.")]
         [Input("decimalPlaces", "The number of decimal places to round to, default 6.")]
         [Output("curve", "The modified BHoM Geometry Arc. As Line if zero length.")]
--- a/Geometry_Engine/Modify/RoundCoordinates.cs
+++ b/Geometry_Engine/Modify/RoundCoordinates.cs
@@ -123,7 +123,7 @@ namespace BH.Engine.Geometry
 
         /***************************************************/
 
-        [Description("Modifies a BHoM Geometry Arc's start and end coordinates to be rounded to the number of provided decimal places.")]
+        [Description("Modifies a BHoM Geometry Arc's start and end coordinates to be rounded to the number of provided decimal places while maintaining the Arc's total angle.")]
         [Input("arc", "The BHoM Geometry Arc to modify.")]
         [Input("decimalPlaces", "The number of decimal places to round to, default 6.")]
         [Output("curve", "The modified BHoM Geometry Arc.")]
@@ -132,7 +132,6 @@ namespace BH.Engine.Geometry
             // do the rounding
             Point start = arc.StartPoint().RoundCoordinates(decimalPlaces);
             Point end = arc.EndPoint().RoundCoordinates(decimalPlaces);
-            Vector normal = arc.CoordinateSystem.Z.RoundCoordinates(decimalPlaces);
 
             double angle = arc.Angle();
             double dist = start.Distance(end);
@@ -157,6 +156,9 @@ namespace BH.Engine.Geometry
                 Math.Pow(dist / 2, 2)   // "half the base"
                 );
 
+            // Align the normal to the new endpoints
+            Vector normal = arc.CoordinateSystem.Z.CrossProduct(end - start).CrossProduct(start - end).Normalise();
+
             Circle startCircle = new Circle() { Normal = normal, Centre = start, Radius = radius };
             Circle endCircle = new Circle() { Normal = normal, Centre = end, Radius = radius };
 
@@ -174,10 +176,11 @@ namespace BH.Engine.Geometry
             {
                 // Ensure that the new centre is at the same side of the start/end points
                 Vector unitNormal = normal.Normalise();
+                unitNormal *= angle > Math.PI ? -1 : 1;
                 foreach (Point pt in intersections)
                 {
                     Vector temp = (start - pt).CrossProduct(end - pt).Normalise();
-                    if ((temp + unitNormal).SquareLength() > 1)
+                    if ((temp + unitNormal).SquareLength() >= 1)
                     {
                         newOrigin = pt;
                         break;
@@ -189,6 +192,7 @@ namespace BH.Engine.Geometry
             oM.Geometry.CoordinateSystem.Cartesian coordClone = Create.CartesianCoordinateSystem(newOrigin, newX, Query.CrossProduct(normal, newX));
 
             double endAngle = (start - newOrigin).Angle(end - newOrigin);
+            endAngle = angle > Math.PI ? 2 * Math.PI - endAngle : endAngle; 
 
             Arc result = new Arc()
             {
--- a/Geometry_Engine/Modify/RoundCoordinates.cs
+++ b/Geometry_Engine/Modify/RoundCoordinates.cs
@@ -77,6 +77,37 @@ namespace BH.Engine.Geometry
 
         /***************************************************/
 
+        [Description("Modifies a BHoM Geometry Vector to be rounded to the number of provided decimal places.")]
+        [Input("vector", "The BHoM Geometry Vector to modify.")]
+        [Input("decimalPlaces", "The number of decimal places to round to, default 6.")]
+        [Output("vector", "The modified BHoM Geometry Vector.")]
+        public static Vector RoundCoordinates(this Vector vector, int decimalPlaces = 6)
+        {
+            return new Vector
+            {
+                X = Math.Round(vector.X, decimalPlaces),
+                Y = Math.Round(vector.Y, decimalPlaces),
+                Z = Math.Round(vector.Z, decimalPlaces),
+            };
+        }
+
+        /***************************************************/
+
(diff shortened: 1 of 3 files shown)


Geometry_Engine/Modify/RoundCoordinates.cs
```diff
@@ -0,0 +1,215 @@
...
+
+namespace BH.Engine.Geometry
+{
+    public static partial class Modify
+    {
+        /***************************************************/
+        /**** Interface Methods                         ****/
+        /***************************************************/
+
+        [Description("Modifies a BHoM IGeometry's Points' coordinates to be rounded to the number of provided decimal places.")]
```
FraserGreenroyd, line 39: What's the difference between `Points'` and `coordinates` in this instance? Would it be better to say `BHoM IGeometry's coordinates` instead?

Geometry_Engine/Modify/RoundCoordinates.cs
```diff
@@ -0,0 +1,215 @@
...
+        [Input("decimalPlaces", "The number of decimal places to round to, default 6.")]
+        [Output("geometry", "The modified BHoM IGeometry.")]
+        public static IGeometry IRoundCoordinates(this IGeometry geometry, int decimalPlaces = 6)
+        {
+            return RoundCoordinates(geometry as dynamic, decimalPlaces);
+        }
+
+        /***************************************************/
+
+        [Description("Modifies a BHoM ICurve's Points coordinates to be rounded to the number of provided decimal places.")]
```
FraserGreenroyd, line 50: As above

Geometry_Engine/Modify/RoundCoordinates.cs
```diff
@@ -0,0 +1,215 @@
...
+            {
+                X = Math.Round(point.X, decimalPlaces),
+                Y = Math.Round(point.Y, decimalPlaces),
+                Z = Math.Round(point.Z, decimalPlaces),
+            };
+        }
+
+        /***************************************************/
+
+        [Description("Modifies a BHoM Geometry Line's end points to be rounded to the number of provided decimal places.")]
```
FraserGreenroyd, line 80: Given a BHoM Line has a definition for an `End` point I wonder if we should change this description to remove the word `end` in the description to avoid confusion given the method is editing both start and end points.

Geometry_Engine/Modify/RoundCoordinates.cs
```diff
@@ -0,0 +1,215 @@
...
+            return new Line()
+            {
+                Start = line.Start.RoundCoordinates(decimalPlaces),
+                End = line.End.RoundCoordinates(decimalPlaces)
+            };
+        }
+
+        /***************************************************/
+
+        [Description("Modifies a BHoM Geometry Arc's end points to be rounded to the number of provided decimal places.")]
```
FraserGreenroyd, line 95: See above

kThorsager: Changed to `start and end coordinates` to clarify which parameters of the `Arc` was modified

kThorsager:
The arc became a lot nicer, but I did diverge from your instructions somewhat:

- mostly vector comparison to pick the correct intersection (distance to old centre was unreliable for "heavy" rounding)
- Made the invariant part the total angle of the arc instead of the radius, which made the transformation more akin to scaling and looked a loot less messy
- Made a brand new Coordinate system as I realised we knew the normal

Not strictly not what you said but good to highlight

Geometry_Engine/Modify/RoundCoordinates.cs
```diff
@@ -0,0 +1,309 @@
...
+                    StartAngle = arc.StartAngle,
+                    EndAngle = arc.StartAngle,
+                };
+            }
+
+            // recalculate the radius based on not changing the total angle
+            //      Consider a equal legged triangle with endpoints at the arc's endpoints
+            //      we know the "top" angle and the "base" length and are solving for the last two sides length
+            double radius = Math.Sqrt(
+                Math.Pow(dist / (2 * Math.Tan(angle / 2)), 2) + // "Heigth"
```
pawelbaran, line 156: height

Geometry_Engine/Modify/RoundCoordinates.cs
```diff
@@ -0,0 +1,309 @@
...
+
+            List<Point> intersections = startCircle.CurveIntersections(endCircle).OrderBy(x => x.SquareDistance(arc.CoordinateSystem.Origin)).ToList();
+
+            Point newOrigin = null;
+            // 180degrees arc where the points got rounded away from eachother
+            if (intersections.Count == 0)
+            {
+                newOrigin = (start + end) / 2;
+                radius = newOrigin.Distance(start);
+            } else
```
pawelbaran, line 173: else in a new line

Geometry_Engine/Modify/RoundCoordinates.cs
```diff
@@ -0,0 +1,309 @@
...
+
+            // recalculate the radius based on not changing the total angle
+            //      Consider a equal legged triangle with endpoints at the arc's endpoints
+            //      we know the "top" angle and the "base" length and are solving for the last two sides length
+            double radius = Math.Sqrt(
+                Math.Pow(dist / (2 * Math.Tan(angle / 2)), 2) + // "Heigth"
+                Math.Pow(dist / 2, 2)   // "half the base"
+                );
+
+            //double radius = arc.Radius;
```
pawelbaran, line 160: delete if not needed

Geometry_Engine/Modify/RoundCoordinates.cs
```diff
@@ -0,0 +1,309 @@
...
+        /***************************************************/
+
+        [Description("Modifies a BHoM Geometry PlanarSurface's defining curves to be rounded to the number of provided decimal places.")]
+        [Input("nurbscurve", "The BHoM Geometry PlanarSurface to modify.")]
+        [Input("decimalPlaces", "The number of decimal places to round to, default 6.")]
+        [Output("surface", "The modified BHoM Geometry PlanarSurface.")]
+        public static PlanarSurface RoundCoordinates(this PlanarSurface planarSurface, int decimalPlaces = 6)
+        {
+            ICurve externalBoundery = planarSurface.ExternalBoundary.IRoundCoordinates(decimalPlaces);
+            List<ICurve> internalBounderies = planarSurface.InternalBoundaries.Select(x => x.IRoundCoordinates(decimalPlaces)).ToList();
```
pawelbaran, line 291: Besides that, this is likely to invalidate the planarity requirement set to `PlanarSurface`. Thinking about it now, I cannot imagine how this could work - I would rather tend to record a warning that would explain the problem and recommend extracting plane => rounding it => projecting the surface on rounded plan as a semi-manual workaround. Return item could simply be the input (or null, not sure tbh).

Plus, boundAry

kThorsager: > extracting plane => rounding it => projecting the surface on rounded plan as a semi-manual workaround.

Don't think that would round the coordinates since every point on a rounded Plane isn't a rounded point.

And I was thinking that for:

- 90% would be in an orthogonal plane (XY, YZ, XZ), and hence work,
- 9% would be vertical quads, and hence work,
- the rest would fail but didn't really know of a way to fix that

Thinking about it some more I think one could essentially transform the plane down to the XY-plane, round there and reverse.
Would be kind of finicky to get the correct transformation where an arbitrary planes would intersect the rounded coordinates to the rounded coordinates on the XY plane.

But think that kind of functionality can wait. (and would need to be discussed around how it would act)

kThorsager:
> plus the Arc looks very complicated (even more complicated that in the first iteration?) - possibly it needs to be like that, but may be a good idea to give it another shot in the morning when cleaning up the typos I've marked in the code comments.

Can't really dispute that, but `Arc` s have proven them self to be complicated from time to time, but I'll give it another shot

Geometry_Engine/Modify/RoundCoordinates.cs
pawelbaran: Hmm, so maybe a solution half way:
- if normal of the surface is parallel to one of the global axes, then round its plane and then round the outline curves
- else record a warning and recommend a semi-manual solution

Could be an easy win for most cases while being safe for others?

Geometry_Engine/Modify/RoundCoordinates.cs
```diff
@@ -0,0 +1,323 @@
...
+        public static PlanarSurface RoundCoordinates(this PlanarSurface planarSurface, int decimalPlaces = 6)
+        {
+            Vector normal = planarSurface.Normal().Normalise();
+
+            if (Math.Abs(Math.Abs(normal.X) - 1) < Tolerance.Distance ||
+                Math.Abs(Math.Abs(normal.Y) - 1) < Tolerance.Distance ||
+                Math.Abs(Math.Abs(normal.Z) - 1) < Tolerance.Distance)
+            {
+                ICurve externalBoundery = planarSurface.ExternalBoundary.IRoundCoordinates(decimalPlaces);
+                List<ICurve> internalBounderies = planarSurface.InternalBoundaries.Select(x => x.IRoundCoordinates(decimalPlaces)).ToList();
```
pawelbaran, line 301: Control points of these curves may be rounded to two different values due to noise, e.g.:
- point A being (1.0051,0,0)
- point B being (1.0049,1,0)

They will have different X coordinates when rounded to 2 digits, resulting in a twisted surface. A bit of an edge case, but slightly tilted surfaces have a long history in BHoM.

Plus, minor one: boundary not boundery ;)

Geometry_Engine/Modify/RoundCoordinates.cs
```diff
@@ -0,0 +1,323 @@
...
+        [Input("nurbscurve", "The BHoM Geometry PlanarSurface to modify.")]
+        [Input("decimalPlaces", "The number of decimal places to round to, default 6.")]
+        [Output("surface", "The modified BHoM Geometry PlanarSurface.")]
+        public static PlanarSurface RoundCoordinates(this PlanarSurface planarSurface, int decimalPlaces = 6)
+        {
+            Vector normal = planarSurface.Normal().Normalise();
+
+            if (Math.Abs(Math.Abs(normal.X) - 1) < Tolerance.Distance ||
+                Math.Abs(Math.Abs(normal.Y) - 1) < Tolerance.Distance ||
+                Math.Abs(Math.Abs(normal.Z) - 1) < Tolerance.Distance)
```
pawelbaran, line 298: Shouldn't these use `Tolerance.Angle`?

kThorsager: Now that you mention it it is an angle (assuming low tolerance)

Geometry_Engine/Modify/RoundCoordinates.cs
```diff
@@ -0,0 +1,324 @@
...
+                Math.Abs(Math.Abs(normal.Z) - 1) < Tolerance.Angle)
+            {
+                Plane plane = new Plane() { Origin = planarSurface.ExternalBoundary.IStartPoint().RoundCoordinates(decimalPlaces), Normal = normal.RoundCoordinates(0) };
+                ICurve externalBoundary = planarSurface.ExternalBoundary.IProject(plane).IRoundCoordinates(decimalPlaces);
+                List<ICurve> internalBoundaries = planarSurface.InternalBoundaries.Select(x => x.IProject(plane).IRoundCoordinates(decimalPlaces)).ToList();
+
+                return Create.PlanarSurface(externalBoundary, internalBoundaries);
+            }
+
+            Reflection.Compute.RecordWarning("Rounding the coordinates of a non-Axis-aligned planar surface while maintaining planarity needs to be done manually. No action has been taken.");
```
pawelbaran, line 307: Could slightly rephrase that to e.g.:

`Reflection.Compute.RecordWarning("Rounding the coordinates of a planar surface that is not aligned with global coordinate system cannot be achieved without risk of losing planarity. No action has been taken.");`

FraserGreenroyd: Is it worth clarifying `the global coordinate system` or `a global coordinate system`?

kThorsager:
For whatever reason I had downloaded my own file somewhere else and worked on it there, but I've uploaded it and refreshed the link

Geometry_Engine/Modify/RoundCoordinates.cs
```diff
@@ -0,0 +1,324 @@
...
+                ControlPoints = nurbscurve.ControlPoints.Select(x => x.RoundCoordinates(decimalPlaces)).ToList(),
+                Knots = nurbscurve.Knots.ToList(),
+                Weights = nurbscurve.Weights.ToList()
+            };
+        }
+
+        /***************************************************/
+
+        [Description("Modifies a BHoM Geometry PlanarSurface's defining curves to be rounded to the number of provided decimal places.")]
+        [Input("nurbscurve", "The BHoM Geometry PlanarSurface to modify.")]
```
FraserGreenroyd, line 289: Should be `planarSurface`

PR #1693, opened 2020-04-23
Versioning_Engine: method to easily generate method keys for versioning

adecler:
Provide helper method/component to go along with this PR: https://github.com/BHoM/BHoM_UI/pull/247

image

This makes it way easier to populate the inputs of a `PreviousVersion` attribute.

### DIFF
--- a/Versioning_Engine/Compute/VersionKey.cs
+++ b/Versioning_Engine/Compute/VersionKey.cs
@@ -48,8 +48,8 @@ namespace BH.Engine.Versioning
 
         [Description("Provide a string representation of a method as it used for versioning by the PreviousVersion attribute.")]
         [Input("declaringType", "Type in which the method is declared. You can use just the name of the type or include a (part of the) namespace in front of it.")]
-        [Input("methodName", "Name of the method. It has to be the exact string")]
-        [Output("Keys", "String representation for each method that matches the input filters.")]
+        [Input("methodName", "Name of the method. It has to be the exact string. If the method is a constructor, you can leave this input blank.")]
+        [Output("keys", "String representation for each method that matches the input filters.")]
         public static List<string> VersioningKey(string declaringType, string methodName = "")
         {
             if (methodName == "")
--- a/Versioning_Engine/Compute/VersionKey.cs
+++ b/Versioning_Engine/Compute/VersionKey.cs
@@ -1,0 +1,66 @@
+/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2020, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 
+ * GNU Lesser General Public License for more details.                          
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using BH.Engine.Reflection;
+using BH.oM.Base;
+using MongoDB.Bson;
+using MongoDB.Bson.IO;
+using MongoDB.Bson.Serialization;
+using MongoDB.Bson.Serialization.Serializers;
+using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Diagnostics;
+using System.IO;
+using System.IO.Pipes;
+using System.Linq;
+using System.Reflection;
+using System.Text;
+using System.Threading.Tasks;
+using BH.oM.Reflection.Attributes;
+
+namespace BH.Engine.Versioning
+{
+    public static partial class Compute
+    {
+        /***************************************************/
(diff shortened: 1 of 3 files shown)


Versioning_Engine/Versioning_Engine.csproj
```diff
@@ -37,6 +37,10 @@
     <Reference Include="MongoDB.Bson, Version=2.4.4.0, Culture=neutral, processorArchitecture=MSIL">
       <HintPath>..\packages\MongoDB.Bson.2.4.4\lib\net45\MongoDB.Bson.dll</HintPath>
     </Reference>
+    <Reference Include="Reflection_oM, Version=3.0.0.0, Culture=neutral, processorArchitecture=MSIL">
+      <SpecificVersion>False</SpecificVersion>
+      <HintPath>..\..\BHoM\Build\Reflection_oM.dll</HintPath>
+    </Reference>
```
FraserGreenroyd, line 43: Reference should not have version, culture, or processor architecture and should be set to `copy local` false

Versioning_Engine/Compute/VersionKey.cs
```diff
@@ -0,0 +1,66 @@
...
+    public static partial class Compute
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Provide a string representation of a method as it used for versioning by the PreviousVersion attribute.")]
+        [Input("declaringType", "Type in which the method is declared. You can use just the name of the type or include a (part of the) namespace in front of it.")]
+        [Input("methodName", "Name of the method. It has to be the exact string")]
+        [Output("Keys", "String representation for each method that matches the input filters.")]
```
FraserGreenroyd, line 52: I think consistency we use lowercase first letters on output names as with input names

Versioning_Engine/Query/VersionKey.cs
```diff
@@ -0,0 +1,75 @@
...
+{
+    public static partial class Query
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Provide a string representation of a method as it used for versioning by the PreviousVersion attribute.")]
+        [Input("methodName", "Method to ge tthe key from")]
+        [Output("Key", "String representation of the method as it will be used by the PreviousVersion attribute.")]
```
FraserGreenroyd, line 51: I think consistency we use lowercase first letters on output names as with input names

Versioning_Engine/Compute/VersionKey.cs
```diff
@@ -0,0 +1,66 @@
...
+{
+    public static partial class Compute
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Provide a string representation of a method as it used for versioning by the PreviousVersion attribute.")]
+        [Input("declaringType", "Type in which the method is declared. You can use just the name of the type or include a (part of the) namespace in front of it.")]
+        [Input("methodName", "Name of the method. It has to be the exact string")]
```
FraserGreenroyd, line 51: Looking at the code, it appears we can have a blank `methodName` to use `.ctor` options - is it worth including this in the description

Versioning_Engine/Query/VersionKey.cs
```diff
@@ -0,0 +1,75 @@
...
+namespace BH.Engine.Versioning
+{
+    public static partial class Query
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Provide a string representation of a method as it used for versioning by the PreviousVersion attribute.")]
+        [Input("methodName", "Method to ge tthe key from")]
```
FraserGreenroyd, line 50: Should probably be:
```suggestion
        [Input("methodName", "Method to get the key from")]
```

FraserGreenroyd:
Looks like everything is working as described for me from this code :smile:

image

adecler:
@FraserGreenroyd , I fixed the problems raised in your review

kThorsager:
Perhaps intended due to `Modify` -> `Compute` not working but if a method is found in both it returns nothing.
image

Worth flagging as that is a planned feature ([think this one](https://github.com/BHoM/Versioning_Toolkit/issues/51))

FraserGreenroyd:
> Perhaps intended due to `Modify` -> `Compute` not working but if a method is found in both it returns nothing.

I think from the demo @adecler gave this morning, the matching starts from the right and works left, so with those inputs it would be looking for `BH.Engine.Geometry.Join` which of course doesn't exist (based on my understanding of it).

Maybe a warning to the user to say that the provided thing couldn't be found might be worthwhile? Or if the issue you link is the right one then solving that at a later date.

kThorsager:
> > Perhaps intended due to `Modify` -> `Compute` not working but if a method is found in both it returns nothing.
> 
> I think from the demo @adecler gave this morning, the matching starts from the right and works left, so with those inputs it would be looking for `BH.Engine.Geometry.Join` which of course doesn't exist (based on my understanding of it).
> 
> Maybe a warning to the user to say that the provided thing couldn't be found might be worthwhile? Or if the issue you link is the right one then solving that at a later date.

No, sorry, re-read the description and sounds right (but thought the demo showed using just a part of the namespace)

FraserGreenroyd:
> No, sorry, re-read the description and sounds right (but thought the demo showed using just a part of the namespace)

It does, but it started from the right rather than the left. So I think something like `Modify` and `Join` would find the right one (or all the ones that match) because there is `*.Modify.Join` methods if that makes sense?

adecler:
A simple way to settle that discussion is to look at the code:
```c#
return Engine.Reflection.Query.AllMethodList()
                .Where(x => x.Name == methodName && x.DeclaringType.FullName.EndsWith(declaringType))
                .Select(x => x.VersioningKey())
                .ToList();
```

I was using `EndsWith` instead of `==` so that users wouldn't have to figure out whether to provide just the name of the declaring type or its full name (including namespace). The nice side effect was that yo could even do partial namespaces. I do however regret showing you that since it has been misinterpreted. The input name is `declaringType`. As far as I know, `BH.Engine.Geometry` is a namespace, not a declaring type.

PR #1708, opened 2020-04-29
Physical_Engine: Implement HasMergeablePropertiesWith

### LINKED ISSUE #1673: Physical_Engine: Implement HasMergeablePropertiesWith
#### Description:
The new required extension method: `HasMergeablePropertiesWith` needs to be implemented for all `IElement` objects

With priority on those which are defining the `IElement1D` edges of a `IElement2D`

##### IElement1D

- [ ] `IFramingElement`

##### IElement2D

- [ ] `IOpening`
- [ ] `ISurface`

kThorsager:

Closes #1673 

No Idea how to check if framing elements pointing towards each other with mirrored versions of each others profiles are correct. So either if there's a suggestion on how to do that or we'll put that  off in a new issue after some mirroring method for Profiles is implemented @IsakNaslundBh 

### Test files

https://burohappold.sharepoint.com/:f:/s/BHoM/Ej0PXCOWAKhIpjcljaqtlDEB_3RPJe3GN7-EHP6YpVVJVw?e=2R3ofC

### DIFF
--- a/Physical_Engine/Query/HasMergeablePropertiesWith.cs
+++ b/Physical_Engine/Query/HasMergeablePropertiesWith.cs
@@ -50,17 +50,20 @@ namespace BH.Engine.Physical
             if (element.GetType() != other.GetType())
                 return false;
 
-            if (!element.Geometry().IIsLinear() || !other.Geometry().IIsLinear())
+            if (element.Location == null || other.Location == null)
+                return false;
+
+            if (!element.Location.IIsLinear() || !other.Location.IIsLinear())
             {
                 Engine.Reflection.Compute.RecordWarning("No merge comparison avalible for non-linear IFramingElements.");
                 return false;
             }
 
-            int parallel = element.Geometry().IStartDir().IsParallel(other.Geometry().IStartDir());
+            int parallel = element.Location.IStartDir().IsParallel(other.Location.IStartDir());
             if (parallel != 1)
                 return false;
 
-            if (element.Normal() != other.Normal())
+            if (element.Normal().Angle(other.Normal()) < BH.oM.Geometry.Tolerance.Angle)
                 return false;
 
             if (element.Property == other.Property)
--- a/Physical_Engine/Query/HasMergeablePropertiesWith.cs
+++ b/Physical_Engine/Query/HasMergeablePropertiesWith.cs
@@ -1,0 +1,92 @@
+﻿/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2020, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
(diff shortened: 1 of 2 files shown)


Physical_Engine/Query/HasMergeablePropertiesWith.cs
```diff
@@ -0,0 +1,119 @@
...
+            {
+                Engine.Reflection.Compute.RecordWarning("No merge comparison avalible for non-linear IFramingElements.");
+                return false;
+            }
+
+            int parallel = element.Geometry().IStartDir().IsParallel(other.Geometry().IStartDir());
+            if (parallel != 1)
+                return false;
+
+            if (element.Normal() != other.Normal())
```
IsakNaslundBh, line 63: Would change to the following **NOTE THIS WILL NOT BUILD WITHOUT UPDATING THE USINGS to include geometry_oM** so might want to update this via visual studio rather than from here.

```suggestion
            if (element.Normal().Angle(other.Normal()) < Tolerance.Angle)
```

Physical_Engine/Query/HasMergeablePropertiesWith.cs
```diff
@@ -0,0 +1,119 @@
...
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Evaluates if the two IFramingElements' non-geometrical data is equal to the point that they could be merged into one object.")]
+        [Input("element", "An IFramingElement to compare the properties of with an other IFramingElement.")]
+        [Input("other", "The IFramingElement to compare with the other IFramingElement.")]
+        [Output("equal", "True if the IFramingElements' non-geometrical property is equal to the point that they could be merged into one object.")]
+        public static bool HasMergeablePropertiesWith(this IFramingElement element, IFramingElement other)
```
IsakNaslundBh, line 48: Now that we check tangents and normals for this, the framing elements need to have a curve defined, to not crash out with null reference exceptions. Need to check for that before those methods are being called. (if any curve is null, return false, as adding curve might make it false when it comes to normals and tangents)

Physical_Engine/Query/HasMergeablePropertiesWith.cs
```diff
@@ -0,0 +1,122 @@
...
+            {
+                Engine.Reflection.Compute.RecordWarning("No merge comparison avalible for non-linear IFramingElements.");
+                return false;
+            }
+
+            int parallel = element.Location.IStartDir().IsParallel(other.Location.IStartDir());
+            if (parallel != 1)
+                return false;
+
+            if (element.Normal().Angle(other.Normal()) < BH.oM.Geometry.Tolerance.Angle)
```
IsakNaslundBh, line 66: Realised my previous suggestion had the signs wrong:

```suggestion
            if (element.Normal().Angle(other.Normal()) > BH.oM.Geometry.Tolerance.Angle)
```

PR #1709, opened 2020-04-29
Geometry_Engine: Removing deprecated code base older than 3.0

### LINKED ISSUE #1687: Geometry_Engine: Remove deprecated code base older than 3.0
#### Description:

As per title.

kThorsager:
This PR has simply removed all methods that we want to remove in the Geometry_Engine, but as we suspect that this might impact a lot of toolkits the CI will be used to evaluate further actions.

The issues were few and have been dealt with, links to those can be found in the issue
Closes #1687 
Old methods have been removed and all were old enough that Versioning is not needed

### Testfiles 
Not needed as only deprecated things have been removed and the code still compiles

### Additional comments
The keen eyed will notice that `DistributeOutlines` have not been removed and that situation is handled here #1713 

use of old methods in Structures and Environment has also been updated to the current ones


Geometry_Engine/Compute/BooleanDifference.cs
```diff
@@ -182,7 +143,7 @@ public static List<Polyline> BooleanDifference(this Polyline region, List<Polyli
             if (allRegions.Count == 1)
                 return new List<Polyline> { region };
 
-            Vector normal = region.Normal();
+            Vector normal = region.INormal();
```
pawelbaran: Why not `Normal`?

Geometry_Engine/Compute/BooleanIntersection.cs
```diff
@@ -20,6 +20,7 @@
  * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
  */
 
+using BH.Engine.Geometry;
```
pawelbaran, line 23: Why are you using namespace in itself?

Geometry_Engine/Compute/BooleanIntersection.cs
```diff
@@ -165,7 +166,7 @@ public static List<Polyline> BooleanIntersection(this Polyline region, Polyline
             }
 
             bool regSameDir = false;
-            if (Math.Abs(region.Normal().DotProduct(refRegion.Normal()) - 1) <= tolerance)
+            if (Math.Abs(region.INormal().DotProduct(refRegion.INormal()) - 1) <= tolerance)
```
pawelbaran: Same as above, why not `Normal`?

Geometry_Engine/Modify/Offset.cs
```diff
@@ -20,6 +20,7 @@
  * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
  */
 
+using BH.Engine.Geometry;
```
pawelbaran: Why are you using namespace in itself?

Geometry_Engine/Modify/Offset.cs
```diff
@@ -159,7 +160,7 @@ public static Polyline Offset(this Polyline curve, double offset, Vector normal
                 if (!isClosed)
                     BH.Engine.Reflection.Compute.RecordError("Normal is missing. Normal vector is not needed only for closed curves");
                 else
-                    normal = curve.Normal();
+                    normal = curve.INormal();
```
pawelbaran: Why not `Normal`?

Geometry_Engine/Compute/BooleanDifference.cs
kThorsager: Same answer to all of them, I trusted VS which complained for whatever reason while batch making changes, something which I thought I reverted in https://github.com/BHoM/BHoM_Engine/pull/1709/commits/f13df3464d3075f8ab5a5ed6716c3fdc92e24e14 but apparently not.

I'll undo these as well

Geometry_Engine/Compute/BooleanIntersection.cs
kThorsager: apart from this one

Geometry_Engine/Compute/BooleanDifference.cs
kThorsager: wait, most these were already removed, think you missed my last commit, and there's an other one coming. just going to double check through it all again first

kThorsager:
I don't feel very guilty off the Modify complaint @FraserGreenroyd and any action on that should be given its own PR.

FraserGreenroyd:
I agree @kThorsager , could you raise an issue to get it addressed in its own PR?

PR #1732, opened 2020-05-02
Reflection_Engine: Fix MethodBase() method

### LINKED ISSUE #1731: Reflection_Engine: MethodBase() can return incorrect method if some parameters are lists
#### Description:
See https://github.com/BHoM/BHoM_UI/pull/248#issuecomment-621038100 for reference.
The issue is that the parameter matching is done through the parameter names. So methods that differ only by the generic argument inside the generic type parameters, will not be differentiated.

A good example of that will be `List<T>`. Any `List<A>` will be confused with any `List<B>` since the type name is simply `List'1`. 
image

#### Steps to reproduce:
Feed this into a `FromJson` component:
```
{ "_t" : "System.Reflection.MethodBase", "TypeName" : "{ \"_t\" : \"System.Type\", \"Name\" : \"BH.Engine.ModelLaundry.Modify, ModelLaundry_Engine, Version=3.0.0.0, Culture=neutral, PublicKeyToken=null\" }", "MethodName" : "ExtendToEachOther", "Parameters" : ["{ \"_t\" : \"System.Type\", \"Name\" : \"System.Collections.Generic.List`1, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089\", \"GenericArguments\" : [{ \"_t\" : \"System.Type\", \"Name\" : \"BH.oM.Dimensional.IElement2D\" }] }", "{ \"_t\" : \"System.Type\", \"Name\" : \"System.Double, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089\" }", "{ \"_t\" : \"System.Type\", \"Name\" : \"System.Double, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089\" }", "{ \"_t\" : \"System.Type\", \"Name\" : \"System.Double, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089\" }"] }
```

#### Test file(s):
You can also use the file provided by @kThorsager on the comment referenced above.

adecler:

Closes #1731

See issue for full details

### DIFF
--- a/Reflection_Engine/Create/MethodBase.cs
+++ b/Reflection_Engine/Create/MethodBase.cs
@@ -20,6 +20,7 @@
  * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
  */
 
+using BH.Engine.Reflection;
 using System;
 using System.Collections.Generic;
 using System.Linq;
@@ -51,7 +52,7 @@ namespace BH.Engine.Serialiser
                     if (parameters.Length == paramTypeNames.Count)
                     {
                         bool matching = true;
-                        List<string> names = parameters.Select(x => x.ParameterType.Name).ToList();
+                        List<string> names = parameters.Select(x => x.ParameterType.ToText(true)).ToList();
                         for (int i = 0; i < paramTypeNames.Count; i++)
                             matching &= names[i] == paramTypeNames[i];
 
@@ -82,7 +83,7 @@ namespace BH.Engine.Serialiser
                 return method;
 
             // So, let's try the other overload
-            return MethodBase(type, methodName, paramTypes.Select(x => x.Name).ToList());
+            return MethodBase(type, methodName, paramTypes.Select(x => x.ToText(true)).ToList());
         }
 
         /*******************************************/


Reflection_Engine/Create/MethodBase.cs
```diff
@@ -51,7 +52,7 @@ public static MethodBase MethodBase(Type type, string methodName, List<string> p
                     if (parameters.Length == paramTypeNames.Count)
                     {
                         bool matching = true;
-                        List<string> names = parameters.Select(x => x.ParameterType.Name).ToList();
+                        List<string> names = parameters.Select(x => x.ParameterType.ToText(true)).ToList();
                         for (int i = 0; i < paramTypeNames.Count; i++)
                             matching &= names[i] == paramTypeNames[i];
```
kThorsager, line 57: These don't match for my case as paramTypeNames is a lot more generic than the other:
image

adecler:
> The components in the test file are now not deserialising at all. Not a problem for all BHoM components, but the ExtendToEachother ones at least.

It works great for me on the file you shared:
image

image

Can you share a file where it doesn't work for you ? If it's the same, then we we should probably do a screen share to figure out why it doesn't work for you.

> Tried some debugging but got lost quickly, but did find that the UI tried to deserialise this:

That's because the json you provided there is the full Custom object, not the method itself. Remember that the method you trying to deserialise needs to be updated first and this needs to be done explicitly (for now) otherwise the deserialisation will fail.

> These don't match for my case as paramTypeNames is a lot more generic than the other:

That is also because you are looking at the code executed before the method is updated through versioning. 

image

kThorsager:
Sounds like I need to make a full rebuild of the BHoM, (but it's strange It did the versioning the old slightly broken way if I moved off this branch)

PR #1740, opened 2020-05-05
Data_Engine: Prototype Data Tree for data with extents with create and query methods based on multiple numerical domains

kThorsager:
## Prototype
This is a prototype and should not be considered in regards to blocking, as these branches won't be merged. This PR is opened to share and discuss code.

### NOTE: Depends on 
https://github.com/BHoM/BHoM/pull/852
   
https://github.com/BHoM/BHoM/issues/843

### Test files

https://burohappold.sharepoint.com/:f:/s/BHoM/EqfcJ1Lyxu5GmWQinRhUYJEBWnMky8nLmSmNoHNPZ5QTZQ?e=QN2cuU


Data_Engine/Create/NTree.cs
```diff
@@ -0,0 +1,179 @@
...
+
+namespace BH.Engine.Data
+{
+    public static partial class Create
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+        
+        public static NTree<T> NTree<T>(IEnumerable<NBound> boundingBoxes, IEnumerable<T> data, int splits = 16)
```
kThorsager, line 41: splits = number of items stored in each branch

Am currently not differentiating between amount of branches per branch and amount of leafs per branch, but could be added.

Data_Engine/Create/NTree.cs
```diff
@@ -0,0 +1,179 @@
...
+                return new NTree<T>() { Items = leaves.ToArray(), Bounds = Bounds(leaves) };
+            }
+        }
+
+
+        /***************************************************/
+        /**** Private Methods                           ****/
+        /***************************************************/
+        
+        private static Output<List<Leaf<T>>, List<Leaf<T>>> SplitList<T>(List<Leaf<T>> list, int sampleSize = 60)
```
kThorsager, line 109: `sampleSize` = amount of object to use for finding the centre for splitting into different boxes.

It's both not exposed for the user atm.

What we'd really want would be the median point of all the `NBounds`, which would be very computational expensive. An OK approximation is the average of a sample of the boxes, which is what I'm taking here.

Data_Engine/Query/ClosestData.cs
```diff
@@ -33,13 +35,76 @@ public static partial class Query
...
+                }
+                else
+                {
+                    result.Add(list[i] as Leaf<T>);
+                    list.RemoveAt(i);
+                }
+                i = list.Count - 1;
+            }
+
+            return result.Select(x => x.Item).ToList();
```
kThorsager, line 92: Could return the `Leave<T>` instead of the Data to align with `PointMatrix`

Data_Engine/Create/NTree.cs
```diff
@@ -109,14 +109,14 @@ public static NTree<T> NTree<T>(IEnumerable<Leaf<T>> leaves, int splits = 16)
...
-            NBound box = new NBound();
+            NBound box = null;
 
             //sampleSize = Math.Max(2, sampleSize);
             sampleSize = Math.Min(list.Count, sampleSize);
 
             int step = (int)Math.Floor((decimal)(list.Count / sampleSize));
             for (int i = 0; i < sampleSize; i++)
-                box += (NBound)list[i * step].Bounds;
+                box += list[i * step].Bounds;
```
FraserGreenroyd, line 119: Does this work in testing @kThorsager ? The box being `null` on the first path would be doing a calculation of `null + box` which I wouldn't expect to work initially?

kThorsager: You'd think so but I was wrote the `+` operator to simply return the second box if the first one is `null`, so yes.

That should perhaps be changed however

kThorsager: It was confusing code so I changed it

kThorsager:
I'm noticing the issue that grasshopper does not like my generic inputs:
image
But only for components which have been copy pasted (or loaded from a file) which draws my eyes to serialisation @adecler , am I doing something wrong? is it a bug?

Geometry_Engine/Create/NTree.cs
```diff
@@ -0,0 +1,47 @@
...
+    public static partial class Create
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        public static NTree<T> NTree<T>(IEnumerable<T> geometry, int split = 16) where T : IGeometry
+        {
+            return Data.Create.NTree(geometry.Select(x => new Leaf<T>() { Bounds = x.IBounds().NBound(), Item = x }), split);
+        }
```
kThorsager, line 42: Managed to avoid the before mentioned bug where one couldn't copy-paste one of the query methods connected to this by using the more explicit Create method in the Data_Engine instead.

Strongly suspect that it's the `where T : IGeometry` which is messing things up since I think that is the only difference.

kThorsager:
@pawelbaran @al-fisher @adecler @rolyhudson 

I'm about as done as I'm going to get on this prototype without feedback.

I've added code comments on things which I'm less sure of in both this and the BHoM PR but am ofc interested in pointers for the rest of the code as well.

And there's also the test-file which makes sure that it works against brute-force methods.

adecler:
> I'm noticing the issue that grasshopper does not like my generic inputs:
> image
> But only for components which have been copy pasted (or loaded from a file) which draws my eyes to serialisation @adecler , am I doing something wrong? is it a bug?

@kThorsager , sorry I missed that. I have done a series of PRs (already merged) this sprint that should fix this problem. Let me know if it doesn't.

kThorsager:
Afraid not, 
I updated the branches from master and made sure to be on updated versions of both BHoM_UI and Grasshopper.

Added a more on point[ test-file for the problem](https://burohappold.sharepoint.com/:u:/s/BHoM/Ea5xdAvCGIFNjNsL0a7vIp0BVnVz5v5_H1XqPuOV2hIFwA?e=vDwxva) But as I've been unable to recreate this with other components it's only for these branches

kThorsager:
@pawelbaran @al-fisher @rolyhudson @adecler 

Would still need some comments on my comments about general approach.
Will remind that a full review is not needed at this stage.

Otherwise this is done apart from a bunch of descriptions (and potentially refinement of method names)

Data_Engine/Create/NTree.cs
```diff
@@ -0,0 +1,174 @@
...
+        /***************************************************/
+
+        public static NTree<T> NTree<T>(IEnumerable<Leaf<T>> leaves, int splits = 16)
+        {
+            splits = splits < 2 ? 2 : splits;
+
+            if (leaves.Count() > splits)
+            {
+                int temp = (int)Math.Ceiling((decimal)(leaves.Count() / splits)) + 1;
+                temp = temp < splits ? temp : splits;
```
pawelbaran, line 60: Math.Min?

Data_Engine/Create/NTree.cs
```diff
@@ -0,0 +1,174 @@
...
+                return null;
+
+            List<Leaf<T>> leaves = boundingBoxes.Zip(data, (x, y) => new Leaf<T>() { Bounds = x, Item = y }).ToList();
+
+            return NTree(leaves, splits);
+        }
+
+        /***************************************************/
+
+        public static NTree<T> NTree<T>(IEnumerable<Leaf<T>> leaves, int splits = 16)
```
pawelbaran, line 53: Looks like you are querying leaves as `IEnumerable` for its count a few times + there is a cast to `List` as well as to `Array`. Possibly could be done smarter?

kThorsager:
> One thing that comes to my mind: would it be possible to use this code for more efficient clustering (e.g. by implementing DBSCAN or other clustering algo with queries on RTree as metric function)? That would be a nice extension to the core functionality we have here.

Yes, I hope so.
One limitation this would have compared to the current one is that it would enforce that the clustering would be based on location, unlike now where it's solely based on an arbitrary metric function.

But no reason to not have both

pawelbaran:
> Yes, I hope so.
> One limitation this would have compared to the current one is that it would enforce that the clustering would be based on location, unlike now where it's solely based on an arbitrary metric function.

I meant a purely geometrical clustering, so sounds like we are good to go!

rolyhudson:
A thought on the name of this. _Spatial_ indicates 3 dimensions - where as this functionality as implemented does not limit to xyz.

pawelbaran:
_Spatial_ does not occur in the code though? Or are you referring to [my comment on the oM](https://github.com/BHoM/BHoM/pull/852#issuecomment-630742745) @rolyhudson?

rolyhudson:
> _Spatial_ does not occur in the code though? Or are you referring to [my comment on the oM](https://github.com/BHoM/BHoM/pull/852#issuecomment-630742745) @rolyhudson?

Sorry @pawelbaran I was unclear. I meant the name of the PR is confusing as it includes _Spatial_ but the PR or at least the conversation and code is about potentially more dimensions.

kThorsager:
This is once again done.

The key changes to how the methods work now in relation to the new objects are:
- Anything inheriting from `INode<T>` should now have the extension methods `Values()` and `Children` in a similar manner to how `IElement` does it. A notable difference is that I'm not using Reflection to call the extension method as that does not work with generics, but that sounds like a later problem.
- As trees are far more about how they treat their data than the data they contain this concept is also kind of flipped on it's head as the times when one does not know which kind of data sturcture one has and need to query it seems low. Hence the "way in" is through each types method, which in turn call the interface method, while providing methods to detail how to traverse itself.

The functionality is the same as before but the methods have had their general concepts (like separating data to create a tree) exposed as inputs.

I'll also mention that my methods with nested generics as input (which require one to specify the type outside of the input) does not work in the UI. Which is probably expected and not really an issue here.
image


But given that these changes get some approval I'll get to adding descriptions, (and presumably opening a fresh branch for a smoother merge) and prepare it for a pre-merge review.

kThorsager:
Closing this as #1833 is the final version

PR #1746, opened 2020-05-07
Serialiser_Engine: Adding serialiser for IntPtr and Bitmap

### LINKED ISSUE #1736: Serialiser_Engine: fails for object containing Bitmap and IntPtr
#### Description:
Those types can be found here: 
Bitmap -> BH.oM.UI.SearchItem
IntPtr -> BH.oM.Excel.Reference

adecler:

Closes #1736

This is a very minor thing as IntPtr and Bitmap will generally not be something you would want to serialise but good for completion sake. 

It is worth noting that there are three issues with Bitmap:
- Generating a Dummy Bitmap leads to a failing `ToJson`  So you have to break in the code where you have access to a `SearchItem` so you can test (de)serialisation: 
image I have fixed that with the commented code at the bottom of the `BitmapDeserializer` file. I decided to stay with the official Bitmap.Save() though because I was not completely sure that my version would not bring it's own set of problems. 
- while Bitmap goes through a serialisation cycle properly now, it fails the `IsEqual` Test. This is because of the `Flags` property in `Palette`. This thing can not be overwritten and is even changing live while I was debugging in VS. This is another reason why I didn't keep my version of the code. Since I couldn't fix the flag issue, I would rather rely on the official version to save/load a bitmap since there is more chance that the flags will be correct. (btw, those flags are supposed to influence the selection of the colour palette).  
- Not all image formats are supported (although all the main ones are: bmp, jpeg, png, tiff,...). My commented solution works with all of them, even the `in memory bmp` format that we get when creating a dummy `SearchItem`. But again, not sure if my solution is reliable on all levels.

Overall, I think I let myself waste too much time on something that is so minor so this PR is probably good enough for now. I am just a bit annoyed that I will have to register bitmaps as an exception when testing equality on serialisation.

### DIFF
--- a/Reflection_Engine/Reflection_Engine.csproj
+++ b/Reflection_Engine/Reflection_Engine.csproj
@@ -67,7 +67,7 @@
       <HintPath>..\packages\Mono.Reflection.1.1.0.0\lib\Mono.Reflection.dll</HintPath>
       <Private>True</Private>
     </Reference>
-    <Reference Include="Quantities_oM, Version=3.0.0.0, Culture=neutral, processorArchitecture=MSIL">
+    <Reference Include="Quantities_oM">
       <SpecificVersion>False</SpecificVersion>
       <HintPath>..\..\BHoM\Build\Quantities_oM.dll</HintPath>
       <Private>False</Private>
--- a/Reflection_Engine/Reflection_Engine.csproj
+++ b/Reflection_Engine/Reflection_Engine.csproj
@@ -29,7 +29,7 @@
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
     <DebugType>pdbonly</DebugType>
     <Optimize>true</Optimize>
-    <OutputPath>bin\Release\</OutputPath>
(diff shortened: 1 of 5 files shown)


IsakNaslundBh:
@adecler do you have a testfile for this? I have not created any obejcts that contain a bitmap or an IntPtr, so not sure what to best test this on for a "real" case.

adecler:
Admittedly, there isn't much to test on. The two classes referenced in the issue are literally the only one that contain those types. 

For `IntPtr`, you can create a `Dummy` `BH.oM.Excel.Reference` and go through a cycle of serialisation before testing equality:
image

For `Bitmap`, the fake Dummy bitmap is not fit for testing so you have to get it from a real one. The only way to do that is to put a break point int the BHoM_UI code as I have shown above and use the Ctrl-Shift-B menu. You can then run the serialisation cycle in `watch`.

IsakNaslundBh:
Compliance issue only related to pre-existing issue with wrong buildfolder for release build, will leave it to you @adecler if this PR is the right place to fix this.

PR #1756, opened 2020-05-09
Serialiser_Engine:  Fix remaining problems with serialisation

### LINKED ISSUE #1620: Serialiser_Engine: Validate the serialisation engine
We need to confirm that serialisation works for every possible object we can create with the BHoM.

Create a full report on result of applying a cycle of ToJson -> FromJson on all possible BHoM objects and other usueful types (such as string). All resulting deserialised object should be equal to the initial object. Each object should be tested both as top object and as property of another object.

@al-fisher FYI

### LINKED ISSUE #1737: Serialiser_Engine: GenericTypeDefinition classes don't deserialise properly if different parameters present
#### Description:
Take for example the case of `public class Tree<T>` in `BH.oM.Data.Collections`

The problem for Tree is that the type is simply stored as `Tree'1` without containing the Generic types. This is because the class itself is a GenericTypeDefinition (i.e. uses `<T>` in the class definition). So trees are deserialisiing properly as long as there is only one type of tree being deserialise. As soon as there is more than one, FromJson fails:

image

Here's another example easier to reproduce:

image

If I would rerun the FromJson of the LibraryTree, it would then give me a CustomObject

### LINKED ISSUE #1755: Serialiser_Engine: Enumerable like FragmentSet and ModuleSet don't serialise as top objects

adecler:
### NOTE: Depends on 

https://github.com/BHoM/BHoM/pull/857
https://github.com/BHoM/Test_Toolkit/pull/247

   
Closes #1620 
Closes #1755 
Closes #1737

Yatta !! Finally got all types working (at least those on my computer).

image

This fixes the serialiser itself but will be accompanied by a few corrections on the non-compliant types themselves to get it all working (see PRs in the `depends on` section)

### Test files
https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/01_Test%20Scripts/BHoM_Engine/Serializer_Engine/serialisationTest.gh?csf=1&web=1&e=1okDsJ

If you have types failing on your machine, please make sure they compliant first. A common issue for example is an incorrect constructor for `IImmutable` classes.

### Changelog
- Move RegisterClassMap and add serialiser for regex
- Take care of the case where a bitmap is null
- Handle Enumerable as top objects and GenericTypeDefinition
- Don't send nulls to versioning

### Additional comments
FYI, A dummy object cannot be created for `MaterialComposition` because it has a restriction inside its constructor that throws an exception if that constraint is not met.

### DIFF
--- a/Diffing_Engine/Create/StreamPointer.cs
+++ b/Diffing_Engine/Create/StreamPointer.cs
@@ -39,7 +39,7 @@ namespace BH.Engine.Diffing
         [Description("Creates new Stream Pointer, generating a new StreamId.")]
         public static StreamPointer StreamPointer(string name = null, string description = null)
         {
-            return new StreamPointer(name, description);
+            return new StreamPointer(Guid.NewGuid(), name, description, DateTime.UtcNow.Ticks);
         }
     }
 }
--- a/Serialiser_Engine/Compute/RegisterClassMap.cs
+++ b/Serialiser_Engine/Compute/RegisterClassMap.cs
@@ -1,0 +1,67 @@
+/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2020, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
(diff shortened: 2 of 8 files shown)


adecler:
@IsakNaslundBh , it works perfectly fine for me so we probably need to do a screen share on this one:

image

PR #1728, opened 2020-05-01
Structure_Engine: Refactor of Bar Reinforcement

IsakNaslundBh:
### NOTE: Depends on 

https://github.com/BHoM/BHoM/pull/844
   
### Issues addressed by this PR

Closes #1609
Closes #1728 
Closes #1742 
Closes #1729 
Closes #1775 

### Test files

https://burohappold.sharepoint.com/:f:/s/BHoM/EpXy3FR30txGs3Q3jTlTwOoBv9zjDu3k-0Z34O9CfSBIBQ?e=OO58dZ

Test of matter engine:

https://burohappold.sharepoint.com/:f:/s/BHoM/EkKKuT_RTYFAoLVZad5J9YkBfn3pm6ElXOonvELiJJqvcg?e=VIvdaP

Test offset:

https://burohappold.sharepoint.com/:f:/s/BHoM/EmsFjWvtqdlFj93FK71gVMwBq7Rou4vYir7Mq5w77UhDIA?e=hhWQ0W

### Changelog

- Refactoring required by changes to Reinforment objects in Structure_oM
- Make use of layout methods in Spacial_Engine to distribute reinforcement
- Pass over Rebar object to FramingElements when created from a structural Bar
- Tweaking `MaterialComposition` for `ConcreteSection`s to account for Reinforcement
- Fixing bug in Geometry_Engine making offsets crash when polycurve contains a circle
- Fixing issue in Matter_Engine, making sure hashes are only calculated once

PR #1759, opened 2020-05-12
Common_Engine: Remove Deprecated Methods

### LINKED ISSUE #1624: Common_Engine: Remove deprecated methods
#### Broken rules:
It has a bunch of Deprecated methods which we're moved to the Spatial_Engine.

- [x] 1. Raise issues about updating the dependencies in other toolkits.
- [x] 2. Close those issues.
- [x] 3. Assure that Versioning works.
- [ ] 4. Remove the methods.

@FraserGreenroyd how are we looking on running a compliance report to find were these Deprecated methods are used?

Most of these toolkit only need to remove their references to Common_Engine, and removing the methods won't break them. But they will break when we eventually remove the Common_Engine all together, so saw no reason not to raise the issues now.

Toolkits with action required before proceeding:

- [x] Revit
- [x]  ETABS

kThorsager:
Closes #1624 
Removes all methods in Common_Engine which were moved to Spatial_Engine and adds Versioning for them.

### Test files

Common_Engines regular testfiles are here and are upgraded with this PR. i.e. the old labels are removed and they still work.
https://burohappold.sharepoint.com/:f:/s/BHoM/EvyCGXcAkK5OqmqvtfZkZMABtuDNXxKAG7YZdY-dI0SWXA?e=rXXt8h

### Additional comments

Won't build as Structure_Engine is currently using these methods and I'm loath to bother #1728 so will leave this in draft for until that's resolved or open


kThorsager:
Changed tag to `do-not-merge` as not work is being done on it, but simply waiting for #1728 to open `Structure_Engine.csproj` for editing

kThorsager:
Recreated in #1784 to avoid annoyance

PR #1761, opened 2020-05-13
Geometry_Engine: Implement necessary methods for immutable PlanarSurface

### LINKED ISSUE #1446: Geometry_Engine: Implement necessary methods for immutable PlanarSurface
#### Description:

Following [BHoM #676](https://github.com/BHoM/BHoM/issues/676).

kThorsager:
### NOTE: Depends on 

https://github.com/BHoM/BHoM/pull/861
   
Closes #1446 
Updates the Create method to use the new constructor and adds extra checks to ensure the validity of the `PlanarSurface`

See oM PR for more info

### Test files

https://burohappold.sharepoint.com/:u:/s/BHoM/Ed2ni1X7x2lOhSlTrobaIZ8BIa56APsehBDEObFLFDHkPA?e=un8SPP

### DIFF
--- a/Geometry_Engine/Create/PlanarSurface.cs
+++ b/Geometry_Engine/Create/PlanarSurface.cs
@@ -67,7 +67,7 @@ namespace BH.Engine.Geometry
 
             if (internalBoundaries.Count != count)
             {
-                Reflection.Compute.RecordWarning("At least one of the internalBoundaries is not closed. And have been disregarded.");
+                Reflection.Compute.RecordWarning("At least one of the internal boundaries is not closed and has been ignored on creation of the planar surface.");
             }
 
             if (internalBoundaries.Count > 0)
@@ -87,7 +87,7 @@ namespace BH.Engine.Geometry
 
             if (internalBoundaries.Count != count)
             {
-                Reflection.Compute.RecordWarning("At least one of the internalBoundaries is selfintersecting. And have been disregarded.");
+                Reflection.Compute.RecordWarning("At least one of the internal boundaries is self-intersecting and has been ignored on creation of the planar surface..");
             }
 
             // Runs Boolean Union to resolve intersections and containemnt
@@ -104,13 +104,13 @@ namespace BH.Engine.Geometry
             //// Checks based around external curve
             if (externalBoundary is NurbsCurve)
             {
-                Reflection.Compute.RecordWarning("Neeseccary checks to ensure vadility of the PlanarSurface is not implemented. The PlanarSurfaces curves relations are not garanteueed.");
+                Reflection.Compute.RecordWarning("External boundary is a nurbs curve. Necessary checks to ensure validity of a planar surface based on nurbs curve cannot be run, therefore correctness of the surface boundaries is not guaranteed.");
                 // External done
                 return new PlanarSurface(externalBoundary, internalBoundaries);
             }
             else if (externalBoundary.IIsSelfIntersecting())
             {
-                Reflection.Compute.RecordError("The provided externalBoundary is selfintersecting.");
+                Reflection.Compute.RecordError("The provided external boundary is self-intersecting.");
                 return null;
             }
             if (externalBoundary is Ellipse)
@@ -179,4 +179,3 @@ namespace BH.Engine.Geometry
         /***************************************************/
     }
 }
-
--- a/Geometry_Engine/Create/PlanarSurface.cs
+++ b/Geometry_Engine/Create/PlanarSurface.cs
@@ -79,6 +79,95 @@ namespace BH.Engine.Geometry
                 }
             }
 
+            bool nonExistantCheckWarning = false;
+
+            try
+            {
+                if (externalBoundary.IIsSelfIntersecting())
+                {
+                    Reflection.Compute.RecordError("The provided externalBoundary is selfintersecting.");
+                    return null;
+                }
+            } catch
+            {
+                nonExistantCheckWarning = true;
+            }
+
+            try
+            {
+                if (internalBoundaries.Any(x => x.IIsSelfIntersecting()))
+                {
+                    Reflection.Compute.RecordError("At least one of the internalBoundaries is selfintersecting.");
+                    return null;
+                }
+            }
+            catch
+            {
+                nonExistantCheckWarning = true;
+            }
+
+            try
+            {
+                if (internalBoundaries.Any(x => !externalBoundary.IIsContaining(x)))
+                {
+                    Reflection.Compute.RecordError("At least one of the internalBoundaries is not contained by the externalBoundary.");
+                    return null;
+                }
+            } catch
+            {
+                nonExistantCheckWarning = true;
+            }
+
+            bool unionDone = false;
+            try
+            { 
+                for (int i = 0; i < internalBoundaries.Count; i++)
+                {
+                    for (int j = i + 1; j < internalBoundaries.Count; j++)
+                    {
+                        if (internalBoundaries[i].ICurveIntersections(internalBoundaries[j]).Count != 0)
+                        {
+                            internalBoundaries = Compute.BooleanUnion(internalBoundaries).Cast<ICurve>().ToList();
+                            Reflection.Compute.RecordWarning("InternalBounderies were overlapping, BooleanUnion has been used to get non overlapping regions.");
+                            i = internalBoundaries.Count;
+                            j = i;
+                            unionDone = true;
+                        }
+                    }
+                }
+            }
+            catch
+            {
+                nonExistantCheckWarning = true;
+            }
+
+            try
+            {
+                if (!unionDone)
+                {
+                    for (int i = 0; i < internalBoundaries.Count; i++)
+                    {
+                        for (int j = 0; j < internalBoundaries.Count; j++)
+                        {
+                            if (internalBoundaries[i].IIsContaining(internalBoundaries[j]))
+                            {
+                                internalBoundaries = Compute.BooleanUnion(internalBoundaries).Cast<ICurve>().ToList();
+                                Reflection.Compute.RecordWarning("InternalBounderies were overlapping, BooleanUnion has been used to get non overlapping regions.");
+                                i = internalBoundaries.Count;
+                                j = i;
+                            }
+                        }
+                    }
+                }
+            }
+            catch
+            {
+                nonExistantCheckWarning = true;
+            }
+
+            if (nonExistantCheckWarning)
+                Reflection.Compute.RecordWarning("Neeseccary checks to ensure vadility of the PlanarSurface is not implemented. The PlanarSurfaces curves relations are not garanteueed.");
+
             return new PlanarSurface(externalBoundary, internalBoundaries);
         }
 
--- a/Geometry_Engine/Create/PlanarSurface.cs
+++ b/Geometry_Engine/Create/PlanarSurface.cs
@@ -55,19 +55,19 @@ namespace BH.Engine.Geometry
 
             internalBoundaries = internalBoundaries ?? new List<ICurve>();
 
-            foreach (ICurve crv in internalBoundaries)
+            if (internalBoundaries.Any(x => !x.IIsPlanar()))
             {
-                if (!crv.IIsPlanar())
-                {
-                    Reflection.Compute.RecordError("At least one of the internal edge curves is not planar");
-                    return null;
-                }
+                Reflection.Compute.RecordError("At least one of the internal edge curves is not planar");
+                return null;
+            }
 
-                if (!crv.IIsClosed())
-                {
-                    Reflection.Compute.RecordError("At least one of the internal edge curves is not closed");
-                    return null;
-                }
+            int count = internalBoundaries.Count;
+
+            internalBoundaries = internalBoundaries.Where(x => x.IIsClosed()).ToList();
+
+            if (internalBoundaries.Count != count)
+            {
+                Reflection.Compute.RecordWarning("At least one of the internalBoundaries is not closed. And have been disregarded.");
             }
 
             if (internalBoundaries.Count > 0)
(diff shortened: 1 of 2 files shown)


kThorsager:
Added various checks in the create method and a test-file with examples.

It became a bit of a mess as I didn't want a check which failed due to a nurbs keeping an other check which would work from executing, but it works

Geometry_Engine/Create/PlanarSurface.cs
```diff
@@ -79,7 +79,96 @@ public static PlanarSurface PlanarSurface(ICurve externalBoundary, List<ICurve>
                 }
             }
 
-            return new PlanarSurface { ExternalBoundary = externalBoundary, InternalBoundaries = internalBoundaries };
+            bool nonExistantCheckWarning = false;
+
+            try
```
epignatelli, line 84: Is there a reason why we are `try`-`catch`ing all the checks, rather than perform the check more safely?

Geometry_Engine/Create/PlanarSurface.cs
```diff
@@ -79,7 +79,96 @@ public static PlanarSurface PlanarSurface(ICurve externalBoundary, List<ICurve>
...
+            {
+                if (externalBoundary.IIsSelfIntersecting())
+                {
+                    Reflection.Compute.RecordError("The provided externalBoundary is selfintersecting.");
```
epignatelli, line 88: If we are returning some kind of objects anyway, would it be better to record a warning instead?

Geometry_Engine/Create/PlanarSurface.cs
```diff
@@ -79,7 +79,96 @@ public static PlanarSurface PlanarSurface(ICurve externalBoundary, List<ICurve>
...
+            catch
+            {
+                nonExistantCheckWarning = true;
+            }
+
+            try
+            {
+                if (internalBoundaries.Any(x => !externalBoundary.IIsContaining(x)))
+                {
+                    Reflection.Compute.RecordError("At least one of the internalBoundaries is not contained by the externalBoundary.");
```
epignatelli, line 113: Can't we remove the surfaces that are not contained in the external boundary and build the planar surface only from the valid ones? Or is this happening in the `BooleanUnion` already?

Geometry_Engine/Create/PlanarSurface.cs
```diff
@@ -79,7 +79,96 @@ public static PlanarSurface PlanarSurface(ICurve externalBoundary, List<ICurve>
...
+            } catch
+            {
+                nonExistantCheckWarning = true;
+            }
+
+            try
+            {
+                if (internalBoundaries.Any(x => x.IIsSelfIntersecting()))
+                {
+                    Reflection.Compute.RecordError("At least one of the internalBoundaries is selfintersecting.");
```
epignatelli, line 100: Can we remove the self-intersecting inner boundaries and build the planar surface only from the valid ones?

Geometry_Engine/Create/PlanarSurface.cs
```diff
@@ -79,7 +79,96 @@ public static PlanarSurface PlanarSurface(ICurve externalBoundary, List<ICurve>
...
+
+            try
+            {
+                if (!unionDone)
+                {
+                    for (int i = 0; i < internalBoundaries.Count; i++)
+                    {
+                        for (int j = 0; j < internalBoundaries.Count; j++)
+                        {
+                            if (i != j && internalBoundaries[i].IIsContaining(internalBoundaries[j]))
```
epignatelli, line 152: Uhm, I didn't get this. Why aren't we doing adding this check in the previous attempt to boolean unite the internal buondaries?

Geometry_Engine/Create/PlanarSurface.cs
kThorsager: My reason is that as this will never fail catastrophically, which I found to be important as there would not be many other ways of constructing `PlanarSurface`s.

The other option would be to exclude all Nurbs (and Ellipses, maybe more) from the checks. which could change both in the good way of them being supported or the bad way where another curve type is added which would break everything.

Not super-sure but went with the way I knew would work.

Geometry_Engine/Create/PlanarSurface.cs
kThorsager: Does `null` count as a object?

Geometry_Engine/Create/PlanarSurface.cs
kThorsager: We could, but it feels like a less clear-cut solution than say the boolean union as it would modify what the user might have meant, but I'm not opposed per say

Geometry_Engine/Create/PlanarSurface.cs
epignatelli: If we exclude the invalid curves from the inner boundaries, and do not skip the entire calculation by returning a `PlanarSurface` built only with some if them, I think we do not always return `null`, don't we?

Geometry_Engine/Create/PlanarSurface.cs
kThorsager: Guess we could, but we'd need to add checks for the special case of internal boundaries intersecting the external one, and perhaps do a boolean difference there?

And no not solved later, returning null

Geometry_Engine/Create/PlanarSurface.cs
kThorsager: The previous check does not catch the case of a internal boundary encapsulating without intersecting another. 

Keeps this from being the curve definition:
image
which would at least mess with the area method

Geometry_Engine/Create/PlanarSurface.cs
kThorsager: Currently I'm returning null after every `RecordError` and using warnings if I thought it could be salvaged.

But could change the methods philosophy to avoid returning null and salvage more things?
- exclude invalid internal boundaries
- Cut the external boundery if it overlapps with a internal boundery

Geometry_Engine/Create/PlanarSurface.cs
pawelbaran: Personally I do not like brute forcing `try/catch` on one of the core `Create` methods..

Instead I would prefer skipping checks for `NurbsCurve` and recording a warning. Fail safe + it will enable creating nurbs-based `PlanarSurfaces`.

Geometry_Engine/Create/PlanarSurface.cs
pawelbaran: I think I would lean towards @epignatelli - ruling out the incorrect curves with warning, using others to create the surface (with `NurbsCurve` not being checked at all, accepted by default). Not the most correct approach, but as a user I would expect something like that I think.

Geometry_Engine/Create/PlanarSurface.cs
pawelbaran: Same as above, leaning towards @epignatelli

Geometry_Engine/Create/PlanarSurface.cs
```diff
@@ -79,7 +79,96 @@ public static PlanarSurface PlanarSurface(ICurve externalBoundary, List<ICurve>
...
+                }
+            } catch
+            {
+                nonExistantCheckWarning = true;
+            }
+
+            bool unionDone = false;
+            try
+            { 
+                for (int i = 0; i < internalBoundaries.Count; i++)
```
pawelbaran, line 124: for (int i = 0; i < internalBoundaries.Count - 1; i++)

Geometry_Engine/Create/PlanarSurface.cs
```diff
@@ -79,7 +79,96 @@ public static PlanarSurface PlanarSurface(ICurve externalBoundary, List<ICurve>
...
+            catch
+            {
+                nonExistantCheckWarning = true;
+            }
+
+            try
+            {
+                if (!unionDone)
+                {
+                    for (int i = 0; i < internalBoundaries.Count; i++)
```
pawelbaran, line 148: for (int i = 0; i < internalBoundaries.Count - 1; i++)

Geometry_Engine/Create/PlanarSurface.cs
pawelbaran: Hmm, this starts to sound a bit philosophical. I would say we either do all:
- `BooleanUnion` on internal boundaries
- `BooleanDifference` internal vs external
- `DistributeOutlines`

And return a clean solution (nurbs can be tricky) OR we do not do any Boolean operations whatsoever. Now I would be a bit confused why did the openings get merged but not cut out of the external boundary?

Geometry_Engine/Create/PlanarSurface.cs
kThorsager: `IsContaining` is not commutative, so the last region must check all the others as well.

kThorsager:
You're both making sense, I'll iron out which curves are unsupported and make it try harder to return a result

Geometry_Engine/Create/PlanarSurface.cs
epignatelli: What is the actual reason we are doing the try catch? 
In case the methods throw an exception, I'd resort to changing those to recording an error and returning the safest condition, e.g. `IsSelfIntersecting(NurbsCurve curve) { RecordError("NotImplemented"); return True`

Geometry_Engine/Create/PlanarSurface.cs
epignatelli: But can't the check happen in the for loops above?

kThorsager: The for loop above is not the same as it only checks each pair once while this one checks every pair in both directions

Geometry_Engine/Create/PlanarSurface.cs
kThorsager: I do not know why we have so many methods with exceptions, but felt out of scope to resolve here.

kThorsager:
So I'm proposing we hardcode which kind of curves we do not check and raise warnings for.

And attempt to resolve every problem which is found along the way to the best of our ability with warning detailing why and what happened.

epignatelli:
If we have another method other than `PlanarSurface` are we going to hardcode the types there as well?

I believe it will just add more debt to the one we have. 

I think the right way to do it is to fix the methods that raise exceptions, turn them into errors, and return the most conservative option. 

I'd feel safer in getting others' opinion on this, and happy to comply with hardcoding them if we believe it's the right way.

@IsakNaslundBh @al-fisher wdyt?

kThorsager:
I do like the thought of getting rid of the throw exception, I do think it would deserve its own issue given a quick check in Geometry_Engine,

image

But would ofc be lower if we constrained our-self's to the relevant methods here.

But I do see your point about building debt, see `Create.FreeFormProfile`

kThorsager:
I'm also saying this as I just finished hard-coding but that can be changed

kThorsager:
> I think the right way to do it is to fix the methods that raise exceptions, turn them into errors, and return the most conservative option.


@epignatelli A thought that struck me, if we do go with the most conservative option in say `IsSelfIntersecting` and return `true` for all curves we can't solve, won't we still need to hard-code here as we want to be lenient and let through things which we do not know the answer to?

Otherwise we'll just end up not being able to create `Nurbs` based `PlanarSurface`s

Geometry_Engine/Create/PlanarSurface.cs
```diff
@@ -55,19 +55,19 @@ public static PlanarSurface PlanarSurface(ICurve externalBoundary, List<ICurve>
...
-                    Reflection.Compute.RecordError("At least one of the internal edge curves is not closed");
-                    return null;
-                }
+            int count = internalBoundaries.Count;
+
+            internalBoundaries = internalBoundaries.Where(x => x.IIsClosed()).ToList();
+
+            if (internalBoundaries.Count != count)
+            {
+                Reflection.Compute.RecordWarning("At least one of the internalBoundaries is not closed. And have been disregarded.");
```
pawelbaran, line 70:
```suggestion
                Reflection.Compute.RecordWarning("At least one of the internal boundaries is not closed and has been ignored on creation of the planar surface.");
```

Geometry_Engine/Create/PlanarSurface.cs
```diff
@@ -55,19 +55,19 @@ public static PlanarSurface PlanarSurface(ICurve externalBoundary, List<ICurve>
 
             internalBoundaries = internalBoundaries ?? new List<ICurve>();
 
-            foreach (ICurve crv in internalBoundaries)
+            if (internalBoundaries.Any(x => !x.IIsPlanar()))
```
pawelbaran, line 58: I would check the planarity as follows:
1. Check external boundary for planarity - if not, return null
2. Run `Where` Linq query to filter out internal outlines that are coplanar with the external one
3. Do same check as you do in lines 68-71

Lines 58-62 and 73-80 could be removed then.

Geometry_Engine/Create/PlanarSurface.cs
```diff
@@ -79,7 +79,75 @@ public static PlanarSurface PlanarSurface(ICurve externalBoundary, List<ICurve>
...
-            return new PlanarSurface { ExternalBoundary = externalBoundary, InternalBoundaries = internalBoundaries };
+            //// Internal checks independent from External
+            // Selfintersections Internal
+            count = internalBoundaries.Count;
+
+            internalBoundaries = internalBoundaries.Where(x => x is NurbsCurve || !x.IIsSelfIntersecting()).ToList();
+
+            if (internalBoundaries.Count != count)
+            {
+                Reflection.Compute.RecordWarning("At least one of the internalBoundaries is selfintersecting. And have been disregarded.");
```
pawelbaran, line 90:
```suggestion
                Reflection.Compute.RecordWarning("At least one of the internal boundaries is self-intersecting and has been ignored on creation of the planar surface..");
```

Geometry_Engine/Create/PlanarSurface.cs
```diff
@@ -79,7 +79,75 @@ public static PlanarSurface PlanarSurface(ICurve externalBoundary, List<ICurve>
...
+
+            internalBoundaries = internalBoundaries.Where(x => x is NurbsCurve || x is Ellipse)
+                                .Concat(internalBoundaries.Where(x => !(x is NurbsCurve) && !(x is Ellipse)).BooleanUnion()).ToList();
+
+            if (internalBoundaries.Count != count)
+            {
+                Reflection.Compute.RecordWarning("At least one of the internalBoundaries was overlapping another one. BooleanUnion was used to resolve it.");
+            }
+
+            //// Checks based around external curve
```
pawelbaran, line 104: Wouldn't it be better to do the external curve checks at the beginning?

Geometry_Engine/Create/PlanarSurface.cs
```diff
@@ -79,7 +79,75 @@ public static PlanarSurface PlanarSurface(ICurve externalBoundary, List<ICurve>
...
+            if (externalBoundary is NurbsCurve)
+            {
+                Reflection.Compute.RecordWarning("Neeseccary checks to ensure vadility of the PlanarSurface is not implemented. The PlanarSurfaces curves relations are not garanteueed.");
```
pawelbaran, line 107:
```suggestion
                Reflection.Compute.RecordWarning("External boundary is a nurbs curve. Necessary checks to ensure validity of a planar surface based on nurbs curve cannot be run, therefore correctness of the surface boundaries is not guaranteed.");
```

Geometry_Engine/Create/PlanarSurface.cs
```diff
@@ -79,7 +79,75 @@ public static PlanarSurface PlanarSurface(ICurve externalBoundary, List<ICurve>
...
+                // External done
+                return new PlanarSurface(externalBoundary, internalBoundaries);
+            }
+            else if (externalBoundary.IIsSelfIntersecting())
+            {
+                Reflection.Compute.RecordError("The provided externalBoundary is selfintersecting.");
```
pawelbaran, line 113:
```suggestion
                Reflection.Compute.RecordError("The provided external boundary is self-intersecting.");
```

Geometry_Engine/Create/PlanarSurface.cs
```diff
@@ -79,7 +79,75 @@ public static PlanarSurface PlanarSurface(ICurve externalBoundary, List<ICurve>
...
+                return null;
+            }
+            if (externalBoundary is Ellipse)
```
pawelbaran, line 116: This should be `else if` together with `NurbsCurve`, see above.

Geometry_Engine/Create/PlanarSurface.cs
```diff
@@ -79,7 +79,75 @@ public static PlanarSurface PlanarSurface(ICurve externalBoundary, List<ICurve>
...
+                return new oM.Geometry.PlanarSurface(externalBoundary, internalBoundaries);
+
+
+            // Intersects External
+            for (int i = 0; i < internalBoundaries.Count; i++)
+            {
+                ICurve intCurve = internalBoundaries[i];
+                if (intCurve is NurbsCurve || intCurve is Ellipse)
```
pawelbaran, line 124: Please check (somewhere close to the beginning) if there are any internal curves that are ellipses or nurbs and record an error that they will be not checked?

Geometry_Engine/Create/PlanarSurface.cs
```diff
@@ -79,7 +79,75 @@ public static PlanarSurface PlanarSurface(ICurve externalBoundary, List<ICurve>
...
+
+            // IsContained by External
+            count = internalBoundaries.Count;
+
+            internalBoundaries = internalBoundaries.Where(x => x is NurbsCurve || x is Ellipse || externalBoundary.IIsContaining(x)).ToList();
+
+            if (internalBoundaries.Count != count)
+            {
+                Reflection.Compute.RecordWarning("At least one of the internalBoundaries is not contained by the externalBoundary. And have been disregarded.");
+            }
```
pawelbaran, line 144: Why is this happening so late?

Geometry_Engine/Create/PlanarSurface.cs
```diff
@@ -79,7 +79,75 @@ public static PlanarSurface PlanarSurface(ICurve externalBoundary, List<ICurve>
...
+            
+
+            if (internalBoundaries.Any(x => x is NurbsCurve || x is Ellipse))
+                Reflection.Compute.RecordWarning("Neeseccary checks to ensure vadility of the PlanarSurface is not implemented. The PlanarSurfaces curves relations are not garanteueed.");
```
pawelbaran, line 148: Spelling, please see above.

Geometry_Engine/Create/PlanarSurface.cs
kThorsager: Thought that I shouldn't skip the IsSelfIntersecting method for Ellipses which works, but then again we know the answer already.

So yeah you're right

Geometry_Engine/Create/PlanarSurface.cs
kThorsager: If a internalCurve intersects the external one it's not contained and needs to be handled by boolean difference, which is what happens before this.

Both of these can only happen if the external curve isn't a NurbsCurve/Ellipse, which happens just before that.

Geometry_Engine/Create/PlanarSurface.cs
kThorsager: I wanted to save what could be saved from the internal curves before returning the planarSurface, which happens here if it's a Nurbs/Ellipse.

Could do another check in the beginning for selfintersections ignoring Nurbs and such, but struck me as even more messy

Geometry_Engine/Create/PlanarSurface.cs
```diff
@@ -55,31 +57,98 @@ public static PlanarSurface PlanarSurface(ICurve externalBoundary, List<ICurve>
...
-                if (!crv.IIsClosed())
-                {
-                    Reflection.Compute.RecordError("At least one of the internal edge curves is not closed");
-                    return null;
-                }
+            //---------------Coplanar-Internal-Boundaries-------------------//
+            count = internalBoundaries.Count;
+
+            Plane p = externalBoundary.IFitPlane();
+            internalBoundaries = internalBoundaries.Where(x => x.IIsInPlane(p)).ToList();
```
pawelbaran, line 74: `IsInPlane` is not implemented for `Ellipse` and `NurbsCurve` - sounds like a 5 min job, possibly could be added in this PR?

Geometry_Engine/Create/PlanarSurface.cs
```diff
@@ -55,31 +57,98 @@ public static PlanarSurface PlanarSurface(ICurve externalBoundary, List<ICurve>
...
+                Reflection.Compute.RecordWarning("External boundary is a nurbs curve or Ellipse. Necessary checks to ensure validity of a planar surface based on nurbs curve cannot be run, therefore correctness of the surface boundaries is not guaranteed.");
+                // External done
+                return new PlanarSurface(externalBoundary, internalBoundaries);
+            }
+
+            //--------------SelfIntersecting-External-Boundary--------------//
+            if (externalBoundary.IIsSelfIntersecting())
+            {
+                Reflection.Compute.RecordError("The provided external boundary is self-intersecting.");
+                return null;
```
pawelbaran, line 118: Hmm, I would put that right after the first two external boundary checks - then we have all 4 external checks at the beginning.

Geometry_Engine/Create/PlanarSurface.cs
```diff
@@ -55,31 +57,98 @@ public static PlanarSurface PlanarSurface(ICurve externalBoundary, List<ICurve>
...
+            count = internalBoundaries.Count;
+
+            internalBoundaries = internalBoundaries.Where(x => x is NurbsCurve || x is Ellipse || externalBoundary.IIsContaining(x)).ToList();
+
+            if (internalBoundaries.Count != count)
+            {
+                Reflection.Compute.RecordWarning("At least one of the internalBoundaries is not contained by the externalBoundary. And have been disregarded.");
+            }
+            
+            //------------------Return-Valid-Curve--------------------------//
```
pawelbaran, line 150: Valid surface

Geometry_Engine/Create/PlanarSurface.cs
pawelbaran: Oke, I trust you are right (cannot find a valid argument against)

Geometry_Engine/Create/PlanarSurface.cs
kThorsager: Just assumed that it would be, but yes ofc

Geometry_Engine/Create/PlanarSurface.cs
pawelbaran: Or, after seeing previous comment, this refers only to self-intersection, which def could be moved to the beginning.

Geometry_Engine/Create/PlanarSurface.cs
pawelbaran: Makes sense, I commented over myself in my latest review.

pawelbaran:
Looks like we are getting to the final solution with @kThorsager, but I am still not sure if the current approach is the best we can get (particularly taken into account @epignatelli's concerns). Personally I would feel more comfortable if someone else could independently confirm if the proposed changes are fine.

@al-fisher @IsakNaslundBh @epignatelli

al-fisher:
Thanks @kThorsager, and others above commenting on this. Sorry it has taken me a while to consider and comment back on this! 🙈 

Following our discussion and reviewing the approach - I do think this is a sensible compromise.
That is in summary:
a) Providing all of the checks on planarity, overlapping and self-intersection on creation of the planar surface as discussed, but
b) _Not_ limiting creation where inputs include the specific `Nurbs` and `Ellipse` curve types where we cannot fully perform checks,, but to instead raise warnings as appropriate.

@epignatelli fair comments raised. Thanks
As discussed @kThorsager and others are considering which additional not implemented methods can be fixed this sprint which could help. But also will not fully address limitations here.

Equally, and I think this was a good point by one of @kThorsager, @IsakNaslundBh, @pawelbaran - it is not trivial to choose the correct conservative response always - thus I do like the approach of returning response with a clear warning - rather than aborting entire workflow with an exception.

Happy to approve this approach.

al-fisher:
@pawelbaran - are you happy (now I have confirmed I am happy!! 😄 ) 
Looking at thread above - just checking before dismissing your requested changes 😃 

Thanks!

al-fisher:
Super - thanks @pawelbaran 😄

PR #1765, opened 2020-05-13
Adding Count method for IOutput

### LINKED ISSUE #1764: Reflection_Engine: Add Count method for IOutput
#### Description:

Add method checking how many items and output have, when working with the base interface

IsakNaslundBh:

Closes #1764 

Adding count method to the IOutput

### Test files

https://burohappold.sharepoint.com/:f:/s/BHoM/EmTK6dEPwhlJk2qpsVRgEGsBdOmcyB_u9Sw5WXMU_KGSNQ?e=qQPl5p


IsakNaslundBh:
> This LGTM, not much to really say about it! Guessing this is to help with the Data-Driven unit tests? 😄 I can see it being useful there at least.

Indeed for the data driven unit tests, to extract the results from the method :)

FraserGreenroyd:
To confirm, I have reviewed the compliance report as well and am happy the report is flagging a known issue that is out of scope for this PR.

IsakNaslundBh:
Pushed up a slight change, making the method for the IOutput named just count and the ones for the Output.

This was an atempt to stop projects that are using Reflection_Engine but not Reflection_oM from complaining when the linq Count() method was being used. Realised, can not have overlapping methodnames at all, so will Keep the ICount, but other methods private.

IsakNaslundBh:
To resolve issue mentioned above, decided to move all code to a separate method instead.

Reflection_Engine/Query/OutputCount.cs
```diff
@@ -0,0 +1,96 @@
...
+
+        /***************************************************/
+        /**** Private Methods - fallback                ****/
+        /***************************************************/
+
+        private static object OutputCount(this object output)
+        {
+            Compute.RecordError("Could not find a suitable OutputCount method for the object provided. Method returns -1");
+            return -1;
+        }
```
adecler, line 90: I would recommend a few changes: 
- input can be an IObject instead of an object since this is the only things that the I method accepts. No big deal here but probably good to set good examples everywhere we can.
- this should return an int, instead of an object. Same for some of the methods above.
- you don't have to give up here. There is a simple way to get the result you want through reflection:
```c#
private static object OutputCount(this IOutput output)
{
    if (output == null)
        return 0;
    else
        return output.GetType().GenericTypeArguments.Count();
}
```
This is quite important since we might end up adding a few more `Output` with more generics in the future and forget to update here. 
- Finally, I am not sure avoiding reflection is the right approach here. Why not use a single method instead ?

IsakNaslundBh: Can get behind using reflection for this. Honestly did not thought of that way of doing it, or I would Have done it. Will replace my code with your method.

adecler:
Given that I pretty much wrote the code that is currently submitted, it would probably be safer for it to be reviewed by someone else.

FraserGreenroyd:
@adecler are you happy to dismiss your 'request changes' review to allow this to be merged?

PR #1781, opened 2020-05-19
Geometry_Engine: Handle Nurbs in PolyCurves for IsClosed & Join

### LINKED ISSUE #1744: Spatial Engine: Translate IElement2D Removes Irregular Planar Surface
#### Description:
Translate fails to move all trimmed surfaces provided as elements.

#### Steps to reproduce:
Feed irregular planar surface geometry to translate tool

#### Expected behaviour:
Translate moves all planar surfaces input to component

#### Test file(s):
spatial_engine_translate_bug

kThorsager:

Closes #1744 
IsClosed for PolyCurves now calls the interface if there's only one sub-curve
and Join returns the sub-Curve if there's only one.

Both of which are used when setting the geometry back on `IPhysical` `ISurfaces` after use in Spatial_Engine.

### Test files

https://burohappold.sharepoint.com/:u:/s/BHoM/EcbPrxtNFaFIgsNGsNphyNUBDcJQBX7CRRBR6nXN3zrfrQ?e=pP7hbU


Geometry_Engine/Compute/Join.cs
```diff
@@ -148,6 +148,9 @@ public static List<Polyline> Join(this List<Polyline> curves, double tolerance =
 
         public static List<PolyCurve> IJoin(this List<ICurve> curves, double tolerance = Tolerance.Distance)
         {
+            if (curves == null)
```
pawelbaran, line 151: Shouldn't a `null` be returned when the subject of operation is `null`?

Geometry_Engine/Compute/Join.cs
```diff
@@ -158,6 +161,9 @@ public static List<PolyCurve> IJoin(this List<ICurve> curves, double tolerance =
                     sections.Add(new PolyCurve { Curves = curve.ISubParts().ToList() });
             }
 
+            if (sections.Count == 1)
```
pawelbaran, line 164: Couldn't this be:

```suggestion
            if (sections.Count < 2)
```

kThorsager:
Fair points @pawelbaran , fixed and updated from master due to BHoM changes

PR #1783, opened 2020-05-20
Structure_Engine: Remove Common_Engine reference

kThorsager:

 #1624 
and allows for #1759 to work


Structure_Engine/Query/Geometry.cs
```diff
@@ -23,7 +23,7 @@
 using BH.oM.Geometry;
 using BH.oM.Structure.Elements;
 using BH.oM.Structure.SectionProperties;
-using BH.Engine.Common;
+using BH.Engine.Spatial;
```
IsakNaslundBh, line 26: This is not even used actually, so might as well get rid of it isntead of replacing

PR #1787, opened 2020-05-21
Spatial_Engine: Restructure ISetElement and INewElement

### LINKED ISSUE #1785: Spatial_Engine: Restructure SetElement and NewElement
#### Description:
The `NewElement` methods are only used to create default elements to be used in `SetElement`.

This can be streamlined by letting a `IGeometry` be seen as a default element by `SetElement`, which would have the added benefit of letting `Bar`, which has properties "connected" to the end stored on the `Bar` instead of the `Node` to default them as well.

This won't be done for `NewInternalElement2D` as we don't have a `IElement` interface on any surface and that would be a bigger exercise.

#### Action
Update `SetElement0D` & `SetElement1D` to default the element if a `IGeometry` is provided.

Add `ToBeRemoved` attribute to all `NewElement0D` & `NewElement1D` methods and return the geometry instead.
(or might version all of them into the interface method to keep the code cleaner.
or just remove them right away as given their low amount of usefulness I don't think there's widespread use)

kThorsager:
Closes #1785
As specified in change-log

### Test files

Tests that the defaulting for the only element (`Bar`) which needs it is working.
https://burohappold.sharepoint.com/:u:/s/BHoM/EXdhou5wJ9xFmUZReD9Inx0B8LoeNgGw6H28RWnIghZkrg?e=yJAyoE

### Changelog
- `ISetElement` `0D` & `1D` methods now defaults if provided a `IGeometry`
- `INewElement` `0D` & `1D` methods are no longer needed and now return the inputted point/curve awaiting removal

### Additional comments

No urgent reason to delete the new methods in each project, but will tag them as `ToBeRemoved and raise separate issues to be resolved at toolkit(project?) leads discretion.
raised in #1788 

Not versioning since it not really anything to version to (maybe the components second input to a `IGeometry` holder component? (don't think that's possible))

### DIFF
--- a/Structure_Engine/Modify/SetElements0D.cs
+++ b/Structure_Engine/Modify/SetElements0D.cs
@@ -56,7 +56,8 @@ namespace BH.Engine.Structure
             if (newElements0D[0] is Point)
             {
                 clone.StartNode = Create.Node(newElements0D[0] as Point);
-                clone.Release.StartRelease = Create.FixConstraint6DOF();
+                if (clone.Release != null)
+                    clone.Release.StartRelease = Create.FixConstraint6DOF();
             } else
                 clone.StartNode = newElements0D[0] as Node;
 
@@ -64,7 +65,8 @@ namespace BH.Engine.Structure
             if (newElements0D[1] is Point)
             {
                 clone.EndNode = Create.Node(newElements0D[1] as Point);
-                clone.Release.EndRelease = Create.FixConstraint6DOF();
+                if (clone.Release != null)
+                    clone.Release.EndRelease = Create.FixConstraint6DOF();
             }
             else
                 clone.EndNode = newElements0D[1] as Node;
--- a/Structure_Engine/Modify/SetElements0D.cs
+++ b/Structure_Engine/Modify/SetElements0D.cs
@@ -37,9 +37,10 @@ namespace BH.Engine.Structure
         /****            IElement1D            ****/
         /******************************************/
 
-        [Description("Sets the IElement0Ds of the Bar, i.e. its two end Nodes. Method required for IElement1Ds.")]
+        [Description("Sets the IElement0Ds of the Bar, i.e. its two end Nodes or Points. Method required for IElement1Ds.")]
         [Input("bar", "The Bar to set the IElement0Ds to.")]
-        [Input("newElements0D", "The new IElement0Ds of the Bar. Should be a list of length two, containing exactly two structural Nodes.")]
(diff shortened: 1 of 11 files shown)


Structure_Engine/Modify/SetElements0D.cs
```diff
@@ -49,9 +50,25 @@ public static Bar SetElements0D(this Bar bar, List<IElement0D> newElements0D)
...
-            Bar clone = bar.GetShallowClone() as Bar;
-            clone.StartNode = newElements0D[0] as Node;
-            clone.EndNode = newElements0D[1] as Node;
+            Bar clone = bar.DeepClone() as Bar;
+
+            // Default the Bars end if the input is an Point
+            if (newElements0D[0] is Point)
+            {
+                clone.StartNode = Create.Node(newElements0D[0] as Point);
+                clone.Release.StartRelease = Create.FixConstraint6DOF();
```
IsakNaslundBh, line 59: Another thing I realised. We will want to null check the release class here before assigning to it. Basically only set the start release here if the Release != null

Structure_Engine/Modify/SetElements0D.cs
```diff
@@ -49,9 +50,25 @@ public static Bar SetElements0D(this Bar bar, List<IElement0D> newElements0D)
...
+            } else
+                clone.StartNode = newElements0D[0] as Node;
+
+            // Default the Bars end if the input is an Point
+            if (newElements0D[1] is Point)
+            {
+                clone.EndNode = Create.Node(newElements0D[1] as Point);
+                clone.Release.EndRelease = Create.FixConstraint6DOF();
```
IsakNaslundBh, line 67: as above

Structure_Engine/Modify/SetElements0D.cs
kThorsager: oh, good catch

Structure_Engine/Modify/SetElements0D.cs
```diff
@@ -49,9 +50,27 @@ public static Bar SetElements0D(this Bar bar, List<IElement0D> newElements0D)
...
-            clone.EndNode = newElements0D[1] as Node;
+            Bar clone = bar.DeepClone() as Bar;
+
+            // Default the Bars end if the input is an Point
+            if (newElements0D[0] is Point)
+            {
+                clone.StartNode = Create.Node(newElements0D[0] as Point);
+                if (clone.Release != null)
+                    clone.Release.StartRelease = Create.FixConstraint6DOF();
+            } else
```
IsakNaslundBh, line 61: final thing, being really picky, can we add a new line between the "}" and the "else"

PR #1809, opened 2020-05-27
Structure_Engine: Update modifier methods to work with new fragment isntead of custom data

### LINKED ISSUE #1808: Structure_Engine: Make ApplyModifiers and GetModifiers method store the data using fragments instead of CustomData
#### Description:

Align with changes made in https://github.com/BHoM/BHoM/issues/578

IsakNaslundBh:
### NOTE: Depends on 

https://github.com/BHoM/BHoM/pull/895
   
Closes #1808

Making modifier methods depend on fragment rather than CustomData

### Additional comments

As the methods using the fragment input and the return the same things as before, I think it should be safe to merge this before dependant PRs have been approved.

### DIFF
--- a/Structure_Engine/Query/Modifiers.cs
+++ b/Structure_Engine/Query/Modifiers.cs
@@ -37,9 +37,9 @@ namespace BH.Engine.Structure
         /***************************************************/
 
         [Description("Gets any modifiers from a property as an array of doubles. The modifiers are used to scale one or more of the property constants for analysis. Constants are multiplied with the modifiers, hence a modifier value of 1 means no change. \n" +
-                     "The values returned are in the following order: FXX, FXY, FYY, MXX, MXY, MYY, VXZ, VYZ, Mass, Mass. Method returns null if no modifiers are found.")]
+                     "The values returned are in the following order: FXX, FXY, FYY, MXX, MXY, MYY, VXZ, VYZ, Mass, Weight. Method returns null if no modifiers are found.")]
         [Input("property", "The property to extract modifiers from.")]
-        [Output("modifiers", "Returns the modifier values of the property as a double array in the following order: f11, f12, f22, m11, m12, m22, v13, v23, mass, weight. Returns null if no modifiers are found.")]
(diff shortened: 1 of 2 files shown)


IsakNaslundBh:
> I think the reference to a weight modifier should be removed, given that it's not implemented in the oM it may lead to confusion. It's an input for the Modify method, but is the the same output (and in the description is called Mass) as the mass modifier.
> 
> I think it would be clearer to remove it until it's actually implemented, if it is at all.

Was considering doing it, but think the method is used, and assuming the double[] to be of a specific length in a couple of toolkits and did not want to break them, but will check where used and see if I can tackle them at the same time. Agree, should not have it in if we do not want it :)

IsakNaslundBh:
Have included Weight for the SurfacePropertyModifier now, so methods should be inputting and outputting exactly the same thing, with no modifications @peterjamesnugent

IsakNaslundBh:
Updating from master to please the CI

Structure_Engine/Query/Modifiers.cs
```diff
@@ -34,38 +36,34 @@ public static partial class Query
         /**** Public Methods                            ****/
         /***************************************************/
 
-        [Description("Gets any modifiers from a property as an array of doubles. The modifiers are used to scale one or more of the property constants for analysis. Constants are multiplied with the modifiers, hence a modifier value of 1 means no change. \n" + 
-                     "The values returned are in the following order: f11, f12, f22, m11, m12, m22, v13, v23, mass, weight. Method returns null if no modifiers are found.")]
+        [Description("Gets any modifiers from a property as an array of doubles. The modifiers are used to scale one or more of the property constants for analysis. Constants are multiplied with the modifiers, hence a modifier value of 1 means no change. \n" +
+                     "The values returned are in the following order: FXX, FXY, FYY, MXX, MXY, MYY, VXZ, VYZ, Mass, Mass. Method returns null if no modifiers are found.")]
```
peterjamesnugent, line 40: Description needs to be updated to include weight

PR #1831, opened 2020-06-08
Result_Engine: Aligning envelope methods with changes in Structure_oM

IsakNaslundBh:
### NOTE: Depends on 

https://github.com/BHoM/BHoM/pull/912 

Closes #

Aligning with changes to structure_oM, making results immutable

Also adding check if any versioning has been applied in the `DeprecatedSerializer` to avoid stack overflow as the method otherwise can call itself indefinitely if the same thing keep on being returned (if something is failing in the de-serialisation/upgrade process). Think this is something we would want, even if we can find a proper fix for the upgraders to kick in appropriately. @adecler


adecler:
@IsakNaslundBh , the changes you made to the serialiser look OK to me. Let's have a quick chat just to make sure though.

PR #1843, opened 2020-06-15
Geometry_Engine: Better handling of corner cases in IsContaining

### LINKED ISSUE #1214: Geometry_Engine: IsContaining returns true incorrectly when tolerance is changed
#### Description:

I have here an example where the opening was picked by panels and is quite far, ...
image

[test script](https://burohappold.sharepoint.com/:f:/r/sites/BHoM/02_Current/12_Scripts/01_Test%20Scripts/BHoM_Engine/Environment_Engine/Environmnet_Engine-Issue-1213%20AddOpening?csf=1&e=Fdc7n0)

kThorsager:
Closes #1214 

`IsContaining`s algorithm assumed that if a line, with end at point A,'s intersection point with an line L was within the distance tolerance, then any line which endpoint is at the point A will intersect the line L.

unfortunately not true,
This PR adds a relative factor based on their angles to query if the other intersection will happen.

This approach only works for linear cases, so added a sanity check for the `PolyCurve` case to avoid the most horrendous errors

### Test files

Generic test:
https://burohappold.sharepoint.com/:u:/s/BHoM/EeLGxufNopJMnOX4g1fMMdoBt8NybP-hh7ytJFvGGoWFRw?e=hfBWhG

Specific test:
See issue

### DIFF
--- a/Geometry_Engine/Query/IsContaining.cs
+++ b/Geometry_Engine/Query/IsContaining.cs
@@ -191,20 +191,29 @@ namespace BH.Engine.Geometry
                             List<Point> intersects = new List<Point>();
                             List<Point> extraIntersects = new List<Point>();
 
-                            foreach (Line subPart in subParts)
+                            Func<double, double, double> ToFactor = (t, n) => (1 - t * t) / (1 - n * n);
+
+                            Line current = subParts[1];
+                            double prevTolFactor = ToFactor(subParts[0].Direction().DotProduct(direction), current.Direction().DotProduct(direction));
+
+                            for (int i = 1; i < subParts.Count + 1; i++)
                             {
-                                Point iPt = subPart.LineIntersection(ray, false, tolerance);
+                                Line next = subParts[(i + 1) % subParts.Count];
+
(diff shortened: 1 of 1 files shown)


pawelbaran:
@kThorsager do we have a problem-specific test script for this one? I am trying to understand the issue but it is not easy without a test file.

kThorsager:
I used this one from the issue: [Test file](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4&viewid=a5a93cba%2Dfcca%2D46cc%2Da31a%2D80b8d5cbfd7b&id=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FTest%20Scripts%2FBHoM%5FEngine%2FEnvironment%5FEngine%2FEnvironmnet%5FEngine%2DIssue%2D1214%20AddOpening)

kThorsager:
Thought I had been good by passing the variable down.
But did not think about how the Bounding Box check is 3d and we're doing a 2d check, so a point in the middle of a closed planar curve is on the edge of the bounding box.

anyway easy fix and good spot @pawelbaran

PR #1880, opened 2020-07-01
Geometry_Engine: Minor method addition

### LINKED ISSUE #1616: Geometry_Engine: Missing methods for PlanarSurface
#### Broken rules:
All `I` methods that take an `IGeometry` or an `ISurface` as first input should have a corresponding method for all possible geometry/surface. This is not the case for `PlanarSurface`.

(This was highlighted in https://github.com/BHoM/Grasshopper_Toolkit/issues/489)

Implement all missing geometry methods for PlanarSurface.

### Comments:
- @FraserGreenroyd, @awakeman , this is/will be part of the CI check you are doing on compiled code. So I'm flagging you up here in case you are already able to provide the full list of missing methods (potentially for more than PlanarSurface).
- Assigning @IsakNaslundBh and @pawelbaran since they were the ones assigned for the creation of the PlanarSurface (see https://github.com/BHoM/BHoM/issues/424) but feel free to reassign appropriately.

### LINKED ISSUE #1771: Geometry_Engine: Add vector from two points method

### LINKED ISSUE #1772: Geometry_Engine: Add IIsNurbsCurve to check if any part of a curve is a nurbs
#### Description:

image

### LINKED ISSUE #1878: Geometry_Engine: implement Flip for NurbsCurve
#### Description:

I've experienced a few issues caused by that one missing. Please correct me if I am wrong, shouldn't it be enough to reverse the order of control points, weights and the knot vector?

### LINKED ISSUE #1879: Geometry_Engine: IsEqual for Basis
#### Description:

Needed for ETABS checks

kThorsager:

Closes #1771 
Closes #1772 
Closes #1878 
Closes #1616 
Closes #1879 

Added a bunch of small requested methods

### Test files

https://burohappold.sharepoint.com/:u:/s/BHoM/EcsuiThG1spMufdZ7gsofIgB7vijPXV-FYZtBkBcNtwenA?e=e6OxIq

### Changelog

- Added Create Vector from two Points
- Added IsNurbsCurve
- Implemented: Clone, ClosestPoint, IsEqual and IsInPlane for PlanarSurface
- Added IsEqual for Basis
- Implemented Flip for NurbsCurve

### DIFF
--- a/Geometry_Engine/Query/Clone.cs
+++ b/Geometry_Engine/Query/Clone.cs
@@ -138,7 +138,7 @@ namespace BH.Engine.Geometry
 
         public static PlanarSurface Clone(this PlanarSurface surface)
         {
-            return new PlanarSurface(surface.ExternalBoundary, surface.InternalBoundaries.ToList());
+            return new PlanarSurface(surface.ExternalBoundary.IClone(), surface.InternalBoundaries.Select(x => x.IClone()).ToList());
         }
 
         /***************************************************/
--- a/Geometry_Engine/Query/Clone.cs
+++ b/Geometry_Engine/Query/Clone.cs
@@ -136,6 +136,13 @@ namespace BH.Engine.Geometry
 
         /***************************************************/
 
+        public static PlanarSurface Clone(this PlanarSurface surface)
+        {
+            return new PlanarSurface(surface.ExternalBoundary, surface.InternalBoundaries.ToList());
+        }
+
+        /***************************************************/
+
         public static Pipe Clone(this Pipe surface)
         {
             return new Pipe { Centreline = surface.Centreline.IClone(), Radius = surface.Radius, Capped = surface.Capped };
--- a/Geometry_Engine/Query/IsNurbsCurve.cs
+++ b/Geometry_Engine/Query/IsNurbsCurve.cs
@@ -42,7 +42,9 @@ namespace BH.Engine.Geometry
         {
             return curve.ISubParts().Any(x => x is NurbsCurve); ;
         }
-        
+
+        /***************************************************/
+
     }
 }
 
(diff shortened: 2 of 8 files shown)


Geometry_Engine/Query/IsNurbsCurve.cs
```diff
@@ -0,0 +1,48 @@
...
+        /***************************************************/
+
+        [Description("Queries if any part of the curve is a nurbs curve.")]
+        [Input("curve", "Curve to query.")]
+        [Output("true if any part of the curve is a nurbs curve.")]
+        public static bool IsNurbsCurve(this ICurve curve)
+        {
+            return curve.ISubParts().Any(x => x is NurbsCurve); ;
+        }
+
```
pawelbaran, line 45: `/***************************************************/` missing 🤖

kThorsager:
> No support for single points in `IsContaining` keeps on bothering me (each time it is called in the code, a single point gets wrapped into a list), but not sure if that is something we could make any better (apart from adding yet another set of methods taking a single `Point`). Anyways, that is out of the scope of this PR, just my personal reflection.

hm, that is annoying, suspect adding a wrapping method which just puts the single point in a list and call the method. As some things are pre-processed,, so flipping is isn't great

pawelbaran:
100% agreed @kThorsager. Then again, not sure if the wrapper method will not cause more confusion than comfort (more methods with same name etc.). What do you think?

kThorsager:
It will be more methods, but it would be enough to add it to the interface one. And the do the same thing, so confusion should be minimal.

pawelbaran:
In such case, I would vote for adding one and replacing all calls taking a list with single points with the actual points. However, very low prio for sure.

Geometry_Engine/Query/Clone.cs
```diff
@@ -136,6 +136,13 @@ public static NurbsSurface Clone(this NurbsSurface surface)
 
         /***************************************************/
 
+        public static PlanarSurface Clone(this PlanarSurface surface)
+        {
+            return new PlanarSurface(surface.ExternalBoundary, surface.InternalBoundaries.ToList());
```
IsakNaslundBh, line 141: Think the boundaries, internal and external, should be cloned here as well. That has at least been the general approach for the rest of the close methods.

kThorsager: Oh, yeah, that makes sense

PR #1883, opened 2020-07-02
Diffing_Engine: Added Newtonsoft.json.dll in xcopy post build event

### LINKED ISSUE #1882: Diffing_Engine: Copy of Newtonsoft.json.dll not in post build
Post build for Diffing_Engine needs to include a copy of Newtonsoft.json.dll.

alelom:

Closes #1882

### Test files

Not needed.

### DIFF
--- a/Diffing_Engine/Diffing_Engine.csproj
+++ b/Diffing_Engine/Diffing_Engine.csproj
@@ -54,7 +54,7 @@
     </Reference>
     <Reference Include="Newtonsoft.Json, Version=12.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
       <HintPath>..\packages\Newtonsoft.Json.12.0.2\lib\net45\Newtonsoft.Json.dll</HintPath>
-      <Private>False</Private>
+      <Private>True</Private>
     </Reference>
     <Reference Include="Reflection_oM">
       <SpecificVersion>False</SpecificVersion>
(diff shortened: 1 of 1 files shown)


adecler:
Can we just get rid of Newtonsoft instead please ?

FraserGreenroyd:
> Can we just get rid of Newtonsoft instead please ?

Seconded!

alelom:
The short answer is no, it's complicated.

FraserGreenroyd:
> The short answer is no, it's complicated.

😿 

What about the long answer? Maybe that is more fun?

alelom:
> > The short answer is no, it's complicated.
> 
> 😿
> 
> What about the long answer? Maybe that is more fun?

It's not like I had not spent the development time trying to avoid adding this dependency. I remember that removing it would require some work. I'd be happier to focus on other things.

FraserGreenroyd:
> It's not like I had not spent the development time trying to avoid adding this dependency. I remember that removing it would require some work. I'd be happier to focus on other things.

That was not more fun 😿 

Maybe we can go for a compromise for now then, we'll review/merge this PR as appropriate but we perhaps do a session with @adecler to see if we can remove this dependency over the course of this milestone? Is that ok? 🙏

alelom:
Ok happy to do but this milestone might be already too full for me, at least its first half for sure. During the last half sounds better, or even better during the next. Will raise issue.

alelom:
Issue here: https://github.com/BHoM/BHoM_Engine/issues/1885

adecler:
Happy to leave the removal of Newtonsoft for later if it's more complicated than it looks from the code itself.
Let's definitely have a workshop on it this milestone though as I am keen to make sure absolutely ALL serialisation happens though the serialisation engine to guarantee consistent behaviour.

PR #1888, opened 2020-07-02
Geometry_Engine: Update PointClustersDBSCAN to use DomainTree

### LINKED ISSUE #1881: Geometry_Engine: Use DomainTree for PointClusters
#### Description:

It would be a lot quicker if it didn't have to loop through the entire list for each new cluster, hence the `DomainTree`

kThorsager:

Closes #1881 
Had some hopes about making this general for DBSCAN and making them use `INode`, but there were annoyances as it i not a `IEnumerable`, there are difficulties in changing the objects in the tree without messing with the data and the DBSCAN method would need to be refactored to look at things per Collection rather than per Item.

And just adding clusters for IGeometries or ICurves was also annoying since no general distance method existed and would add a extra layer to this method. And there doesn't seem to be any huge need either so.

But the speed up is considerable, especially for large sets of points. 😁
(40000 points, 1.3m to 3.4s in the test file)

### Test files

https://burohappold.sharepoint.com/:u:/s/BHoM/Ec8Vk7lcL1dCgXq-QdewJokBfJ4s8jwT5hIATIwiA-3WWg?e=RqtVBu

### DIFF
--- a/Data_Engine/Compute/DomainTreeClusters.cs
+++ b/Data_Engine/Compute/DomainTreeClusters.cs
@@ -40,42 +40,40 @@ namespace BH.Engine.Data
 
             List<T> items = new List<T>(data);
             List<bool> check = items.Select(x => false).ToList();
-            DomainTree<int> indexTree = Data.Create.DomainTree(items.Select((x, i) => Data.Create.DomainTreeLeaf(i, toDomainBox(x))));
+            List<DomainBox> domainBoxes = data.Select(x => toDomainBox(x)).ToList();
+            DomainTree<int> indexTree = Data.Create.DomainTree(domainBoxes.Select((x, i) => Data.Create.DomainTreeLeaf(i, x)));
 
             List<List<T>> result = new List<List<T>>();
-            List<T> toEvaluate = new List<T>();
+            List<int> toEvaluate = new List<int>();
 
             int count = -1;
             for (int i = 0; i < items.Count; i++)
             {
-                T pivot = items[i];
-
                 if (check[i])
                     continue;
 
                 result.Add(new List<T>());
                 count++;
-                toEvaluate.Add(pivot);
+                toEvaluate.Add(i);
                 check[i] = true;
 
                 while (toEvaluate.Count > 0)
                 {
                     // Find all the neighbours for each item in toEvaluate, and add them in toEvaluate
-                    foreach (int index in Data.Query.ItemsInRange<DomainTree<int>, int>(indexTree, x => TreeEvaluator(x.DomainBox, toDomainBox(toEvaluate[0]))))
+                    foreach (int index in Data.Query.ItemsInRange<DomainTree<int>, int>(indexTree, x => TreeEvaluator(x.DomainBox, domainBoxes[toEvaluate[0]])))
                     {
                         if (!check[index])
                         {
-                            T item = items[index];
-                            if (itemEvaluator(toEvaluate[0], item))
+                            if (itemEvaluator(items[toEvaluate[0]], items[index]))
                             {
-                                toEvaluate.Add(item);
+                                toEvaluate.Add(index);
                                 check[index] = true;
                             }
                         }
                     }
 
                     // move the checked items from toEvaluate to result
-                    result[count].Add(toEvaluate[0]);
+                    result[count].Add(items[toEvaluate[0]]);
                     toEvaluate.RemoveAt(0);
                 }
             }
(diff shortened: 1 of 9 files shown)


kThorsager:
> Works fine, LGTM, but two questions ring in my head:
> 
> * before the proposed change, `PointClusetersDBSCAN` was and example of implementation of the generic DBSCAN algorithm, while now it does not have anything to do with it any more

It doesn't have anything to do with that specific method in Data_Engine, but it is still using the DBSCAN algorithm

> * if that can work on vectors in 3-dimensional space, couldn't it be generalized a bit cheaply?
> 
> These translate into: couldn't the tree become a part of the generic DBSCAN algorithm? I have read the PR description, but still, that would be soo nice!

I agree that that would be super nice, so I'll give it another go and see if I can be less annoyed and/or put in the extra work to make it feasible

kThorsager:
So it is still annoying.

To make use of something similar to the existing version one would have to copy the tree structure over with an other collection of data, and iterate over that tree, and make sure the queries does not remove the object reference.

First one can't be done with the interface as it is structured as it would by necessity return an interface, which breaks everything as it's recursively defined. The second one is possible but exceedingly annoying without making it an `IEnumerable` which would break the UI. And could at least not form a quick look make the third one work.

So to copy the one previously here where I keep the incidences in a tree. This does not work for the interface as it's kind of a special case that it can be traversed without the data present, which would need to be exchanged for `int`s.

So we got a method for any thing which one can construct a `DomainBox` from and wish to cluster based on those in some manner.


Will add documentation if this is general enough. @pawelbaran

Data_Engine/Compute/DomainTreeClusters.cs
```diff
@@ -0,0 +1,89 @@
...
+        /***************************************************/
+
+        public static List<List<T>> DomainTreeClusters<T>(this List<T> data, Func<T, DomainBox> toDomainBox, Func<DomainBox, DomainBox, bool> TreeEvaluator, Func<T, T, bool> itemEvaluator, int minPointCount = 1)
+        {
+            if (data.Count == 0)
+                return new List<List<T>>();
+
+            List<T> items = new List<T>(data);
+            List<bool> check = items.Select(x => false).ToList();
+            DomainTree<int> indexTree = Data.Create.DomainTree(items.Select((x, i) => Data.Create.DomainTreeLeaf(i, toDomainBox(x))));
```
pawelbaran, line 43: Wouldn't that be better to call `toDomainBox` only once (now it is called here and in line 64)? Not sure whether any improvement in practice, but assuming that the method can be as heavy as one can imagine, that could help avoiding overhead.

Geometry_Engine/Compute/PointClustersDBSCAN.cs
```diff
@@ -33,12 +34,14 @@ public static partial class Compute
...
         /***************************************************/
-
+        
         public static List<List<Point>> PointClustersDBSCAN(this List<Point> points, double maxDist, int minPointCount = 1)
         {
-            double maxSqrDist = maxDist * maxDist;
-            Func<Point, Point, bool> metricFunction = (x, y) => x.SquareDistance(y) <= maxSqrDist;
-            return points.ClusterDBSCAN(metricFunction, minPointCount);
+            double sqDist = maxDist * maxDist;
+            Func<Point, DomainBox> toDomainBox = a => a.IBounds().DomainBox();
```
pawelbaran, line 41: Can't we simply create a `DomainBox` without calling a chain of methods here?

Data_Engine/Compute/DomainTreeClusters.cs
kThorsager: Excellent point, can just pre-process them and store in a list 👍

Geometry_Engine/Compute/PointClustersDBSCAN.cs
kThorsager: Yes, (but it's a bit ugly)

kThorsager:
> It works nice and super fast, I like the way of generalization (could possibly be used to other data types than 3-dimensional goemetry - correct me if I am wrong @kThorsager).

Yes, anything which one can search for in a DomainTree and which can be clustered by its abstract DomainBoxes proximity (which is in a arbitrary dimension).

The one thing is that this only supports the one Create method for the tree, DomainBoxes encompassing DomainBoxes, which should be fine. But then again, who knows how this might be used one day.

PR #1898, opened 2020-07-09
Geometry_Engine: Fix for centroid and normal

### LINKED ISSUE #1896: Geometry_Engine: Centroid does not always return a point in a polygon
#### Description:
Some polylines which represent a closed polygon are not getting the correct centroids returned. This causes issues in workflows relying on the centroid of a polyline. Because the `PointInRegion` method also uses a centroid to begin with, this does not provide a workaround. `Centre` does provide a workaround, but does not work for the reasons `Centroid` exists in the first place.

#### Steps to reproduce:
I have isolated 4 polylines that have this issue in the linked test file. Run the test file and see that the centroid is no where near the original polyline.

#### Expected behaviour:
That the centroid is actually somewhere vaguely near the polyline it's for (ideally, somewhere in the centre of it...)

#### Test file(s):
[Download here](https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/01_Test%20Scripts/BHoM_Engine/Geometry_Engine/Centroid%20Bug/TestFile.gh?csf=1&web=1&e=Y3m7vP)

FYI: @rboulton-BH

IsakNaslundBh:

Closes #1896 

Fixes the Normal method, using average of points instead of just looking at the first. As IsClockwise is being called, the direction of the normal will be fixed there.

Simplifying the Centroid methods for Polyline and PolyCurve. Given that the normal is now checked, and ensured to be in the right direction, should be able to rely on that, and not have to preform the "first normal" checks and direction checks, which was the cause of the problem in the issue.

As this is changing some quite core methods, would like to see quite rigorous testing on methods making use of this. I have tried testing through multiple generic cases, and have so far not had any issues, but would be thankful if more edgecases could be thought of that could have the potential to go wrong.

### Test files

Issue test:
https://burohappold.sharepoint.com/:f:/s/BHoM/EoVbTyb6WHpEoomPOjc9JVUBwCGfK-_zrBfzlIQkNTyJKg?e=VRsqVQ

Boolean methods test:
https://burohappold.sharepoint.com/:f:/s/BHoM/EqCVTBONrQ9Bs-kJ1b22-2EB2Q_opuba0IkcEnYBE81G1Q?e=PAogpj

Any offset tests

Also, @FraserGreenroyd , can you check that this does not affect your projects scripts in any negative way.

### DIFF
--- a/Geometry_Engine/Query/Normal.cs
+++ b/Geometry_Engine/Query/Normal.cs
@@ -116,13 +116,14 @@ namespace BH.Engine.Geometry
             }
 
 
-            //Get out normal, by cross product between the average of points and 
+            //Get out normal, by cross product between the average of points and first points of the curve
             Point avg = curve.ControlPoints.Average();
             Point pA = curve.ControlPoints[0];
 
             foreach (Point pt in curve.ControlPoints.Skip(1))
             {
                 Vector normal = CrossProduct(avg - pA, avg - pt);
+                //If normal is non-zero (if the first points are not on a line with the average point) use this as the normal
                 if (normal.SquareLength() > tolerance * tolerance)
                 {
                     normal = normal.Normalise();
@@ -197,6 +198,7 @@ namespace BH.Engine.Geometry
                 foreach (Point pt in points.Skip(1))
                 {
                     Vector normal = CrossProduct(avg - pA, avg - pt);
+                    //If normal is non-zero (if the first points are not on a line with the average point) use this as the normal
                     if (normal.SquareLength() > tolerance * tolerance)
                     {
                         normal = normal.Normalise();
--- a/Geometry_Engine/Query/Normal.cs
+++ b/Geometry_Engine/Query/Normal.cs
@@ -117,24 +117,20 @@ namespace BH.Engine.Geometry
             }
 
 
-            //Get out normal, by cross product between the average of points and first points of the curve
             //Get out normal, from cross product of first points that are not colinear
-            Vector normal = Normal(curve.ControlPoints, tolerance);
+            Point avg = curve.ControlPoints.Average();
+            Vector normal = new Vector();
 
-            if (normal != null)
-            {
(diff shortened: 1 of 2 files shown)


FraserGreenroyd:
I have performed an initial test on the project which caused the issue and it seems to be working slightly more robustly at the moment. I want to perform some more checks, but will try and fit this in later tonight if that's ok. Thanks for jumping on the issue so quick @IsakNaslundBh ! :smile:

FYI: @michaelhoehn  @rboulton-BH in case you're able to do some testing from Environments side too?

Geometry_Engine/Query/Normal.cs
```diff
@@ -188,20 +183,12 @@ public static Vector Normal(this PolyCurve curve, double tolerance = Tolerance.D
...
-                int i = 0;
-                do
-                {
-                    Point pA = points[i];
-                    Point pB = points[(i + 1) % points.Count];
-                    Point pC = points[(i + 2) % points.Count];
-
-                    normal = CrossProduct(pB - pA, pC - pB);
-                    i++;
+                //Get out normal, from cross product of first points that are not colinear
```
kThorsager, line 186: these comments

Geometry_Engine/Query/Normal.cs
```diff
@@ -115,20 +116,13 @@ public static Vector Normal(this Polyline curve, double tolerance = Tolerance.Di
...
-            int i = 0;
-            do
-            {
-                Point pA = curve.ControlPoints[i];
-                Point pB = curve.ControlPoints[(i + 1) % curve.ControlPoints.Count];
-                Point pC = curve.ControlPoints[(i + 2) % curve.ControlPoints.Count];
-
-                normal = CrossProduct(pB - pA, pC - pB);
-                i++;
+            //Get out normal, from cross product of first points that are not colinear
```
kThorsager, line 120: these comments

Issue #952, opened 2019-05-10
Environment_Engine: Set Construction for Panel


michaldengusiak:
allow to set custom construction

PR #1912, opened 2020-07-29
Composite ShapeProfiles - supporting bidirectional offsets

### LINKED ISSUE #1909: Geometry_Engine: ShapeProfile support for multi-directional offsets
#### Description:

In connection with recent MEP_oM + MEP_Engine methods, we need to create/evaluate composite SectionProfile. For example, an MEP Duct's box SectionProfile could consist of three distinct entities; the element itself, an inner lining, and outer insulation. Each of these sections consists of different materials and/or thicknesses. 

This level of geometrical representation may not always be required in typical system planning-oriented workflows, but at least in LCA-oriented scopes, understanding the full material composition is always required. Therefore a single element SectionProfile will always be insufficient for some analysis efforts. 

I understand that this could potentially be useful for other disciplines such as structures for layered beam/column assemblies for example. 

Link to MEP_Engine: https://github.com/BHoM/BHoM_Engine/issues/1908

michaelhoehn:

Closes #1909 

Commented two error catches for BoxProfile and one for TubeProfile as these two are current test cases that will be immediately useful for MEP_oM. Removing these allows for multi-directional offsets, which I imaging could introduce conflicts to other toolkits + workflows. 

I'd like to simply start the discussion for how to best handles the concept of composite sections.

### Test files

Requires the following branches from both BHoM and BHoM Engine: ```MEP_oM-#952-Object-Creation```

https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/01_Test%20Scripts/BHoM/MEP_oM/200728_MEP_oM%20Testing.gh?csf=1&web=1&e=1rmXkx

### Additional comments

CompositeSectionProfiles

### DIFF
--- a/Geometry_Engine/Create/ShapeProfile.cs
+++ b/Geometry_Engine/Create/ShapeProfile.cs
@@ -109,17 +109,17 @@ namespace BH.Engine.Geometry
                 return null;
             }
 
-            if (Math.Sqrt(2) * thickness <= Math.Sqrt(2) * outerRadius - outerRadius - Math.Sqrt(2) * innerRadius + innerRadius)
-            {
-                InvalidRatioError("thickness", "outerRadius and innerRadius");
-                return null;
-            }
+            //if (Math.Sqrt(2) * thickness <= Math.Sqrt(2) * outerRadius - outerRadius - Math.Sqrt(2) * innerRadius + innerRadius)
+            //{
+            //    InvalidRatioError("thickness", "outerRadius and innerRadius");
+            //    return null;
+            //}
 
-            if (height <= 0 || width <= 0 || thickness <= 0 || outerRadius < 0 || innerRadius < 0)
-            {
-                Engine.Reflection.Compute.RecordError("Input length less or equal to 0");
-                return null;
-            }
+            //if (height <= 0 || width <= 0 || thickness <= 0 || outerRadius < 0 || innerRadius < 0)
+            //{
+            //    Engine.Reflection.Compute.RecordError("Input length less or equal to 0");
+            //    return null;
+            //}
 
             List<ICurve> curves = BoxProfileCurves(width, height, thickness, thickness, innerRadius, outerRadius);
             return new BoxProfile(height, width, thickness, outerRadius, innerRadius, curves);
@@ -542,11 +542,11 @@ namespace BH.Engine.Geometry
                 return null;
             }
 
-            if (diameter <= 0 || thickness <= 0)
-            {
-                Engine.Reflection.Compute.RecordError("Input length less or equal to 0");
-                return null;
-            }
+            //if (diameter <= 0 || thickness <= 0)
+            //{
+            //    Engine.Reflection.Compute.RecordError("Input length less or equal to 0");
+            //    return null;
+            //}
 
             List<ICurve> curves = TubeProfileCurves(diameter / 2, thickness);
             return new TubeProfile(diameter, thickness, curves);


IsakNaslundBh:
@michaelhoehn reading it as you want the inner diameter to become the outer and vice versa? I might be missing something, but I fear I fail to see what the benefit of this would be?

Regarding composite sections, we have started some discussions around them, but not gotten it to far as of yet. The general idea we have had in the discussions in structures though would be to have multiple profiles stored on the property, each representing some part/materiality. But, again, have not taken it so far as of yet..

michaelhoehn:
@IsakNaslundBh thanks for the input. Hoping to add a bit of clarity, apologies my description was pretty simplistic. I think we are actually talking about the same solution if I'm not mistaken. I've currently prototyped this on the latest MEP_oM and Engine PRs. This image should show the DuctSectionProperty hosting the SectionProfile. The SectionProfile instantiates the base ShapeProfile as both lining (offset inwards) and the insulation (offset outwards) and is controlled via the respective thickness parameters. 

image
image

In this way, each shape profile is capable of returning accurate solid/void areas. This method discrete ShapeProfiles is, of course, preferred to an offset of inner/outer edges (which is not what I'm proposing here). The functionality demonstrated in these images is currently not possible with the amount of error handling options within the ShapeProfile Create methods. 

FYI @kayleighhoude @FraserGreenroyd

al-fisher:
The shape profiles currently just store a thickness as a double. 

The convention to date has been to keep these geometrical profiles a simple as possible - with dimensional properties not taking negative values. This is possible as, for the tube as an example, you can define completely equivalent geometry by assuming the thickness is always outside of the Diameter reference 

PNG image 2020-08-04 08_34_05

My default would be to propose we do keep "Thickness as absolute values only" and therefore handle any necessary modification of the given Diameter (and other dimensions) to keep Thickness as positive in the Create methods.

Could have a composite Profile Create Method that takes a single `Diameter` and `List<Doubles>` and returns a list of Profiles.
This list of doubles _could_ allow negative values. And a compatible list of Diameters (dimensions) could then be calculated for creation of the Profiles

This approach, I think will help avoid confusion of the terms _Outer_ Radius and _Inner_ Radius on the Box say.

If we prefer to allow negative dimensions - I think this would require renaming of the properties to help clarity.

Either way - I think now might be a good time to add some useful Descriptions to the profiles to help guide usage. (I think it was me that skipped annotating descriptions on profiles as I was adding attributes throughout rest of Geometry oM 😳!)

@michaelhoehn @kayleighhoude @IsakNaslundBh does that makes sense?

IsakNaslundBh:
Thanks @michaelhoehn , that makes more sense now!

Would second @al-fisher s comment above 100% though, including all the corner cases such as the inner and outer radius for the boxes (pun intended).
IMO that is the best way to handle it, keeping all values on the profiles as positive (and for some allowing 0), but never negative. After all, all of the values are representing physical lengths/dimensions/distances and those should never be below 0. Simply letting engine methods such as the one in your screenshot do the maths for you, calculating the appropriate thicknesses and diameters I think is a better approach. For a general case, I would not expect this maths to be that complicated anyways.

> Either way - I think now might be a good time to add some useful Descriptions to the profiles to help guide usage. (I think it was me that skipped annotating descriptions on profiles as I was adding attributes throughout rest of Geometry oM 😳!)

Agree on this as well. Good to get this in asap. There are some descriptions on some create methods in the structures engine for some steel/concrete sections that potentially could be nicked if helpful!

al-fisher:
Now is probably a good time to add the Length QuantityAttribute too to the appropriate properties. Lengths can not be negative (at least for representing physical space).

This would distinguish "Thicknesses" etc. which are Lengths from "Offsets" which are Vectors (and can have "negative" directions)

michaelhoehn:
@al-fisher and @IsakNaslundBh thanks for the input! For clarity, I'll add a short summary of actionable items in order to close this draft PR and move forward. Feel free to tick them off or not, whatever works! 

- [ ] Useful descriptions on ShapeProfiles
- [ ] Length QuantityAttribute 
- [ ] Discussion for additional ShapeProfile create methods to expose the offset vector direction if desired
- [ ] Align maths in referenced PRs to work with Diameter and Thickness constraints MH

PR #1913, opened 2020-07-31
Geometry_Engine: Not Implemented Exception Review

### LINKED ISSUE #957: Geometry_Engine: Correct interface name in IsOnCurve
Interface of IsOnCurve is missing 'I' in the beginning

### LINKED ISSUE #1779: Geometry_Engine: Not Implemented Exception Review
#### Broken rules:
We are trying to keep away from trowing exceptions and instead add a error through reflection.

We're also using the `[NotImplemented]` attribute, which does hides the methods from the UI, but not from the code. i.e. If I'm coding another method I need to go to the methods definition to see if it works.

Currently it's thrown for some different reasons:
- Part of a method which does not work with it.
- The internals of any method with the attribute `[NotImplemented]`
- Some methods which do not seem to have solutions, `ToPolyline(Arc)`

We also have methods like `DiscontinuityPoints(PolyCurve)` which has the attribute but looks like it should work? And is presumably called from the interface all the time.

- Use private fallback methods opposed to keeping a to do list in the code as not implemented methods.
- Return `null` or `NaN` and log a error in Reflection if no solution can be found.
- Don't know how to handle methods who return booleans as there, the safest solution is to throw exception as it can be difficult to determine what the safe option is. So would opt to leave those as is unless a good solution to that could be found.

FYI @pawelbaran @epignatelli @al-fisher

LMarkowski:

Closes #1779 
Closes #957

EDIT
All methods now have fallback methods that  record errors and return nulls/NaNs. The exception are methods returning booleans and some other cases listed in additional comments.
/EDIT

As said in [the original issue](https://github.com/BHoM/BHoM_Engine/issues/1779) there are many methods in `Geometry_Engine` with `[NotImplemented]` attribute and throwing exceptions. Ideally all exceptions should be replaced with errors and methods returned nulls. I think as a lot of toolkit leaders are now on holidays it could be to much of a change (some errors are needed to be catch downstream).
To stay on the safe side and yet push the issue a step forward I've replaced most of non-implemented methods with private fallback methods catching non-implemented inputs which go through the interfaces. Although they are still throwing `NotImplementedException`s. If we agree in this PR that they can be replaced with errors and nulls I'll be happy to do that before merging.

### Test files

No test file needed.

### Changelog

- Replaced most of `[NonImplemented]` cases of methods in `GeometryEngine` with one private fallback method per file.

### Additional comments

Some of `[NotImplemented]` methods couldn't be handled for a few different reasons. Here's a list:
#### <s> Already having a fallback method logging error and returning null, yet still having some [NotImplemented] cases left in the code
- Area
- EndPoint
- Project
- ProjectAlong
- StartPoint</s>
#### <s>[NotImplemented] with 0 references
- BasisFunction
- DerivativeFunction</s>
#### <s>Not sure if the issue is applicable:
- ShapeProfile
- ShapeProfileCurves</s>
#### All or almost all cases are not implemented (will be resolved in a follow-up issue)
- Create.NurbsCurve
- Create.NurbsSurface
#### [NotImplemented] and doesn't match the interface or general method's scheme
- MeshIntersections
- PlaneIntersections


LMarkowski:
There are cases of `MeshIntersections` and `PlaneIntersections` that are marked as `[NotImplemented]` and doesn't match the interface or general method's scheme. @pawelbaran and @epignatelli are you okay with deleting them and raising an issue to implement them or should they be left in code?

epignatelli:
Hi @LMarkowski Is this ready for review already?

LMarkowski:
@epignatelli Yes 😃

Geometry_Engine/Compute/CurveProximity.cs
```diff
@@ -1034,22 +1034,6 @@ public static partial class Compute
             return result;
         }
 
-        /***************************************************/
-
-        [NotImplemented]
-        public static Output<Point, Point> CurveProximity(this ICurve curve1, Ellipse curve2, double tolerance = Tolerance.Distance)
```
epignatelli, line 1040: Are we removing this in favour of an interface method or just removing it completely?

epignatelli: Sorry - just found it - agreed.

LMarkowski:
@epignatelli Thanks for the review!
Good point about nulls, I didn't think of that.
I think as it won't be an downgrade to the master, it might be a bit out of the scope. But definitely worth getting covered as the next step. Do we actually have any standard of dealing with null inputs?

LMarkowski:
Done test and even if fallback method would take `object` we get an ambiguous call. Nulls won't pass through the interface method.
I guess next best solution would be null-checking at the level of interface, but I would opt for doing it in separate PR.
@al-fisher @adecler @epignatelli @IsakNaslundBh

al-fisher:
> Done test and even if fallback method would take object we get an ambiguous call. Nulls won't pass through the interface method.
I guess next best solution would be null-checking at the level of interface, but I would opt for doing it in separate PR.

Shame was a neat solution.
Yes - agreed. Further options out of scope of this PR I would say too. Thanks @LMarkowski 

Worth raising an issue though - and referencing back to the comments above for picking up and prioritising later and so knowledge is capture. @LMarkowski do you mind doing that? cheers

Geometry_Engine/Query/EndPoint.cs
```diff
@@ -68,7 +68,10 @@ public static Point EndPoint(this NurbsCurve curve)
             if (!curve.IsPeriodic())
                 return curve.ControlPoints.LastOrDefault();
             else
-                throw new NotImplementedException();
+            {
+                Reflection.Compute.RecordError("EndPoint is not implemented for non-periodic NurbsCurves");
```
IsakNaslundBh, line 72:
```suggestion
                Reflection.Compute.RecordError("EndPoint is not implemented for periodic NurbsCurves");
```

Geometry_Engine/Query/IsClockwise.cs
```diff
@@ -85,7 +85,7 @@ public static bool IsClockwise(this PolyCurve curve, Vector normal, double toler
                     cPts.Add((c as Circle).EndPoint());
                 }
                 else
-                    throw new NotImplementedException();
+                    throw new NotImplementedException("PolyCurve connsisting of type: " + c.GetType().Name + " is not implemented for IsClockwise.");
```
IsakNaslundBh, line 88:
```suggestion
                    throw new NotImplementedException("PolyCurve consisting of type: " + c.GetType().Name + " is not implemented for IsClockwise.");
```

Geometry_Engine/Query/StartPoint.cs
```diff
@@ -63,14 +63,17 @@ public static Point StartPoint(this Line line)
...
+
         public static Point StartPoint(this NurbsCurve curve)
         {
             //TODO: This should be based on the basis function?
             if (!curve.IsPeriodic())
                 return curve.ControlPoints.FirstOrDefault();
             else
-                throw new NotImplementedException();
+            {
+                Reflection.Compute.RecordError("StartPoint is not implemented for non-periodic NurbsCurves");
```
IsakNaslundBh, line 74:
```suggestion
                Reflection.Compute.RecordError("StartPoint is not implemented for periodic NurbsCurves");
```

IsakNaslundBh (changes requested):
'connsisting' should be 'consisting'. Might be quicker with a find-replace than me adding request changes  :)

Apart from that and the other review changes think this is looking good!

IsakNaslundBh:
Accidentally closed this when merging another PR. Re-opening.

Geometry_Engine/Query/DiscontinuityPoints.cs
```diff
@@ -55,15 +55,6 @@ public static List<Point> DiscontinuityPoints(this Line curve)
...
 
-        [NotImplemented]
-        public static List<Point> DiscontinuityPoints(this NurbsCurve curve)
-        {
-            throw new NotImplementedException();
-        }
-
-        /***************************************************/
-
-        [NotImplemented]
```
IsakNaslundBh, line 66: Not sure I would remove this tbh. DO not think this method is up to scratch. Tangents needs to be checked between the start/end of a segment:

image

top is as PolyCurve, bottom is as Polyline.

Would rather return an exception here.

LMarkowski: Great point!
I will keep the attribute for now for the sake of PR's clarity and fix that one together with correcting inputs for this and other `DiscontinuityPoints` methods right after this PR got merged.

LMarkowski:
Opened two minor follow-up issues which are to be resolved right after this one will get merged.
#1954 - DiscontinuityPoints(PolyCurve)
#1955 - Create NurbsCurve and NurbsSurface
They could be added to this PR but I think it'd be better to separate them to not prolong this one too much.
From my end I hope it is ready for final reviewing and merging.

Geometry_Engine/Query/ControlPoints.cs
```diff
@@ -72,14 +72,6 @@ public static List<Point> ControlPoints(this Line curve)
 
         /***************************************************/
 
-        [NotImplemented]
```
IsakNaslundBh, line 75: Wonder if we should just remove the ` [NotImplemented]` here rather than wipe the method? I mean, it is correct?

Is if we want controlpoints in a query sense to always be on the curve, but, for nurbs I am not sure if that makes sense?

Geometry_Engine/Query/IsEqual.cs
```diff
@@ -117,14 +117,6 @@ public static bool IsEqual(this Line line, Line other, double tolerance = Tolera
 
         /***************************************************/
 
-        [NotImplemented]
-        public static bool IsEqual(this NurbsCurve curve, NurbsCurve other, double tolerance = Tolerance.Distance)
```
IsakNaslundBh, line 121: anopther one I think we just should implement rather than remove, but might be out of scope for this PR

IsakNaslundBh (commented):
Generally looking good. Would keep the controlpoints method for Nurbs though, as it actually looks correct to me? @pawelbaran @al-fisher @FraserGreenroyd @epignatelli ?

Geometry_Engine/Query/ControlPoints.cs
LMarkowski: Whoops. I must've thought it just throws an error, otherwise I wouldn't wipe that. I probably rushed too much on that one 🙈 
Having another look, this whole file has a lot of TODOs and general uncertainty. For now I would remove `[NotImplemented]` and left the nurbs method as it is, as it is technically correct. And maybe raise an issue to sort the whole method out.

Geometry_Engine/Query/IsEqual.cs
LMarkowski: Sure. I think I can implement that one together with #1955

Geometry_Engine/Query/ControlPoints.cs
IsakNaslundBh: as agreed in the call, lets just keep

Geometry_Engine/Query/IsEqual.cs
LMarkowski: Given another thought i decided to raise another issue for that (#1958) and resolve it after merging to keep code more consistent about not making `[NotImplemented]` methods a "todo list".

Geometry_Engine/Query/ControlPoints.cs
al-fisher: Yep - as per our Description in the oM - this is correct. So all looks good removing `[NotImplemented]`

IsakNaslundBh (approved):
Now Happy with issues raised to sort out concerns!

PR #1934, opened 2020-08-12
Environment_Engine: Add ability to convert Brep and Surface to Environment Panels

### LINKED ISSUE #1929: Environment_Engine: BRep to Panels method
#### Description:

We have an extrude to volume and convert to Panels method, but it would be nice to have a method of converting directly from a BRep (making sure it's a closed polysurface kind-of Brep in the first place). This should shorten a load of workflows for creating geometry in one tool to pass to another for testing simple concepts.

FraserGreenroyd:
Fixes #1929 

### Test files
[BrepToPanel.zip](https://github.com/BHoM/BHoM_Engine/files/5063228/BrepToPanel.zip)


Environment_Engine/Create/Panel.cs
```diff
@@ -40,6 +43,52 @@ public static partial class Create
...
+
+            return panels;
+        }
+
+        [Description("Create an Environments Panel from a BHoM Surface")]
+        [Input("surface", "A BHoM surface to convert into an Environment Panel")]
+        [Input("connectedSpaceName", "A name for the space which this panel is connected to. If no name is provided, a randomised default will be generated")]
+        [Input("angleTolerance", "The angle tolerance for collapsing to polylines used when generating the external edges of the surfaces")]
+        [Output("panel", "An Environment Panels representing a closed space generated from the provided Brep geometry")]
+        public static Panel Panel(this ISurface surface, string connectedSpaceName = null, double angleTolerance = BH.oM.Geometry.Tolerance.Angle)
```
IsakNaslundBh, line 80: One more comment here, is that for at least PlanarSurfaces, good if the openings of it are turned into openings of the panel as well? That is at least what we have been doing for structures (we only have creates for planar surface, not for a generic surface)

FraserGreenroyd: Cheers for the suggestion @IsakNaslundBh , have implemented using `IInternalEdges` from the surface.

vbeardow (approved):
LGTM 👍

PR #1936, opened 2020-08-13
Environment engine #1935 fibonacci sphere

### LINKED ISSUE #1935: Environment_Engine: Add fibonacci sphere method
#### Description:

Generation of a fibonacci sphere equidistantly placing points on a sphere. Useful for sampling contextual geometry and basic ray tracing.

tg359:

Closes #1935 

Adds a method for calculating sample vectors equidistantly spaced around an origin point for context sampling and ray tracing.

### Test files

[unnamed (2).zip](https://github.com/BHoM/BHoM_Engine/files/5068811/unnamed.2.zip)

### Changelog

- `Compute.FibonacciSphere()` method added to the `Environment_Engine`

### DIFF
--- a/Environment_Engine/Compute/FibonacciSphere.cs
+++ b/Environment_Engine/Compute/FibonacciSphere.cs
@@ -1,0 +1,75 @@
+﻿/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2020, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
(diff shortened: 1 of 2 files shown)


BingWangUS:
Just curious, is there a reason why we are not including FibonacciSphereCartesian.cs?

tg359:
> Just curious, is there a reason why we are not including FibonacciSphereCartesian.cs?

The original `FibonacciSphereCartesian` method referenced a `FibonacciSphereSpherical` method and converted the spherical coordinates into cartesian ones. This method effectively combines the two into one with a single `FibonacciSphere` name that outputs both cartesian vectors (to the surface points on the sphere - which can be interpreted as either `Vector` or `Point`) and the spherical coordinates.

This mainly came about to save on creating different files for different methods, and the fact that we generally always want vectors, but may only sometimes want spherical coordinates, so it made sense to just output both in one method and simplify the workflow for generating rays.

tg359:
Now captured within #1939.

PR #1938, opened 2020-08-18
Geometry_Engine: Add MarchingSquares & MarchingCubes

theThorsager:

Closes #

Implements the MarchingSquares and MarchingCubes for meshes and mesh3d's respectively.

In essence it allows one to extract a constant value range from a linearly approximated scalar field.
This implementation is general in the sense that it does not require neither squares nor cubes but is general to the degree that the BHoM mesh allows. (only 3 to 4 vertices per face)
And it is also not bound to the world axis but placed in the local space provided by the meshes.

Example: the scalar field = distance from a point.
a1f3270b-019f-4cad-8af5-ea2918d85b7b

Anyway could probably also be useful for result display.

### Test files

[MarchingSquares.zip](https://github.com/BHoM/BHoM_Engine/files/5091004/MarchingSquares.zip)

[MarchingCubes.zip](https://github.com/BHoM/BHoM_Engine/files/5091000/MarchingCubes.zip)

 ### Additional comments

### DIFF
--- a/Geometry_Engine/Compute/MarchingSquares.cs
+++ b/Geometry_Engine/Compute/MarchingSquares.cs
@@ -1,0 +1,155 @@
+/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2020, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 
+ * GNU Lesser General Public License for more details.                          
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using BH.Engine.Data;
+using BH.oM.Data.Collections;
+using BH.oM.Geometry;
+using BH.oM.Reflection.Attributes;
+using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Linq;
+
+namespace BH.Engine.Geometry
+{
+    public static partial class Compute
+    {
+        /***************************************************/
+        /****   Public Methods                          ****/
+        /***************************************************/
+
+        [Description("Computes contour lines of the iso-values in the scalar field defined by the mesh and the values. \n" +
+                     "The values are assumed to change linearly between vertices.")]
+        [Input("mesh3d", "Mesh which defines the positions of the values in the scalar field.")]
+        [Input("vertexValues", "The value for of scalar field for each vertex in the mesh.")]
+        [Input("isoValues", "Values in the scalar field to produce the iso-lines for.")]
+        [Output("isoLines", "Lines covering the iso-values in the scalar field defined by the mesh and values.")]
+        public static List<List<Line>> MarchingSquares(this Mesh mesh, List<double> vertexValues, List<double> isoValues)
+        {
+            if (mesh?.Vertices == null || vertexValues == null || mesh.Vertices.Count != vertexValues.Count)
+            {
+                Reflection.Compute.RecordError("Number of vertexValues must match the number of vertices in the mesh.");
+                return new List<List<Line>>();
+            }
+            if (isoValues == null)
+                return new List<List<Line>>();
+
+            List<List<Line>> results = new List<List<Line>>();
+            for (int i = 0; i < isoValues.Count; i++)
+                results.Add(new List<Line>());
+
+            // Look at every face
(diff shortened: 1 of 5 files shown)


IsakNaslundBh (commented):
First: Great to see some more stuff added by you @theThorsager !!! (also love the account name!) 🚀 

Have tested through the functionality a bit, focusing firstly on the marching squares, and was able to produce this with some GH fiddeling:

image

which is pretty awsome!

Also checked the numbers on that more closely in your testscript and it seem to be working fine, with high precision. 

Also, quickly tested the MarchingCube testscript you added. It generally seem to be working really quite well, but I found some discrepancies there between the values of the vertecies in the produced meshes and the distances:

image

Would assume those values to all be bang on 3.7? Unless I am missing some quirk here. When testing the same for marching squares all values came out exact at least.

Anyhow, again, functionalitywise it generally is looking great, and fantastic to see you committing! 😄

Geometry_Engine/Compute/SubMesh.cs
```diff
@@ -0,0 +1,108 @@
...
+
+        /***************************************************/
+
+        [Description("Create a mesh from selected faces of an existing mesh.")]
+        [Input("mesh", "Mesh to get a sub-section from.")]
+        [Input("faces", "The faces of the old mesh to carry over to the new mesh.")]
+        [Input("vertexRelatedData", "A list where each item is related to the vertex at the same index in the mesh. \n" +
+                                    "Will be changed to relate to the new mesh's vertex list.")]
+        [Output("mesh", "Mesh composed of the faces and the vertices in those faces.")]
+        public static Mesh SubMesh<T>(this Mesh mesh, List<Face> faces, ref List<T> vertexRelatedData)
```
pawelbaran, line 56: Wouldn't it be possible to avoid `ref` argument here? Besides making @FraserGreenroyd unhappy, I would expect it being potentially harmful when accessed from the UI level.

I would either try to split the method into two if possible without massive overhead, or, alternatively, make it `private`.

Geometry_Engine/Compute/SubMesh.cs
```diff
@@ -0,0 +1,108 @@
...
+        {
+            List<Point> vertices = new List<Point>();
+            List<Face> resultFaces = new List<Face>();
+
+            List<int> indecies = faces.SelectMany(x => x.ToArray()).Distinct().ToList();
+            List<Tuple<int, int>> mapping = indecies.Select((x, i) => new Tuple<int, int>(x, i)).ToList();
+
+            Dictionary<int, int> map = mapping.ToDictionary(x => x.Item1, x => x.Item2);
```
pawelbaran, line 64: You could either wrap these 2 lines into one or change the variable type in the upper one to `IEnumerable<Tuple<int, int>>` + remove the call to `ToList()`.

Geometry_Engine/Compute/SubMesh.cs
```diff
@@ -0,0 +1,108 @@
...
+            if (vertexRelatedData != null && 
+                vertexRelatedData.Count == mesh.Vertices.Count)
+            {
+                List<T> newList = new List<T>();
+                foreach (int i in indecies)
+                {
+                    newList.Add(vertexRelatedData[i]);
+                }
+                vertexRelatedData = newList;
+            }
```
pawelbaran, line 80: That could be wrapped in a single line using `Select` Linq query (don't be afraid of the deferred execution 🤣)

pawelbaran (changes requested):
Good to see you back, sir @theThorsager! 🥳 

I went through the code and left a few comments, most of them purely cosmetical. The only bigger issue I have code wise is `ref` method argument that I always considered an _argumenta non grata_.

Geometry_Engine/Compute/MarchingSquares.cs
```diff
@@ -0,0 +1,155 @@
...
+                            continue;
+
+                        // As a Faces edges are a closed loop, there will always be an even number of results.
+                        if (prePoint[k] == null)
+                            prePoint[k] = pt;
+                        else
+                        {
+                            results[k].Add(new Line()
+                            {
+                                Start = prePoint[k].Clone(),
```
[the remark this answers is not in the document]
theThorsager: Leftover from aggressive debugging, was not needed

Geometry_Engine/Compute/SubMesh.cs
theThorsager: You're right, it became less strange than I imagined

theThorsager:
> (also love the account name!) 🚀

It was the most time consuming part 😜 

> Would assume those values to all be bang on 3.7? Unless I am missing some quirk here. When testing the same for marching squares all values came out exact at least.

We're only getting a few exact values for the scalar field (the vertices of the original mesh), all other values are approximated to be changing linearly between connected points on/in the mesh. Hence the MarchingSquare testfile had accurate results as the scalar field was linear (= Z), but the MarchingCubes one is an approximation, as the distance to an point does not vary linearly between two arbitrary points.

So yes it's to be expected

And think I've dealt with your points as well @pawelbaran

pawelbaran (approved):
LGTM by the looks of code 👍 but would be nice to have @IsakNaslundBh's confirmation that we are happy about the functionality before it gets merged.

IsakNaslundBh:
> We're only getting a few exact values for the scalar field (the vertices of the original mesh), all other values are approximated to be changing linearly between connected points on/in the mesh. Hence the MarchingSquare testfile had accurate results as the scalar field was linear (= Z), but the MarchingCubes one is an approximation, as the distance to an point does not vary linearly between two arbitrary points.
>
> So yes it's to be expected

Ah, yeah, ofc. Makes full sense. Will test through the PR again after the changes, if still working as before, happy with the functionality!

IsakNaslundBh (approved):
Happy with the functionality. Nice one @theThorsager

PR #1939, opened 2020-08-19
Environment_Engine: Refactoring, tidying, and aligning

FraserGreenroyd:
### NOTE: Depends on 
https://github.com/BHoM/BHoM/pull/971


Environment_Engine/Query/PointsMatch.cs
```diff
@@ -51,7 +51,8 @@ public static bool PointsMatch(this List<Point> controlPoints, List<Point> measu
             foreach (Point p in controlPoints)
             {
                 Point ptInMeasure = measurePoints.Where(x => x.X == p.X && x.Y == p.Y && x.Z == p.Z).FirstOrDefault();
```
IsakNaslundBh, line 53: not related tot he PR, but spotted scrolling through. Could be safer doing this with a distance and tolerance to avoid equal for doubles? Maybe with a square distance even.
```
                Point ptInMeasure = measurePoints.Where(x => p.SquareDistance(x) < squareTolerance).FirstOrDefault();
```

where the tolerance is input/defined above to whatever is suitable. (Please ignore if to off-topic for this PR)

FraserGreenroyd: Thanks for the spot @IsakNaslundBh - have modified accordingly! :smile:

PR #1951, opened 2020-08-25
Analytical_Engine: Added methods for IsOutlineRectangular, IsOutlineSquare and IsOutlinequad to query Panels

### LINKED ISSUE #1946: Analytical_Engine: Method for IsSquare Panel check
#### Description:

Check if a panel is rectangular, i.e. its external edges forms a perfect square, within tolerance. Should also not contain any openings.

### LINKED ISSUE #1947: Analytical_Engine: Method for IsRectangular Panel
#### Description:

Check if a panel is rectangular, i.e. its external edges forms a perfect rectangle. Should also not contain any openings.

### LINKED ISSUE #1948: Analytical_Engine: Method for IsQuad Panel
#### Description:

Check if a panel is quad, i.e. its external edges forms 4 straight lines and the panel does not have any openings.

peterjamesnugent:

Closes #1946 
Closes #1947 
Closes #1948 

### Test files

https://burohappold.sharepoint.com/:f:/s/BHoM/Es8zCkf7as5FobVL8WpGvYoBBTKIA5iks6EtVzB7VIcJUw?e=lz9bMS


Analytical_Engine/Analytical_Engine.csproj
```diff
@@ -50,6 +50,10 @@
       <HintPath>C:\ProgramData\BHoM\Assemblies\Reflection_oM.dll</HintPath>
       <Private>False</Private>
     </Reference>
+    <Reference Include="Spatial_Engine">
+      <HintPath>C:\ProgramData\BHoM\Assemblies\Spatial_Engine.dll</HintPath>
+	  <Private>False</Private>
+    </Reference>
```
FraserGreenroyd, line 56: This should be a project reference.

Analytical_Engine/Analytical_Engine.csproj
```diff
@@ -89,6 +94,9 @@
       <Name>Reflection_Engine</Name>
     </ProjectReference>
   </ItemGroup>
+  <ItemGroup>
+    <Compile Include="Query\IsSquare.cs" />
+  </ItemGroup>
```
FraserGreenroyd, line 99: Why is this separate from the other `Compile Include`s?

Analytical_Engine/Query/IsRectangle.cs
```diff
@@ -0,0 +1,113 @@
...
+        [Description("Determines whether a panel is a rectangle")]
+        [Input("panel", "The Panel to check if it is a rectangle")]
+        [Output("bool", "True for rectangular panels or false for non-rectangular panels")]
+        public static bool IsRectangle<TEdge, TOpening>(this IPanel<TEdge, TOpening> panel)
+            where TEdge : IEdge
+            where TOpening : IOpening<TEdge>
+        {
+            //Check if there are any openings
+            if (panel.Openings.Count != 0)
+                return false;
```
FraserGreenroyd, line 52: So if a panel has openings, it is immediately not rectangular?

Analytical_Engine/Query/IsRectangle.cs
```diff
@@ -0,0 +1,113 @@
...
+            var groupedCurves = curves.GroupBy(x => x.IStartDir());
+            foreach (var groupedCurve in groupedCurves)
+            {
+                ICurve jointCurve = Engine.Geometry.Compute.IJoin(groupedCurve.Select(x => x).ToList()).First();
+                points.Add(jointCurve.IStartPoint());
+            }
+
+            //Check there are four discontinuity points present (start and end same point)
+            if (points.Count != 4)
+                return false;
```
FraserGreenroyd, line 79: If there are 4 points, and the first and last point at the same, is that not now a triangle rather than a rectangle?

Analytical_Engine/Query/IsRectangle.cs
```diff
@@ -0,0 +1,113 @@
...
+            }
+
+            //Check the three angles are pi/2 degrees within tolerance
+            if (angles.Any(x => Math.Abs(Math.PI / 2 - x) > Tolerance.Angle))
+                return false;
+
+            //Check if all sides are the same length
+            double length = vectors.First().Length();
+            if (vectors.Skip(0).All(x => (Math.Abs(x.Length() - length) < Tolerance.Distance)))
+                return false;
```
FraserGreenroyd, line 103: If all the sides are the same length surely that's now a square not a rectangle?

Analytical_Engine/Query/IsSquare.cs
```diff
@@ -0,0 +1,110 @@
...
+        [Description("Determines whether a panel is a square")]
+        [Input("panel", "The IPanel to check if it is a square")]
+        [Output("bool", "True for square panels or false for non-square panels")]
+        public static bool IsSquare<TEdge, TOpening>(this IPanel<TEdge, TOpening> panel)
+            where TEdge : IEdge
+            where TOpening : IOpening<TEdge>
+        {
+            //Check if there are any openings
+            if (panel.Openings.Count != 0)
+                return false;
```
FraserGreenroyd, line 53: So if a panel has any openings it suddenly can't be square?

Analytical_Engine/Query/IsSquare.cs
```diff
@@ -0,0 +1,110 @@
...
+            var groupedCurves = curves.GroupBy(x => x.IStartDir());
+            foreach (var groupedCurve in groupedCurves)
+            {
+                ICurve jointCurve = Engine.Geometry.Compute.IJoin(groupedCurve.Select(x => x).ToList()).First();
+                points.Add(jointCurve.IStartPoint());
+            }
+
+            //Check there are four discontinuity points present (start and end same point)
+            if (points.Count != 4)
+                return false;
```
FraserGreenroyd, line 80: See previous comment

FraserGreenroyd (dismissed):

Analytical_Engine/Query/IsRectangle.cs
IsakNaslundBh: Think this will be checked below, this is just a first step checking if it is worth to continue.

Analytical_Engine/Query/IsRectangle.cs
```diff
@@ -0,0 +1,113 @@
...
+
+            //Get the angles in the panel, only three are needed
+            List<double> angles = new List<double>() { vectors[3].Angle(vectors[0]) };
+            for (int i = 0; i < 3; i++)
+            {
+                angles.Add(vectors[i].Angle(vectors[i + 1]));
+            }
+
+            //Check the three angles are pi/2 degrees within tolerance
+            if (angles.Any(x => Math.Abs(Math.PI / 2 - x) > Tolerance.Angle))
```
IsakNaslundBh, line 97: Actually think this test is enough. As long as the curves are planar (which they should be) and all angles are 90 degrees, I this is is a rectangle (or a square), and I would say that a square should return true for this as well, being a special case of a rectangle.

Analytical_Engine/Query/IsRectangle.cs
FraserGreenroyd: But we return false if the points are not equal to 4, when something being equal to 4 is a triangle, so what would be the point of the check below? Maybe I'm missing something?

Analytical_Engine/Query/IsRectangle.cs
IsakNaslundBh: Could see arguments both for and against this. Say for example this panel (which we deem valid in BHoM):

image

Should probably not be seen as a rectangle, but then, maybe this one should:

image


Might be worth adding a boolean argument, something like "acceptOpenings" or something like that?

Analytical_Engine/Query/IsRectangle.cs
IsakNaslundBh: If you have 3 or less points, you can be sure it is not a rectangle. If it is 5 or more points, you can be sure it is not a rectangle. -> return false.

If you have 4 points, it might be a rectangle, do more check to determine if this is the case.

That is at least the logic I see behind this.

Analytical_Engine/Query/IsRectangle.cs
FraserGreenroyd: In that first picture, the host panel is still rectangular, that looks to be the green line to me? The opening having an edge coincident with the panel doesn't change the shape of the host panel. I imagine this is because in structures you wouldn't care if that was a window or not, to you that is a hole and is affecting the shape of the panel from a structural loading basis? Whereas in Environments, the existence of that window makes no difference to the host panel and thus the outline is still square.

IsakNaslundBh: yeah, outline the same for both panels here.

And, yes, in structures an opening is basically stating "there is nothing here", same as not having a panel there at all, so would want to treat the top image as a non-rectangle. An U-shaped panel with only external edges would be equivalent for structural analysis/design.

But maybe the suggested boolean argument then could be a nice compromise? Also some potential to add an extension method to our opening classes, like `IsVoid()` or similar that could determine how the openings should be treated? Maybe both of these?

Analytical_Engine/Query/IsRectangle.cs
IsakNaslundBh: And now I am reading what you are reading for the comment above the if-statement. Agree, if the comment is right, the if-statement should say `!= 5`. If not, the comment should be fixed.

Analytical_Engine/Query/IsRectangle.cs
FraserGreenroyd: But this is the Analytical Engine, not the Structural Engine. Structural Analysis is different from generic Analysis, so should a panel with an opening that could cause a U-Shaped panel be non-rectangular in the Analytical_Engine? I don't think it should, I feel that is an edge case which should be handled by an override in the Structural_Engine because, on the face of it, the panel itself is square. The openings have no bearing on that (in those examples)?

A boolean argument could resolve it, but I would still be wanting to question (and discuss or maybe reach agreement on) how that compromise doesn't compromise the Analytical Engine itself for future issues (with other disciplines maybe choosing different ways to categorise it)?

For example, another way round could be for Analytical Engine to say the first panel is rectangular, which Structural_Engine calls to determine, but then has an additional check. So Structures Engine can use this method to quickly rule out definite non-square panels, but then has additional checks for structural squareness?

IsakNaslundBh: what You wrote, or, two separate methods?

`HasSquareOutline()` and `IsSolidSquare` or something like that? Adding it to Structure_Engine straight away kind of defeats the purpose of putting it in the Analytical_Engine in the first place IMO. Much more prefer a solution with a compromise in terms of arguments/method name.

Would say that simply adding an argument regarding if the geometry of all of the pieces of the panel should be accounted for or not to determine the squareness of a panel does not compromise the Analytical_Engine itself?

FraserGreenroyd: It's more a case of how many future cases might we have to consider for other disciplines which leads me to want to look at the logic of squareness without the bits in the middle.

I would support `HasSquareOutline()` and `IsSolidSquare` - though the second might need a boolean argument to offer flexibility on the openings but then we're back to that argument again, which I don't think is needed. I would agree with this suggestion as is as a better alternative than the current proposal :smile:

peterjamesnugent: Ok I will make the change for `IsOutlineSquare()` and `IsOutlineRectangular` - perhaps we have a separate method for `ContainsOpenings` in the Analytical_Engine?

Analytical_Engine/Query/IsRectangle.cs
peterjamesnugent: I willl ammend the comment.

It would be `!=5` if we used `.IDiscontinuityPoints()` to get the points, but for a `PolyCurve` this returns intermediate points that exist on the same line (see test script where square/rectangle is constructed from >4 lines).

So I've had to group the curves by direction, joined them and taken the start point of each edge resulting in four points being returned.

Analytical_Engine/Query/IsRectangle.cs
peterjamesnugent: It returns `false` if all the sides are the same length, because it's not a rectangle. 

This check is needed because the check below (checking opposing sides are equal length) is still true for a square, so needed an additional check to filter out any squares.

Analytical_Engine/Query/IsRectangle.cs
peterjamesnugent: Yeah agree on square being a special rectangle.

Analytical_Engine/Query/IsOutlineRectangular.cs
```diff
@@ -0,0 +1,103 @@
...
+
+            PolyCurve polycurve = polycurves.First();
+
+            //Check that all subparts of the curve are linear
+            if (polycurve.SubParts().Any(x => !x.IIsLinear()))
+                return false;
+
+            //Group curves by direction vector to obtain discontinuity points
+            List<Point> points = new List<Point>();
+            var groupedCurves = curves.GroupBy(x => x.IStartDir());
```
IsakNaslundBh, line 65: This fails if two parallel edges has the same direction. The Join command should be flipping the edges accordingly, so would use the subcurves of that instead of the original curves.

```suggestion
            var groupedCurves = polycurve.SubParts().GroupBy(x => x.IStartDir());
```

Analytical_Engine/Query/IsOutlineSquare.cs
```diff
@@ -0,0 +1,105 @@
...
+
+            PolyCurve polycurve = polycurves.First();
+
+            //Check that all subparts of the curve are linear
+            if (polycurve.SubParts().Any(x => !x.IIsLinear()))
+                return false;
+
+            //Group curves by direction vector to obtain discontinuity points
+            List<Point> points = new List<Point>();
+            var groupedCurves = curves.GroupBy(x => x.IStartDir());
```
IsakNaslundBh, line 66: Reason stated above

```suggestion
            var groupedCurves = polycurve.SubParts().GroupBy(x => x.IStartDir());
```

IsakNaslundBh (changes requested):
Some issues with panels that have been created with manually generated edges. Have added some examples to the script.

IsakNaslundBh:
@peterjamesnugent wonder if https://github.com/BHoM/BHoM_Engine/pull/1960 could help simplify this method a bit. With properly implemented discontinuity points, I think could actually use them. Requre that PR to get merged first ofc, but will ahve a look at that asap + already approved by pawel.

Analytical_Engine/Query/IsOutlineSquare.cs
```diff
@@ -0,0 +1,73 @@
...
+        [Input("panel", "The IPanel to check if the outline is a square")]
+        [Output("bool", "True for panels with a square outline or false for panels with a non square outline")]
+        public static bool IsOutlineSquare<TEdge, TOpening>(this IPanel<TEdge, TOpening> panel)
+            where TEdge : IEdge
+            where TOpening : IOpening<TEdge>
+        {
+            bool isOutlineSquare = true;
+            PolyCurve polycurve = GetPolycurve(panel, out isOutlineSquare);
+            if (!isOutlineSquare)
+                return isOutlineSquare;
```
IsakNaslundBh, line 53: Would add check here that all segements of the curve are linear

IsakNaslundBh (changes requested):
A few comments below!

Analytical_Engine/Query/IsOutlineRectangular.cs
```diff
@@ -0,0 +1,71 @@
...
+        [Description("Determines whether a panel's outline is a rectangular")]
+        [Input("panel", "The IPanel to check if the outline is a rectangular")]
+        [Output("bool", "True for panels with a rectangular outline or false for panels with a non rectangular outline")]
+        public static bool IsOutlineRectangular<TEdge, TOpening>(this IPanel<TEdge, TOpening> panel)
+            where TEdge : IEdge
+            where TOpening : IOpening<TEdge>
+        {
+            bool isOutlineRectanglular = true;
+            PolyCurve polycurve = GetPolycurve(panel, out isOutlineRectanglular);
+            if (!isOutlineRectanglular)
```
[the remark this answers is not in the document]
peterjamesnugent: This is done within the `GetPolycurve` method which is now `ExternalPolyCurve`

Analytical_Engine/Query/IsOutlineSquare.cs
peterjamesnugent: This is done within the `GetPolycurve` method which is now `ExternalPolyCurve`

Analytical_Engine/Query/IsOutlineQuad.cs
```diff
@@ -0,0 +1,68 @@
...
+        /***************************************************/
+
+        [Description("Determines whether a panel's outline is a quadilateral.")]
+        [Input("panel", "The IPanel to check if the outline is a quadilateral.")]
+        [Output("bool", "True for panels with a quadilateral outline or false for panels with a non-quadilateral outline.")]
+        public static bool IsOutlineQuad<TEdge, TOpening>(this IPanel<TEdge, TOpening> panel)
+            where TEdge : IEdge
+            where TOpening : IOpening<TEdge>
+        {
+            PolyCurve polycurve = ExternalPolyCurve(panel);
```
IsakNaslundBh, line 49: need to null check the polycurve here as method returns null if nothing is found. If null, return false. added test to the testfile.

Analytical_Engine/Query/IsOutlineQuad.cs
```diff
@@ -0,0 +1,68 @@
...
+
+            if (polycurve.SubParts().Any(x => !x.IIsLinear()))
+            {
+                Reflection.Compute.RecordError("At least one of the external Panel edges is not linear.");
```
IsakNaslundBh, line 53: Would not raise error here, but just return false. After all, this is one of the things that should be checked, so nothing to treat as error/warning.

IsakNaslundBh (changes requested):
Some final comments. All seem fine besides those points. Changes requested need to be added to all three methods.

IsakNaslundBh (approved):
Works as expected now! LGTM!

IsakNaslundBh:
image

PR #1956, opened 2020-09-01
Structure_Engine: Adding method for checking objects on loads id assignment

### LINKED ISSUE #875: Structure_Engine: Add method to check loads for id assignment
Add method to check loads for id assignment to prevent Exception raised when loads are defined before having done an initial push of the geometry.

For example, when exporting bar with loads, if you try  to export both bar and load simultaneously with the same Push component you will get a `the given key is not present in the Dictionary` error on it. This is not informative to newcomers. Potential actions:

1. The exception should be caught or, better, prevented. A method should be checking loads to see if they are to be applied to bar which do not yet have an Id assigned (`bar.CustomData[AdapterId]` not existing).

2. A more meaningful error message should appear on the Push component, explaining that a first Push is needed in order to assign Ids to the bars, and then a second push component may be used to export Loads.

### Test file

This is an example file with an ETABS export.

https://burohappold.sharepoint.com/:u:/s/BHoM/ES01T4axRFxMjHx_CJ0UzhEBNebevX0PjQ_sWKSp5Xm_Lw?e=XuIIbC

The exception mentioned above happens in the `SetLoad` method in the `Load.cs` file when `SetLoadDistributed` is called:
``` 
ret = model.FrameObj.SetLoadDistributed(bar.CustomData[AdapterId].ToString(), ...
```
as `CustomData` is not already defined.

IsakNaslundBh:

Closes #875

Adding method checking if obejcts on the load all contain an specific adapter id.

### Test files

https://burohappold.sharepoint.com/:f:/s/BHoM/Ei47BMo0kzhNpmJDXFRIHlABU8sbHtLiiE-ELNQ12IwLXw?e=NAdF1b

### DIFF
--- a/Structure_Engine/Query/HasAssignedObjectIds.cs
+++ b/Structure_Engine/Query/HasAssignedObjectIds.cs
@@ -1,0 +1,51 @@
+﻿/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2020, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 
+ * GNU Lesser General Public License for more details.                          
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using System;
+using System.Linq;
+using System.Collections.Generic;
+using System.ComponentModel;
+using BH.oM.Reflection.Attributes;
+using BH.oM.Structure.Loads;
+using BH.oM.Base;
+
+namespace BH.Engine.Structure
+{
+    public static partial class Query
+    {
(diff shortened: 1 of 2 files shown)


IsakNaslundBh:
@peterjamesnugent @JosefTaylor would any of you be able to have a quick look at this? Postscript exists. Should hopefully be a 5 min job.

Structure_Engine/Query/HasAssignedObjectIds.cs
```diff
@@ -0,0 +1,51 @@
...
+
+namespace BH.Engine.Structure
+{
+    public static partial class Query
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Checks if the objects assigned to the IElementLoad all have an identifier matching the 'adapterIdName' assigned.\n" +
```
peterjamesnugent, line 39: AdapterIdName

Structure_Engine/Query/HasAssignedObjectIds.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Checks if the objects assigned to the IElementLoad all have an identifier matching the 'adapterIdName' assigned.\n" + 
+                     "Loads without correctly assigned ids to the objects can generally not be pushed through to structural packages.")]
+        [Input("load", "The load to check for id assignment.")]
+        [Input("adapterIdName", "The identifier key to check. For example 'Robot_id' for Autodesk Robot.")]
+        [Output("result", "Returns true if all objects assigned to the load has an id matching the 'adapterIdName' assigned.")]
```
peterjamesnugent, line 43: AdapterIdName - no need for '

peterjamesnugent (changes requested):
Functionality works as expected, uploaded a new version of the script that works with the current UI.

Changes for description.

Structure_Engine/Query/HasAssignedObjectIds.cs
IsakNaslundBh: was refering to the input name, but do get your point. Have updated

peterjamesnugent (approved):
LGTM.

PR #1960, opened 2020-09-03
Geometry_Engine: implementing DiscontinuityPoints(PolyCurve) and adding tolerances to inputs

### LINKED ISSUE #1954: Geometry_Engine: implement DiscontinuityPoints(PolyCurve) and give all methods proper inputs
#### Broken rules:

Following [#1919](https://github.com/BHoM/BHoM_Engine/pull/1913#discussion_r481020849):
- `DiscontinuityPoints(PolyCurve)` has a `[NotImplemented]` attribute despite being partially implemented.
- Most of `DiscontinuityPoints` methods are missing tolerance in their inputs.

- Implement `DiscontinuityPoints(PolyCurve)`.
- Add tolerance to every method.

LMarkowski:

Closes #1954 

### Test files

[SharePoint](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?viewid=a5a93cba%2Dfcca%2D46cc%2Da31a%2D80b8d5cbfd7b&id=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FTest%20Scripts%2FBHoM%5FEngine%2FGeometry%5FEngine%2FGeometry%5FEngine%2D%231954%2DimplementDiscontinuityPoints%28PolyCurve%29) - there are two "failing" cases in the test file. They are caused by errors from a grasshopper method. It is pointed in the file. **EDIT: Added another test file to prove versioning works fine.**

### Changelog

- updated `DiscontinuityPoints()` method for the `PolyCurve` class to return real discontinuity points.
- added tolerances to the inputs of every `DiscontinuityPoints()` method.

### Additional comments

`[PreviousVersion]` attribute has been used to update methods' inputs so rebuilding `Versioning_Toolkit` is needed for scripts using outdated methods to work.

### DIFF
--- a/Geometry_Engine/Query/DiscontinuityPoints.cs
+++ b/Geometry_Engine/Query/DiscontinuityPoints.cs
@@ -34,31 +34,57 @@ namespace BH.Engine.Geometry
         /**** Public Methods - Curves                   ****/
         /***************************************************/
 
-        public static List<Point> DiscontinuityPoints(this Arc curve)
+        [PreviousVersion("3.3", "BH.Engine.Geometry.Query.DiscontinuityPoints(BH.oM.Geometry.Arc)")]
+        public static List<Point> DiscontinuityPoints(this Arc curve, double distanceTolerance = Tolerance.Distance, double angleTolerance = Tolerance.Angle)
         {
             return new List<Point> { curve.StartPoint(), curve.EndPoint() };
         }
 
         /***************************************************/
 
-        public static List<Point> DiscontinuityPoints(this Circle curve)
+        [PreviousVersion("3.3", "BH.Engine.Geometry.Query.DiscontinuityPoints(BH.oM.Geometry.Circle)")]
+        public static List<Point> DiscontinuityPoints(this Circle curve, double distanceTolerance = Tolerance.Distance, double angleTolerance = Tolerance.Angle)
         {
             return new List<Point>();
         }
 
         /***************************************************/
 
-        public static List<Point> DiscontinuityPoints(this Line curve)
+        [PreviousVersion("3.3", "BH.Engine.Geometry.Query.DiscontinuityPoints(BH.oM.Geometry.Line)")]
+        public static List<Point> DiscontinuityPoints(this Line curve, double distanceTolerance = Tolerance.Distance, double angleTolerance = Tolerance.Angle)
         {
             return new List<Point> { curve.Start, curve.End };
         }
 
         /***************************************************/
 
-        [NotImplemented]
-        public static List<Point> DiscontinuityPoints(this PolyCurve curve)
+        [PreviousVersion("3.3", "BH.Engine.Geometry.Query.DiscontinuityPoints(BH.oM.Geometry.PolyCurve)")]
+        public static List<Point> DiscontinuityPoints(this PolyCurve curve, double distanceTolerance = Tolerance.Distance, double angleTolerance = Tolerance.Angle)
         {
-            return curve.Curves.SelectMany((x, i) => x.IDiscontinuityPoints().Skip((i > 0) ? 1 : 0)).ToList();
+            List<Point> result = new List<Point>();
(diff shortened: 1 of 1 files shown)


Geometry_Engine/Query/DiscontinuityPoints.cs
```diff
@@ -34,31 +34,57 @@ public static partial class Query
         /**** Public Methods - Curves                   ****/
         /***************************************************/
 
-        public static List<Point> DiscontinuityPoints(this Arc curve)
+        [PreviousVersion("3.3", "BH.Engine.Geometry.Query.DiscontinuityPoints(BH.oM.Geometry.Arc)")]
+        public static List<Point> DiscontinuityPoints(this Arc curve, double distanceTolerance = Tolerance.Distance, double angleTolerance = Tolerance.Angle)
         {
             return new List<Point> { curve.StartPoint(), curve.EndPoint() };
```
pawelbaran, line 40: Wouldn't it be worth checking if the arc's angle is not a multiplication of 2*Pi? Happy to resolve it in a separate thread as out of the scope of this PR.

pawelbaran (approved):
Happy about the code changes as well as the test file, therefore approving. However, I would feel safer if there was a test file proving that versioning works fine.

On top of that, I might have spotted the bug in the method for `Arc` - please see the code comment.

Geometry_Engine/Query/DiscontinuityPoints.cs
LMarkowski: Hm yes, if we assume that any 2*Pi Arc is a Circle it shouldn't return any discontinuity points. But do we?

Also I've taken a look at the `Arc` to `Circle` casting operation:      
```
public static explicit operator Circle(Arc curve)
{
    if (System.Math.Abs((curve.EndAngle - curve.StartAngle) - System.Math.PI * 2) > Tolerance.Distance)
        return null;

        return new Circle()
    {
        Centre = curve.CoordinateSystem.Origin,
        Normal = curve.CoordinateSystem.Z,
        Radius = curve.Radius,
    };
}
```
Shouldn't it record any `couldn't convert arc to circle` error?
If we try to cast a regular arc to a circle it can mess results without letting the user know that something went wrong.
image

LMarkowski:
>I would feel safer if there was a test file proving that versioning works fine.

@pawelbaran done 😉

Geometry_Engine/Query/DiscontinuityPoints.cs
pawelbaran: > Hm yes, if we assume that any 2*Pi Arc is a Circle it shouldn't return any discontinuity points. But do we?

In my understanding, if tangents of two meeting segments (or one, in case of 360 deg arc) are same, then there is no discontinuity at that point. However, it all comes down to how we define the discontinuity.

Concerning the `Arc` => `Circle` cast, shouldn't the component turn red due to `null` argument?

IsakNaslundBh (approved):
Had a look as well, and all generally seem fine.

Only concern would be for "invalid" polycurves, that is, curves that have segments in unsorted order etc, but I think that is a larger issue that do not need to be tackled just for this case. I ahve raised a separate issue for it here https://github.com/BHoM/BHoM_Engine/issues/1963

So, apart from that, all looks great!

Geometry_Engine/Query/DiscontinuityPoints.cs
LMarkowski: > shouldn't the component turn red due to null argument? 

It doesn't. The method always just returns an empty list so the null argument is never used (and stays undiscovered).
image

pawelbaran: Right, good point. Then there should be a null check I guess

LMarkowski: Moving to #1964

PR #1970, opened 2020-09-08
Analytical_Engine: Adding support for IRegion

IsakNaslundBh:
### NOTE: Depends on 

https://github.com/BHoM/BHoM/pull/985
   
https://github.com/BHoM/BHoM/issues/970

Adding IElement2D methods required for IRegion.

https://github.com/BHoM/documentation/wiki/IElement-required-extension-methods

### Test files

https://burohappold.sharepoint.com/:f:/s/BHoM/ElcGHiplcxxJuRsZYz9ewx8BkczFetyKcjcxAhclljhVSw?e=Pdksu3

### DIFF
--- a/Analytical_Engine/Modify/SetOutlineElements1D.cs
+++ b/Analytical_Engine/Modify/SetOutlineElements1D.cs
@@ -28,6 +28,7 @@ using System.Collections.Generic;
 using System.Linq;
 using BH.oM.Reflection.Attributes;
 using System.ComponentModel;
+using BH.Engine.Geometry;
 using BH.Engine.Base;
 
 
@@ -79,7 +80,27 @@ namespace BH.Engine.Analytical
         public static IRegion SetOutlineElements1D(this IRegion region, IEnumerable<IElement1D> outlineElements)
         {
             IRegion r = region.GetShallowClone(true) as IRegion;
-            r.Perimeter = BH.Engine.Geometry.Compute.IJoin(outlineElements.Cast<ICurve>().ToList())[0];
+
+            if (outlineElements.Count() == 1)
+                r.Perimeter = outlineElements.First() as ICurve;
+            else
+            {
+                List<PolyCurve> joinedCurves = Engine.Geometry.Compute.IJoin(outlineElements.Cast<ICurve>().ToList());
+
+                if (joinedCurves.Count == 1)
+                {
+                    if (!joinedCurves[0].IsClosed())
+                        Engine.Reflection.Compute.RecordWarning("The outline elements assigned to the region do not form a closed loop.");
+
+                    r.Perimeter = joinedCurves[0];
+                }
+                else
+                {
+                    Engine.Reflection.Compute.RecordWarning("The outline elements assigned to the region are disjointed.");
+                    r.Perimeter = new PolyCurve { Curves = outlineElements.Cast<ICurve>().ToList() };
+                }
+            }
+
             return r;
         }
 
--- a/Analytical_Engine/Modify/SetOutlineElements1D.cs
+++ b/Analytical_Engine/Modify/SetOutlineElements1D.cs
@@ -73,7 +73,7 @@ namespace BH.Engine.Analytical
         /***************************************************/
 
         [Description("Sets the Outline Element1Ds of an IRegion, i.e. the perimiter. Method required for all IElement2Ds.")]
-        [Input("region", "The Opening to update the Edges of.")]
+        [Input("region", "The IRegion to update the Perimeter of.")]
         [Input("outlineElements", "A list of IElement1Ds which all should be Geometrical ICurves.")]
         [Output("region", "The region with updated perimiter.")]
         public static IRegion SetOutlineElements1D(this IRegion region, IEnumerable<IElement1D> outlineElements)
@@ -112,4 +112,3 @@ namespace BH.Engine.Analytical
         /***************************************************/
     }
 }
-
--- a/Analytical_Engine/Modify/SetOutlineElements1D.cs
+++ b/Analytical_Engine/Modify/SetOutlineElements1D.cs
@@ -81,24 +81,21 @@ namespace BH.Engine.Analytical
         {
             IRegion r = region.GetShallowClone(true) as IRegion;
 
-            if (outlineElements.Count() == 1)
-                r.Perimeter = outlineElements.First() as ICurve;
+            IEnumerable<ICurve> joinedCurves = outlineElements.Cast<ICurve>();
+            if (outlineElements.Count() != 1)
+                joinedCurves = Engine.Geometry.Compute.IJoin(outlineElements.Cast<ICurve>().ToList());
+
+            if (joinedCurves.Count() == 1)
+            {
+                if (!joinedCurves.First().IIsClosed())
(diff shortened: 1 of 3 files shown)


Analytical_Engine/Modify/SetOutlineElements1D.cs
```diff
@@ -70,6 +70,19 @@ public static IOpening<TEdge> SetOutlineElements1D<TEdge>(this IOpening<TEdge> o
             return pp;
         }
 
+        /***************************************************/
+
+        [Description("Sets the Outline Element1Ds of an IRegion, i.e. the perimiter. Method required for all IElement2Ds.")]
+        [Input("region", "The Opening to update the Edges of.")]
```
pawelbaran, line 76:
```suggestion
        [Input("region", "The IRegion to update the Perimeter of.")]
```

Analytical_Engine/Query/Geometry.cs
```diff
@@ -171,5 +171,17 @@ public static Face Geometry(this IFace face)
             return geomFace;
         }
 
+        /***************************************************/
+
+        [Description("Gets the geometry of a IRegion as its Perimiter curve. Method required for automatic display in UI packages.")]
```
pawelbaran, line 176: Not only for that, but also for `RunExtensionMethod` pattern - may be worth mentioning

Analytical_Engine/Query/OutlineElements1D.cs
```diff
@@ -60,6 +61,16 @@ public static List<IElement1D> OutlineElements1D<TEdge>(this IOpening<TEdge> ope
         }
 
         /***************************************************/
+
+        [Description("Gets the boundary from an IRegion defining the boundary of the element as the subparts of the perimiter curve. Method required for all IElement2Ds.")]
+        [Input("region", "The IRegion to get outline elements from.")]
+        [Output("elements", "Outline elements of the IOpening, i.e. the subparts of the Perimiter curve.")]
```
pawelbaran, line 67:
```suggestion
        [Output("elements", "Outline elements of the IRegion, i.e. the subparts of the Perimiter curve.")]
```

pawelbaran (changes requested):
Descriptions need improvement - please see the code comments.

Analytical_Engine/Query/Geometry.cs
IsakNaslundBh: hmm, true. Not sure if that adds info to the user or not though? Saying that, not sure the current comment does either!

pawelbaran: Not sure about the user, but could be good to capture that in the API? I am not sure how many people will reach the `RunExtensionMethod` Wiki page.

The above may be a more general question to all get/set geometry methods, therefore happy to approve the PR as is.

Analytical_Engine/Modify/SetOutlineElements1D.cs
```diff
@@ -70,6 +70,19 @@ public static IOpening<TEdge> SetOutlineElements1D<TEdge>(this IOpening<TEdge> o
...
+        /***************************************************/
+
+        [Description("Sets the Outline Element1Ds of an IRegion, i.e. the perimiter. Method required for all IElement2Ds.")]
+        [Input("region", "The IRegion to update the Perimeter of.")]
+        [Input("outlineElements", "A list of IElement1Ds which all should be Geometrical ICurves.")]
+        [Output("region", "The region with updated perimiter.")]
+        public static IRegion SetOutlineElements1D(this IRegion region, IEnumerable<IElement1D> outlineElements)
+        {
+            IRegion r = region.GetShallowClone(true) as IRegion;
+            r.Perimeter = BH.Engine.Geometry.Compute.IJoin(outlineElements.Cast<ICurve>().ToList())[0];
```
pawelbaran, line 82: I would consider adding a check if the output of `IJoin` contains only 1 item and if that item is closed - if not, raise a warning that the output may be wrong.

pawelbaran (changes requested):
Sorry for blocking again, but looks like one of my previous comments did not get recorded for some mysterious reason 👻 Please see the code comment below.

Analytical_Engine/Modify/SetOutlineElements1D.cs
```diff
@@ -79,7 +80,27 @@ public static IOpening<TEdge> SetOutlineElements1D<TEdge>(this IOpening<TEdge> o
         public static IRegion SetOutlineElements1D(this IRegion region, IEnumerable<IElement1D> outlineElements)
         {
             IRegion r = region.GetShallowClone(true) as IRegion;
-            r.Perimeter = BH.Engine.Geometry.Compute.IJoin(outlineElements.Cast<ICurve>().ToList())[0];
+
+            if (outlineElements.Count() == 1)
```
pawelbaran, line 84: I would remove this `if` statement - what if the `outlineElements` contains one curve that is not closed? Better let it pass through `IJoin` regardless the number of curves and then do the checks.

pawelbaran (changes requested):

Analytical_Engine/Modify/SetOutlineElements1D.cs
IsakNaslundBh: thinking here was if the curve was a single circle for example. Then no real point wrapping it in a polycurve. still ok with changing to what you suggest, but any thoughts about that?

pawelbaran: Right, I was convinced that `IJoin` returns the original curve in case of single item input.

I take back my comment, but now when I am thinking about it I am not sure if it would not be better to refine that in `Join` (redo it to return `ICurve` and return single item inputs without wrapping into `PolyCurve`) instead of adding such check here?

And if we are meant to keep the code as is, it would still be good to check if the input curve is closed. To wrap it a bit, maybe:

```
IEnumerable<ICurve> joinedCurves = outlineElements;
            if (outlineElements.Count() != 1)
                joinedCurves = Engine.Geometry.Compute.IJoin(outlineElements.Cast<ICurve>().ToList());

            if (joinedCurves.Count() == 1)
            {
                if (!joinedCurves.First().IIsClosed())
                    Engine.Reflection.Compute.RecordWarning("The outline elements assigned to the region do not form a closed loop.");

                r.Perimeter = joinedCurves[0];
            }
            else
            {
                Engine.Reflection.Compute.RecordWarning("The outline elements assigned to the region are disjointed.");
                r.Perimeter = new PolyCurve { Curves = outlineElements.Cast<ICurve>().ToList() };
            }
```

Not sure if I did not float with thoughts too far though 🤣

IsakNaslundBh: updated

pawelbaran (approved):
Code LGTM now, last question: are we happy to merge without a test file? Looks safe, but double checking.

FraserGreenroyd (approved):
Approved in line with the oM :smile:

IsakNaslundBh:
> Code LGTM now, last question: are we happy to merge without a test file? Looks safe, but double checking.

Good point pawel. Missed that I never uploaded the testfile. Have put it on sharepoint, link in description, if you want to have a quick test!

LMarkowski (approved):
Checked the test file. Works fine and record proper warnings. lgtm

PR #1985, opened 2020-09-10
Structure_Engine: Update Torsional constant calculation for BoxProfiles

### LINKED ISSUE #1977: Structure_Engine: Improve Torsion constant calculation for Box profiles with corner radii
#### Description:

Updating torsion constant calculation for box profiles to use formulation from BS EN 10210-2:2019

IsakNaslundBh:

Closes #1977

Updating definition to match https://eurocodeapplied.com/design/en1993/shs-design-properties which is the same as used in BS EN 10210-2:2019

### Test files

https://burohappold.sharepoint.com/:f:/s/BHoM/EjDM8rvuDEFLtM7SyVXbeCABNmrhszeNgbCetHjliwFhMw?e=KAYjzR

### DIFF
--- a/Structure_Engine/Query/TorsionalConstant.cs
+++ b/Structure_Engine/Query/TorsionalConstant.cs
@@ -107,20 +107,20 @@ namespace BH.Engine.Structure
 
         /***************************************************/
 
-        [Description("Calcualtes the Torsinal constant for the profile. Note that this is not the polar moment of inertia.")]
+        [Description("Calcualtes the Torsinal constant for the profile. Note that this is not the polar moment of inertia. Formulation taken from BS EN 10210-2:2019.")]
         [Input("profile", "The ShapeProfile to calculate the torsional constant for.")]
         [Output("J", "Torsional constant of the profile. Note that this is not the polar moment of inertia.", typeof(TorsionConstant))]
         public static double TorsionalConstant(this BoxProfile profile)
         {
-            double tf1 = profile.Thickness;
-            double tw = profile.Thickness;
+            double t = profile.Thickness;
             double width = profile.Width;
             double height = profile.Height;
 
-
-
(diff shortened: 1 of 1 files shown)


Structure_Engine/Query/TorsionalConstant.cs
```diff
@@ -107,20 +107,20 @@ public static double TorsionalConstant(this FabricatedBoxProfile profile)
 
         /***************************************************/
 
-        [Description("Calcualtes the Torsinal constant for the profile. Note that this is not the polar moment of inertia.")]
+        [Description("Calcualtes the Torsinal constant for the profile. Note that this is not the polar moment of inertia. Formulation taken from BS EN 10210-2:2019.")]
```
peterjamesnugent, line 110: Calculates the torsional constant for the profile in accordance with BS EN 10210-2:2019. Note that this is not the polar moment of intertia.

peterjamesnugent (changes requested):
Happy with the formulation and the percentage differences based on:
- The old formulation does not take in to account the additional section provided by the radii
- The new formulation is very close (<0.5%) to the catalogued sections 
- Even though the old formulation (with 0 radius is 5% different) and more conservative does not necessarily mean it is more accurate
- I would feel more comfortable with the codified approach than a formula used by a specific software

I have just made a changed so that it's abundantly clear where the formulation has come from - would it be too much adding the Equation number?

Structure_Engine/Query/TorsionalConstant.cs
IsakNaslundBh: Done

IsakNaslundBh:
> I have just made a changed so that it's abundantly clear where the formulation has come from - would it be too much adding the Equation number?

Did not actually have an equation number in the document, but added reference to the section.

Also, fixed description typo on the other methods.

peterjamesnugent (approved):
LGTM!

PR #1994, opened 2020-09-15
Data_Engine: Adding CleanList method

### LINKED ISSUE #1993: Data_Engine: Add clean list method
#### Description:

Method for cleaning list (removing nulls) was correctly removed from the Excel_Engine in https://github.com/BHoM/Excel_Toolkit/pull/243 .

This method was however not replaced in the Data_Engine during this refactoring. Having the ability to clean up lists of objects in excel is critical and hard to achieve in Excel without a method of this sort, with no simple workaround, and this is functionality that has now gone missing from previous beta.

This is for those reasons proposed to be pushed as a critical issue to be included in the beta release, even if feature of nature, as critical for usability of the Excel_Toolkit for several workflows.

IsakNaslundBh:

Closes #1993 

Adding `CleanList` method critical for excel workflows.

### Test files

Test functionality:
https://burohappold.sharepoint.com/:f:/s/BHoM/Eo3XCg9b5rlNteTRdW0oT0oBnfGJaotIe7sKZlWIHZb7OA?e=Z3fdi8

Test in workflow:
Cell I2 in the supported nodes tab:
https://burohappold.sharepoint.com/:f:/s/BHoM/EszNjShmWNFJpiMmQ1EdM7ABAa3i7rfMe8oa54uHFtbwxg?e=O1XXQR

### DIFF
--- a/Data_Engine/Query/RemoveNulls.cs
+++ b/Data_Engine/Query/RemoveNulls.cs
@@ -34,7 +34,7 @@ namespace BH.Engine.Data
         /***************************************************/
 
         [Description("Removing any null objects from the input collection.")]
-        [Input("collection", "The collection to clean from null objects.")]
+        [Input("collection", "The collection to remove nulls from.")]
         [Output("list", "The collection without nulls in it as a list.")]
         public static List<T> RemoveNulls<T>(this IEnumerable<T> collection)
         {
--- a/Data_Engine/Query/RemoveNulls.cs
+++ b/Data_Engine/Query/RemoveNulls.cs
@@ -32,10 +32,10 @@ namespace BH.Engine.Data
         /**** Public Methods                            ****/
         /***************************************************/
 
-        [Description("Cleans the list by removing any null objects from it.")]
+        [Description("Removing any null objects from the input list.")]
         [Input("list", "The list to clean from null objects.")]
         [Output("list", "The list without nulls in it.")]
-        public static List<T> CleanList<T>(this List<T> list)
+        public static List<T> RemoveNulls<T>(this List<T> list)
         {
             return list.FindAll(x => x != null);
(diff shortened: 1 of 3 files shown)


FraserGreenroyd (dismissed):
LGTM following discussion, understand why the code is what it is, so am happy, thanks @IsakNaslundBh

adecler:
Happy for this to be merged but worth noting that I would have preferred the method to be named `RemoveNulls` as `CleanList` has two issues in my opinion:
- the main one:  high chance to guess incorrectly what the method does by just looking at the name
- side one: why have `List` in the method name while we already know it applies to list ?

IsakNaslundBh:
Fair point @adecler , have updated. Think the method, whilst in excel, might initially have had a few more checks in it, but agree with the name change for this here!

Data_Engine/Query/RemoveNulls.cs
```diff
@@ -0,0 +1,45 @@
...
+    public static partial class Query
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Removing any null objects from the input list.")]
+        [Input("list", "The list to clean from null objects.")]
+        [Output("list", "The list without nulls in it.")]
+        public static List<T> RemoveNulls<T>(this List<T> list)
```
FraserGreenroyd, line 38: Would it be more powerful to have this work on `IEnumerable` instead of strict `List`?

IsakNaslundBh: Yeah, good point. Updated

FraserGreenroyd (approved):
Sorry for the picky comments on chat @IsakNaslundBh - but LGTM now :smile:

PR #1996, opened 2020-09-16
Physical_Engine: Removing reference to Common_oM

### LINKED ISSUE #1995: Physical_Engine: Still has Common_oM referenced
#### Broken rules:

Common_oM does no longer exist and should not be referenced

Remove the reference

IsakNaslundBh:

Closes #1995 

Common_oM does no longer exist and should not be referenced.

Not sure how this snuck by the major purge of Common_oM/Engine, but better late than never.

### Test files

No needed. Only removing reference to a dll no longer used/existing.


pawelbaran (approved):
The number of changed lines looks like a test of perceptiveness 😃

FraserGreenroyd (approved):
Kanske...

PR #2002, opened 2020-09-21
Reflection_Engine: Record message when failing to load methods

### LINKED ISSUE #2001: Reflection_Engine: Show exception messages on failing to load methods
#### Description:

Error message should be recorded to give better insight into errors on loading methods

IsakNaslundBh:

Closes #2001 

Recording exception message when failing to load methods to give more detail of what is failing.

### DIFF
--- a/Reflection_Engine/Query/MethodList.cs
+++ b/Reflection_Engine/Query/MethodList.cs
@@ -124,9 +124,17 @@ namespace BH.Engine.Reflection
                         }
                     }
                 }
-                catch (Exception)
+                catch (Exception e)
                 {
-                    Compute.RecordWarning("Cannot load types from assembly " + asm.GetName().Name);
+
+                    string message = "Cannot load types from assembly " + asm.GetName().Name + ". Exception message: " + e.Message;
+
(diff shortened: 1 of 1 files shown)


Reflection_Engine/Query/MethodList.cs
```diff
@@ -124,9 +124,17 @@ private static void ExtractAllMethods()
...
                     }
                 }
-                catch (Exception)
+                catch (Exception e)
                 {
-                    Compute.RecordWarning("Cannot load types from assembly " + asm.GetName().Name);
+
+                    string message = "Cannot load types from assembly " + asm.GetName().Name + ". Exception message: " + e.Message;
+
+                    if (e.InnerException != null)
```
alelom, line 132: It feels safer if we do:
```suggestion
                    if (!string.IsNullOrEmpty(e.InnerException.Message))
```
because the exception message can be empty in some cases, where we would end up having a trailing `InnerException: ` text with nothing after it.

alelom (approved):
LGTM!

al-fisher (approved):
Agreed. Helpful already
LGTM.

PR #2004, opened 2020-09-21
Reflection_Engine: Adding in pad locks for method and type extraction

### LINKED ISSUE #2003: Reflection_Engine: MethodList and TypeLists are not thread safe
#### Description:

Methods extracting types and/or methods are not thread safe. Can in rare cases lead to race conditions when two threads tries to extract methods/types.

Padlocks to be put in place to protect against this.

#### Steps to reproduce:

Could be linked to https://github.com/BHoM/BHoM_UI/issues/305

IsakNaslundBh:

Closes #2003

Adding padlocks around methods calling ExtractAllMethods and ExtractAllTypes to make methods threadsafe.

This is an attempt of solving the https://github.com/BHoM/BHoM_UI/issues/305 issue.

Note that the `m_GetMethodsLock` in the Engine types is intentional, as engine types currently is extracted when extracting methods. Something we might want to change in the future, but outside the scope of this PR.

**EDIT:** Had a look through and added locks to other potentially unsafe places for the BHoMAnalytics as well as making sure the whole Reflection_Engine got covered. This includes LoadAllAssemblies, AssemblyList and RegisterTypes in BSON converts. Not sure if the one is BSON is needed, as Mongo might be threadsafe and able to cope with it, but rather safe than sorry.

### Test files

No real testfile needed. Everything should work as before, but hopefully stops the crashes of GH at startup.

### DIFF
--- a/Reflection_Engine/Compute/LoadAllAssemblies.cs
+++ b/Reflection_Engine/Compute/LoadAllAssemblies.cs
@@ -38,53 +38,54 @@ namespace BH.Engine.Reflection
 
         public static void LoadAllAssemblies(string folder = "")
         {
-            if (!m_AssemblyAlreadyLoaded)
+            lock (m_LoadAssembliesLock)
             {
-                m_AssemblyAlreadyLoaded = true;
-                HashSet<string> loaded = new HashSet<string>(AppDomain.CurrentDomain.GetAssemblies().Select(x => x.FullName.Split(',').First()));
-
-                if (string.IsNullOrEmpty(folder))
-                    folder = Query.BHoMFolder();
-
-                if (!Directory.Exists(folder))
+                if (!m_AssemblyAlreadyLoaded)
                 {
-                    RecordWarning("The folder provided to load the assemblies from does not exist: " + folder);
-                    return;
-                }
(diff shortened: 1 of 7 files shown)


al-fisher:
FYI - Been running on this branch for the last 4 hrs or so. All behaving as expected.

adecler:
Looks good to me. 
Out of curiosity, why didn't you put the lock directly inside `ExtractAllMethods` and `ExtractAllTypes` ?

IsakNaslundBh:
> Out of curiosity, why didn't you put the lock directly inside `ExtractAllMethods` and `ExtractAllTypes` ?

Thought about doing that, but then extraction and reflection work done in  `ExtractAllMethods` and `ExtractAllTypes` would have been triggered one more time unnecessarily. Also, could mean that the list was again changed if a second run of `ExtractAllMethods` was running when at the same time a call was made to `BHoMMethodList()` for example. So thought safer to put the protection around all public checks and entry points, rather than just around the private methods.

adecler (approved):
This looks good to me.

Happy for it to be merge as it purely an improvement without downsides. This is not aimed to make the code multithreading safe but purely to fix the GH freeze we experienced lately and the locks are well isolated from each others.

PR #2018, opened 2020-10-01
Spatial_Engine: ElementVertices fixed to return actual vertices

### LINKED ISSUE #2017: Spatial_Engine: ElementVertices returns incomplete discontinuity points instead of actual vertices
#### Description:

As in the title: `ElementVertices` returns discontinuity points of an `IElement1D` or discontinuity points of an outline of `IElement2D`, plus it skips the 1st point. Using the example in the picture below, 5 points should be returned instead of 3:

image

A short explanation about 3 types of point queries and how I understand them:
- control points - control points of a curve in its nurbs form (currently simplified for `Arc` and `Circle` in the BHoM)
- discontinuity points - points at geometrical discontinuities of a curve (discrete change in the tangent)
- vertices - control points in case of a `Polyline` (see the example above) and discontinuity points for all other types - in general, all points that are either a discontinuity or are explicitly specified in the curve definition

I will open a PR following the above in a sec, FYI @al-fisher @LMarkowski @IsakNaslundBh  @isak

pawelbaran:

Closes #2017

### Test files

On [SharePoint](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?RootFolder=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FTest%20Scripts%2FBHoM%5FEngine%2FSpatial%5FEngine%2FIssues%2FSpatial%5FEngine%5F%232017%5FElementVerticesBug&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4) - both issue specific test as well as updated general test

### Additional comments

Following #2017, my understanding of different types of points in relationship to their parent curve:
> * control points - control points of a curve in its nurbs form (currently simplified for `Arc` and `Circle` in the BHoM)
> * discontinuity points - points at geometrical discontinuities of a curve (discrete change in the tangent)
> * vertices - control points in case of a `Polyline` (see the example above) and discontinuity points for all other types - in general, all points that are either a discontinuity or are explicitly specified in the curve definition

The method subject to this PR should return the third option, as far as I see it.

### DIFF
--- a/Spatial_Engine/Query/ElementVertices.cs
+++ b/Spatial_Engine/Query/ElementVertices.cs
@@ -56,7 +56,11 @@ namespace BH.Engine.Spatial
         public static List<Point> ElementVertices(this IElement1D element1D)
         {
             ICurve curve = element1D.IGeometry();
-            List<Point> vertices = curve.IDiscontinuityPoints();
+            List<Point> vertices;
+            if (curve is Polyline)
+                vertices = ((Polyline)curve).ControlPoints;
+            else
+                vertices = curve.IDiscontinuityPoints();
 
             if (curve.IIsClosed())
                 vertices.RemoveAt(vertices.Count - 1);
--- a/Spatial_Engine/Query/ElementVertices.cs
+++ b/Spatial_Engine/Query/ElementVertices.cs
@@ -56,11 +56,19 @@ namespace BH.Engine.Spatial
         public static List<Point> ElementVertices(this IElement1D element1D)
         {
             ICurve curve = element1D.IGeometry();
-            List<Point> vertices;
-            if (curve is Polyline)
-                vertices = ((Polyline)curve).ControlPoints;
-            else
-                vertices = curve.IDiscontinuityPoints();
+
+            List<Point> vertices = new List<Point>();
+            List<ICurve> subParts = curve.ISubParts().ToList();
+            if (subParts.Count == 0 || subParts.All(x => x is Circle))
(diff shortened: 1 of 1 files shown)


LMarkowski:
Hm. IMO returning cPts for Polyline and discontinuity points for everything else seems a bit inconsistent.
Especially if we imagine a `PolyCurve` consisting of lines only. In such a case a point between two tangent segments won't be returned as an vertex, but would be for geometrically identical `Polyline`.

Correct me if I misunderstood the rule. Haven't done any tests yet.

pawelbaran:
Ha, yeah, `PolyCurve` with cotangent segments is exactly what I am struggling with right  now 🐒 The PR is incomplete code wise, but I would still stick to the description, where vertices would be all kinks + end points of every segment.

pawelbaran:
Code fixed + test files updated - ready to review!

LMarkowski (approved):
Code looks good, tests run as they should and the vertices as the ends of an element's every segment sound logical. LGTM

pawelbaran:
@al-fisher agreed on the ambiguity, the concept of a vertex can also be understood differently, as discussed with @IsakNaslundBh. Maybe it is worth writing it down in the Wiki and then trickle down by checking whether every relevant method aligns to the definitions set?

al-fisher:
Yes - agreed. Sounds like a good idea @pawelbaran

PR #2020, opened 2020-10-02
Spatial_Engine: Missing methods added

### LINKED ISSUE #2019: Spatial_Engine: Add missing methods
#### Description:

A few low hanging fruits to pick in the Spatial_Engine:
- [ ] IsSelfIntersecting (`IElement0D`, interface)
- [ ] ElementCurves(`IElement0D`)
- [ ] ControlPoints(`IElement0D`, interface)
- [ ] Centroid(`IElement0D`, interface)
- [ ] Area(`IElement0D`, `IElement1D`, interface)

pawelbaran:

Closes #2019

### Test files

Not needed - obvious cases added. But I encourage to read the code carefully and randomly test the changes!

### Changelog

Missing Spatial_Engine methods added:
- IsSelfIntersecting (`IElement0D`, interface)
- ElementCurves(`IElement0D`)
- ControlPoints(`IElement0D`, interface)
- Centroid(`IElement0D`, interface)
- Area(`IElement0D`, `IElement1D`, interface)

### DIFF
--- a/Spatial_Engine/Query/Centroid.cs
+++ b/Spatial_Engine/Query/Centroid.cs
@@ -36,10 +36,10 @@ namespace BH.Engine.Spatial
         /****            IElement0D            ****/
         /******************************************/
         
-        [Description("Queries the centre of weight of an IElement0D. Always returns the point location due to zero-dimensionality of an IElement0D.")]
+        [Description("Queries the centre of weight for the homogeneous geometrical representation of an IElement0D. Always returns the point location due to zero-dimensionality of an IElement0D.")]
         [Input("element0D", "The IElement0D with the geometry to get the centre of weight of.")]
         [Output("centroid", "The Point at the centre of weight for the homogeneous geometrical representation of the IElement0D.")]
-        public static Point Centroid(this IElement0D element0D)
+        public static Point Centroid(this IElement0D element0D, double tolerance = Tolerance.Distance)
         {
             return element0D.IGeometry();
         }
@@ -79,10 +79,10 @@ namespace BH.Engine.Spatial
         /****   Public Methods - Interfaces    ****/
         /******************************************/
         
-        [Description("Queries the centre of weight for a representation of the IElement.")]
+        [Description("Queries the centre of weight for the homogeneous geometrical representation of an IElement.")]
         [Input("element", "The IElement with the geometry to get the centre of mass of.")]
         [Input("tolerance", "Distance tolerance used in geometry processing, default set to BH.oM.Geometry.Tolerance.Distance")]
-        [Output("centroid", "The Point at the centre of weight for the homogeneous geometrical representation of the IElement.")]
+        [Output("centroid", "The Point at the centre of weight for the homogeneous geometrical representation of an IElement.")]
         public static Point ICentroid(this IElement element, double tolerance = Tolerance.Distance)
         {
             return Centroid(element as dynamic, tolerance);
--- a/Spatial_Engine/Query/Centroid.cs
+++ b/Spatial_Engine/Query/Centroid.cs
@@ -83,7 +83,7 @@ namespace BH.Engine.Spatial
         [Input("element", "The IElement with the geometry to get the centre of mass of.")]
         [Input("tolerance", "Distance tolerance used in geometry processing, default set to BH.oM.Geometry.Tolerance.Distance")]
         [Output("centroid", "The Point at the centre of weight for the homogeneous geometrical representation of the IElement.")]
-        public static bool ICentroid(this IElement element, double tolerance = Tolerance.Distance)
+        public static Point ICentroid(this IElement element, double tolerance = Tolerance.Distance)
         {
             return Centroid(element as dynamic, tolerance);
         }
--- a/Spatial_Engine/Query/Centroid.cs
+++ b/Spatial_Engine/Query/Centroid.cs
@@ -33,6 +33,19 @@ namespace BH.Engine.Spatial
     public static partial class Query
     {
         /******************************************/
+        /****            IElement0D            ****/
+        /******************************************/
+        
+        [Description("Queries the centre of weight of an IElement0D. Always returns the point location due to zero-dimensionality of an IElement0D.")]
+        [Input("element0D", "The IElement0D with the geometry to get the centre of weight of.")]
+        [Output("centroid", "The Point at the centre of weight for the homogeneous geometrical representation of the IElement0D.")]
+        public static Point Centroid(this IElement0D element0D)
+        {
+            return element0D.IGeometry();
+        }
+
+
+        /******************************************/
         /****            IElement1D            ****/
         /******************************************/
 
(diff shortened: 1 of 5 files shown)


Spatial_Engine/Query/ControlPoints.cs
```diff
@@ -30,6 +30,19 @@ namespace BH.Engine.Spatial
 {
     public static partial class Query
     {
+        /******************************************/
+        /****            IElement0D            ****/
+        /******************************************/
+
+        [Description("Queries the control points of the geometrical representation of the IElement0D. Always returns the point location due to zero-dimensionality of an IElement0D.")]
+        [Input("element0D", "The IElement0D with the geometry to get the control points from.")]
+        [Output("cPoints", "The control points of the geometrical representation of an IElement0D.")]
```
LMarkowski, line 39:
```suggestion
        [Description("Queries the control points of the geometrical representation of an IElement0D. Always returns the point location due to zero-dimensionality of an IElement0D.")]
        [Input("element0D", "The IElement0D with the geometry to get the control points from.")]
        [Output("cPoints", "The control points of the geometrical representation of an IElement0D.")]
```
Changing the article to match the general pattern. (Also I think both are correct, but I like "an" better in these cases. Maybe a native speaker should decide).

Spatial_Engine/Query/ControlPoints.cs
```diff
@@ -64,6 +77,19 @@ public static List<Point> ControlPoints(this IElement2D element2D, bool external
...
         }
 
+
+        /******************************************/
+        /****             IElement             ****/
+        /******************************************/
+
+        [Description("Queries the control points of the geometrical representation of the IElement.")]
+        [Input("element", "The IElement with the geometry to get the control points from.")]
+        [Output("cPoints", "The control points of the geometrical representation of an IElement.")]
```
LMarkowski, line 87:
```suggestion
        [Description("Queries the control points of the geometrical representation of an IElement.")]
        [Input("element", "The IElement with the geometry to get the control points from.")]
        [Output("cPoints", "The control points of the geometrical representation of an IElement.")]
```
same as above

Spatial_Engine/Query/ElementCurves.cs
```diff
@@ -33,11 +33,25 @@ namespace BH.Engine.Spatial
...
     public static partial class Query
     {
+        /******************************************/
+        /****            IElement0D            ****/
+        /******************************************/
+
+        [Description("Queries the defining curves of the IElement0D. Always returns empty collection due to zero-dimensionality of an IElement0D.")]
+        [Input("element0D", "The IElement0D to extract the defining curves from.")]
+        [Input("recursive", "Has no effect for IElement0D. Input here to unify inputs between all IElements.")]
+        [Output("elementCurves", "The curves defining the base geometry of the IElement0D.")]
```
LMarkowski, line 43:
```suggestion
        [Description("Queries the defining curves of an IElement0D. Always returns empty collection due to zero-dimensionality of an IElement0D.")]
        [Input("element0D", "The IElement0D to extract the defining curves from.")]
        [Input("recursive", "Has no effect for IElement0D. Input here to unify inputs between all IElements.")]
        [Output("elementCurves", "The curves defining the base geometry of the IElement0D.")]
```
Same as above, I guess? Maybe I'm to picky at this point.. 🙈

Spatial_Engine/Query/IsSelfIntersecting.cs
```diff
@@ -20,19 +20,30 @@
...
     public static partial class Query
     {
+        /******************************************/
+        /****            IElement0D            ****/
+        /******************************************/
+
+        [Description("Returns if the one dimensional representation of the IElement0D is closer to itself than the tolerance at any two points. Always false due to zero-dimensionality of an IElement0D.")]
+        [Input("element0D", "The IElement0D to evaluate self intersections from.")]
+        [Input("tolerance", "Minimum distance to be considered intersecting.", typeof(Length))]
+        [Output("A boolean which is true if the IElement0D is self intersecting.")]
```
LMarkowski, line 40:
```suggestion
        [Description("Returns if the geometrical representation of an IElement0D is closer to itself than the tolerance at any two points. Always false due to zero-dimensionality of an IElement0D.")]
        [Input("element0D", "The IElement0D to evaluate self intersections from.")]
        [Input("tolerance", "Minimum distance to be considered intersecting.", typeof(Length))]
        [Output("A boolean which is true if the geometrical representation of an IElement0D is self intersecting.")]
```
I don't think the `IElementd0D` becomes 1D at any point here. I think the following sentence makes the description clear enough.

Spatial_Engine/Query/IsSelfIntersecting.cs
```diff
@@ -69,6 +80,20 @@ public static bool IsSelfIntersecting(this IElement2D element2D, double toleranc
...
 
+
+        /******************************************/
+        /****   Public Methods - Interfaces    ****/
+        /******************************************/
+
+        [Description("Returns if the one dimensional representation of the IElement is closer to itself than the tolerance at any two points.")]
+        [Input("element", "The IElement to evaluate self intersections from.")]
+        [Input("tolerance", "Minimum distance to be considered intersecting.", typeof(Length))]
+        [Output("A boolean which is true if the IElement is self intersecting.")]
```
LMarkowski, line 91:
```suggestion
        [Description("Returns if the one dimensional representation of an IElement is closer to itself than the tolerance at any two points.")]
        [Input("element", "The IElement to evaluate self intersections from.")]
        [Input("tolerance", "Minimum distance to be considered intersecting.", typeof(Length))]
        [Output("A boolean which is true if the IElement is self intersecting.")]
```
I'm not sure about that one but is "one dimensional" suitable for `IElement2D`s? Probably yes, if we consider outlines as `IElement1D`s.
Also I think if we go the through interface we might lose an important information from `IElement2D` description ("Does not check for intersections between external and internal curves, or between different internal curves").

Spatial_Engine/Query/ElementCurves.cs
```diff
@@ -33,11 +33,25 @@ namespace BH.Engine.Spatial
...
+            return new List<ICurve>();
+        }
+
+
         /******************************************/
         /****            IElement1D            ****/
         /******************************************/
 
-        [Description("Queries the geometricly defining curve of the IElement1D.")]
+        [Description("Queries the defining curves of the IElement1D.")]
```
LMarkowski, line 54:
```suggestion
        [Description("Queries the defining curves of an IElement1D.")]
```
🤭

Spatial_Engine/Query/Centroid.cs
```diff
@@ -32,6 +32,19 @@ namespace BH.Engine.Spatial
...
     public static partial class Query
     {
+        /******************************************/
+        /****            IElement0D            ****/
+        /******************************************/
+        
+        [Description("Queries the centre of weight of an IElement0D. Always returns the point location due to zero-dimensionality of an IElement0D.")]
+        [Input("element0D", "The IElement0D with the geometry to get the centre of weight of.")]
+        [Output("centroid", "The Point at the centre of weight for the homogeneous geometrical representation of the IElement0D.")]
+        public static Point Centroid(this IElement0D element0D)
```
LMarkowski, line 42:
```suggestion
        public static Point Centroid(this IElement0D element0D, double tolerance = Tolerance.Distance)
```
Tolerance is needed to be picked up by the interface.
Similar issue is in the ControlPoints method, where `IElement2D` have different arguments then others, but that might be out of scope.

LMarkowski (changes requested):
Generally lgtm.
I took a deeper look into the descriptions. Sorry for all those change requests 🙊 
Maybe it's worth raising to create a clear template for the descriptions and then clean the whole.. project, and then maybe others.

Spatial_Engine/Query/IsSelfIntersecting.cs
pawelbaran: Good point, I will add an extra definition explaining how does the check work for the IElement2D. One-dimensional representation makes similar sense to me as to you - I see it as _wireframe_ of the multi-dimensional object.

Spatial_Engine/Query/Centroid.cs
pawelbaran: Nice catch 👍

LMarkowski (approved):
LGTM now 👍

al-fisher (approved):
Makes sense to me also

PR #2040, opened 2020-10-08
Environment_Engine: Adds features to the orientation method 

### LINKED ISSUE #1976: Environment_Engine: Filter elements by orientation
#### Description:

A filter method to return objects where their normal is oriented in the given direction - ideally by angle from north (with tolerance).

The way I see this method working would be:
- Set the north vector (by default this would be [0, 1, 0]).
- Access the underlying geometry of the panel and find the surface normal as a Vector. 
- Calculate the clockwise angle between the normal and north (excluding the Z-component) - where -90° would be equal to +270° rotation.

linaeriksson:
Adds features to the orientation method, such as azimuthAngle, and also works when the  northangle is not 90 degrees

Closes #1976 
[Normal Orientation test.zip](https://github.com/BHoM/BHoM_Engine/files/5349372/Normal.Orientation.test.zip)

### DIFF
--- a/Environment_Engine/Query/Orientation.cs
+++ b/Environment_Engine/Query/Orientation.cs
@@ -31,6 +31,7 @@ using BH.oM.Geometry;
 
 using BH.oM.Reflection.Attributes;
 using System.ComponentModel;
+using System.Security.Cryptography.X509Certificates;
 
 namespace BH.Engine.Environment
 {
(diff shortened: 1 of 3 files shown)


Environment_Engine/Query/Orientation.cs
```diff
@@ -42,17 +43,64 @@ public static partial class Query
 
         [Description("Returns the orientation of a given environmental object")]
         [Input("environmentObject", "Any object implementing the IEnvironmentObject interface that can have its orientation queried")]
-        [Output("orientation", "The orientation of the Environment Object")]
```
linaeriksson, line 45: I just realized that I deleted the output, will fix that

Environment_Engine/Query/Orientation.cs
```diff
@@ -42,17 +43,64 @@ public static partial class Query
...
+            Vector right = BH.Engine.Geometry.Create.Vector(1, 0, 0);
+
+
+            if (normal == up || normal == down)
+            {
+                BH.Engine.Reflection.Compute.RecordError("This method cannot successfully evaluate the orientation");
+                return (-1);
+            }
+
+            else if (normal.X > 0 && northAngle == 90.0 && azimuthAngle == true)
```
FraserGreenroyd, line 68: If the element we're evaluating is a boolean, like `azimuthAngle`, we don't need to do an `== true` - we can just do `if(azimuthAngle)` for example. If we want to evaluate whether it is false, we can use the `!` operator (sometimes referred to as the bang operator, or not operator) to indicate a `not` value. For example, `!azimuthAngle` will ask whether the variable is not true (i.e. false). This is the accepted convention for boolean evaluations rather than `== true` or `== false` :smile:

So the change needed for this line, and all the other lines below it, is to simply remove `== true` from the statement :smile:

FraserGreenroyd (changes requested):
Some minor comments on the code, and a major issue on the CSProject to resolve. Otherwise looks good, but still need to run full tests. @linaeriksson could you provide an example Grasshopper script to aid with testing if possible? :smile:

Environment_Engine/Query/Orientation.cs
```diff
@@ -42,17 +42,58 @@ public static partial class Query
 
         [Description("Returns the orientation of a given environmental object")]
         [Input("environmentObject", "Any object implementing the IEnvironmentObject interface that can have its orientation queried")]
+        [Input("northAngle", "The angle in degrees for north. Default is 90.0.")]
+        [Input("azimuthAngle", "Set to true to return the azimuth angle from north instead of the angle between north and the object normal. Default is false.")]
```
FraserGreenroyd, line 46: The input name on this documentation needs to match the variable name on the input - `returnAzimuthAngle`

Environment_Engine/Query/Orientation.cs
```diff
@@ -42,17 +42,58 @@ public static partial class Query
...
         [Output("orientation", "The orientation of the Environment Object")]
-        public static double Orientation(this IEnvironmentObject environmentObject)
+        [PreviousVersion("4.0", "BH.Engine.Environment.Query.Orientation(BH.oM.Environment.IEnvironmentObject)")]
+        public static double Orientation(this IEnvironmentObject environmentObject, double northAngle = 90.0, bool returnazimuthAngle = false)
```
FraserGreenroyd, line 49: `returnazimuthAngle` should be `returnAzimuthAngle` for camel casing :smile:

FraserGreenroyd (changes requested):
1 minor change and then this should be good to go for me :smile:

@tg359 are you happy with the functionality given you requested this from the issue? :smile:

tg359:
A couple of changes made to harness the Plane oriented angle method available in Geometry_Engine, but otherwise the logic is the same.

[Test script here that contains both the Orientation and Cardinality methods](https://github.com/BHoM/BHoM_Engine/files/5355855/BHoM_Orientation_Cardinality_testScript.zip)

FraserGreenroyd (approved):
This LGTM now, thanks both!

Issue #2044, opened 2020-10-09
Geometry_Engine: Restore coherency and add tolerance to PointAtParameter methods


LMarkowski:
#### Broken rules:

Some of `PointAtParameter` methods allow the parameter to be outside of 0 to 1 domain while others in those cases return nulls.
Also `Polyline` and `PolyCurve` methods has hard coded check to return endpoint while parameter is equal 1 (to avoid computational errors). This could be replaced by using tolerance. 

#### Suggestions to restore compliance:

- Make all methods prohibit input parameter from being outside of the 0 to 1 domain. Ensure that 0 means the startpoint and 1 the endpoint of the curve.
- Add proper descriptions and return null with logging an error when parameter is outside the 0 to 1 domain.

#### Additional notes:
- Circles and other closed curves are discussable.
- In case of Nurbs parameter could be scaled to meet the requirement from first suggestion.

pawelbaran:
I am removing the point regarding Nurbs from here as to be resolved in a separate issue (https://github.com/BHoM/BHoM_Engine/issues/2228).

PR #2028, opened 2020-10-06
Geometry_Engine: BasisFunction, Points, Tangents, Derivatives and Curvatures for NurbsCurve/Surfaces

theThorsager:
### Issues addressed by this PR
Closes #2016

So I fell down the Nurbs hole.
Not a horrendus place if one ignores all documentation about it.
And got to learn some neat things about surfaces.

Anyway, heres a skeleton for NurbsCurves and -Surfaces.

### Test files

[NurbsCurve.zip](https://github.com/BHoM/BHoM_Engine/files/5331849/NurbsCurve.zip)
[NurbSurface.zip](https://github.com/BHoM/BHoM_Engine/files/5331851/NurbSurface.zip)

### Additional comments

- There is a bug for un-clamped curves of degree 2 at their max parameter for the magnitude of derivatives. 
- Also that this is not very optimised. i.e. the basis functions are evaluated for points which impossibly can affect the parameter.
- Feel free to commit on spelling and general compliance

PR #2045, opened 2020-10-09
Geometry Engine: PointAtParameter(Polyline) sometimes returns null on parameter equal 1.0

### LINKED ISSUE #2043: Geometry_Engine: PointAtParameter(Polyline) sometimes return null when given parameter equals 1.0
#### Description:

Due to the computational error point at 1.0 parameter is sometimes outside the curve and then the method returns null.

#### Steps to reproduce:

The only known case occurred deep inside other method so it's hard to reproduce.

#### Expected behaviour:

Should return endpoint.

#### Test file(s):

As in Steps to reproduce.

LMarkowski:

Closes #2043 

Adding a quick fix to the PointAtParameter method for `Polyline` class which matches the one added once to the `PolyCurve` one.

Both will be probably replaced by using tolerance while resolving #2044.

### Test files

The issue itself occurred only once, deep inside another method so it's hard to isolate it to the specific test file.
[Here](https://burohappold.sharepoint.com/:f:/s/BHoM/Esd2eZj9Hc1Mkh5tar8m-NsBlIxJtyJ40zXatKwF4D3uOQ?e=aAy2n2) is the generic `PointAtParameter` test to ensure that the fix doesn't break anything. (Test for the Circle class need to be fixed, out of scope this PR)

### Changelog

- Modified `PointAtParameter` method for `Polyline` class in the `Geometry_Engine`
 
### Additional comments

This will unfortunately interfere with #2028 by editing the same file, but is necessary to resolve [#48](https://github.com/BHoM/TriangleNet_Toolkit/issues/48)


pawelbaran (approved):
LGTM, happy as an interim fix

pawelbaran:
I would consider adding same checks to `parameter == 0` for `Polyline` and `PolyCurve` too - @LMarkowski, does that make sense?

LMarkowski:
@pawelbaran hm, I can't see a case for which it could crash for `parameter == 0` but it could save as a tiny bit on computation time and make the code more transparent.
I'm pushing a commit then.

pawelbaran (changes requested):
Looks like you've pushed too much @LMarkowski 🙈

LMarkowski:
@pawelbaran, yes. Accidentally pushed all my stashed changes 🤦
fixed.

pawelbaran (approved):
LGTM as interim fix 👍

al-fisher (approved):
Makes sense as discussed

PR #2051, opened 2020-10-13
Reflection_Engine: Make sure outputs always have different names

### LINKED ISSUE #2050: Creating methods with Output<T1, T2> with same first letter and no MultiOutput does not work
#### Description:
Found during development of the MapSpaces method with @JohannaOlin1 

When we create a method with multiple outputs of the same type, the UI fails to load that component, saying that `index[1]` (and others) are out of range for the component.

On investigation, this seems to only happen when there are multiple output types that start with the same first letter, and no documentation `MultiOutput` attribute providing different output names. The result seems to be that the outputs get confused and thus become 1 output of the same name, rather than 2 or more. An example method which causes this can be seen below.

#### Steps to reproduce:
Create a method in any engine that looks like this - I recommend Environment_Engine to be able to copy and paste the below without many other changes - but the important parts are the return types in the `Output<>` should start with the same letter, even if they're different types. This example is using `BH.oM.Environment.Elements.Panel` and `BH.oM.Geometry.Pipe`.

```
public static Output<Panel, Pipe> Test()
{
    return new Output<Panel, Pipe>
    {
        Item1 = new Panel(),
        Item2 = new Pipe()
    };
}
```

When loading that method as a component in the UI, the component will only show one output `P`.
If you then add `[MultiOutput(0, "output1", "")]` and `[MultiOutput(1, "output2", "")]` to the top of the method, the component will then load correctly. The component will also load if only one `MultiOutput` tag is added - it merely needs to provide differentiation between the different types of `P` in the output from what I can see.

#### Expected behaviour:
I would expect to be able to create a method returning an `Output<>` without having to provide `MultiOutput` documentation for it to work in the UI.

adecler:

Closes #2050

See issue for details and testing procedure

### DIFF
--- a/Reflection_Engine/Query/OutputAttributes.cs
+++ b/Reflection_Engine/Query/OutputAttributes.cs
@@ -60,7 +60,14 @@ namespace BH.Engine.Reflection
                         outputs.Add(new OutputAttribute(outputDefs[i].Name, desc));
                     }
                     else
-                        outputs.Add(new OutputAttribute(types[i].UnderlyingType().Type.Name.Substring(0, 1), ""));
+                    {
+                        string name = types[i].UnderlyingType().Type.Name.Substring(0, 1);
+                        int nbSame = outputs.Where(x => x.Name.StartsWith(name)).Count();
+                        if (nbSame > 0)
+                            name += (nbSame + 1).ToString();
+                        outputs.Add(new OutputAttribute(name, ""));
+                    }
+                        
                 }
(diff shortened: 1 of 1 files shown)


JohannaOlin1 (approved):
image
Tried the same method from yesterday, now getting a component with multiple outputs.😄

FraserGreenroyd (approved):
LGTM, works as desired, thanks @adecler 😄

PR #2057, opened 2020-10-14
Reflection_Engine: Adding attributes to better control auto-create components

adecler:
### NOTE: Depends on 
https://github.com/BHoM/BHoM/pull/1020

See the oM PR for more details


IsakNaslundBh (approved):
Works well with other PRs.

The only comment I could give would be to add Descriptions, but will not let that hold up as Reflection_Engine needs more of those anyway if there is a plan to batch fix the engine at one point.

If not, good to add now 😄

PR #2062, opened 2020-10-15
Reflection_Engine: StoreAllMethods() to catch exception on individual method

### LINKED ISSUE #2061: Reflection_Engine: StoreAllMethods() to catch exception on individual method
#### Description:

While using Versioning, I realised I had a failure because `StoreAllMethods()` was throwing an exception when attempting to do `method.GetMethodBody()` on an assembly that could not load correctly. Failing that, no Constructor and no other methods were collected.

The assembly was `System.Threading.Thread.Extension, version 4.2.0.0`. I am still investigating as per why that was not loaded correctly, because the .dll is present in the BHoM folder. 

In any case, to add robustness, it would be a good idea to try-catch individual MethodInfo. 
This way at least the execution continues for valid methods.

alelom:

Closes #2061

While using Versioning, I realised it failed to work in a scenario where `StoreAllMethods()` was interrupted by an exception when attempting to do `GetMethodBody()`. The exception happened because of on an assembly that could not load correctly (see issue for details). In any case, a single assembly that cannot load shouldn't prevent the correct retrieval of methods from all other assemblies.

To add robustness, `StoreAllMethods()` can use a foreach and try-catch individual `MethodInfo`s. 
This way at least all valid methods are retrieved.


Reflection_Engine/Query/MethodList.cs
```diff
@@ -156,9 +156,29 @@ private static void StoreAllMethods(Type type)
...
+                foreach (var method in methods)
+                {
+                    try
+                    {
+                        if (method.GetMethodBody() != null && !method.IsAutoGenerated())
+                            m_AllMethodList.Add(method);
+                    }
+                    catch(Exception e)
+                    {
+                        string message = "Cannot load methods from type  " + type.Name + ". Exception message: " + e.Message;
```
adecler, line 171: The catch is for a single method so the error message should be the same.

adecler (changes requested):
Looking fine to me (except for the error message).

I don't have any `GetMethodBody` failing on my machine but the code makes sense, compiles, and doesn't break anything when tested.

Reflection_Engine/Query/MethodList.cs
alelom: Not sure I understand, however I corrected the text to say "Cannot load method" (singular, not plural), see commit.
The rest of the text seems fine to me?

adecler: Thanks but I meant actually telling the user which method cannot be loaded. Right now, they will only know it is from a given type.

adecler: I wouldn't require a change just for an extra `s` 😋

alelom: Oh I see! Done!

adecler:
Perfect, thanks 👍

adecler (approved):
LGTM

PR #2066, opened 2020-10-16
Reflection_Engine: Remove calls to Clone methods and Clean SetProperty methods

### LINKED ISSUE #2065: Reflection_Engine: Remove use of Clone methods
#### Broken rules:
Now that immutability is enforced on the level of the UI, we shouldn't be cloning our input in `Modify` methods. 

See https://github.com/BHoM/admin/issues/9 for more details

Delete the calls to Clone methods when possible

adecler:
### NOTE: Depends on 

One PR on the BHoM_UI depends on this: https://github.com/BHoM/BHoM_UI/pull/326
   
Closes #2065 

As discussed in https://github.com/BHoM/admin/issues/9, I have removed the cloning inside the only reflection method that was doing it.

Note that the BHoM_UI is already doing the cloning for the `Modify` and `SetProperty`. 

I have also cleaned the methods here because a lot of it was outdated legacy code:
- I have removed the `Modify.PropertyValue` because it was 
   - non-compliant in term of name
   - only working on BHoM object (for no reason since it was calling `SetProperty(this object,...)` anyways
   - was redundant with `SetProperty` method
- I have changed the output of `SetProperty` method to be compliant with the current state of the code. Note that that methods was not used anywhere in any repo. Ironically, we will later have all `Modify` methods to return void but it needs to be compliant until then. 
- Delete the `SetProperty` method that was taking list as first input. There was quite a few things that was incorrect in that method and, since it wasn't used anywhere and was not compliant for the UI, I decided to remove it. If you want a trip down to memory lane, this methods was migrated from teh BHoM_Adapter repo which is not using it anymore (see https://github.com/BHoM/BHoM_Engine/commit/1364a20af3443d13fddd73de76a752b5e6fd770e).

Make sure to also switch to the PR in BHoM_UI since the method name has changed from `PropertyName` to `SetPropertyName`. Because the `SetProperty` component is not a dynamic component, the change of name doesn't cause incompatibility with old saved components.

### Additional comments
A bit more changes that I initially intended but feels good to have that old legacy code finally cleaned up and compliant 😄

### DIFF
--- a/Reflection_Engine/Modify/SetPropertyValue.cs
+++ b/Reflection_Engine/Modify/SetPropertyValue.cs
@@ -60,6 +60,12 @@ namespace BH.Engine.Reflection
 
             if (prop != null)
             {
+                if (!prop.CanWrite)
+                {
+                    Engine.Reflection.Compute.RecordError("This property doesn't have a public setter so it is not possible to modify it.");
+                    return obj;
+                }
+
                 if (value.GetType() != prop.PropertyType && value.GetType().GenericTypeArguments.Length > 0 && prop.PropertyType.GenericTypeArguments.Length > 0)
                 {
                     value = Modify.CastGeneric(value as dynamic, prop.PropertyType.GenericTypeArguments[0]);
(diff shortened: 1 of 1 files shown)


alelom:
See https://github.com/BHoM/BHoM_UI/pull/326#issuecomment-709986186

adecler:
FYI, I had to do a rebase of the branch for CI to be happy

alelom (approved):
LGTM, works fine now:
https://github.com/BHoM/BHoM_UI/pull/326#pullrequestreview-511625477

PR #2070, opened 2020-10-17
Geometry_Engine: Replace calls to Clone, IClone, and GetShallowClone methods with calls to DeepClone and ShallowClone

### LINKED ISSUE #2067: Geometry_Engine: Replace calls to Clone, IClone, and GetShallowClone methods with calls to DeepClone and ShallowClone
#### Broken rules:
See https://github.com/BHoM/admin/issues/9 for details

- Replace calls to `Clone` and `IClone` to calls to `Engine.Base.DeepClone()`. Feel free to use `Engine.Base.ShallowClone()` where you feel appropriate.
- Replace calls to `GetShallowClone` with calls to `Engine.Base.ShallowClone()`
- Remove the `as X` conversion after the cloning since casting is not necessary anymore

adecler:

Closes #2067

Here's the procedure that I have followed:
- Replace calls to `Geometry.Clone` and `Geometry.IClone` to calls to `Engine.Base.DeepClone()`. Feel free to use `Engine.Base.ShallowClone()` where you feel appropriate.
- Replace calls to `GetShallowClone` with calls to `Engine.Base.ShallowClone()`
- Remove the `as X` conversion after the cloning since casting is not necessary anymore

### Test files
- Review the code itself to make sure it compiles and makes sense
- Run the standard testing procedure that involve the geometry engine

### Additional comments
After a few back and forth regarding the scope of this task, I have decided to keep it really simple for now. Turning the `Modify` methods into mutable methods has been pushed to later as it needs more coordination than expected. So I have instead focused on making sure that only the two methods `Engine.Base.DeepClone()` and `Engine.Base.ShallowClone()` are used for cloning. This solves the critical and original point of constant `BHoMGuid` which is what matters for now.

### DIFF
--- a/Geometry_Engine/Compute/BooleanDifference.cs
+++ b/Geometry_Engine/Compute/BooleanDifference.cs
@@ -20,6 +20,7 @@
  * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
  */
 
+using BH.Engine.Base;
 using BH.oM.Geometry;
 using BH.oM.Reflection.Attributes;
 using System;
@@ -37,7 +38,7 @@ namespace BH.Engine.Geometry
         public static List<Line> BooleanDifference(this Line line, Line refLine, double tolerance = Tolerance.Distance)
         {
             if (refLine.Length() <= tolerance)
-                return new List<Line> { line.Clone() };
+                return new List<Line> { line.DeepClone() };
 
             if (line.IsCollinear(refLine, tolerance))
             {
@@ -61,14 +62,14 @@ namespace BH.Engine.Geometry
                     Point aRPt = refLine.ControlPoints().Average();
 
                     if (aRPt.SquareDistance(splitLine[0]) > sqTol && aPt.SquareDistance(refLine) > sqTol)
-                        splitLine = new List<Line> { line.Clone() };
+                        splitLine = new List<Line> { line.DeepClone() };
                     else
                         splitLine = new List<Line>();
                 }
 
                 return splitLine;
             }
-            return new List<Line> { line.Clone() };
+            return new List<Line> { line.DeepClone() };
         }
 
         /***************************************************/
@@ -79,7 +80,7 @@ namespace BH.Engine.Geometry
 
             foreach (Line line in lines)
             {
-                List<Line> splitLine = new List<Line> { line.Clone() };
+                List<Line> splitLine = new List<Line> { line.DeepClone() };
                 int k = 0;
                 bool split = false;
                 do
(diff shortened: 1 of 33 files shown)


al-fisher (approved):
LGTM
Reviewing code - all makes sense.
Ran some geometry test scripts and tested a few specific methods.

@LMarkowski @pawelbaran - worth a quick look to check you are happy with the philosophy and get your quick approval too if possible

pawelbaran (commented):
LGTM code wise as a blanket-cover approach, I've also checked a few random methods. However, I would question whether it would not make sense to spend a bit of time on doing the changes properly, once we are doing them. Simply looking at the code:
- `LineIntersections` should be happy with shallow clone
- `IsCoplanar` should not need cloning at all
- should `SortCollinear` and `SortAlongCurve` clone all input points or simply sort them? My intuition says only sort, same as you sort collections
- `ClusterCollinear` same as above?

Another observation I just made: in case of returning same object as input (say, a bounding box of a bounding box), some `Query` methods return the original object, while some return its clone. Since we are switching to mutable `Modify` methods, the question is what should happen to those, i.e. how would one know when the original object (or its property) is returned and when it is a clone?

al-fisher:
Thanks @pawelbaran - your suggested improvements above do all make good sense.

However I would be happy to keep this set of PRs all focused on simply replacing existing calls to Clone to all be the correct Deep and Shallow versions.
As this will be the first of a number of PRs to close out https://github.com/BHoM/admin/issues/9 - so worry now about scope creep and making this a potentially large exercise. 
Once all instances of GetShallowClone are purged we are then good to revisit and make those surgical improvements. Might suggest raising specific separate issue so we do not forget already. 

Does that make sense @pawelbaran @adecler ?

al-fisher:
> Another observation I just made: in case of returning same object as input (say, a bounding box of a bounding box), some Query methods return the original object, while some return its clone. Since we are switching to mutable Modify methods, the question is what should happen to those, i.e. how would one know when the original object (or its property) is returned and when it is a clone?

In answer to this - if I understand your question correctly. 

A Query method should never return itself. You are asking for derived properties so in the edge case of BoundingBox of BoundingBox - the returned box should be a new (cloned) box.

Are there any other examples where this is the case and this does not make sense?

adecler:
Will now merge based on reviews above.

I agree with @al-fisher response to @pawelbaran. This PR is only meant to replace the clone methods in the safest and simplest way possible. An incorrect use of `ShallowClone` instead of `DeepClone` could potentially inject a new bug in the current state of the code while a `DeepClone` never hurts. Since we are going to refactor the `Modify` methods next quarter, a lot of those cloning calls will be removed anyways so no need to focus on minor performance gains right now.

pawelbaran:
@al-fisher I've raised https://github.com/BHoM/BHoM_Engine/issues/2100, I believe it would be good to capture your comment on `Query` never returning itself as one of the checkboxes that need to be ticked off (same as `Modify` never returning same type).

What is more, it would be good to emphasize whether returning e.g. property of or an item contained by the input object is OK or if it would also need to be cloned (think of `Line.EndPoint()` for the property or `List<Point>.ClosestPoint(Point`) for contained item - they both return uncloned objects).

PR #2080, opened 2020-10-19
Structure_Engine: Remove guid from comparers

### LINKED ISSUE #2029: Structure_Engine: Update object comparers to not rely on GUID
#### Broken rules:

BHoM_Guid is being used to check equality of objects in comparers. This need to be avoided as the GUID will be persistent across modify methods etc.

Remove any checking of the GUID for comparers.

IsakNaslundBh:

Closes #2029 

- Removes references to BHoM_Guids from comparers
- Additional null checks added
- Fixes wrong GetHashCode method used for the BarEndNodeDistanceComparer

### DIFF
--- a/BHoM_Engine/Objects/EqualityComparers/BHoMObjectNameComparer.cs
+++ b/BHoM_Engine/Objects/EqualityComparers/BHoMObjectNameComparer.cs
@@ -41,10 +41,6 @@ namespace BH.Engine.Base.Objects
             if (Object.ReferenceEquals(x, null) || Object.ReferenceEquals(y, null))
                 return false;
 
-            //Check if the GUIDs are the same
-            if (x.BHoM_Guid == y.BHoM_Guid)
-                return true;
-
             return x.Name == y.Name;
         }
 
--- a/BHoM_Engine/Objects/EqualityComparers/BHoMObjectNameOrToStringComparer.cs
+++ b/BHoM_Engine/Objects/EqualityComparers/BHoMObjectNameOrToStringComparer.cs
@@ -41,10 +41,6 @@ namespace BH.Engine.Base.Objects
             if (Object.ReferenceEquals(x, null) || Object.ReferenceEquals(y, null))
                 return false;
 
-            //Check if the GUIDs are the same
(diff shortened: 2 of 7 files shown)


adecler (commented):
Looks ok. 
My only problem is the change to `GetHashCode`. I know it is not there yet but this should use the `Base.Query.HashCode` that @alelom will soon add. 
In the mean time, why change to using the `GetHashCode` from the comparer ?

IsakNaslundBh:
> My only problem is the change to GetHashCode. I know it is not there yet but this should use the Base.Query.HashCode that @alelom will soon add.

Can not use the generic has in the comparers here. Basically, the `GetHashCode` method in an `IEqualityComparer` should return the same hash for two objects that are equal in the sense of the `Equal` method. This is 

- Objects are Equal -> `GetHashCode` should return the same hash
- `GetHashCode` is not equal -> objects are not equal
- `GetHashCode` is equal -> objects _might_ be Equal, but no guarantee that they are.

which leads to

> In the mean time, why change to using the GetHashCode from the comparer ?

There was actually a bug in the method. Should had been using the same EqualityComparer for getting out the hash as in the Equals method in the first place.

Should have put this in the description, sorry for missing that. Raised the PR just before I left yesterday.

Also fine with reverting that change and raising as a separate PR if you think it is too out of scope, but the change will need to happen as it is an old mistake by myself, that so far has not caused any issues (hence staying in for so long).

alelom:
Looks ok for me too.

> Also fine with reverting that change and raising as a separate PR if you think it is too out of scope, but the change will need to happen as it is an old mistake by myself, that so far has not caused any issues (hence staying in for so long).
> 
> EDIT Adding link to developer docs describing what I tried to describe above:
> 
> https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.equalitycomparer-1.gethashcode?view=netcore-3.1#System_Collections_Generic_EqualityComparer_1_GetHashCode__0_

Makes sense. For me this can stay.

alelom (approved):
LGTM.

IsakNaslundBh:
Had another chat with @adecler who said he agreed with the change, so will merge this!

Issue #2086, opened 2020-10-20
Spatial_Engine: There should be no Area method for IElement1D and IElement0D


IsakNaslundBh:
#### Broken rules:

We have methods for calculating the Area for an `IElement1D` and `IElement0D`. This has been shown to cause issues such as 

https://github.com/BHoM/BHoM_Engine/issues/2081
https://github.com/BHoM/BHoM_Engine/issues/2085

Unless we have a clear need to have an area method that works for any `IElement` I suggest we simply remove the methods for `IElement1D`, `IElement0D` as well as the interface method for `IElement` as it is royally confusing to call Area method as it stands in the Spatial_Engine, which is one of the core places we want to deal with geometry.

#### Suggestions to restore compliance:

Remove the `Area` methods for `IElement1D`, `IElement0D` as well as the interface method for `IElement`

Any objections to why this should not be done? @pawelbaran @al-fisher @LMarkowski @FraserGreenroyd @michaelhoehn

FraserGreenroyd:
No objections from me :smile:

pawelbaran:
Not sure if it is as confusing to me (in the end, it was me who added the methods), but I do not mind keeping `Area` only for `IElement2D`.

PR #2087, opened 2020-10-20
Spatial_Engine: Updating Area calls to Geometry_Engine instead of Spatial_Engine

### LINKED ISSUE #2085: Spatial_Engine: Wrong Area method used
#### Description:

Similar to https://github.com/BHoM/BHoM_Engine/issues/2081 the wrong Area method is used in a few more places in the Spatial_Engine, where the IElement1D method is reached instead of the Geometry method causing issues with the algorithm.

Methods affected:
- DistributeOutlines
- FreeFormProfile
- PointLayout

#### Steps to reproduce:

Use any of the above methods

#### Expected behaviour:

For the correct area method to be used.

IsakNaslundBh:

Closes #2085 

Fixing wrong area methods being called in Spatial_Engine. Putting as critical as this is currently stopping some quite key methods from at all working, so would be good to get this merged ASAP.

Even more incentive to get rid of the Area method for `IElement1D` as I am sure we will stumble upon this again in the future. Issue for it raised here:

https://github.com/BHoM/BHoM_Engine/issues/2086

before the above is resolved, resorting to explicit calls to the geometry engine.

### DIFF
--- a/Spatial_Engine/Compute/DistributeOutlines.cs
+++ b/Spatial_Engine/Compute/DistributeOutlines.cs
@@ -47,7 +47,7 @@ namespace BH.Engine.Spatial
             
             outlineCurves.Sort(delegate (Tuple<PolyCurve, List<IElement1D>> t1, Tuple<PolyCurve, List<IElement1D>> t2)
             {
-                return t1.Item1.Area().CompareTo(t2.Item1.Area());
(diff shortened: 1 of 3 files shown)


pawelbaran (approved):
LGTM

FraserGreenroyd (approved):
Code LGTM

Issue #1949, opened 2020-08-25
Base_Engine: implement GetHash method


alelom:
#### Description:

As discussed with @adecler, a way to get around the Hash generation difficulties posed by the needed serialisation would be to write the hash computation ourselves.

This could follow the structure of the method `GetValue()` used by `Compute.GetDummy` in the Test_Toolkit:
https://github.com/BHoM/Test_Toolkit/blob/6d230a0ef83bd5a5eebd8b625badb0e2cdd2802d/Test_Engine/Compute/DummyObject.cs#L92-L266

This means essentially to build a big recursive if-else method that returns a big string with all values relevant in generating the hash. This method will be also able to pick and choose certain properties if Exceptions are specified.

The resulting string can then be used to generate a byteArray and passed to the SHA256 hash algorithm as usual.

This would enable us to remove the dependancy on the serialisers from the Diffing_Engine and to make the Hash a core mechanism.

PR #2090, opened 2020-10-20
BHoM_Engine: add `Query.Hash()` extension method

### LINKED ISSUE #2060: BHoM_Engine: add `Query.Hash()` method
#### Description:

As title. This method should call the `Compute.Hash()` with default values.

alelom:

Closes #2060


BHoM_Engine/Query/Hash.cs
```diff
@@ -0,0 +1,57 @@
...
+    public static partial class Query
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/ 
+
+        [Description("Computes a Hash code for the object which depends solely on its properties. The full method with additional options is in `Compute.Hash()`.")]
+        [Input("obj", "Objects the hash code should be calculated for.")]
+        public static string Hash(
+            this IObject iObj)
```
adecler, line 49: The name of the parameters doesn't correspond to the one provided in the input attribute.

Also minor thing but wouldn't it be cleaner to have the method parameter on the same line as the rest ?

BHoM_Engine/Compute/Hash.cs
```diff
@@ -51,7 +51,7 @@ public static partial class Compute
         [Input("typeExceptions", "(Optional) e.g. `typeof(Guid)`. Any corresponding type is ignored.")]
         [Input("maxNesting", "(Optional) e.g. `100`. If any property is nested into the object over that level, it is ignored.")]
         public static string Hash(
-            this IObject iObj,
+            IObject iObj,
```
adecler, line 54: Is there a reason why this input has to be an `IObject` instead of an `object` ?
The `DefiningString` method is already taking in an `object` so might be better to enable it here too.

adecler (changes requested):
A few minor things but otherwise looks good.

I wonder if it wouldn't make more sense to have the `Hash` method that takes all the arguments to be in `Query` though. It kind of feels weird that one is in Query and one is in `Compute` since the only difference is their number of arguments.

BHoM_Engine/Compute/Hash.cs
alelom: Just for the record (after our discussion yesterday), `DefiningString()` works with `System.Object` because it is made to work recursively and deal with types that may be part (properties) of an `IObject`. It was developed following the structure of the method [`GetValue()`](https://github.com/BHoM/Test_Toolkit/blob/6d230a0ef83bd5a5eebd8b625badb0e2cdd2802d/Test_Engine/Compute/DummyObject.cs#L92-L266) used by `Compute.GetDummy` in the Test_Toolkit (as explained in https://github.com/BHoM/BHoM_Engine/issues/1949). It is therefore guaranteed to work with the limited amount of types included in it (which means, it should work fine with IObjects).  
Extending the parent caller method `Hash()` to object, however, would imply that the Hash could work with any object, which we cannot guarantee.

BHoM_Engine/Query/Hash.cs
alelom: Corrected the input description param.
I actually prefer the multiline parameters – find it much simpler to read – but if it's a matter of unblocking the PR I will change it.

BHoM_Engine/Query/Hash.cs
```diff
@@ -37,14 +37,23 @@
...
+        public static string Hash(this IObject iObj)
+        {
+            return Hash(iObj, null);
+        }
+
+        /***************************************************/
 
-        [Description("Computes a Hash code for the object which depends solely on its properties.")]
-        [Input("obj", "Objects the hash code should be calculated for.")]
+        [Description("Computes a Hash code for the iObject. The hash uniquely identifies the iObject based on its properties and their values.")]
```
al-fisher, line 55:
```suggestion
        [Description("Computes a Hash code for the iObject with exceptions. The hash uniquely identifies the iObject based on its included properties and their values.")]
```

al-fisher:
Looks good @alelom - added a small suggested change to one of the descriptions to distinguish

al-fisher (changes requested):
As discussed @alelom - unexpected behaviour between the two hash methods returning different results for same input
image

alelom:
@al-fisher thanks - fixed in https://github.com/BHoM/BHoM_Engine/pull/2090/commits/6ba007563de8f919914d69d4273712174a1d92cf

image

al-fisher:
Perfect - this is working as expected now! Thanks

BHoM_Engine/Query/Hash.cs
```diff
@@ -37,14 +37,24 @@
...
 {
-    public static partial class Compute
+    public static partial class Query
     {
         /***************************************************/
         /**** Public Methods                            ****/
-        /***************************************************/ 
+        /***************************************************/
+
+        [Description("Computes a Hash code for the iObject. The hash uniquely identifies the iObject based on its properties and their values.")]
```
al-fisher, line 46:
```suggestion
        [Description("Computes a Hash code for the iObject. The hash uniquely represents an object's state based on its combined properties and their values.")]
```

al-fisher: @alelom - also suggested change to description. Wonder if this use of the work "state" here will help going forward in explaining use vs guid?
Happy to get your alternate thoughts

BHoM_Engine/Query/Hash.cs
```diff
@@ -37,14 +37,24 @@
...
+        {
+            return Hash(iObj, null, null, null, null, 100);
+        }
 
-        [Description("Computes a Hash code for the object which depends solely on its properties.")]
-        [Input("obj", "Objects the hash code should be calculated for.")]
+        /***************************************************/
+
+        [Description("Computes a Hash code for the iObject with exceptions and custom options.\n"
+            + "The hash uniquely identifies the iObject based on the included properties and their values.")]
```
al-fisher, line 56:
```suggestion
            + "The hash uniquely represents an object's state based on the included properties and their values.")]
```

al-fisher: Sorry to mirror the other change pushed. Sorry 🙈 

Otherwise looking good now - and neat, to help explain the core principles behind hashes, guids etc. in BHoM going forward

adecler:
Isn't it a bit problematic that a call to `Query.Hash(item)` is actually undetermined between the two methods ?

alelom:
> Isn't it a bit problematic that a call to `Query.Hash(item)` is actually undetermined between the two methods ?

Sorry, I can't see where the problem is – the overload with the closest number of argument is always called by C#, so it's not undetermined?

Am I missing something?
image

al-fisher (approved):

al-fisher:
Thanks for mopping up those descriptions @alelom 😄

adecler (approved):

adecler:
@alelom , sorry I was just bothered by this: `return Hash(iObj, null, null, null, null, 100);`
The fact that you have to provide the default arguments to the method for it to be differentiated from `Hash(iObj)` is not ideal (for example, what if you change the default value of those arguments later ?).

But that's not a big deal at all so that's fine.

PR #2091, opened 2020-10-21
Environment_Engine: Add ability to group openings and panels by their type

### LINKED ISSUE #2047: Environment_Engine: Ability to group panels by their type in a multi output
We currently have a component which can split panels into those matching a type and those not matching a type, but this is limited to the type the user selects.

It would be nice and beneficial to model building to plug in a list of panels to a component, and have the output group the panels matching each available type, giving an output of maybe 10 types listing each panel that has that type.

Similar to the component from Honeybee below.

image

FYI: @rboulton-BH @charliebrookerBH @tg359 @vbeardow @LivAndersson @MarthaVoul

JohannaOlin1:
Closes #2047

Makes it possible to group a collection of panels or openings by their type.


Environment_Engine/Environment_Engine.csproj
```diff
@@ -139,6 +139,7 @@
     <Compile Include="Query\Absorptance.cs" />
     <Compile Include="Query\BottomLeft.cs" />
     <Compile Include="Query\BottomRight.cs" />
+    <Compile Include="Query\GroupingPanels.cs" />
```
FraserGreenroyd, line 142: File name should be `Group`

FraserGreenroyd (changes requested):

Environment_Engine/Environment_Engine.csproj
FraserGreenroyd: File name here needs to be `Group` too - so probably best to compile and recommit the CSProj changes :smile:

FraserGreenroyd (changes requested):

FraserGreenroyd (approved):
LGTM now, tack @JohannaOlin1 :smile:

PR #2093, opened 2020-10-21
Spatial_Engine: Add Query.IsNearGrid() method

### LINKED ISSUE #2055: Spatial_Engine: Is on Grid
#### Description:

Check if an element is on a particular grid, within a maximum distance.

Method accepting an IElement and a grid

IsakLarbornBH:

Closes #2055

Adds Query.IsNearGrid() method to Spatial_Engine. This method accepts an IElement, a Grid, and a double maxDistance as input, and tests the IElement for proximity to the Grid, returning true if the distance is lower than or equal to maxDistance.

### Test files

[IsOnGrid-Test.zip](https://github.com/BHoM/BHoM_Engine/files/5417078/IsOnGrid-Test.zip)

### Changelog

- `Query.IsNearGrid()` Query method added in the `Spatial_Engine` for `IElement` class


IsakNaslundBh (changes requested):
Generally looks great, including great descriptions!

Will test through tomorrow, in the mean time found a minor thing that needs fixing, see below:

Spatial_Engine/Query/IsOnGrid.cs
```diff
@@ -0,0 +1,103 @@
...
+        /******************************************/
+        /****   Public Methods - Interfaces    ****/
+        /******************************************/
+
+        [Description("Checks if the geometrical representation of an IElement, projected onto the XY Plane, is within a set distance from a grid line.")]
+        [Input("element", "The IElement that will be checked whether it is on Grid.")]
+        [Input("grid", "The Grid to use for evaulation.")]
+        [Input("maxDistance", "The maximum distance allowed from the Grid for this method to return true.", typeof(Length))]
+        [Output("isOnGrid", "A boolean which is true if the geometrical representation of the IElement is within a set distance from the grid line.")]
+        public static bool IsOnGrid(this IElement element, Grid grid, double maxDistance)
```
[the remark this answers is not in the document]
IsakLarbornBH: Ah, that makes a lot of sense! I added that method simply because it seemed like most other methods had one, while not asking anybody first and not being aware of the norms beyond that.

IsakNaslundBh (dismissed):
This looks good to me now!

Before merge though, just want to make sure that the base logic of this makes sense to @al-fisher @FraserGreenroyd and @pawelbaran .

This method will return true (the element is on the grid) if _any_ point on the element is within the maximum distance to any point on the gridline. For example, a Bar not aligned with the grid, but crossing it will return true. Same, a tilted wall/column with any part within distance of the grid will return true. 

To me, I think this makes sense, but wanted to get your points of view as well!

Well done @IsakLarbornBH !

Spatial_Engine/Query/IsOnGrid.cs
```diff
@@ -0,0 +1,103 @@
...
+        [Input("grid", "The Grid to use for evaulation.")]
+        [Input("maxDistance", "The maximum distance allowed from the Grid for this method to return true.", typeof(Length))]
+        [Output("isOnGrid", "A boolean which is true if the geometrical representation of an IElement2D is within a set distance from the grid line.")]
+        public static bool IsOnGrid(this IElement2D element2D, Grid grid, double maxDistance)
+        {
+            List<IElement1D> elements1D = element2D.IOutlineElements1D();
+
+            foreach (IElement1D e1D in elements1D)
+                if (IsOnGrid(e1D, grid, maxDistance))
+                    return true;
```
FraserGreenroyd, line 81: Convention in BHoM is for `{` and `}` on the outer loops where things like this can be done. So this should look like:

```
foreach (IElement1D e1D in elements1D)
{
    if (IsOnGrid(e1D, grid, maxDistance))
        return true;
}
```

FraserGreenroyd (changes requested):
Code change request - then will add to @IsakNaslundBh discussion on the other bit as I don't entirely agree, but don't disagree enough to hold up the PR beyond this code change.

FraserGreenroyd:
> This method will return true (the element is on the grid) if _any_ point on the element is within the maximum distance to any point on the gridline. For example, a Bar not aligned with the grid, but crossing it will return true. Same, a tilted wall/column with any part within distance of the grid will return true.

I would say this doesn't make a lot of sense given the name of the method and the description provided. The `IElement2D` method check allows for any edge to be on the grid and return true for the element being on the grid even if 75% of the element is not on the grid.

I don't know what the use case is which might have prompted this issue, so I'm happy for this to be merged as is (though perhaps a better description to make it clear it's _any part_ of the element and not the whole element), but I wonder if we could make it clearer through a different method name, or a change to the logic for the whole element rather than part of it.

Maybe it's a method called `IsOnGrid` which does as this PR offers, and checks whether _any part_ of an element is on the grid, and then a second method called `IsEntirelyOnGrid` which then checks if the whole of the element is on the grid?

IsakNaslundBh:
> I don't know what the use case is which might have prompted this issue, so I'm happy for this to be merged as is (though perhaps a better description to make it clear it's any part of the element and not the whole element), but I wonder if we could make it clearer through a different method name, or a change to the logic for the whole element rather than part of it.

This method first spawned from different checks we wanted to make available for floor slab panels, and then, forcing the full element to be on grid would render this method not useable, as it then only would be possible for perfectly vertical elements, i.e. walls.

What we wanted to be able to check for the panels in particular is basically "This panel lies between grid C and D and 7 and 8" kind of check, and then, checking closest proximity felt like the most straight forward way to do it.

There are ofc some middle ground to the two as well, like "Is at least one of the edges entirely on the grid" or similar, but that might just make things harder, not sure. Could potentially change this method name to `IsNearGrid` or similar, which might be a more accurate name when I think of it.

Spatial_Engine/Query/IsOnGrid.cs
IsakLarbornBH: Fixed!

FraserGreenroyd:
> There are ofc some middle ground to the two as well, like "Is at least one of the edges entirely on the grid" or similar, but that might just make things harder, not sure. Could potentially change this method name to `IsNearGrid` or similar, which might be a more accurate name when I think of it.

I think `IsNearGrid` is a good name, and more closely describes what the method is doing than `IsOnGrid` with the distance checks, etc. so I'd be happy with this :smile:

Spatial_Engine/Query/IsOnGrid.cs
```diff
@@ -0,0 +1,104 @@
...
+        /******************************************/
+        /****  Public Methods                  ****/
+        /******************************************/
+
+        [Description("Checks if the geometrical representation of an IElement0D, projected onto the XY Plane, is within a set distance from a grid line.")]
+        [Input("element0D", "The IElement0D that will be checked whether it is on Grid.")]
+        [Input("grid", "The Grid to use for evaulation.")]
+        [Input("maxDistance", "The maximum distance allowed from the Grid for this method to return true.", typeof(Length))]
+        [Output("isOnGrid", "A boolean which is true if the geometrical representation of the IElement0D is within a set distance from the grid line.")]
+        public static bool IsOnGrid(this IElement0D element0D, Grid grid, double maxDistance)
```
FraserGreenroyd, line 45: Change to `IsNearGrid` per discussion

Spatial_Engine/Query/IsOnGrid.cs
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        /******************************************/
+
+        [Description("Checks if the geometrical representation of an IElement1D, projected onto the XY Plane, is within a set distance from a grid line.")]
+        [Input("element1D", "The IElement1D that will be checked whether it is on Grid.")]
+        [Input("grid", "The Grid to use for evaulation.")]
+        [Input("maxDistance", "The maximum distance allowed from the Grid for this method to return true.", typeof(Length))]
+        [Output("isOnGrid", "A boolean which is true if the geometrical representation of an IElement1D is within a set distance from the grid line.")]
+        public static bool IsOnGrid(this IElement1D element1D, Grid grid, double maxDistance)
```
FraserGreenroyd, line 60: Change to `IsNearGrid` per discussion

Spatial_Engine/Query/IsOnGrid.cs
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        /******************************************/
+
+        [Description("Checks if the geometrical representation of an IElement2D, projected onto the XY Plane, is within a set distance from a grid line.")]
+        [Input("element2D", "The IElement2D that will be checked whether it is on Grid.")]
+        [Input("grid", "The Grid to use for evaulation.")]
+        [Input("maxDistance", "The maximum distance allowed from the Grid for this method to return true.", typeof(Length))]
+        [Output("isOnGrid", "A boolean which is true if the geometrical representation of an IElement2D is within a set distance from the grid line.")]
+        public static bool IsOnGrid(this IElement2D element2D, Grid grid, double maxDistance)
```
FraserGreenroyd, line 75: Change to `IsNearGrid` per discussion

Spatial_Engine/Query/IsOnGrid.cs
```diff
@@ -0,0 +1,104 @@
...
+        /******************************************/
+        /****   Public Methods - Interfaces    ****/
+        /******************************************/
+
+        [Description("Checks if the geometrical representation of an IElement, projected onto the XY Plane, is within a set distance from a grid line.")]
+        [Input("element", "The IElement that will be checked whether it is on Grid.")]
+        [Input("grid", "The Grid to use for evaulation.")]
+        [Input("maxDistance", "The maximum distance allowed from the Grid for this method to return true.", typeof(Length))]
+        [Output("isOnGrid", "A boolean which is true if the geometrical representation of the IElement is within a set distance from the grid line.")]
+        public static bool IIsOnGrid(this IElement element, Grid grid, double maxDistance)
```
FraserGreenroyd, line 96: Change to `IIsNearGrid` per discussion

FraserGreenroyd (changes requested):

IsakNaslundBh:
> I think IsNearGrid is a good name, and more closely describes what the method is doing than IsOnGrid with the distance checks, etc. so I'd be happy with this 😄

Cool, yes, lets update to that then! :)

IsakNaslundBh (approved):
Now all looks fine to me! @FraserGreenroyd Happy as well?

FraserGreenroyd (approved):
Ja jag ar lycklig, tack @IsakNaslundBh och @IsakLarbornBH

PR #1805, opened 2020-05-26
Reflection_Engine: Adding IsAutoExplode methods

IsakNaslundBh:
### NOTE: Depends on 

https://github.com/BHoM/BHoM/pull/894
   
### Issues addressed by this PR

Closes #1804

Add method checking if a returntype should be automatically exploded.

PR #2098, opened 2020-10-21
Reflection_Engine: Extraction of MemberInfo/ParameterInfo description without type description enabled

### LINKED ISSUE #2097: Reflection_Engine: Allow extraction of MemberInfo/ParameterInfo description without type description
#### Description:

Easy to do by adding a defaulted Boolean.

pawelbaran:

Closes #2097

### Test files

Not needed I think. Rather a simple change, seems not to break things in GH (one simply needs to remember about rebuilding BHoM_UI and Grasshopper_Toolkit @IsakNaslundBh 😆), but would be good to double check by more educated ones.

### Additional comments

This PR should not conflict neither with #1805 nor with #2062.

I would like to get the approval of @adecler before this gets merged - just to make sure we do not blow up the UI 😉

### DIFF
--- a/Reflection_Engine/Query/Description.cs
+++ b/Reflection_Engine/Query/Description.cs
@@ -38,8 +38,9 @@ namespace BH.Engine.Reflection
         /**** Public Methods                            ****/
         /***************************************************/
 
+        [PreviousVersion("4.0", "BH.Engine.Reflection.Query.Description(System.Reflection.MemberInfo)")]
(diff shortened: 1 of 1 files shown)


adecler (commented):
The code change itself is fine but I am not sure I understand with the principle behind this PR.
The whole point of having centralised methods like `Description` is to guarantee a consistent behaviour across all UIs. So having additional input parameters for such a method goes against that principle.

adecler (approved):
Based on our discussion, I am happy to approve this PR.

PR #2099, opened 2020-10-22
Geometry_Engine: Addition of Query.Volume methods for ISolids

### LINKED ISSUE #2046: Geometry_Engine: Addition of Query.Volume() methods for ISolids
#### Description:

Following https://github.com/BHoM/BHoM/issues/1016 - this can be exposed additonally here, but along with new methods or all other `BH.oM.Geometry.ISolids` based on analytical solutions

al-fisher:

Closes #2046 

### Test files

See test file here:
https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?RootFolder=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FTest%20Scripts%2FBHoM%5FEngine%2FGeometry%5FEngine%2FGeometryEngine%2D%232046%2DVolumeQueryForISolids&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4

### Changelog

- Volume for BoundaryRepresentation is a straight exposure of the Get only Volume Property stored on the object and set at conversion
- All other ISolid volumes are derived from the analytical formulae comprised from defining properties.

### DIFF
--- a/Geometry_Engine/Geometry_Engine.csproj
+++ b/Geometry_Engine/Geometry_Engine.csproj
@@ -247,6 +247,7 @@
     <Compile Include="Modify\Orient.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
     <Compile Include="Query\PointAtX.cs" />
+    <Compile Include="Query\Volume.cs" />
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\BHoM_Engine\BHoM_Engine.csproj">
--- a/Geometry_Engine/Query/Volume.cs
+++ b/Geometry_Engine/Query/Volume.cs
@@ -33,20 +33,6 @@ namespace BH.Engine.Geometry
     public static partial class Query
     {
         /***************************************************/
-        /**** Public Methods - Interfaces               ****/
-        /***************************************************/
-
-
-        [Description("Gets the enclosed volume of a solid.")]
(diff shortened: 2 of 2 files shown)


al-fisher:
@michaelhoehn FYI - this completes, I think, the geometrical foundations for your latest LCA work. 
Speak later on all things related in any case 😄

IsakNaslundBh (changes requested):
Have not tested, but formulas look correct!

Some small things I might comment on:

1. I have generally been putting the interface method at the bottom of the file. This is not consistent across all engine files, but I think it is the most common place for them. Would not block for this, but think we need to agree on a consistent convention for this.
2. We should add a fallbackmethod, a private method `Volume(ISolid solid)` that raises an error message and returns double.NaN. Similar to what has been done to the `Length` methods for curves in this engine. This I think we should get in!

al-fisher:
Thanks @IsakNaslundBh 
Both good comments - will address 

On this
> I have generally been putting the interface method at the bottom of the file. 

Agree actually - think this is the better convention. 👍

michaelhoehn (approved):
Tested and works great!

IsakNaslundBh (approved):
Code now looks good, and relying on @michaelhoehn s test of functionality!

PR #2105, opened 2020-10-27
Diffing_Engine: clarification of concept of Hash vs Diff; leverage persistentId

### LINKED ISSUE #2095: Diffing_Engine: diffing to leverage the new IPersistentId and IAdapterId fragment interfaces
#### Description:

As explained in https://github.com/BHoM/BHoM/issues/1037.

### LINKED ISSUE #2139: Diffing_Engine: formalise distinction between Diffing and Hash.
#### Description:

- Current `HashFragment` that stores both current/old hash must be renamed to `RevisionFragment`. It was designed to be used with a "diffing-revision workflow" anyway (so you can track objects back in time and tell who is new/modified/old).
- Introduce new `HashFragment` that only stores the **current** hash. This is to be used for "diffing-current state workflow".
- Review the `IDiffing()` method:
   - to make sure that by default computes `HashFragment` for objects and uses that (do a "diffing-current state");
   - if instead a RevisionFragment is found on the objects, do a "diffing-revision";
   - make sure that in the "diffing-current state", the property-by-property diffing is off by default (priority must be speed).
   - make sure that in the "diffing-revision", the property-by-property diffing is true by default (priority must be complete comparison).
   
Also consider:
- Introduce new HashConfig? To be used as input to the `BH.Engine.Base.Query.Hash()` method instead of the long list of inputs.
- Allow conversion/mapping between DiffConfig-HashConfig? 
- Allow conversion between RevisionFragment and HashFragment?

@rolyhudson @al-fisher

alelom:
### NOTE: Depends on 

https://github.com/BHoM/BHoM/pull/1084
   
Closes #2095
Closes #2139


alelom:
@BHoMBot check core

Analytical_Engine/Analytical_Engine.csproj
```diff
@@ -75,7 +75,7 @@
     <Compile Include="Convert\Graph\ToDataVizObjects.cs" />
     <Compile Include="Convert\Graph\ToDotFormat.cs" />
     <Compile Include="Convert\Graph\ToRelation.cs" />
-    <Compile Include="Create\Graph\DiffConfig.cs" />
+    <Compile Include="Create\Graph\DistinctConfig.cs" />
     <Compile Include="Create\Graph\Graph.cs" />
```
rolyhudson, line 79: This file should be ComparisoConfig

Analytical_Engine/Analytical_Engine.csproj
```diff
@@ -95,12 +95,12 @@
     <Compile Include="Query\ExternalPolycurve.cs" />
     <Compile Include="Query\Geometry.cs" />
     <Compile Include="Query\Graph\Adjacency.cs" />
+    <Compile Include="Query\Graph\DistinctEntities.cs" />
     <Compile Include="Query\Graph\EntityNeighbourhood.cs" />
```
rolyhudson, line 99: Need to rename this CompareEntities

Analytical_Engine/Query/Graph/DistinctEntities.cs
```diff
@@ -43,36 +43,27 @@ public static partial class Query
...
-
-            //find decimal places from numeric tolerance
-            int decimalPlaces = DecimalPlaces(diffConfig.NumericTolerance);
-
-            //generate hashes
-            entities.ForEach(ent => objectHash.Add(ent, ent.HashEntity(diffConfig, decimalPlaces)));
+            HashComparer<object> hashComparer = new HashComparer<object>(cc);
 
             foreach (KeyValuePair<IBHoMObject, string> entityA in objectHash)
             {
```
rolyhudson, line 58: Probably I am ignorant of how to use the HashComparer. So maybe I missing something but here no comparisons take place.

rolyhudson (commented):
From the Graph point of view looks like this will work and the changes in naming make a lot of sense. One thing I am missing in the comparison of Graph entities is the generation of hashes. 
Could be just needing to loop the entities passed into the method? Rather than the Dict of hashes in the old version.

Which I may have misunderstood.

Analytical_Engine/Query/Graph/DistinctEntities.cs
alelom: Comparison should happen on line 65 (`if (hashComparer.Equals(entityA.Value, entityB.Value))`) by calling Equals.

Behaviour in my mind should've been the same as before, at least that was my intention when I refactored the code. However I did not test this function (`DistinctEntities`) but I did test the `HashComparer.Equals` and that works.

However I did expect I might have done something wrong here, so happy to set up a call to see together what the problem might be 😄

Analytical_Engine/Analytical_Engine.csproj
alelom: Actually I think the name could stay as it is (DistinctEntities) because you do not only compare the entities, but the end result of the function is getting the unique entities (output is a Dictionary).

So either rename it to UniqueEntities() or leave it as DistinctEntities(), but I would avoid the name "compare" as not fully descriptive of the function.

FraserGreenroyd:
@BHoMBot check compliance

Analytical_Engine/Create/Graph/ComparisonConfig.cs
```diff
@@ -1,4 +1,5 @@
-﻿using BH.oM.Diffing;
+﻿using BH.oM.Base;
```
FraserGreenroyd, line 1: Needs copyright header please

BHoM_Engine/Create/IObject/RandomObject.cs
```diff
@@ -215,7 +215,11 @@ private static object GetValue(Type type, Random rnd, int depth)
             else if (type.IsInterface || type.IsAbstract)
             {
                 if (depth > 50) return null;
-                return GetValue(m_ImplementingTypes[type], rnd, depth + 1);
+
+                Type obj = null;
+                if (!m_ImplementingTypes.TryGetValue(type, out obj)) return null;
```
FraserGreenroyd, line 220:
```suggestion
                if (!m_ImplementingTypes.TryGetValue(type, out obj))
                    return null;
```

FraserGreenroyd: For consistency of style across the BHoM.

BHoM_Engine/Query/Hash.cs
```diff
@@ -173,53 +192,56 @@ private static string SHA256Hash(string str)
...
+                    else 
+                        propertyPath += "." + prop.Name;
 
-                    if (isInPropertyNameExceptions || isInPropertyFullNameExceptions)
+                    bool isInPropertyExceptions = cc.PropertyExceptions?.Count > 0 && cc.PropertyExceptions.Where(ex => new WildcardPattern(ex).IsMatch(propertyPath)).Any();
+                    if (isInPropertyExceptions)
                         continue;
 
+                    //if (cc.PropertyNamesToConsider?.Count() > 0 && !cc.PropertyNamesToConsider.Any(pn => (propertyPath).Contains(pn))) //!cc.PropertyNamesToConsider.Contains(prop.Name))
+                    //    continue;
```
FraserGreenroyd, line 213: If this isn't needed it should be deleted

Diffing_Engine/Compute/_IDiffing.cs
```diff
@@ -43,7 +43,7 @@ namespace BH.Engine.Diffing
     public static partial class Compute
     {
         [Description("Dispatches to the most appropriate Diffing method, depending on the provided inputs.")]
-        public static Diff Diffing(IEnumerable<object> pastObjs, IEnumerable<object> followingObjs, DiffConfig diffConfig = null, string customDataIdKey = null)
+        public static Diff Diffing(IEnumerable<object> pastObjs, IEnumerable<object> followingObjs, DiffingType diffingType = DiffingType.Automatic, DiffingConfig diffConfig = null)
```
FraserGreenroyd, line 46: File name probably shouldn't be starting with an `_` - is there a reason why it does @alelom? I think removing `_` will allow compliance to be happy, as it takes into account `I` files I think.

Diffing_Engine/Compute/_IDiffing.cs
```diff
@@ -52,62 +52,99 @@ public static Diff Diffing(IEnumerable<object> pastObjs, IEnumerable<object> fol
...
 
-            // Check if the BHoMObjects all have a hashfragment assigned.
-            // If so, we may attempt the Revision diffing.
-            if (bHoMObjects_past.AllHaveHashFragment() && bHoMObjects_following.AllHaveHashFragment())
-            {
-                BH.Engine.Reflection.Compute.RecordNote($"Calling the diffing method '{nameof(DiffRevisionObjects)}'.");
-                return DiffRevisionObjects(pastObjs, followingObjs, diffConfigCopy);
-            }
+            // Check if the bhomObjects have a persistentId assigned.
+            List<object> reminder_past, reminder_following;
```
FraserGreenroyd, line 119:
```suggestion
            List<object> reminder_past;
            List<object> reminder_following;
```

For consistency with BHoM practices.

Diffing_Engine/Create/Delta.cs
```diff
@@ -45,10 +45,10 @@ public static partial class Create
         [Description("Returns a Delta object with the Diff between the two input Revisions, also called `Diff-based Delta`.")]
         [Input("previousRevision", "A previous Revision")]
```
FraserGreenroyd, line 46: Should probably be `pastRevision` by the looks of the method?

Diffing_Engine/Modify/RemoveDuplicatesByHash.cs
```diff
@@ -33,16 +33,21 @@
...
 using BH.oM.Reflection.Attributes;
 using System.ComponentModel;
+using BH.Engine.Base;
 
 namespace BH.Engine.Diffing
 {
     public static partial class Modify
     {
-        [Description("Removes duplicates from a collection of objects. The comparison is made through their Diffing Hash.")]
+        [Description("Removes duplicates from a collection of objects. The comparison is made using their Hash. If hash is missing, it is computed.")]
```
FraserGreenroyd, line 42:
```suggestion
        [Description("Removes duplicates from a collection of objects. The comparison is made using their Hash. If the hash is missing, it will be computed.")]
```

FraserGreenroyd: Will the hash be stored on the object that is returned? (Apologies if that's explained somewhere else - but if the answer is yes, probably good to tell the user that in the description as well.)

Diffing_Engine/Query/ListModifiedProperties.cs
```diff
@@ -39,8 +39,8 @@ public static partial class Query
 
         [MultiOutput(0, "identifier", "Identifier of the objects which have some modified properties.\nWhen using Revisions, this is the Hash of the objects. When Diffing using CustomData, this is the specified Id.")]
         [MultiOutput(1, "propNames", "List of properties changed per each object.")]
-        [MultiOutput(2, "value_Current", "List of current values of the properties.")]
-        [MultiOutput(3, "value_Past", "List of past values of the properties.")]
+        [MultiOutput(2, "value_obj1", "List of current values of the properties.")]
+        [MultiOutput(3, "value_obj2", "List of past values of the properties.")]
```
FraserGreenroyd, line 43:
```suggestion
        [MultiOutput(2, "valueObj1", "List of current values of the properties.")]
        [MultiOutput(3, "valueObj2", "List of past values of the properties.")]
```

FraserGreenroyd, line 43:
```suggestion
        [MultiOutput(2, "currentValues", "List of current values of the properties.")]
        [MultiOutput(3, "pastValues", "List of past values of the properties.")]
```

FraserGreenroyd: This suggestion is more compliant.

FraserGreenroyd: This suggestion is more helpful to the user I think in terms of what the outputs are (based on their descriptions)

Physical_Engine/Query/UniqueConstructions.cs
```diff
@@ -43,22 +44,20 @@ public static partial class Query
...
-                PropertiesToIgnore = new List<string>
-                {
-                    "BHoM_Guid",
-                    "CustomData",
-                },
-                NumericTolerance = BH.oM.Geometry.Tolerance.Distance,
+                PropertyExceptions = new List<string>
+                    {
+                        "CustomData"
+                    },
```
FraserGreenroyd, line 52:
```suggestion
                PropertyExceptions = new List<string>
                {
                    "CustomData"
                },
```

FraserGreenroyd (commented):
Any compliance result which starts `Modify methods must return the same type as their first parameter` can be ignored owing to the soon to be implemented change that Modify methods will return void in 4.1.

The rest should try to be addressed as appropriate, however if they are not in your engine (diffing?) then it is ok to ignore them.

On the whole this is looking good, it just needs aligning with the requested changes on the oM and then some of the changes requested here :smile:

Great work though @alelom, by no means easy looking at some of the code!

FraserGreenroyd (changes requested):
Hit submit before hitting request changes 🤦‍♂️

Diffing_Engine/Compute/_IDiffing.cs
alelom: Oh, I was using it simply to have the file on top in my folder in VS. Sorry I didn't notice it was a problem. Removing it.

Diffing_Engine/Modify/RemoveDuplicatesByHash.cs
alelom: The idea is: if it's not stated anywhere and it's not obvious from the method name, then it's not done 😄

FraserGreenroyd:
Just checking the new annotations workflow @alelom as a heads up! :smile:

alelom:
@IsakNaslundBh just to confirm that the computed Hash for the same objects did not change with this PR. I used GSA - PushPullCompare to check:

<details>
All good for Node, Bar and MaterialFragments:

image

Not sure about:
image

image

<\details>

I think this should be sufficient to determine that the Datasets are ok.

alelom:
@rolyhudson a change I made to simplify the code had adverse impact on the performance, I should've thought of that. Just leaving it for the record:
image
Once brought back to what it was:
image
as restored performance to the same order of magnitude as before.

rolyhudson (dismissed):
Looks good. Ran the graph performance test script and it is fast. Happy to help work through the performance issue with the next round of upgrades and additional control to define property comparisons.

alelom:
Regarding my comment above https://github.com/BHoM/BHoM_Engine/pull/2105#issuecomment-734297995

I've raised an issue: https://github.com/BHoM/BHoM_Engine/issues/2193

alelom:
@BHoMBot check compliance

alelom:
@BHoMBot check compliance

alelom:
@BHoMBot check installer

FraserGreenroyd (approved):
I'm happy with what I've seen, I would hope others are too before merge :smile:

alelom:
@BHoMBot checkinstaller

alelom:
@BHoMBot check installer

PR #2107, opened 2020-10-28
BHoM_Engine: Added global & property-specific fractional precision in `Query.Hash()`

### LINKED ISSUE #2103: BHoM_Engine: add options to set numerical tolerance in `Query.Hash()`
#### Description:

Use case from @rolyhudson :
> I've been using the DifferentProperties() method on a fairly complex structure with 670 nodes - with occasional tiny errors in node position that I need to ignore - diff with tolerances would be very helpful

Should be an easy addition to the current method.
Consider:
- a new parameter `numericTolerance` that applies globally to all numbers
- also to provide another parameter `Dictionary<string,double> roundingPerProperty` that allows to set a rounding per specific property

### LINKED ISSUE #2104: BHoM_Engine: consider adding wildcard/Regex to the property name identification to `Query.Hash()`
#### Description:

Right now, to identify a property (e.g. in `propertyNameExceptions`) you can only specify the exact name of a property to be excluded.

It could be useful to add the same mechanism that we have for the File_Toolkit where you can specify wildcards (e.g. Windows-explorer-style asterisks `*`) and/or C# Regexes to the input text when looking for specific properties.  

This way we could input things like e.g.:
- BH.oM.*.Material.*.Name to find all Names of any material
- *.Position.* to specify any position coordinate.
- etc

Perhaps this could be limited to the asterisk wildcard to avoid the complexity that may result from allowing both `*` wildcard and Regexes (see File_Toolkit - https://github.com/BHoM/Filing_Toolkit/pull/61).

Also, this can initially be limited to the new `roundingPerProperty` parameter (see https://github.com/BHoM/BHoM_Engine/issues/2103).  
Then later we can evaluate whether to aggregate `propertyNameExceptions`, `propertyFullNameExceptions` and `namespaceExceptions` into a same parameter by leveraging this mechanism.

alelom:

Closes #2103 
Closes #2104

### Test files

https://burohappold.sharepoint.com/:f:/s/BHoM/EqKWsGEj8cNCuWm1KsWSMSIBiRBvjfLusj572G0zq62exA?e=9AdD5w

### DIFF
--- a/BHoM_Engine/BHoM_Engine.csproj
+++ b/BHoM_Engine/BHoM_Engine.csproj
@@ -52,6 +52,7 @@
     <Reference Include="System" />
     <Reference Include="System.Core" />
     <Reference Include="System.Drawing" />
+    <Reference Include="System.Management.Automation" />
     <Reference Include="System.Xml.Linq" />
     <Reference Include="System.Data.DataSetExtensions" />
     <Reference Include="Microsoft.CSharp" />
--- a/BHoM_Engine/Query/Hash.cs
+++ b/BHoM_Engine/Query/Hash.cs
@@ -34,6 +34,7 @@ using System.ComponentModel;
 using BH.Engine.Base;
 using System.Collections;
 using System.Data;
+using System.Management.Automation;
 
 namespace BH.Engine.Base
 {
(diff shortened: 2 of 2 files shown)


rolyhudson (commented):
Nice. Tested on the supplied test file and on comparing 2056 Structural Nodes on the use case I am working with. Works as expected. ++ Fast 👍 👍 
Is there a reference for the wildcard usage? It would be useful to include it in the description. Not sure for other users but it was not clear how a single * can be used to match a series of other characters. Otherwise great.

rolyhudson (approved):
Nice work!

PR #2108, opened 2020-10-28
MEP_Engine: Remove CustomData

### LINKED ISSUE #2075: MEP_Engine: Remove use of CustomData in the code
The files listed below have been detected as having code using `CustomData`. This detection has been automatic, and may have picked up on code comments or documentation. Please review each file carefully and remove all instances of `CustomData`.

This should be completed by the 30th October 2020.

<details>
<summary>Full list</summary>
<p>

 - [x] BHoM_Engine\MEP_Engine\Query\EquipmentLengthByPart.cs

</p>
</details>

michaelhoehn:

Closes #2075 

### Additional comments

- I am unsure what this method was using prior to the changes as the description references a dataset of airside parts by length that I am not familiar with. I changed the input requirements to `IFlow` object, from which `Length()` can be called from any object. @kayleighhoude @FraserGreenroyd Please let me know if this approach seems logical and useful, or if I'm missing a potentially larger picture here 😄


MEP_Engine/Query/EquipmentLengthByPart.cs
```diff
@@ -49,14 +35,21 @@ public static partial class Query
         [Description("Returns the length of the equipment based on the input of a list of parts")]
         [Input("airsidePartsByLength", "Collection of airside parts from a dataset that contains parts by length")]
         [Output("length", "The total length of the equipment based on the length of the parts")]
-        public static double EquipmentLengthByPart (this List<CustomObject> airsidePartsByLength)
+        public static double EquipmentLengthByPart (this List<IFlow> airsidePartsByLength)
```
FraserGreenroyd, line 38: Looks fine to me, just needs some versioning I think :smile:

FraserGreenroyd (changes requested):

kayleighhoude:
@michaelhoehn I'm afraid this would only work if we had implemented the IFlow interface on the mechanical equipment objects. The idea was to plug in a series of cooling coils and fans (sections of an AHU) and sum the lengths.

michaelhoehn:
> @michaelhoehn I'm afraid this would only work if we had implemented the IFlow interface on the mechanical equipment objects. The idea was to plug in a series of cooling coils and fans (sections of an AHU) and sum the lengths.

@kayleighhoude yeah understood. I was thinking that the method was ideally looking for `Parts` instead of `Elements`, but currently there's no capacity for calculating the length for the `Parts` objects that have no location information or length properties. Which is why CustomData was being used. So how would we like to resolve this?

We could edit `IParts` to include a length property?

FraserGreenroyd:
Coming back to the original problem then, the use of CustomData in this method needs removing, @kayleighhoude would you be happy with us just removing the method if it hasn't yet made its way to the workflows it was intended?

michaelhoehn:
I support removal as well if this isn't being used. Happy to revisit after the upcoming namespace conversation and further building of the MEP_oM.

kayleighhoude:
Agreed on removal @michaelhoehn @FraserGreenroyd

MEP_Engine/Query/EquipmentLengthByPart.cs
michaelhoehn: Version = deleted

FraserGreenroyd (approved):
LGTM, thanks @michaelhoehn

PR #2113, opened 2020-10-29
Physical_Engine: BulkMaterials Volume Query

### LINKED ISSUE #2112: Physical_Engine: Volume Query Method for BulkMaterials
#### Description:

For use in other toolkits making use of BulkMaterials.

michaelhoehn:

Closes #2112


Physical_Engine/Query/MaterialComposition.cs
```diff
@@ -68,6 +69,21 @@ public static MaterialComposition MaterialComposition(this ISurface surface)
...
+        [Input("bulkMaterial", "The BulkMaterial to get the MaterialComposition from")]
+        [Output("materialComposition", "The kind of matter the BulkMaterial is composed of and in which ratios", typeof(Ratio))]
+        public static MaterialComposition MaterialComposition(this BulkMaterial bulkMaterial)
+        {
+            if (bulkMaterial.MaterialComposition == null)
+            {
+                Engine.Reflection.Compute.RecordError("The BulkMaterial MaterialComposition could not be calculated as no Materials have been assigned.");
+                return null;
+            }
+            return Matter.Create.MaterialComposition(bulkMaterial.MaterialComposition().Materials, bulkMaterial.MaterialComposition().Ratios);
```
IsakNaslundBh, line 84: Can just return the property here? (also think the current implementation will lead to stack-overflow if I am not missing something).

Might want to keep the warning as well, if usefull.

```suggestion
            return bulkMaterial.MaterialComposition;
```

If you want to create a new material composition, should at least change to:

```suggestion
            if (bulkMaterial.MaterialComposition == null)
            {
                Engine.Reflection.Compute.RecordError("The BulkMaterial MaterialComposition could not be calculated as no Materials have been assigned.");
                return null;
            }
            return Matter.Create.MaterialComposition(bulkMaterial.MaterialComposition.Materials, bulkMaterial.MaterialComposition.Ratios);
```

To avoid the stack overflow.

IsakNaslundBh (changes requested):
Tweaks to the material composition method.

Physical_Engine/Query/MaterialComposition.cs
michaelhoehn: Thanks @IsakNaslundBh, indeed the last commit led to stack-overflow. I've changed to simply use the properties to work as expected.

michaelhoehn:
Closing this draft PR as this is dependant and needed to be re-branched. I initially intended for the PRs to take place in sequence, but find them easier to pair together.

PR #2115, opened 2020-10-30
Spatial_Engine: Find top/ bottom centreline of IFramingElement

### LINKED ISSUE #2079: Physical_Engine: Add Query methods returning the centerline of BH.oM.Physical.Elements.Beam aligned to top/bottom
#### Description:

As in the title, `TopCenterLine` and `BottomCenterLine` would be a nice to have, particularly for the purpose of Revit interaction (at least until https://github.com/BHoM/BHoM/issues/935 and https://github.com/BHoM/BHoM/issues/936 get properly resolved).

FYI @JosefTaylor

annastigenberg:

Closes #2079

### Test files

[Spatial_Engin#2115_Test File.zip](https://github.com/BHoM/BHoM_Engine/files/5466069/Spatial_Engin.2115_Test.File.zip)

### Additional comments

Created two methods to query an IFramingElement of its top/ bottom centreline.


.github/CODEOWNERS
```diff
@@ -1,30 +0,0 @@
-# BHoM_Engine REPO CODEOWNERS
```
IsakNaslundBh, line 1: Not sure what happened here, but these files should not be deleted.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,27 +0,0 @@
-<!-- PLEASE ENSURE YOU REVIEW THE CONTENT OF EACH PR CAREFULLY, INCLUDING SUBSEQUENT COMMENTS BY YOURSELF OR OTHERS. -->
```
IsakNaslundBh, line 1: Not sure what happened here, but these files should not be deleted.

Physical_Engine/Query/BottomCentreline.cs
```diff
@@ -0,0 +1,65 @@
...
+            Vector normal;
+            
+            if(location.IIsLinear())
+            { 
+                normal = BH.Engine.Physical.Query.Normal(element);
+            }
+            else
+            {
+                Engine.Reflection.Compute.RecordError("IFramingElement must have linear location line");
+                normal = null;
```
IsakNaslundBh, line 39: Would return null here instead, letting it run will just let it crash on line 50

Physical_Engine/Query/BottomCentreline.cs
```diff
@@ -0,0 +1,65 @@
...
+
+            Vector normal;
+            
+            if(location.IIsLinear())
+            { 
+                normal = BH.Engine.Physical.Query.Normal(element);
+            }
+            else
+            {
+                Engine.Reflection.Compute.RecordError("IFramingElement must have linear location line");
```
IsakNaslundBh, line 38: Could be worth specifying a bit more what method is failing. This method could be called from somewhere else, and then having an error message that refers more to were things went wrong can be helpful.
```suggestion
                Engine.Reflection.Compute.RecordError("Can only extract bottom centrelines from linear IFramingElements.");
```

Physical_Engine/Query/TopCentreline.cs
```diff
@@ -0,0 +1,69 @@
...
+            
+            if(location.IIsLinear())
+            { 
+                normal = BH.Engine.Physical.Query.Normal(element);
+            }
+            else
+            {
+                Engine.Reflection.Compute.RecordError("IFramingElement must have linear location line");
+                normal = null;
+            }
```
IsakNaslundBh, line 40: Would either change to retuning null in the else statement (as above) or change to something like

```suggestion
            normal = BH.Engine.Physical.Query.Normal(element);
            
            if(normal == null)
                return null;
```

This would make use of the checks and warnings in the normal method, and return null if it failed.

Not saying this have to be changed to this though, what you have with suggested changes to the other file works fine.

Physical_Engine/Query/BottomCentreline.cs
```diff
@@ -0,0 +1,65 @@
...
+            {
+                height = ((IConvertible)propValue).ToDouble(null);
+
+                ICurve bottomCentreline = location.ITranslate(normal * -0.5 * height);
+
+                return bottomCentreline;
+            }
+            else
+            {
+                Engine.Reflection.Compute.RecordError("Failed extracting height");
```
IsakNaslundBh, line 56: Same as the above error message. Good if this could be expanded a bit to include why failing to extract the height was a problem, as in, what were you trying to do, but were unable to do because of the fact that you failed to extract the height.

Physical_Engine/Query/BottomCentreline.cs
```diff
@@ -0,0 +1,65 @@
...
+            }
+            else
+            {
+                Engine.Reflection.Compute.RecordError("IFramingElement must have linear location line");
+                normal = null;
+            }
+
+            double height = 0;
+           
+            object propValue = element.PropertyValue("Property.Profile.Height");
```
IsakNaslundBh, line 44: mostly as an FYI here. Think this case actually can be ok for extracting height like this, using the PropertyValue method, but should be used with quite a bit of care as it is significantly slower than straight access to the properties. Again, think this might be one of the corner cases that it is ok to use it, given how many levels of abstraction you have to go through, but FYI. 😄

Physical_Engine/Query/BottomCentreline.cs
```diff
@@ -0,0 +1,65 @@
...
+using System.Linq.Expressions;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace BH.Engine.Physical
+{
+    public static partial class Query
+    {
+
+        [Description("Returns the bottom centreline of an IFramingElement")]
```
IsakNaslundBh, line 22: Really picky comment, but try to end the descriptions with a `.` 

Could also expand a bit, short, how this is done, as in describing what methodology you are using.

Physical_Engine/Query/BottomCentreline.cs
```diff
@@ -0,0 +1,65 @@
...
+        [Input("BHoM.Physical.Elements.IFramingElement", "The IFramingEment to query the bottom centreline of")]
```
IsakNaslundBh, line 23: Input descriptions should have their first argument as the name of the parameter rather than the type. (also needs a dot)
```suggestion
        [Input("element", "The IFramingEment to query the bottom centreline of.")]
```

Physical_Engine/Query/BottomCentreline.cs
```diff
@@ -0,0 +1,65 @@
...
+        [Output("ICurve", "The bottom centreline of the IFramingElement")]
```
IsakNaslundBh, line 24: Would change the name of the output to something closer to the name of a parameter (also needs a dot)

```suggestion
        [Output("curve", "The bottom centreline of the IFramingElement.")]
```

Physical_Engine/Query/TopCentreline.cs
```diff
@@ -0,0 +1,69 @@
...
+using System.Threading.Tasks;
+
+namespace BH.Engine.Physical
+{
+    public static partial class Query
+    {
+
+        [Description("Returns the top centreline of an IFramingElement")]
+        [Input("BHoM.Physical.Elements.IFramingElement", "The IFramingEment to query the top centreline of")]
+        [Output("ICurve", "The top centreline of the IFramingElement")]
```
IsakNaslundBh, line 24: Same comments as descriptions above

IsakNaslundBh (changes requested):
Ok, sorry for what looks like an absolute bombardment of comments, most of it is descriptions and similar. 

Think the general gist of it looks good and correct and good! 👍

.github/CODEOWNERS
FraserGreenroyd: Seconded

.github/PULL_REQUEST_TEMPLATE.md
FraserGreenroyd: Seconded

Physical_Engine/Query/BottomCentreline.cs
```diff
@@ -0,0 +1,65 @@
+﻿using BH.Engine.Geometry;
```
FraserGreenroyd, line 1: File needs copyright header

Physical_Engine/Query/BottomCentreline.cs
FraserGreenroyd: How many centrelines can a framing element have? Surely it should be `Can only extra bottom centreline from linear IFramingElements.` - singular `centreline`?

Physical_Engine/Query/BottomCentreline.cs
FraserGreenroyd: The level of abstraction is only 3 layers, I thought all `IFramingElements` had a property called `Property` which would then make this easier to just do `double height = element.Property.Profile.Height`?

FraserGreenroyd, line 44: I would also not call this `propValue` - it isn't a very descriptive variable name for what you're planning to get out of this, irregardless of which method you use to obtain it.

Physical_Engine/Query/TopCentreline.cs
```diff
@@ -0,0 +1,69 @@
+﻿using BH.Engine.Geometry;
```
FraserGreenroyd, line 1: Copyright header needed

Physical_Engine/Query/TopCentreline.cs
```diff
@@ -0,0 +1,69 @@
...
+            {
+                Engine.Reflection.Compute.RecordError("Failed extracting height");
+
+                return null;
+            }
+            
+         
+
+            
+
```
FraserGreenroyd, line 64: Delete these blank lines

Physical_Engine/Query/TopCentreline.cs
```diff
@@ -0,0 +1,69 @@
...
+            }
+            else
+            {
+                Engine.Reflection.Compute.RecordError("IFramingElement must have linear location line");
+                normal = null;
+            }
+
+            double height = 0;
+           
+            object propValue = element.PropertyValue("Property.Profile.Height");
```
FraserGreenroyd, line 44: Same comment as above regarding variable name, and data extraction.

FraserGreenroyd (changes requested):

Physical_Engine/Query/BottomCentreline.cs
annastigenberg: @FraserGreenroyd I think since different types of IFramingElements have different properties writing element.Property.Profile.Height will not work, I did not get it to work anyways.

FraserGreenroyd: Just double checked the code, it seems all `IFramingElement` have an `IFramingElementProperty` but that interface does not enforce a `IProfile` property. However, we only have one object which implements `IFramingElementProperty` which is `ConstantFramingElement`. @IsakNaslundBh should `IFramingElementProperty` enforce an `IProfile` property?

If not then, this approach is fine, but I would probably go for a quicker convert than the `if`/`else` you've done instead.

IsakNaslundBh: >  @IsakNaslundBh should IFramingElementProperty enforce an IProfile property?

No, can not. need to be more flexible.

Also, all profiles do not have a `Height` property. Look at tube and circle profiles for example.

Discussed offline with @annastigenberg that a solution around it would be to add `FullHeight` query methods for both `IFramingProperty` as well as `IProfile`. Something that I am not necessarily against, but for sure expands the scope of this task.

Another way to do it, that just came to mind, is to check if the property is a `ConstantFramingProeprty`, if so, extract the profile, then extract the edges of the profile, then check the bounds of them, and then you can get the height from the difference in Y coordinates of the min and the max.

Might actually be a better interim solution than relying on the reflection in PropertyValue

FraserGreenroyd: > Another way to do it, that just came to mind, is to check if the property is a ConstantFramingProeprty, if so, extract the profile, then extract the edges of the profile, then check the bounds of them, and then you can get the height from the difference in Y coordinates of the min and the max.

I think this is a good fix for this PR, then agree a new issue raised and further work on `FullHeight` queries - which I agree is out of scope for this PR.

annastigenberg:
Closing and reopening this PR due to issues with OneDrive.

Physical_Engine/Query/BottomCentreline.cs
annastigenberg: The definition of the IFramingElement has an ICurve and an IFramingElementProperty, which in its turn does not have ConstantFramingProperty. Therefore not sure how to or if it is possible to access the edges of the profile.

IsakNaslundBh: The IFramingElementProperty can _be_ a ConstantFramingProperty.

What you can check is something like

```
If(element.Property is ConstantFramingProperty)
{
     ConstantFramingProperty constantProperty = element.Property as ConstantFramingProperty;
IProfile profile = constantProperty.Profile;

......
}
else
{
//Raise warning stating this is only wokring for constatn proeprties
}
```
Sure I have some typos in the above, but that is the principles of it.

PR #2118, opened 2020-11-02
Physical_Engine: Query the top/bottom centreline of an IFramingElement

### LINKED ISSUE #2079: Physical_Engine: Add Query methods returning the centerline of BH.oM.Physical.Elements.Beam aligned to top/bottom

annastigenberg:

Closes #2079 

[https://github.com/BHoM/BHoM_Engine/pull/2115](url)

### Test files

[Spatial_Engin#2115_Test File.zip](https://github.com/BHoM/BHoM_Engine/files/5475190/Spatial_Engin.2115_Test.File.zip)

### Additional comments

Creating two query methods returning the top/ bottom centreline of an IFramingElement.

### DIFF
--- a/Physical_Engine/Query/BottomCentreline.cs
+++ b/Physical_Engine/Query/BottomCentreline.cs
@@ -45,12 +45,11 @@ namespace BH.Engine.Physical
         [Description("Returns the bottom centreline of an IFramingElement.")]
         [Input("element", "The IFramingElement to query the bottom centreline of.")]
         [Output("curve", "The bottom centreline of the IFramingElement.")]
-
         public static ICurve BottomCentreline(this BH.oM.Physical.Elements.IFramingElement element)
         {
             ICurve location = element.Location;
 
-            Vector normal;
+            Vector normal = null;
 
             try
             {
@@ -88,7 +87,7 @@ namespace BH.Engine.Physical
             }
             else
             {
-                Engine.Reflection.Compute.RecordError("Element does not have ConstantFramingProperty, was not able to compute section height.");
+                Engine.Reflection.Compute.RecordError("Element does not have a suitable framing property, so the section height could not be calculated.");
                 return null;
             }
         }
--- a/Physical_Engine/Query/BottomCentreline.cs
+++ b/Physical_Engine/Query/BottomCentreline.cs
@@ -1,0 +1,84 @@
+/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2020, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 
+ * GNU Lesser General Public License for more details.                          
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using BH.Engine.Geometry;
+using BH.Engine.Reflection;
+using BH.Engine.Spatial;
+using BH.oM.Base;
+using BH.oM.Geometry;
+using BH.oM.Physical.FramingProperties;
+using BH.oM.Reflection.Attributes;
+using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Configuration;
+using System.Linq;
+using System.Linq.Expressions;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace BH.Engine.Physical
+{
+    public static partial class Query
+    {
+
+        [Description("Returns the bottom centreline of an IFramingElement.")]
+        [Input("element", "The IFramingElement to query the bottom centreline of.")]
+        [Output("curve", "The bottom centreline of the IFramingElement.")]
+
+        public static ICurve BottomCentreline(this BH.oM.Physical.Elements.IFramingElement element)
+        {
+            ICurve location = element.Location;
+
+            Vector normal;
+
+            normal = BH.Engine.Physical.Query.Normal(element);
+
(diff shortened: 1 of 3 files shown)


Physical_Engine/Query/BottomCentreline.cs
```diff
@@ -0,0 +1,96 @@
...
+
+namespace BH.Engine.Physical
+{
+    public static partial class Query
+    {
+
+        [Description("Returns the bottom centreline of an IFramingElement.")]
+        [Input("element", "The IFramingElement to query the bottom centreline of.")]
+        [Output("curve", "The bottom centreline of the IFramingElement.")]
+
```
FraserGreenroyd, line 48: Delete this blank line

Physical_Engine/Query/BottomCentreline.cs
```diff
@@ -0,0 +1,96 @@
...
+        public static ICurve BottomCentreline(this BH.oM.Physical.Elements.IFramingElement element)
+        {
+            ICurve location = element.Location;
+
+            Vector normal;
```
FraserGreenroyd, line 53: Default this to `null` to be explicit

Physical_Engine/Query/BottomCentreline.cs
```diff
@@ -0,0 +1,96 @@
...
+
+                double height = profileMax.Y - profileMin.Y;
+
+                ICurve bottomCentreline = location.ITranslate(normal * -0.5 * height);
+
+                return bottomCentreline;
+            }
+            else
+            {
+                Engine.Reflection.Compute.RecordError("Element does not have ConstantFramingProperty, was not able to compute section height.");
```
FraserGreenroyd, line 91: In the future, further framing properties might be added. To avoid this error needing updating every time, I'd advise we go for `Element does not have a suitable framing property, so the section height could not be calculated`.

Physical_Engine/Query/TopCentreline.cs
```diff
@@ -0,0 +1,89 @@
...
+using System.Threading.Tasks;
+
+namespace BH.Engine.Physical
+{
+    public static partial class Query
+    {
+        [Description("Returns the top centreline of an IFramingElement.")]
+        [Input("element", "The IFramingElement to query the top centreline of.")]
+        [Output("curve", "The top centreline of the IFramingElement.")]
+
```
FraserGreenroyd, line 47: Delete this blank line

Physical_Engine/Query/TopCentreline.cs
```diff
@@ -0,0 +1,89 @@
...
+        public static ICurve TopCentreline(this BH.oM.Physical.Elements.IFramingElement element)
+        {   
+            ICurve location = element.Location;
+
+            Vector normal;
```
FraserGreenroyd, line 52: Default this to `null` as above

Physical_Engine/Query/TopCentreline.cs
```diff
@@ -0,0 +1,89 @@
...
+
+                double height = profileMax.Y - profileMin.Y;
+
+                ICurve topCentreline = location.ITranslate(normal * -0.5 * height);
+
+                return topCentreline;
+            }
+            else
+            {
+                Engine.Reflection.Compute.RecordError("Element does not have ConstantFramingProperty, was not able to compute section height.");
```
FraserGreenroyd, line 84: As above

FraserGreenroyd (changes requested):
Do we not have a `CentreLine` query method for an `IFramingElement`? If not, maybe it's worth adding one, and then calling that and modifying it slightly to get the top/bottom? Cause at the moment, I'm seeing a lot of duplicated code across both methods that we could tidy by doing the common bits in another method?

Otherwise this is looking great @annastigenberg :smile: sorry you've ended up with one with lots of discussion!

Physical_Engine/Query/BottomCentreline.cs
annastigenberg: *Commenting your main comment*

At least in the Revit to BHoM workflow when pulling IFramingElements from Revit the location line is always the elements centreline regardless of how it is defined in Revit. A tapered element will have a tilted location line a.s.o.

FraserGreenroyd: That makes sense, but in a BHoM workflow we usually wouldn't mind too much about Revit at this stage of the code (even if those are the workflows in mind). @IsakNaslundBh what do you think regarding a method for `Centreline` of an `IFramingElement`? I'm mostly looking at a reduction of duplicated code, but if there's no benefit then this is acceptable :smile:

IsakNaslundBh: As it stands right now, and as mentioned by @annastigenberg , the Location property of the FramingElements is the centreline already. We have had discussion of adding in offsets, insertionPoints and similar additional properties to the FramingElements, but until that is done, there is no need for a CentreLine method.

Not sure was the initial comment was about though?

Physical_Engine/Query/TopCentreline.cs
```diff
@@ -0,0 +1,88 @@
...
+            try
+            {
+                normal = BH.Engine.Physical.Query.Normal(element);
+            }
+            catch
+            {
+                Engine.Reflection.Compute.RecordError("IFramingElement must have linear location line.");
+                return null;
+            }
+
```
IsakNaslundBh, line 62: As you have done for the BottomCentreLine, this would need a null check for the normal vector.

Physical_Engine/Query/BottomCentreline.cs
```diff
@@ -0,0 +1,95 @@
...
+
+namespace BH.Engine.Physical
+{
+    public static partial class Query
+    {
+
+        [Description("Returns the bottom centreline of an IFramingElement.")]
+        [Input("element", "The IFramingElement to query the bottom centreline of.")]
+        [Output("curve", "The bottom centreline of the IFramingElement.")]
+        public static ICurve BottomCentreline(this BH.oM.Physical.Elements.IFramingElement element)
```
IsakNaslundBh, line 48: Details here, but would personally add a `using` of `BH.oM.Physical.Elements` so you can shorten the input here to just be `IFramingElement`. Not as critical for a single parameter method like this, but still think that is clearer.

Same for the other method.

IsakNaslundBh (changes requested):
Additional nullcheck and some minor details

FraserGreenroyd (approved):
Ok given the discussion on the centreline, I'm happy with this as is, though accept some code is duplicated for now, we can always refactor later rather than hold up for too long :smile:

IsakNaslundBh (approved):
Agree with @FraserGreenroyd comments, now looks good to me!

PR #2120, opened 2020-11-03
BHoM_Engine: Correctly assign CustomObject properties like Name and Tags

### LINKED ISSUE #1981: BHoM_Engine: Name is saved in CustomData instead of replacing Name property. Same for other properties
Can't find object properties to input on my custom objects.

not a custom object:
image

CreateCustom object:
image

adecler:

Closes #1981

Assigning properties like `Name`, `BHoM_Guid`, `Tags`, and `Fragments` now tries to assign it to the corresponding property (if type is valid) instead of directly into the `CustomData` dictionary.

### Test files

image

### DIFF
--- a/BHoM_Engine/Create/CustomObject.cs
+++ b/BHoM_Engine/Create/CustomObject.cs
@@ -22,6 +22,7 @@
 
 using BH.oM.Base;
 using System.Collections.Generic;
+using System.Linq;
 
 namespace BH.Engine.Base
 {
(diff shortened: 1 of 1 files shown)


BHoM_Engine/Create/CustomObject.cs
```diff
@@ -44,21 +45,89 @@ public static CustomObject CustomObject(Dictionary<string, object> data, string
...
             }
             else
                 throw new System.Exception("The list of property names must be the same length as the list of property values when creating a Custon object.");
 
-            return new CustomObject
-            {
-                CustomData = customData,
-                Name = name
-            };
+            if (name != "")
```
pawelbaran, line 72: Shouldn't it be `if (!string.IsNullOrWhiteSpace(name))`?

pawelbaran (dismissed):
Tested, seems to be working fine, LGTM 👍 One question/remark in the code comment - not sure if relevant, please feel free to ignore.

BHoM_Engine/Create/CustomObject.cs
adecler: It is safer, yes. Now corrected.

pawelbaran (approved):

PR #2124, opened 2020-11-04
Reflection_Engine: Adding check for Fragments in the DistinctProperties

### LINKED ISSUE #1250: Reflection_Engine: DistinctProperties to look for fragments
#### Description:

Once the fragments have been updated to the FragmentSet (https://github.com/BHoM/BHoM/pull/574), the DistinctProperties method can start looking through the fragments as well for any property of the specified type.

Will increase flexibility in the DependencyObjects and allow toolkits to add fragments to obejcts and still rely on the standard behaviour of the adapter.

#### Comments:

@rwemay think this can be a neat way to finally solve being able to handle for example FramingElementDesignProperties in Robot that I know you asked for quite some time ago.

FYI @JosefTaylor @peterjamesnugent @adecler @FraserGreenroyd

IsakNaslundBh:

Closes #1250 

Making DistinctProperties check if P is a fragment type, and if so, try to fetch it from the fragment List of the objects.

### DIFF
--- a/Reflection_Engine/Query/DistinctProperties.cs
+++ b/Reflection_Engine/Query/DistinctProperties.cs
@@ -59,6 +59,7 @@ namespace BH.Engine.Reflection
             // Get the list of properties corresponding to type P
             Dictionary<Type, List<PropertyInfo>> propertyDictionary = typeof(T).GetProperties().GroupBy(x => x.PropertyType).ToDictionary(x => x.Key, x => x.ToList());
             List<PropertyInfo> propertiesSingle, propertiesList, propertiesGroups;
+            bool isFragment = typeof(IFragment).IsAssignableFrom(typeof(P));
             bool singleExist, listExists;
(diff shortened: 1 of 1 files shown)


adecler (commented):
Looks good to me on the side of the code. 

That will obviously need to be reviewed from the perspective of the Adapter (as it is the only part of the code calling this method).

Not a deal breaker but it is good to keep in mind that this PR creates a inconsistent behaviour among the property accessor methods in the reflection Engine as it is the only one that now treats fragment as a primary property. 

This, for example, raises the question on what should be done on the level of the `Explode` component. This is an important one because all objects of sa same type are not guaranteed to have the same fragments on them (as opposed to a guaranteed list of properties). So would potentially need a reevaluation of the warning system in `Explode` as well.

FraserGreenroyd:
@BHoMBot check project compliance

FraserGreenroyd:
@BHoMBot check code compliance

alelom (approved):
Approving from the Adapter perspective.

PR #2132, opened 2020-11-06
Reflection_Engine: Code updated to the changes introduced in BHoM #1068

pawelbaran:
### NOTE: Depends on 

https://github.com/BHoM/BHoM/pull/1068
   
Closes https://github.com/BHoM/BHoM/issues/1067

### Test files

Create one of the components that take inputs with quantities, e.g. `BH.Engine.Data.Query.ClosestData`, then check if the input descriptions are OK.

### Changelog

- code updated to the changes made in https://github.com/BHoM/BHoM/pull/1068

### Additional comments

This PR will not conflict with #2124.

### DIFF
--- a/Reflection_Engine/Query/Description.cs
+++ b/Reflection_Engine/Query/Description.cs
@@ -223,6 +223,18 @@ namespace BH.Engine.Reflection
 
 
         /***************************************************/
+        /**** Fallback Methods                          ****/
+        /***************************************************/
+
+        [Description("Fallback returning an empty string in case a type-specific Description method is missing for a given subtype of InputClassificationAttribute.")]
+        [Input("classification", "Input classification attribute to be queried for description.")]
+        private static string Description(this InputClassificationAttribute classification)
+        {
+            return "";
+        }
+
+
+        /***************************************************/
         /**** Private Methods                           ****/
         /***************************************************/
 
--- a/Reflection_Engine/Query/Description.cs
+++ b/Reflection_Engine/Query/Description.cs
@@ -205,7 +205,7 @@ namespace BH.Engine.Reflection
 
         /***************************************************/
 
-        [Description("Return the custom description of a folder path attribute.")]
+        [Description("Return the description of a folder path attribute.")]
         [Input("folderPath", "Folder path attribute to be queried for description.")]
         public static string Description(this FolderPathAttribute folderPath)
         {
@@ -214,7 +214,7 @@ namespace BH.Engine.Reflection
 
         /***************************************************/
 
-        [Description("Return the custom description of a file path attribute.")]
+        [Description("Return the description of a file path attribute.")]
         [Input("filePath", "File path attribute to be queried for description.")]
         public static string Description(this FilePathAttribute filePath)
         {
--- a/Reflection_Engine/Query/Description.cs
+++ b/Reflection_Engine/Query/Description.cs
@@ -43,7 +43,7 @@ namespace BH.Engine.Reflection
         public static string Description(this MemberInfo member, bool addTypeDescription = true)
         {
             DescriptionAttribute descriptionAttribute = member.GetCustomAttribute<DescriptionAttribute>();
-            QuantityAttribute quantityAttribute = member.GetCustomAttribute<QuantityAttribute>();
+            InputClassificationAttribute classification = member.GetCustomAttribute<InputClassificationAttribute>();
 
             string desc = "";
             if (descriptionAttribute != null && !string.IsNullOrWhiteSpace(descriptionAttribute.Description))
@@ -51,7 +51,7 @@ namespace BH.Engine.Reflection
 
             if (addTypeDescription && member is PropertyInfo && (typeof(IObject).IsAssignableFrom(((PropertyInfo)member).PropertyType)))
             {
-                desc += ((PropertyInfo)member).PropertyType.Description(quantityAttribute) + Environment.NewLine;
+                desc += ((PropertyInfo)member).PropertyType.Description(classification) + Environment.NewLine;
             }
 
             return desc;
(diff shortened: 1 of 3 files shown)


Reflection_Engine/Query/Description.cs
```diff
@@ -185,6 +185,43 @@ public static string IDescription(this object item)
...
+            return "This is a " + quantity.GetType().Name + " [" + quantity.SIUnit + "]";
+        }
+
+        /***************************************************/
+
+        [Description("Return the custom description of a folder path attribute.")]
+        [Input("folderPath", "Folder path attribute to be queried for description.")]
+        public static string Description(this FolderPathAttribute folderPath)
+        {
+            return "This is a folder path.";
```
FraserGreenroyd, line 212: I thought it was an elephant?

Reflection_Engine/Query/Description.cs
```diff
@@ -185,6 +185,43 @@ public static string IDescription(this object item)
...
+        [Description("Return the custom description of a quantity attribute.")]
+        [Input("quantity", "Quantity attribute to be queried for description.")]
+        public static string Description(this QuantityAttribute quantity)
+        {
+            return "This is a " + quantity.GetType().Name + " [" + quantity.SIUnit + "]";
+        }
+
+        /***************************************************/
+
+        [Description("Return the custom description of a folder path attribute.")]
```
FraserGreenroyd, line 208: "custom"

"This is a folder path."

One of these is not like the other... :wink: can we update the description to be accurate to the code? :smile:

Reflection_Engine/Query/Description.cs
```diff
@@ -185,6 +185,43 @@ public static string IDescription(this object item)
...
+        /***************************************************/
+
+        [Description("Return the custom description of a file path attribute.")]
+        [Input("filePath", "File path attribute to be queried for description.")]
+        public static string Description(this FilePathAttribute filePath)
+        {
+            return "This is a file path.";
+        }
+
+
```
FraserGreenroyd, line 224: We need a fall back method in case someone later adds a `InputClassificationAttribute` but does not add the method here to get the description, otherwise `IDescription` will fall into a stack overflow.

Reflection_Engine/Query/OutputAttributes.cs
```diff
@@ -54,8 +54,8 @@ public static List<OutputAttribute> OutputAttributes(this MethodBase method)
                         if (types[i] != null)
                         {
                             desc += Environment.NewLine;
-                            QuantityAttribute quantityAttribute = outputDefs[i].Quantity;
-                            desc += types[i].UnderlyingType().Type.Description(quantityAttribute);
+                            InputClassificationAttribute clasification = outputDefs[i].Classification;
```
FraserGreenroyd, line 57: Should be `classification` for the variable name - 2 `s` :smile:

FraserGreenroyd (changes requested):

Reflection_Engine/Query/Description.cs
pawelbaran: _custom_ keyword removed 🙈

Reflection_Engine/Query/OutputAttributes.cs
pawelbaran: Fixed

Reflection_Engine/Query/Description.cs
pawelbaran: Please correct me if I am wrong, but I think the above should be possible taken that `InputClassificationAttribute` is abstract?

FraserGreenroyd: The case still remains that later, someone can make a new class that implements/inherits `InputClassificationAttribute` that does not have a method here. The bit I've put wrong cause this was one of the last PRs I looked at yesterday so was quite tired, is that `IDescription` will fall into a stack overflow. It won't, but it will fail if it cannot find a suitable method for `InputClassificationAttribute as dynamic`. So probably just something like this should work.

```
public static string Description(this InputClassificationAttribute attribute)
{
    return ""; //Default fall back case
}
```

If it helps, it is BHoM compliance to provide a fall back method in situations where we're doing this, it's just one of the compliance things BHoMBot currently doesn't check so has to be done manually.

FraserGreenroyd (changes requested):

pawelbaran:
> If it helps, it is BHoM compliance to provide a fall back method in situations where we're doing this, it's just one of the compliance things BHoMBot currently doesn't check so has to be done manually.

Yeah, I agree that the code may crash, but that is the case for many Engine methods that are not implemented for all possible types. A fallback method could be desirable, but I am not sure if it is better to return invalid description (empty string) or let it crash and consequently force the developer to spend 2 mins on adding a type-specific implementation of `Description` method?

IsakNaslundBh:
> > If it helps, it is BHoM compliance to provide a fall back method in situations where we're doing this, it's just one of the compliance things BHoMBot currently doesn't check so has to be done manually.
> 
> Yeah, I agree that the code may crash, but that is the case for many Engine methods that are not implemented for all possible types. A fallback method could be desirable, but I am not sure if it is better to return invalid description (empty string) or let it crash and consequently force the developer to spend 2 mins on adding a type-specific implementation of `Description` method?

Have to think of the case that this is not caught in the PR though. This would crash on trying to generate the component, which potentially could lead to a full crash of the UI.

Say for example someone adds another InputClassificationAttribute in Plaxis_Toolkit for whatever reason and adds it to a property, then it will crash out here.

Ofc, this should be handled by whoever adds the attribute, but agree with @FraserGreenroyd , we most certainly need a fallback method here.

FraserGreenroyd:
> Have to think of the case that this is not caught in the PR though. This would crash on trying to generate the component, which potentially could lead to a full crash of the UI.
> 
> Say for example someone adds another InputClassificationAttribute in Plaxis_Toolkit for whatever reason and adds it to a property, then it will crash out here.
> 
> Ofc, this should be handled by whoever adds the attribute, but agree with @FraserGreenroyd , we most certainly need a fallback method here.

@IsakNaslundBh answered before I could - we cannot guarantee (no matter how much we might wish to force the developer to do something) that a developer will do something. There is high risk that code in the oM, or a toolkit like Plaxis, is merged without any update here that causes breakages across the UI.

> Yeah, I agree that the code may crash, but that is the case for many Engine methods that are not implemented for all possible types.

Just because others are doing it wrong, is no excuse for us not to do it right :wink:

pawelbaran:
> Just because others are doing it wrong, is no excuse for us not to do it right 😉 and is the reason we're trying to re-add them as we go. We have an issue open to try and find all the places it's not implemented so we can fix it, but that'll take time. This PR can fix this one though, and save us some time on finding it later 😄

Fair enough, makes sense in this particular case - fixed!

adecler (approved):
Looks good to me.
Fully agree with @FraserGreenroyd and @IsakNaslundBh. Happy to see the fallback added 👍 .

FraserGreenroyd (approved):
Happy with the changes

PR #2134, opened 2020-11-09
Spatial_Engine: Add fallback method to Query.IsNearGrid()

IsakLarbornBH:

Adds a fallback method to IsNearGrid, which was originally merged into master with pull request #2093, and currently does not have one. See that and issue #2055 for general details.


IsakNaslundBh (approved):
LGTM

PR #2142, opened 2020-11-12
Reflection_Engine: Make it easy to clear the value of a property

adecler:
### NOTE: Depends on 

Needed for this PR in BHoM_UI:  https://github.com/BHoM/BHoM_UI/pull/334

   
### Issues addressed by this PR

See BHoM_UI PR

PR #2143, opened 2020-11-12
Reflection_Engine: Access fragment properties

### LINKED ISSUE #477: Geometry_Engine: Missing Test for all Compute.DistributedOutlines methods
### This issue has been archived 

Historic issue archive can be found [here](https://burohappold.sharepoint.com/sites/BHoM/01_Master%20File/Forms/AllItems.aspx?id=%2Fsites%2FBHoM%2F01_Master%20File%2F999_BHoM_GitHub_History_Archive)

adecler:

Closes https://github.com/BHoM/BHoM/issues/477

We can now access fragments through `GetProperty` by providing the name of the fragment type.

I have also mirrored that behaviour in SetProperty. This is a bit weirder because of the `name` input but this, at least, provide a functioning solution for those that don't know about `AddFragment`.

### Test files
Definitely recommend trying your own thing but here's what I did:

image

### Additional comments
I would prefer if https://github.com/BHoM/BHoM_Engine/pull/2142 was merged first as the two PRs are modifying the same file (although not the same lines) and this one is easier to fix if rebase doesn't work.

### DIFF
--- a/Reflection_Engine/Modify/SetPropertyValue.cs
+++ b/Reflection_Engine/Modify/SetPropertyValue.cs
@@ -108,9 +108,19 @@ namespace BH.Engine.Reflection
             if (obj == null) return false;
 
             if (!obj.CustomData.ContainsKey(propName))
-                Compute.RecordWarning("The objects does not contain any property with the name " + propName + ". The value is being set as custom data");
+            {
+                if (value is IFragment)
+                {
+                    obj.Fragments.Add(value as IFragment);
+                    Compute.RecordWarning("The objects does not contain any property with the name " + propName + ". The value is being set as a fragment.");
+                }
+                else
+                {
+                    obj.CustomData[propName] = value;
+                    Compute.RecordWarning("The objects does not contain any property with the name " + propName + ". The value is being set as custom data.");
+                }
+            }
 
-            obj.CustomData[propName] = value;
             return true;
         }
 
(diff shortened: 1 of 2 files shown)


adecler:
It would be great if one of you could review this PR. I need this to be on the master to continue working around a better support for fragments.

adecler:
Actually, don't worry about merging this right now. 

I think the whole idea of fragments as properties needs a bit more polishing before it makes complete sense:

image

The problem is that using the fragment type as the property name causes a few issues:
- What about the case where multiple fragments from different namespaces have the same name ?
- What about the case in the picture where using the fragment type is not what would make the most sense from a user perspective ?

adecler:
All, 

I have modified the way `GetProperty` works with fragments based on last discussion. This should hopefully be ready for review and merging now.

image

al-fisher (approved):
LGTM.

The GetProperty for Fragments now works really neat - and I think is really intuitive.
Return the first with a warning when multiple results when querying by interface is also fine I think.

The SetProperty workflow for Fragments is a bit unusual - having to assign "Fragment" as the propName.
Wondering if we would enable setting no value for the propName in the specific case of a Fragment - as it is unambiguous without providing a value?
Not critical - as think the improvements to GetProperty for Fragments much more important in UI than the SetProperty Component.
So happy to approve with current behaviour

PR #2149, opened 2020-11-13
Structure_Engine: Add proper nullchecks for orientation methods

### LINKED ISSUE #2109: Structure_Engine: Proper nullchecks to be added for orientation methods
#### Broken rules:

The local orientation methods are a bit to sensitive and can crash out. Proper null checks should be put in place to stop this from happening.

The following methods should be made to never crash, whatever is thrown at them, but instead give meaningful errors and return nulls when failing:

- LocalOrientation
- CoordinateSystem
- IsVertical
- Normal
- Tangent

IsakLarbornBH:

Closes #2109 

This pull request adds thorough null checks to:
`Query.IsVertical()`
`Query.Tangent()`
`Query.Normal()` 
`Query.LocalOrientation()` 
`Query.CoordinateSystem()` 
`Query.OrientationAngleAreaElement()`

In order to accomplish this in a simple, non-duplicate way, it adds a new method to `Structure_Engine` - `Compute.NullCheck()`. This method verifies elements and their vital properties for types:
`Node`
`Bar`
`FEMesh`
`FEMeshFace`
`Panel` 

The method has been placed in `Structure_Engine` as it only handles structural elements, and in `Compute` as it does more than your average `Query` method.

Care has been taken to ensure that as much data as possible is output from methods where null checks have been implemented. This means that, for example, you can run `Query.Normals()` on an `FEMesh` with null nodes or faces and will still get as many Normals output as possible. Where null checks fail, methods will output null, or in the case of `IsVertical()`, false.

### Test files

[NullCheck.zip](https://github.com/BHoM/BHoM_Engine/files/5573570/NullCheck.zip)

### Additional comments

Removing draft status, as everything is now in place. Still think there is some valuable room for feedback, especially regarding the Compute.NullCheck() method (Where should it be placed, and should it be made unavailable for end users? Etc.)


IsakLarbornBH:
Checked after merging master and this still works fine for me. I'm getting no build errors so I don't know why the build check is failing here, and there's not much to go on for me to try to fix it. 

Would appreciate some review as I've done everything I can from my end, @IsakNaslundBh and whoever else might have time :)

If somebody wants an Etabs file that crashes the current 4.0 alpha but pulls in fine with this, contact me (can't share it publicly.)

IsakLarbornBH:
> I'm getting no build errors so I don't know why the build check is failing here, and there's not much to go on for me to try to fix it.

Managed to find the issue after looking through the build logs. Turns out that I was using a fairly new C# feature that apparently does not exist in the version used for the build checks. After changing `List<int> nodeListIndices = default` to `List<int> nodeListIndices = default(List<int>)` it now passes the build checks.

Structure_Engine/Compute/NullCheck.cs
```diff
@@ -0,0 +1,186 @@
...
+        }
+
+        [Description("Checks if an FEMesh or one of its Nodes are null and outputs relevant error message.")]
+        [Input("mesh", "The FEMesh to test for null.")]
+        [Input("methodName", "Optional name of the method to reference in the error message.")]
+        [Input("checkFaces", "Optional bool to tell the method whether to check FEMeshFaces or not.")]
+        [Input("checkNodes", "Optional bool to tell the method whether to check mesh Nodes or not.")]
+        [Input("nodeListIndices", "Optional list of nodes to limit check to")]
+        [Output("pass", "A boolean which is true if the FEMeshFace passes the null check.")]
+        public static bool NullCheck(this FEMesh mesh, string methodName = "Method", bool checkFaces = true, bool checkNodes = true, List<int> nodeListIndices = default(List<int>))
```
FraserGreenroyd, line 83:
```suggestion
        public static bool NullCheck(this FEMesh mesh, string methodName = "Method", bool checkFaces = true, bool checkNodes = true, List<int> nodeListIndices = null)
```

Is more appropriate than `default(type)`. Then obviously handle the null in the method.

FraserGreenroyd (changes requested):

Structure_Engine/Compute/NullCheck.cs
IsakLarbornBH: It's been weeks, but I seem to remember trying null at first but it not working, and changing to this because of it. Will try again and see what happens.

FraserGreenroyd: Feel free to gimme a shout if you need some help, happy to do a quick session over teams if needed :smile:

IsakLarbornBH: Worked fine now. Not sure what happened when I originally tried it but since it seems to work now I've implemented this suggestion. Thank you for the offer, we'll see if I need it ;)

Structure_Engine/Query/CoordinateSystem.cs
```diff
@@ -54,9 +54,9 @@ public static Cartesian CoordinateSystem(this Node node)
         [Output("coordinateSystem", "The local cartesian coordinate system of the Bar.")]
         public static Cartesian CoordinateSystem(this Bar bar)
         {
-            Vector tan = bar.Tangent(true);
-            Vector ax = bar.Normal().CrossProduct(tan);
-            return Engine.Geometry.Create.CartesianCoordinateSystem(bar.StartNode.Position, tan, ax);
+            Vector tan = bar?.Tangent(true);
+            Vector ax = bar?.Normal()?.CrossProduct(tan);
+            return tan != null && ax != null ? Engine.Geometry.Create.CartesianCoordinateSystem(bar.StartNode.Position, tan, ax) : null;
```
IsakLarbornBH, line 59: Having a look through this myself again, hoping to catch some issues to make the review easier. At first I wondered why I had forgot to include error messages at so many points in CoordinateSystem. After looking into it further I realised that this is because Tangent and Normal already use NullCheck and therefore already show error messages. The check in return is just to check whether these null checks failed or passed.

IsakNaslundBh (approved):
Generally LGTM. Happy for this to get merged to as it improves stability and given clearer messages to the users. Any further improvement can be added in 4.1

Structure_Engine/Query/LocalOrientation.cs
```diff
@@ -45,7 +45,9 @@ public static partial class Query
         [Output("orienation", "The local orientation of the Panel as a vector Basis.")]
         public static Basis LocalOrientation(this Panel panel)
         {
-            return LocalOrientation(Engine.Spatial.Query.Normal(panel), panel.OrientationAngle);
+            Vector normal = panel.NullCheck("LocalOrientation") ? Engine.Spatial.Query.Normal(panel) : null;
+
+            return normal != null ? LocalOrientation(Engine.Spatial.Query.Normal(panel), panel.OrientationAngle) : null;
```
IsakNaslundBh, line 50: Can just use the calculated normal

```suggestion
            return normal != null ? LocalOrientation(normal, panel.OrientationAngle) : null;
```

Structure_Engine/Compute/NullCheck.cs
```diff
@@ -0,0 +1,186 @@
...
+            return true;
+        }
+
+        /***************************************************/
+        /**** Private Methods                           ****/
+        /***************************************************/
+
+        private static void ErrorMessage(string methodName, string type)
+        {
+            Engine.Reflection.Compute.RecordError($"Cannot run {methodName} because {type} failed a null check");
```
IsakNaslundBh, line 182:
```suggestion
            Engine.Reflection.Compute.RecordError($"Cannot evaluate {methodName} because {type} failed a null check");
```

IsakNaslundBh (changes requested):
After looking through again, some minor comments

IsakNaslundBh (approved):
Re-approving

IsakNaslundBh (approved):
Again, re-approving

PR #2160, opened 2020-11-16
Structure_Engine: Updates to BarVaryingDistributedLoad

### LINKED ISSUE #1030: Structure_Engine: BarVaryingDistributedLoad Create method accepting parameter
Create method for `BarVaryingDistributedLoad` that accepts the length parameter (between 0 and 1) rather than distance. 

Method will need to check through length of each bar sent to the create method and create one load object for each unique length found in the list of bars.

FYI @jdarcyBHE

IsakNaslundBh:
### NOTE: Depends on 

https://github.com/BHoM/BHoM/pull/1075
   
Closes #1030 

### Test files

Test versioning.

Note, the versioning of the objects themselves is not working, as it gets quite convoluted and requires for many cases that one object is turned into multiple objects, which is currently not supported. The script below tests through the the versioning for the create methods is working though.

https://burohappold.sharepoint.com/:f:/s/BHoM/EhIZk-booZNHhaTeGJLSuMUB0TS-q8UA6_WxZH8crV4fWg?e=WkN1Ni


IsakNaslundBh:
@BHoMBot check all

FraserGreenroyd:
@BHoMBot check installer

FraserGreenroyd:

FraserGreenroyd:
@IsakNaslundBh - the documentation compliance check has been updated to account for `InputFromProperty` as per https://github.com/BHoM/Test_Toolkit/issues/159 - the remaining warnings now highlighted by the compliance check are valid for the PR in its current state I believe (though I've only glanced, not done an in depth).

Take a look and double check, but I think you should find the warnings more appropriate for you going forward :smile:

IsakNaslundBh:
@BHoMBot check all

IsakNaslundBh:
@BHoMBot check compliance

IsakNaslundBh:
@BHoMBot check all

peterjamesnugent (changes requested):
**Engine
- I am not keen on the storing the two `.cs` files in a seperate folder, it makes them more difficult to find
- Some minor comments
**Versioning**
- Engine methods version fine
- Noted in PR objects won't version

Structure_Engine/Create/Loads/BarVaryingDistributedLoad/BarVaryingDistributedLoad.cs
```diff
@@ -0,0 +1,135 @@
...
+        [Output("barVarLoad", "The created BarVaryingDistributedLoad.")]
+        public static BarVaryingDistributedLoad BarVaryingDistributedLoad(Loadcase loadcase, BHoMGroup<Bar> group, double startPosition = 0, Vector forceAtStart = null, Vector momentAtStart = null, double endPosition = 1, Vector forceAtEnd = null, Vector momentAtEnd = null, bool relativePositions = true, LoadAxis axis = LoadAxis.Global, bool projected = false, string name = "")
+        {
+            if ((forceAtStart == null || forceAtEnd == null) && (momentAtStart == null || momentAtEnd == null))
+            {
+                Engine.Reflection.Compute.RecordError("Bar varying load requires either the force at start and end and/or the moment at start and end to be defined");
+                return null;
+            }
+
+            if (relativePositions && (startPosition < 0 || startPosition > 1 || endPosition < 0 || endPosition > 1))
```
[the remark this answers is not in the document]
IsakNaslundBh: true, will update

Structure_Engine/Create/Loads/BarVaryingDistributedLoad/BarVaryingDistributedLoad.cs
```diff
@@ -0,0 +1,135 @@
...
+
+            Dictionary<double, List<Bar>> dict = new Dictionary<double, List<Bar>>();
+            foreach (var group in bars.GroupBy(x => (int)Math.Round(x.Length() / tolerance)))
+            {
+                dict[group.Key * tolerance] = group.ToList();
+            }
+            return dict;
+        }
+
+        /***************************************************/
```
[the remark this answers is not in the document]
IsakNaslundBh: Good shout, will move

IsakNaslundBh:
> **Engine**
> 
> * I am not keen on the storing the two `.cs` files in a seperate folder, it makes them more difficult to find

This comes from the compliance rules, not really something I really prefer either, but the create methods need to both match filename with method name _and_ they need to sit in a file/folder with the same name as the object they are creating.

https://github.com/BHoM/documentation/wiki/IsValidCreateMethod

> **Versioning**
> 
> * Engine methods version fine
> * Noted in PR objects won't version

Yeah, might try to have a look at the objects again if I find some time, but it gets quite complicated...

> @IsakNaslundBh just to add to this, the `Create` method versions correctly but the default value for `relativePositions` is not used, is this intentional to flag to the user that a change has occured?

Yeah, my reasoning here was that we are adding a new option to the load, and making this new option default in the oM and the new constructors. If we were to use this new default for old components, it would change their behaviour in pre-existing scripts. What we could do is to set relative positions to false as a default in the old components, but that would mean a missmatch between the old creates and the new creates, which I think would be very confusing. So, my solution became to add the boolean to the old create method, but not provide a default value. Not perfect, but felt like the best of the options I could think of.

Structure_Engine/Create/Loads/BarVaryingDistributedLoad/BarVaryingDistributedLoad.cs
```diff
@@ -38,36 +39,52 @@ public static partial class Create
...
+        public static BarVaryingDistributedLoad BarVaryingDistributedLoad(Loadcase loadcase, BHoMGroup<Bar> group, double startPosition = 0, Vector forceAtStart = null, Vector momentAtStart = null, double endPosition = 1, Vector forceAtEnd = null, Vector momentAtEnd = null, bool relativePositions = true, LoadAxis axis = LoadAxis.Global, bool projected = false, string name = "")
         {
-            if ((forceA == null || forceB == null) && (momentA == null || momentB == null))
-                throw new ArgumentException("Bar varying load requires either the force at A and B and/or the moment at A and B to be defined");
+            if ((forceAtStart == null || forceAtEnd == null) && (momentAtStart == null || momentAtEnd == null))
+            {
+                Engine.Reflection.Compute.RecordError("Bar varying load requires either the force at start and end and/or the moment at start and end to be defined.");
+                return null;
+            }
```
peterjamesnugent, line 62: I think this requires a rethink, because if you have:
```
momentAtStart = {0, 0, 0};
momentAtEnd = {-1000, 0, 0};
```

This will return an error - when it is actually a valid load.

FraserGreenroyd: Apologies if this is a dumb question, I am basing it purely on the subset of code I can see following @peterjamesnugent comments, and please ignore if it's derailing the conversation, but, what happens if you have a `forceAtStart` and a `forceAtEnd` but no `momentAtStart` or `momentAtEnd`? Based on the `if` statement here it appears that this would be acceptable, but the error message says it requires `either the force at start and end and/or the moment at start and end`.

Reading the error message suggests having `forceAtStart` and `forceAtEnd` but no `momentAtStart/End` is problematic, I think because of the use of `and/or` - reading it with the `and` suggests all the options are needed, but reading with the `or` suggests (what I suspect is the desired effect) that either `forceAtStart/End` OR `momentAtStart/End` are needed but not both.

Maybe the error message could be improved with commas rather than an `and` to read something like:

`Bar varying load requires either the force at start and end, or the moment at start and end, to be defined`

As I say though, very much only based on reading 3 lines highlighted in @peterjamesnugent snippet above, have not read the rest of the code, don't even understand overly what it's doing, so please do feel free to entirely ignore this late night rambling :smile:

IsakNaslundBh: @peterjamesnugent , good found. Was actually no problem with the logic here, but was a problem with wrong parameters being passed from the create method accepting an `IEnumerable<Bar>` 🤦 

Have fixed that issue now, very glad that got caught!

peterjamesnugent (approved):
LGTM.

PR #2158, opened 2020-11-16
Spatial_Engine: Add Query.IsFullyOnLevel() method

IsakLarbornBH:
### Issues addressed by this PR

Closes #2111 

Adds Query.IsFullyOnLevel() method to Spatial_Engine. This method accepts an IElement, a Level, and a double maxDistance as input, and tests the IElement for proximity to the Level, returning true if the IElement is fully contained within the set maxDistance from the level Elevation.

### Test files

[IsFullyOnLevel.zip](https://github.com/BHoM/BHoM_Engine/files/5644115/IsFullyOnLevel.zip)

### Changelog

`Query.IsFullyOnLevel()` Query method added in the `Spatial_Engine` for `IElement` class

### Additional comments

This method is identical to Query.IsNearLevel() apart from the following two points:

* IElement1Ds: Bounding box must now be fully contained within max distance from level elevation, rather than partially as with IsNearLevel.

* IElement2Ds: Will now only return true if every constituent IElement1D returns true, rather than just one as with IsNearLevel.

PR #2176, opened 2020-11-20
Spatial_Engine: Add Query.IsFullyOnGrid() method

### LINKED ISSUE #2110: Spatial_Engine: Add IsFullyOnGrid method for all IElement types
#### Description:

Similar to what was added in https://github.com/BHoM/BHoM_Engine/issues/2055, but a method to check if all parts of an element is within a given distance from a gridline.

IsakLarbornBH:

Closes #2110 

Adds Query.IsFullyOnGrid() method to Spatial_Engine. This method accepts an IElement, a Level, and a double tolerance as input, and tests the IElement for proximity to the Grid, returning true if the IElement is fully contained within the set tolerance from the Grid curve.

### Test files

[IsFullyOnGrid.zip](https://github.com/BHoM/BHoM_Engine/files/5651932/IsFullyOnGrid.zip)
(Still don't have my sharepoint access back, so the test file is here)

### Changelog

`Query.IsFullyOnGrid()` Query method added in the `Spatial_Engine` for `IElement` class

### Additional comments

Currently the distance input is called `tolerance` and defaults to `BH.oM.Geometry.Tolerance.Distance` pending a final decision on #2158 IsFullyOnLevel. These two methods should have the same name and behaviour for these inputs before merging. (Update: IsFullyOnLevel is now merged, using tolerance in this way.)

This method does not properly support NurbsCurves, as it uses control points for evaluation, which works well for Lines and Arcs. To properly support NurbsCurves a more sophisticated algorithm capable of finding the curve's farthest point in distance from the Grid curve would be required. As it currently stands, this method throws a warning if it detects that its input is a NurbsCurve, but is not prevented from running.


Spatial_Engine/Query/IsFullyOnGrid.cs
```diff
@@ -0,0 +1,132 @@
...
+                BH.Engine.Reflection.Compute.RecordWarning("IsFullyOnGrid does not fully support NurbsCurves. Results may be inaccurate.");
+            }
+
+            foreach (Point pt in ctrlPts)
+            {
+                if (pt.IDistance(gridCurve) >= tolerance)
+                    return false;
+            }
+
+            return ctrlPts.Count > 0;
```
FraserGreenroyd, line 83: Won't this will always be true? And might thus be easier to just do `return true;`?

FraserGreenroyd (changes requested):
Minor comment but otherwise this looks good @IsakLarbornBH - we should probably have a catch up with @IsakNaslundBh next week to bottom out the points you're raising regarding the other PR as well to unblock you :smile:

Spatial_Engine/Query/IsFullyOnGrid.cs
IsakLarbornBH: You are probably right. It's just I've been working on null checks, and that made me wonder what would happen if you fed it an empty curve object with no curve segments. Since I didn't want it to return true in that case or some other when there are no control points returned, I added this check. But I don't think there's currently a way to trigger this that doesn't just crash in one of one of the earlier methods.

Spatial_Engine/Query/IsFullyOnGrid.cs
```diff
@@ -0,0 +1,132 @@
...
+
+        [Description("Checks if the geometrical representation of an IElement1D, projected onto the XY Plane, is fully within a set tolerance from a grid line.")]
+        [Input("element1D", "The IElement1D that will be checked for proximity to the grid.")]
+        [Input("grid", "The Grid to use for evaulation.")]
+        [Input("tolerance", "The maximum distance allowed from the Grid for this method to return true.", typeof(Length))]
+        [Output("isFullyOnGrid", "A boolean which is true if the geometrical representation of an IElement1D is fully within a set tolerance from the grid line.")]
+        public static bool IsFullyOnGrid(this IElement1D element1D, Grid grid, double tolerance = BH.oM.Geometry.Tolerance.Distance)
+        {
+            ICurve curve = element1D.IGeometry().IProject(Plane.XY);
+            List <Point> ctrlPts = curve.ControlPoints();
```
IsakNaslundBh, line 63: Think we are better off projecting the controlpoints rather than the curve, as projecting the curve actually might turn it into a nurbs curve. (arcs and circles for example). Could help with your check below.

This is, ofc, unless you had it thew way you had it before intentionally

```suggestion
            ICurve curve = element1D.IGeometry();
            List <Point> ctrlPts = curve.ControlPoints().Select(x => x.Project(Plane.XY)).ToList();
```

Can also add, though, that checking the control points might not be fully rigorous (arc vs a polyline for example, might give you a false "true", but lets catch up on that next week.)

IsakNaslundBh (commented):
Initial comment regarding the projection

Spatial_Engine/Query/IsFullyOnGrid.cs
IsakLarbornBH: That does sound like a good idea. There's no particular reason why the curve is projected rather than the points.

You're right about the potential for false positives. I initially had the idea of using gradient descent to find local maximi in distance, but I figured it might be too heavy and complex for a method like this. I'm sure there are many more ways of making better algorithms than what is there currently. Let's catch up on that.

IsakLarbornBH:
>Initial comment regarding the projection

Points, rather than curves, are now projected to the XY plane for checking. I have also disabled support for curved grids as discussed. It would be possible to reduce the likelihood of false positives for curved grids by doing an identical check for the grid against the projected curve, but a more sophisticated algorithm would be even better and be able to support NurbsCurves so if curved grid support becomes necessary that would be the way to go.

IsakLarbornBH:
I've now added a new test file based on the one I made for IsFullyOnGrid. I still don't have my Sharepoint access back, or even a response from IT a week after I contacted them, so the file is uploaded here.

Also changed the E1D method to use the E0D method for checking the distance of control points, rather than having its own identical comparison.

Spatial_Engine/Query/IsFullyOnGrid.cs
IsakLarbornBH: I got rid of this check.

IsakNaslundBh:
FYI @IsakLarbornBH I have put your testfile on sharepoint here:

https://burohappold.sharepoint.com/:f:/s/BHoM/EjkDVhX6IA9Gu2CLJHgHDBQBl6y3_XKQx28ed8NXIOwiFA?e=6KZ5vh

IsakNaslundBh:
@BHoMBot check compliance

IsakLarbornBH:
> FYI @IsakLarbornBH I have put your testfile on sharepoint here:

Thank you :)

I've done some tests of a few other algorithms but not really found anything that works better than this, at least not without being slow. Therefore I think it's better that this gets merged as is, so I've removed WIP status.

IsakNaslundBh (approved):
Think this looks good now and seem to be working fine. Approved!

IsakNaslundBh:
@FraserGreenroyd think your comment has been resolved, if you want to have another look?

FraserGreenroyd (approved):
My comment has been addressed, I'm happy, thanks @IsakNaslundBh and @IsakLarbornBH

PR #2178, opened 2020-11-20
Serialiser_Engine: add support for Json Arrays

### LINKED ISSUE #2177: Serialiser_Engine: add support for Json Arrays
#### Description:

I've been having a considerable amount of trouble producing and consuming valid JSONs, while working in the File Adapter and trying to leverage our methods ToJson/FromJson.

99% of the trouble is due to this piece of code:

https://github.com/BHoM/BHoM_Engine/blob/a3e39b120828103670fbaaac42ad4a4b1dd53b38/Serialiser_Engine/Convert/Json.cs#L77-L90

This is because, when we have a List of BHoMObjects, we never serialise the list as a proper JSON array (which should be the correct way); instead, we have been "improperly" serialising it, by simply concatenating individually-serialised objects separated by a newline, which is not JSON. I have named this "format" as _DatasetJson_ because it's commonly used in Datasets. **I know this "format" may be deprecated in the near future, but this issue does not regard this behaviour**.

**The core of this issue is:** in addition to the above, the piece of code adds the behaviour of "_if a json array is input, then deserialise it to a CustomObject_".
Is this latter behaviour used anywhere? Does it really make sense?
@al-fisher @adecler @IsakNaslundBh @FraserGreenroyd 

**I am proposing that we rather add correct support of Serialization/Deserialization of Json Arrays.**
This change would not affect how we deal with _DatasetJson_. The serialization "as DatasetJson" is has always been done by another, separate piece of code in the File_Adapter.

My proposal is to replace the above code with:
```cs
            else if (json.StartsWith("["))
            {
                BsonArray array = MongoDB.Bson.Serialization.BsonSerializer.Deserialize<BsonArray>(json);

                return array.Select(b => b.IsBsonDocument ? BH.Engine.Serialiser.Convert.FromBson(b.AsBsonDocument) : BH.Engine.Serialiser.Convert.FromJson(b.ToString())).ToList();
            }
```

which would allow to deserialise valid Json Arrays. If we do this change, we could still "improperly" serialise list of objects as we have always done, thanks to a setting in the File_Adapter.

Additionally, we can simply add a method to serialise Lists of objects into a JsonArray:
<details>

```cs

        [Description("Convert a List of objects to a Json array.")]
        [Input("obj", "Object to be converted.")]
        [Output("json", "String representing the object in json.")]
        public static string ToJson(this IEnumerable<object> objs)
        {
            List<string> allLines = new List<string>();
            string json = "";

            // Join all individually serialised objects with a comma and a newline
            json = string.Join(",\n", objs.Where(c => c != null).Select(obj => obj.ToJson()));

            // Put between square brackets, to make a valid JSON array.
            json = "[" + json + "]";

            return json;
        }
```

</details>

This would hugely help me simplifying the code in the File_Adapter and it could be helpful in other places were you might just need the serialised JSON array of a list of object, rather than having to write to a File with the File_Adapter to obtain it.

alelom:

Closes #2177

### Test files

https://burohappold.sharepoint.com/:f:/r/sites/BHoM/02_Current/12_Scripts/02_Pull%20Request/BHoM/BHoM_Engine/Serialiser_Engine/%232178-AddJsonArraySupport?csf=1&web=1&e=X5O769

image

### Changelog

- do not return a CustomObject upon deserialising a valid Json Array;
- allow ToJson to serialise json arrays.


Serialiser_Engine/Convert/Json.cs
```diff
@@ -98,6 +91,27 @@ public static object FromJson(string json)
...
+        public static string ToJson(this IEnumerable<object> objs)
+        {
+            List<string> allLines = new List<string>();
+            string json = "";
+
+            // Append a comma to the invididually serialized objects
+            allLines.AddRange(objs.Where(c => c != null).Select(obj => obj.ToJson() + ","));
+
+            // Remove the trailing comma for the last element.
+            allLines[allLines.Count - 1] = allLines[allLines.Count - 1].Remove(allLines[allLines.Count - 1].Length - 1);
```
adecler, line 107: You should use `Aggregate` instead to better handle the `,`

adecler (changes requested):
This breaks the UI for me:

image

Serialiser_Engine/Convert/Json.cs
```diff
@@ -98,6 +96,44 @@ public static object FromJson(string json)
...
+            // Put between square brackets, to make a valid JSON array.
+            json = "[" + json + "]";
+
+            return json;
+        }
+
+        /*******************************************/
+
+
+        [Description("Convert a Json string to a BHoMObject")]
```
FraserGreenroyd, line 120: This description doesn't seem to match the function? Should it be:
```suggestion
        [Description("Convert a Json string to a collection of objects")]
```

Name as it's returning an IEnumerable, and of object rather than BHoMObject?

FraserGreenroyd (changes requested):
Minor documentation suggestion. Have not tested functionality wise yet, need to do that. @adecler will also need to review before merging anyhow :smile:

adecler (approved):
Looks good to me now.

I took the liberty to do a few minor cosmetic fixes so make sure you check them before merging:
- Fix incorrect names and descriptions for input/output of `ToJsonArray`/`FromJsonArray`. That was causing the description to not show in the UI for the case of the input.
- Adding missing final separator comment
- Removed unnecessary empty line
- Removed unnecessary namespace in front of method from the same type as it was reducing readability of the code
- Reverted `RecordWarning` to `RecordNote` as it was originally. The component should not turn orange for something that works perfectly as intended.

All pretty minor things but I like to keep my code tidy 😄

alelom:
Perfect, thanks @adecler !

FraserGreenroyd (approved):
This works well for me, thanks @alelom

FraserGreenroyd:

PR #2180, opened 2020-11-20
Geometry_Engine: Normal() returns null with a warning instead of an error

### LINKED ISSUE #2179: Geometry_Engine: in case of self intersecting Normal() returns null with a warning instead of an error
#### Broken rules:

`Normal()` returns nothing and raises only a warning instead of an error.

Change warning into an error or make it return something.

LMarkowski:

Closes #2179

Changing the `Normal()` for `Polyline` and `PolyCurve` so it'll return a results even for self intersecting curves (with a proper warning).

The reason for that is because there exist some cases when the curve is distorted and is read as self-intesecting (because of e.g. tiny overlapping segments) but calculating correct normal vector is possible despites the distortion. 

On the other hand, failing other checks (closure and planarity) makes it always impossible to calculate normal. In these cases error is raised and null returned.

### Test files

[Issue specific](https://burohappold.sharepoint.com/:f:/s/BHoM/Ev0hASsgYEFMtKvjgKEDinQB8e5CMzNdL_kl9JiuU7cwgw?e=ucs0iE) - BooleanUnion was crashing because of missing Normal of a distorted region. Now should work as expected.
[General test for Normal](https://burohappold.sharepoint.com/:f:/s/BHoM/Esd2eZj9Hc1Mkh5tar8m-NsBlIxJtyJ40zXatKwF4D3uOQ?e=9Sseg9) (Normal.gh).

### Changelog

- Changing the `Normal()` for `Polyline` and `PolyCurve` so it'll return a results even for self intersecting curve (with a proper warning)

### DIFF
--- a/Geometry_Engine/Query/Normal.cs
+++ b/Geometry_Engine/Query/Normal.cs
@@ -102,16 +102,16 @@ namespace BH.Engine.Geometry
 
             if (!curve.IsPlanar(tolerance))
             {
-                Reflection.Compute.RecordError("Input must be planar.");
+                Reflection.Compute.RecordError("Input curve must be planar.");
                 return null;
             }
             else if (!curve.IsClosed(tolerance))
             {
-                Reflection.Compute.RecordError("Curve is not closed. Input must be a polygon");
+                Reflection.Compute.RecordError("Input curve is not closed. Input must be a polygon");
                 return null;
             }
             else if (curve.IsSelfIntersecting(tolerance))
-                Reflection.Compute.RecordWarning("Input curve is self intersecting. Resulting normal vector might be incorrect.");
+                Reflection.Compute.RecordWarning("Input curve is self-intersecting. Resulting normal vector might be flipped.");
 
             Point avg = curve.ControlPoints.Average();
             Vector normal = new Vector();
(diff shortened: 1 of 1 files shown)


FraserGreenroyd:
@BHoMBot check compliance

LMarkowski:
@al-fisher @IsakLarbornBH @pawelbaran 
Updated issue specific test file.
In case of a centrally symmetric self-intersecting curve the method would return a zero vector (cross products from one part add up to zero with products from other part). Decided to add another check for normal length for this and similar cases. I think zero vector is the same result as failing to calculate normal.

Will raise another issue for IsClosed. From my end that one is ready for a final review.

pawelbaran (approved):
LGTM now 👍 

Concerning the open curves, I would opt for being a bit cautious: what does it mean for an open curve to be clockwise or counterclockwise? What does it mean for a line? For a polyline consisting of 2 almost collinear segments? As far as I see it, the best `Normal` of an open curve we can give to the user is `FitPlane()?.Normal`, which will only give a false impression that there is some intelligence in it (plus it can be easily done using the existing functionality).

What do others think?

LMarkowski:
@BHoMBot check installer

Issue #895, opened 2019-04-08
Create method to create a new object from the UI


epignatelli:
Since most of the `BHoMObject`s are just a collection of key value pairs, I think it would ease to job for newcomers to create schemas from the UI.

Given a project name, a namespace, a class name and a series of properties, we should be able to create the cs file straight from these info, and save it in the right place.

PR #2186, opened 2020-11-24
Physical_Engine: Add query method moving location line to the centre of the IFramingElement's section bounding box

annastigenberg:
### Test files

[Test File](https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/01_Test%20Scripts/BHoM_Engine/Physical_Engine/Query/GeometricalCentreline.gh?csf=1&web=1&e=lFOaJv)

### Additional comments
Additional method to be used in the [Revit_Toolkit PR #895](https://github.com/BHoM/Revit_Toolkit/pull/895). Creating an additional method similar to the TopCentreline and BottomCentreline methods. Returns the element's section bounding box geometrical centreline.

### DIFF
--- a/Physical_Engine/Query/GeometricalCentreline.cs
+++ b/Physical_Engine/Query/GeometricalCentreline.cs
@@ -33,7 +33,7 @@ namespace BH.Engine.Physical
 {
     public static partial class Query
     {
-        [Description("Returns the geometrical centreline of an IFramingElement.")]
+        [Description("Returns the centreline of an IFramingElement's section bounding box.")]
         [Input("element", "The IFramingElement to query the geometrical centreline of.")]
         [Output("curve", "The geometrical centreline of the IFramingElement.")]
         public static ICurve GeometricalCentreline(this IFramingElement element)
--- a/Physical_Engine/Query/GeometricalCentreline.cs
+++ b/Physical_Engine/Query/GeometricalCentreline.cs
@@ -38,25 +38,22 @@ namespace BH.Engine.Physical
         [Output("curve", "The geometrical centreline of the IFramingElement.")]
         public static ICurve GeometricalCentreline(this IFramingElement element)
         {   
-            ICurve location = element.Location;
+            ICurve location = element?.Location;
+            if (location == null)
+                return null;
 
             if (element.Property is ConstantFramingProperty)
             {
-                ConstantFramingProperty constantProperty = element.Property as ConstantFramingProperty;
+                Point geometricalCentre = (element.Property as ConstantFramingProperty)?.Profile?.Edges?.Bounds()?.Centre();
 
-                IProfile profile = constantProperty.Profile; 
+                if (geometricalCentre == null)
+                    return null;
 
-                Point geometricalCentre = profile.Edges.Bounds().Centre();
-
-                Vector translate = geometricalCentre - Point.Origin;
-
-                ICurve geometricalCentreline = location.ITranslate(translate);
-
-                return geometricalCentreline;
+                return location.ITranslate(geometricalCentre - Point.Origin);
             }
             else
             {
-                Engine.Reflection.Compute.RecordError("Element does not have a suitable framing property, was not able to extract section bounding box.");
+                Engine.Reflection.Compute.RecordError("Only suitable framing property for the IFramingElement is ConstantFramingProperty.");
                 return null;
             }
         }
--- a/Physical_Engine/Query/GeometricalCentreline.cs
+++ b/Physical_Engine/Query/GeometricalCentreline.cs
@@ -49,10 +49,7 @@ namespace BH.Engine.Physical
                 return null;
             }
 
-            Vector tangent = null;
-            tangent = location.ITangentAtPoint(location.ICentroid());
-            if (tangent == null)
-                return null;
+            Vector tangent = (location.IEndPoint() - location.IStartPoint()).Normalise();
 
             Vector thirdAxis = tangent.CrossProduct(normal);
 
@@ -70,7 +67,7 @@ namespace BH.Engine.Physical
             }
             else
             {
-                Engine.Reflection.Compute.RecordError("Only suitable framing property for the IFramingElement is ConstantFramingProperty.");
+                Engine.Reflection.Compute.RecordError("Only suitable framing property for the action is ConstantFramingProperty.");
                 return null;
             }
         }
--- a/Physical_Engine/Query/GeometricalCentreline.cs
+++ b/Physical_Engine/Query/GeometricalCentreline.cs
@@ -65,7 +65,7 @@ namespace BH.Engine.Physical
 
                 Vector sectionTranslate = geometricalCentre - Point.Origin;
 
-                return location.ITranslate(tangent * sectionTranslate.X + thirdAxis * sectionTranslate.Y + normal * sectionTranslate.Z);
+                return location.ITranslate(- thirdAxis * sectionTranslate.X + normal * sectionTranslate.Y);
 
             }
             else
--- a/Physical_Engine/Query/GeometricalCentreline.cs
+++ b/Physical_Engine/Query/GeometricalCentreline.cs
@@ -1,0 +1,84 @@
+/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2020, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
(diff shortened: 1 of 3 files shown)


Physical_Engine/Query/GeometricalCentreline.cs
```diff
@@ -0,0 +1,64 @@
...
+namespace BH.Engine.Physical
+{
+    public static partial class Query
+    {
+        [Description("Returns the geometrical centreline of an IFramingElement.")]
+        [Input("element", "The IFramingElement to query the geometrical centreline of.")]
+        [Output("curve", "The geometrical centreline of the IFramingElement.")]
+        public static ICurve GeometricalCentreline(this IFramingElement element)
+        {   
+            ICurve location = element.Location;
```
pawelbaran, line 41: Here I would add a few null checks:
```suggestion
            ICurve location = element?.Location;
            if (location == null)
                return null;
```

Physical_Engine/Query/GeometricalCentreline.cs
```diff
@@ -0,0 +1,64 @@
...
+
+            if (element.Property is ConstantFramingProperty)
+            {
+                ConstantFramingProperty constantProperty = element.Property as ConstantFramingProperty;
+
+                IProfile profile = constantProperty.Profile; 
+
+                Point geometricalCentre = profile.Edges.Bounds().Centre();
```
pawelbaran, line 49: You could do it in one line, then it would be good to add a null check:
```suggestion
                Point geometricalCentre = (element.Property as ConstantFramingProperty)?.Profile?.Edges?.Bounds()?.Centre();
                if (geometricalCentre == null)
                    return null;
```

Alternatively, you can keep the code structure you have but you would need to add a null check after every line.

Physical_Engine/Query/GeometricalCentreline.cs
```diff
@@ -0,0 +1,64 @@
...
+
+                Vector translate = geometricalCentre - Point.Origin;
+
+                ICurve geometricalCentreline = location.ITranslate(translate);
+
+                return geometricalCentreline;
+            }
+            else
+            {
+                Engine.Reflection.Compute.RecordError("Element does not have a suitable framing property, was not able to extract section bounding box.");
```
pawelbaran, line 59: Here I would write that the only suitable framing property is `ConstantFramingProperty`.

Physical_Engine/Query/GeometricalCentreline.cs
```diff
@@ -0,0 +1,64 @@
...
+
+                IProfile profile = constantProperty.Profile; 
+
+                Point geometricalCentre = profile.Edges.Bounds().Centre();
+
+                Vector translate = geometricalCentre - Point.Origin;
+
+                ICurve geometricalCentreline = location.ITranslate(translate);
+
+                return geometricalCentreline;
```
pawelbaran, line 55: Again, to be more concise, you could remove the blanks and simply write the code as follows:
```suggestion
                Vector translate = geometricalCentre - Point.Origin;
                return location.ITranslate(translate);
```

If you'd like to be even more concise, you could skip declaring `translate` variable and simply pass `geometricalCentre - Point.Origin` as the argument of `ITranslate`.

pawelbaran (changes requested):
The functionality itself looks good 👍 However, I have a few comments on code quality and general compliance:
- you are missing a few null checks plus the code could be a bit more concise (please see the code comments)
- please change the PR title to start with _Physical_Engine:_ and then describe what had been implemented
- please fill in the changelog
- the test file could be improved a bit:
    - Revit adapter and Pull components could be removed - they are irrelevant to this PR
    - you could compare the vectors extracted using the method and in your GH alternative to prove that they are the same (simply take the one from each group and compare them, e.g. using `IsEqual` BHoM method)
    - personally I put all test files in [Test Scripts SharePoint folder](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?id=%2Fsites%2FBHoM%2F02_Current%2F12_Scripts%2F01_Test%20Scripts) (less hassle for you to pack the files, easier to update, easier to download) - if you navigate to BHoM_Engine, you will quickly find out where and how to create a folder for your PR 

Please feel free to ping me in case of any questions :)

IsakNaslundBh (commented):
Would update the method name of this to something more explicit. The location curve is already the geometrical centreline I would argue, with respect to the centroid of the profile. Basically, I would assume GeometricalCentre to mean the centroid, not boundingbox centre.

I think also the description need to clarify this, what centre is being referred to.

Physical_Engine/Query/GeometricalCentreline.cs
FraserGreenroyd: It would be beneficial to explain what `?` is doing in this context @pawelbaran because not everyone has our knowledge of coding - this is not usually seen in the early tutorials for coding, so if you're going to suggest it, can you provide your understanding of it to help people learn - otherwise you might as well just write the code yourself :wink:

annastigenberg: Googled - learned :)

pawelbaran:
@annastigenberg I just realized that the functionality itself is incorrect at the moment as you are not transforming the translation vector from the global coordinate system to beam's local one - I have slightly updated your test file to highlight the issue.

To get the necessary transformation, I would recommend using the `Normal` query method from Physical_Engine, extracting beam's tangent and then getting the 3rd axis of the local coordinate system by simple cross product of the two above. This gives you the target coordinate system (translation not needed as you are working with a vector) to which you need to transform.

Physical_Engine/Query/GeometricalCentreline.cs
```diff
@@ -0,0 +1,78 @@
...
+            if (element.Property is ConstantFramingProperty)
+            {
+                Point geometricalCentre = (element.Property as ConstantFramingProperty)?.Profile?.Edges?.Bounds()?.Centre();
+
+                if (geometricalCentre == null)
+                    return null;
+
+                Vector sectionTranslate = geometricalCentre - Point.Origin;
+
+                return location.ITranslate(tangent * sectionTranslate.X + thirdAxis * sectionTranslate.Y + normal * sectionTranslate.Z);
```
pawelbaran: I think `normal * sectionTranslate.Z` term is not needed here - we do not want to move the curve along the element.
```suggestion
                return location.ITranslate(tangent * sectionTranslate.X + thirdAxis * sectionTranslate.Y);
```

Physical_Engine/Query/GeometricalCentreline.cs
```diff
@@ -0,0 +1,78 @@
...
+            if (normal == null)
+            {
+                Engine.Reflection.Compute.RecordError("IFramingElement must have linear location line.");
+                return null;
+            }
+
+            Vector tangent = null;
+            tangent = location.ITangentAtPoint(location.ICentroid());
+            if (tangent == null)
+                return null;
```
pawelbaran, line 55: We already know that the curve is linear, so we can pick a vector between start and enpoint, which requires less computations (minor tweak, more to make the code simpler).
```suggestion
            Vector tangent = (location.IEndPoint() - location.IStartPoint()).Normalise();
```

Physical_Engine/Query/GeometricalCentreline.cs
```diff
@@ -0,0 +1,78 @@
...
+                    return null;
+
+                Vector sectionTranslate = geometricalCentre - Point.Origin;
+
+                return location.ITranslate(tangent * sectionTranslate.X + thirdAxis * sectionTranslate.Y + normal * sectionTranslate.Z);
+
+            }
+            else
+            {
+                Engine.Reflection.Compute.RecordError("Only suitable framing property for the IFramingElement is ConstantFramingProperty.");
```
pawelbaran, line 73: `ConstantFramingProperty` is not the only suitable property for `IFramingElement` as such, it is the only suitable one for this query :)
```suggestion
                Engine.Reflection.Compute.RecordError("Only suitable framing property for the action is ConstantFramingProperty.");
```

pawelbaran (changes requested):
A bit of polishing requested, besides that LGTM!

pawelbaran (approved):
LGTM now 👍 Would be nice if @IsakNaslundBh could have a final look and merge if he is happy too.

Physical_Engine/Query/GeometricalCentreline.cs
```diff
@@ -0,0 +1,75 @@
...
+using System.ComponentModel;
+
+namespace BH.Engine.Physical
+{
+    public static partial class Query
+    {
+        [Description("Returns the centreline of an IFramingElement's section bounding box.")]
+        [Input("element", "The IFramingElement to query the geometrical centreline of.")]
+        [Output("curve", "The geometrical centreline of the IFramingElement.")]
+        public static ICurve GeometricalCentreline(this IFramingElement element)
```
IsakNaslundBh, line 38: Would suggest to change the name of this to `BoundingBoxCentreline` instead as I think that better captures what this does.

pawelbaran: After a short follow-up with @IsakNaslundBh, I agree on the above. I would also think of adding a few words about the bounding box being aligned to the local coordinate system of the bar.

pawelbaran (approved):
LGTM! 👍

PR #2192, opened 2020-11-26
Geometry_Engine: Adding descriptions to the Normal() methods

### LINKED ISSUE #2184: Geometry_Engine: Let Normal() calculate normals for open curves
#### Broken rules:

Inconsistent handling of checks in `Normal()` methods. (See #2180)

Let `Normal()` calculate normals for open curves.

LMarkowski:

Closes #2184

Adding descriptions to all `Normal()` methods.

### Test files

Only the descriptions and arguments names are changed, however [here](https://burohappold.sharepoint.com/:f:/s/BHoM/Esd2eZj9Hc1Mkh5tar8m-NsBlIxJtyJ40zXatKwF4D3uOQ?e=2cGnWq) is the general tet file.

### Changelog

- Adding descriptions to all `Normal()` and `Normals()` methods in the `Geometry_Engine`.

### Additional comments

As this one concerns mostly descriptions, any suggestions from more experienced English speakers will be very welcomed 😉

### DIFF
--- a/Geometry_Engine/Query/Normal.cs
+++ b/Geometry_Engine/Query/Normal.cs
@@ -35,6 +35,10 @@ namespace BH.Engine.Geometry
         /**** Public Methods - Surfaces                 ****/
         /***************************************************/
 
+        [Description("Returns the normal vector to the specified face in the mesh.")]
+        [Input("face", "The Face to get the normal to.")]
+        [Input("mesh", "The Mesh containing the face to get the normal to.")]
+        [Output("normal", "The normal vector to the mesh's specified face.")]
         public static Vector Normal(this Face face, Mesh mesh)
         {
             List<Point> vertices = mesh.Vertices;
@@ -57,6 +61,9 @@ namespace BH.Engine.Geometry
 
         /***************************************************/
 
+        [Description("Returns the list of normal vectors to the mesh's faces.")]
+        [Input("mesh", "The Mesh to get the normals to.")]
+        [Output("normals", "The List of normal vectors to the given mesh's faces.")]
         public static List<Vector> Normals(this Mesh mesh)
         {
             List<Vector> normals = new List<Vector>(mesh.Faces.Count);
@@ -87,6 +94,9 @@ namespace BH.Engine.Geometry
 
         /***************************************************/
 
+        [Description("Returns the list of normal vectors to the mesh's faces.")]
+        [Input("mesh", "The Mesh3D to get the normals to.")]
+        [Output("normals", "The List of normal vectors to the given mesh's faces.")]
         public static List<Vector> Normals(this Mesh3D mesh)
         {
             return mesh.ToMesh().Normals();
@@ -97,17 +107,20 @@ namespace BH.Engine.Geometry
         /**** Public Methods - Curves                   ****/
         /***************************************************/
 
+        [Description("Returns the normal vector to the curve's plane, oriented according to the rigth hand rule. Works only for closed and planar curves.")]
+        [Input("curve", "The Polyline to get the normal to.")]
+        [Input("tolerance", "Distance tolerance used in geometry processing, default set to BH.oM.Geometry.Tolerance.Distance.")]
+        [Output("normal", "The normal vector to the curve's plane.")]
         public static Vector Normal(this Polyline curve, double tolerance = Tolerance.Distance)
         {
-
             if (!curve.IsPlanar(tolerance))
             {
-                Reflection.Compute.RecordError("Input curve must be planar.");
+                Reflection.Compute.RecordError("A single normal vector is not unambiguously definable for non-planar curves.");
                 return null;
             }
             else if (!curve.IsClosed(tolerance))
             {
-                Reflection.Compute.RecordError("Input curve is not closed. Input must be a polygon");
+                Reflection.Compute.RecordError("A single normal vector is not unambiguously definable for open curves.");
                 return null;
             }
             else if (curve.IsSelfIntersecting(tolerance))
@@ -122,7 +135,7 @@ namespace BH.Engine.Geometry
 
             if (normal.Length() < tolerance)
             {
-                Reflection.Compute.RecordError("Couldn't calculate a normal vector of the given curve");
+                Reflection.Compute.RecordError("Couldn't calculate a normal vector of the given curve.");
                 return null;
             }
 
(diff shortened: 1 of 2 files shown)


Geometry_Engine/Query/Normal.cs
```diff
@@ -35,6 +35,10 @@ public static partial class Query
         /**** Public Methods - Surfaces                 ****/
         /***************************************************/
 
+        [Description("Returns the normal vector to the specified face in the mesh.")]
```
pawelbaran, line 38:
```suggestion
        [Description("Returns a vector normal to a given face of the mesh.")]
```

Geometry_Engine/Query/Normal.cs
```diff
@@ -35,6 +35,10 @@ public static partial class Query
...
+        [Input("face", "The Face to get the normal to.")]
+        [Input("mesh", "The Mesh containing the face to get the normal to.")]
+        [Output("normal", "The normal vector to the mesh's specified face.")]
```
pawelbaran, line 41:
```suggestion
        [Output("normal", "Vector normal to the given mesh face.")]
```

Geometry_Engine/Query/Normal.cs
```diff
@@ -57,6 +61,9 @@ public static Vector Normal(this Face face, Mesh mesh)
 
         /***************************************************/
 
+        [Description("Returns the list of normal vectors to the mesh's faces.")]
```
pawelbaran, line 64:
```suggestion
        [Description("Returns the list of vectors normal to each face of the mesh.")]
```

Geometry_Engine/Query/Normal.cs
```diff
@@ -57,6 +61,9 @@ public static Vector Normal(this Face face, Mesh mesh)
...
+        [Input("mesh", "The Mesh to get the normals to.")]
+        [Output("normals", "The List of normal vectors to the given mesh's faces.")]
```
pawelbaran, line 66:
```suggestion
        [Output("normals", "List of vectors normal to the faces of a given mesh.")]
```

Geometry_Engine/Query/Normal.cs
```diff
@@ -87,6 +94,9 @@ public static List<Vector> Normals(this Mesh mesh)
 
         /***************************************************/
 
+        [Description("Returns the list of normal vectors to the mesh's faces.")]
```
pawelbaran, line 97:
```suggestion
        [Description("Returns the list of vectors normal to each face of the mesh.")]
```

Geometry_Engine/Query/Normal.cs
```diff
@@ -87,6 +94,9 @@ public static List<Vector> Normals(this Mesh mesh)
...
+        [Input("mesh", "The Mesh3D to get the normals to.")]
+        [Output("normals", "The List of normal vectors to the given mesh's faces.")]
```
pawelbaran, line 99:
```suggestion
        [Output("normals", "List of vectors normal to the faces of a given mesh.")]
```

Geometry_Engine/Query/Normal.cs
```diff
@@ -97,17 +107,20 @@ public static List<Vector> Normals(this Mesh3D mesh)
         /**** Public Methods - Curves                   ****/
         /***************************************************/
 
+        [Description("Returns the normal vector to the curve's plane, oriented according to the rigth hand rule. Works only for closed and planar curves.")]
```
pawelbaran, line 110: I would apply it here and to all other methods taking curves.
```suggestion
        [Description("Returns the normal vector to the plane of a given curve, oriented according to the right hand rule. Works only for closed, planar curves.")]
```

Geometry_Engine/Query/Normal.cs
```diff
@@ -97,17 +107,20 @@ public static List<Vector> Normals(this Mesh3D mesh)
...
+        [Input("curve", "The Polyline to get the normal to.")]
+        [Input("tolerance", "Distance tolerance used in geometry processing, default set to BH.oM.Geometry.Tolerance.Distance.")]
```
pawelbaran, line 112: Not sure if we should refer to `BH.oM.Geometry.Tolerance.Distance` or to 1e-6? (refers to all methods that have `tolerance` parameter)

Geometry_Engine/Query/Normal.cs
```diff
@@ -97,17 +107,20 @@ public static List<Vector> Normals(this Mesh3D mesh)
...
+        [Output("normal", "The normal vector to the curve's plane.")]
```
pawelbaran, line 113: Minor change, would sound cleared to me (applies to other methods too).
```suggestion
        [Output("normal", "The normal vector to the plane of a curve.")]
```

pawelbaran (commented):
I am not a native speaker, but let myself do a quick check 🙊 Please see the code comments.

Geometry_Engine/Query/Normal.cs
LMarkowski: Had the same thought. It keeps as on the safe side, but isn't very useful for the end user.
It might a bit out of scope here but worth discussing anyway.
Having a permanent value in a string seems.. wrong. Maybe when we'll have the tolerance object we could somehow access it from the descriptions to set it dynamically?

pawelbaran: Maybe we could consider `typeof(ToleranceAttribute)` that would extract the actual value when `BH.Engine.Reflection.Query.Description` gets called? @al-fisher @IsakNaslundBh

pawelbaran (commented):
LGTM 👍  Although would be good to have an action plan regarding `tolerance` before this gets merged.

FraserGreenroyd:
@BHoMBot check compliance

FraserGreenroyd (changes requested):
Needs an appropriate label before merge please.

Geometry_Engine/Query/Normal.cs
FraserGreenroyd: Convention is to use `BH.oM.Geometry.Tolerance.Distance` because that's what the actual default is set to. If we change the value of that tolerance in the oM, then we would have to go through and update all of the descriptions if we used the value `1e-6`.

Hovering over the input will also give the value (at least in Grasshopper, dunno about Dynamo/Excel) when it is an optional input.

FraserGreenroyd (approved):
All LGTM now on brief review.

PR #2195, opened 2020-11-28
Reflection_Engine: Fix Enum display in UI

### LINKED ISSUE #2194: Reflection_Engine: Enum display not displaying value
image

Should be:

image

Seems to be the case for any enum, it is displaying the type of the enum rather than the value. Second screenshot is from Monday's Alpha, first one from today's (Friday).

FraserGreenroyd:
Fixes #2194 

### Test files
image

### Additional comments
Originally was a PR on [Grasshopper_Toolkit](https://github.com/BHoM/Grasshopper_Toolkit/pull/570) but moved here at @adecler request.

@adecler not sure if this is the right fix, this seemed to work (as seen in screenshot above) after compiling this branch against today's alpha (thus discarding the Grasshopper branch).

### DIFF
--- a/Reflection_Engine/Convert/ToText.cs
+++ b/Reflection_Engine/Convert/ToText.cs
@@ -146,7 +146,12 @@ namespace BH.Engine.Reflection
             }
         }
 
+        /***************************************************/
 
+        /*public static string ToText(this Enum e)
+        {
+            
+        }*/
 
         /***************************************************/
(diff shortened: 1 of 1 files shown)


IsakNaslundBh (approved):
Solves it for me. Saw you tried fixing with dynamic casting, but guess that does not work for all enum types, so think this is the right solution then.

PR #2212, opened 2020-12-03
BHoM_Enigne: Remove duplicate FindFragment method

### LINKED ISSUE #2211: BHoM_Engine: Find fragment is not functioning from the UI
#### Description:

https://github.com/BHoM/BHoM_Engine/pull/2105 added another FindFragment method. Having two, with essentially the same method signature, makes non of them work in the UI, leading to pre-exisiting components to behave like this:

image

And adding a new component gives this, picking any of the two options:

image

image

This is something that long term needs to be fixed in the UI but for now, think it needs to be fixed in the Engine.

#### Steps to reproduce:

try create a FindFragment method in a UI

#### Expected behaviour:

For the method do appear/deserialise

#### Test file(s):

Can not provide

IsakNaslundBh:

Closes #2211

Removing the added duplicate non-generic method for FindFragment. Having both causes the UI to crash.

@alelom could not find it used anywhere?

### Test files

Any script that uses find fragment, or try to create a ne one

### DIFF
--- a/BHoM_Engine/Query/FindFragment.cs
+++ b/BHoM_Engine/Query/FindFragment.cs
@@ -52,16 +52,7 @@ namespace BH.Engine.Base
             return (T)System.Convert.ChangeType(fragment, fragmentType);
         }
 
-        [Description("Returns an instance of a BHoM Fragment if it exists on the object")]
-        [Input("iBHoMObject", "A generic IBHoMObject object")]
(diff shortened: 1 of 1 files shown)


IsakNaslundBh:
@BHoMBot check installer

FraserGreenroyd (approved):
As discussed with @IsakNaslundBh LGTM

PR #2217, opened 2020-12-08
Serialiser_Engine: Fix serialisation of non upgraded objects with modified properties

### LINKED ISSUE #2200: Serialiser_Engine: Unable to deserialise non-upgraded space
#### Description:
The Environments Space object has been changed in [this PR](https://github.com/BHoM/BHoM/pull/1108) and is being versioned to a new object in [this PR](https://github.com/BHoM/Versioning_Toolkit/pull/108). However, when testing the versioning, the object fails to deserialise correctly in the `FromJson` component, returning the warning that it has come back out as custom objects.

However, if the JSON is run through the `ToNewVersion` component from Versioning_Engine, and the returned JSON is then passed through the `FromJson` component, the returned object is the correct Space object is provided. This suggests the versioning is happy, but `FromJson` is not correctly calling it.

A brief investigation with @IsakNaslundBh suggests the problem lies in `catch` statement at line `325` of `ObjectSerializer.cs` within Serialiser_Engine, however, I accept this may not be accurate.

During our investigation, it appears that the Serialiser is expecting the `List<LightingGain>` property, however, the JSON contains only the previous `LightingGain` object (not a list - it became a list in the PR mentioned above).

#### Steps to reproduce:
Run the attached script - the `FromJson` component fails to deserialise unless it is plugged through the `ToNewVersion` option.

#### Expected behaviour:
Correct deserialisation of the space object.

#### Test file(s):
[SpaceUpgrade.zip](https://github.com/BHoM/BHoM_Engine/files/5621851/SpaceUpgrade.zip)

Grasshopper file attached has the JSON in a panel. Image below shows current operation.

image

adecler:

Closes #2200

The serialiser was not calling the versioning toolkit for classes that had only seen a change of their properties. This PR fixes that.

### Test files
See issue

### DIFF
--- a/Serialiser_Engine/Objects/BsonSerializers/CustomObjectSerializer.cs
+++ b/Serialiser_Engine/Objects/BsonSerializers/CustomObjectSerializer.cs
@@ -120,7 +120,7 @@ namespace BH.Engine.Serialiser.BsonSerializers
                                 {
                                     Dictionary<string, object> customData = value as Dictionary<string, object>;
                                     if (customData.Count > 0)
-                                        dic["AdditionalData"] = value;
+                                        dic["CustomData"] = value;
                                 }
                                 break;
                             default:
--- a/Serialiser_Engine/Objects/BsonSerializers/DeprecatedSerializer.cs
+++ b/Serialiser_Engine/Objects/BsonSerializers/DeprecatedSerializer.cs
@@ -57,7 +57,18 @@ namespace BH.Engine.Serialiser.BsonSerializers
             BsonDocument doc = bSerializer.Deserialize(context, args) as BsonDocument;
             BsonDocument newDoc = Versioning.Convert.ToNewVersion(doc);
 
-            if (newDoc == null || doc == newDoc)
+            bool overflow = false;
+            if (doc.Contains("BHoM_Guid"))
+            {
+                Guid objectId = doc.GetElement("BHoM_Guid").Value.AsGuid;
+                if (m_StackCounter.ContainsKey(objectId))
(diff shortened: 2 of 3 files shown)


adecler:
I have added a second layer of protection on the scenario covered by this PR to avoid a stackoverflow exception when trying to deserialise an older version of a `Space`.

Here's what was happening: Another change than the latest switch to List properties had to be versioned as well for the json to be valid. To problem is that the custom upgrader that was created is only focussing on taking all the properties that need turning into arrays and place them into one, regardless of whether they are arrays already or not (I hope that makes sense). Basically that custom upgrader is applying the switch to arrays regardless of that fix being the one needed or not. 

That's a fair mistake and the serialiser engine should not crash the UI when that error occurs. Hence this additional commit.

adecler:
Do not merge until I am 100% sure that this PR doesn't break anything else

adecler:
I am now happy with this. 

The case of a `CustomObject` containing a `CustomData` key in their `CustomData` property is still mildly weird but nothing too bad for such an exceptional case. And I believe it should be fixed in the Explode (where the problem happens), not in the serialiser. I have raised an issue here: https://github.com/BHoM/BHoM_Engine/issues/2236

image

At least, it works as expected in the serialisation cycle.

IsakNaslundBh (approved):
Approving based on that this has been the basis of the versioning testing, and me agreing with the results it is giving.

This PR also resolves the issues seen in https://github.com/BHoM/ETABS_Toolkit/pull/366 with the Bars failing.

PR #2218, opened 2020-12-08
Reflection_Engine: Adding ability to remove events from log

adecler:

Change requires by this BHoM_UI PR: https://github.com/BHoM/BHoM_UI/pull/339

### DIFF
--- a/Reflection_Engine/Compute/RemoveEvent.cs
+++ b/Reflection_Engine/Compute/RemoveEvent.cs
@@ -1,0 +1,68 @@
+/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2020, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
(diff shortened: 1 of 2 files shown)


Reflection_Engine/Compute/RemoveEvent.cs
```diff
@@ -0,0 +1,68 @@
...
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        public static bool RemoveEvent(Event newEvent)
+        {
+            Log log = Query.DebugLog();
+            bool success = log.AllEvents.Remove(newEvent);
+            success &= log.CurrentEvents.Remove(newEvent);
+            return success;
```
FraserGreenroyd, line 47: A super minor query, but is it worth using  `return RemoveEvent(new List<Event> { newEvent } );` here so that in the future, if `Log` ends up with more events than just `All` or `Current`, only one method needs updating for the removal of the event? In the long run the code isn't that harsh in this file so I'd like to think people wouldn't miss doing it in both places but as a 'rather safe than sorry' approach?

FraserGreenroyd (approved):
LGTM, minor suggestion that can be ignored if not worth the overhead making it into a list would be :smile:

Have not tested in conjunction with the UI PR though. As a standalone bit of functionality though this LGTM.

Reflection_Engine/Compute/RemoveEvent.cs
adecler: Believe it or not, it was actually a conscious decision. One of my hard core optimisation state of mind moments. Not that I still necessary agree with it the next day but it's fine as it is I think. Very little chance we will ever get more than `All` and `Current` but, if I am wrong, we can change the code then.

PR #2222, opened 2020-12-09
CI: Add CI test for Serialiser_Engine

### LINKED ISSUE #2221: CI: Add test script for the serialiser engine
#### Definition of the test :
This test should run a serialisation cycle on all types available on the BHoM. 
This should be written in C# in the `.ci\code` folder.

adecler:
### NOTE: Depends on 
- https://github.com/BHoM/BHoM/pull/1122
- https://github.com/BHoM/Test_Toolkit/pull/307
   
Closes #2221

This provides the Verification of the `Serialiser_Engine` in a format that can be executed by the BHoMBot.

### Test files

Just run the TestRunner in `ProgramData/BHoM/Assemblies` by providing `Serialiser` as first argument.
If the BHoMBot is already capable of running this, you can also test it that way.

image

### Additional comments
@FraserGreenroyd , I suppose the key thing here is to make sure that you are happy with the general framework on this PR and the two it depends on. From our discussion yesterday, I feel this should be quite close but happy to modify as needed.

### DIFF
--- a/.ci/code/Serialiser_Test/Helpers/TestObjects.cs
+++ b/.ci/code/Serialiser_Test/Helpers/TestObjects.cs
@@ -1,0 +1,55 @@
+﻿using BH.Engine.Reflection;
+using BH.Engine.Serialiser;
+using BH.Engine.Test;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace BH.Test.Serialiser
+{
+    public static partial class Helpers
+    {
+        /*************************************/
+        /**** Public Methods              ****/
+        /*************************************/
+
+        public static List<object> TestObjects(Type type)
+        {
+            if (m_TestObjects.ContainsKey(type))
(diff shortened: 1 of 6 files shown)


FraserGreenroyd (approved):
I like the simplicity here. This is the first of its kind within BHoM so I imagine we might improve as we iterate as we do on most things, but as a starting framework I like this. The use of the `Verify` class helps make it clear what's going to happen, and gives a clear structure to the use of these kinds of files. Obviously we're hoping these will be rare (used for the real nitty-gritty framework level things like Serialisation, Versioning, and Diffing) where data-driven unit tests (and others) aren't as applicable, so keeping the framework structured but simple works for the rare occasion when we will be adding these types of tests.

I've tested this PR in conjunction with the other 2, with different DLLs so different results naturally, but valid results from what I was using and expecting. I'm happy for us to merge this to unblock the next phase of testing work as we wrap up 4.0 and look to 4.1 :smile:

FraserGreenroyd:
@BHoMBot check installer

PR #2224, opened 2020-12-09
BHoM_Engine: Query.Hash() to return a Hash also when an object with all empty or null properties is input

### LINKED ISSUE #2223: BHoM_Engine: Query.Hash fails when an object with all empty or null properties is input.
#### Description:

As title.
Quite an edge case, but of fundamental importance.
Discovered by @FraserGreenroyd and @JohannaOlin1 when using the XML_Toolkit. The Push there triggers the BHoM_Engine method `UniqueConstructions` that failed when constructions with all null or empty properties were input.

#### Steps to reproduce:

See above.

#### Expected behaviour:

A Hash should always be returned, even if all properties are empty or null.

alelom:

Closes #2223

### DIFF
--- a/BHoM_Engine/Query/Hash.cs
+++ b/BHoM_Engine/Query/Hash.cs
@@ -108,7 +108,13 @@ namespace BH.Engine.Base
             string hashString = DefiningString(iObj_copy, cc, fractionalDigits, 0);
 
             if (string.IsNullOrWhiteSpace(hashString))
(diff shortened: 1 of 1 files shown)


JohannaOlin1 (approved):
Worked for the XML test that found the initial error

FraserGreenroyd (approved):
Tested through with @JohannaOlin1 and this works fine for the edge case of objects with empty/no properties! Thanks for the speedy fix @alelom :smile:

PR #2227, opened 2020-12-10
Reflection_Engine: Fix Adapter list to include adapters that don't inherit directly from BHoM_Adapter

### LINKED ISSUE #2226: Reflection_Engine: Adapter are not available in the menus if they don't inherit directly from BHoMAdapter
#### Description:
For example, the GitHub admin adapter inherits from the GitHub adapter

#### Steps to reproduce:

Just try to find teh GitHub admin adapter in teh global menu or do this:
image

#### Expected behaviour:
The GitHub admin adapter should be listed as an adapter

adecler:

Closes #2226

Since the engine cannot know about the adapter type, it was just listing a type as adapter if its direct parent name is `BHoMAdapter` . The problem is that GitHubAdminAdapter is inheriting from GitHubAdapter so it was not listed correctly. This PR fixes that by looking at the full inheritance chain.


FraserGreenroyd (approved):
Have tested with the workflow being prepared for next weeks beta release and this has worked well. Could not see any issue with other adapters during the test. Approving and merging :smile:

PR #2263, opened 2021-01-11
Geometry_Engine: Removing Centre(Polyline)

### LINKED ISSUE #944: Geometry_Engine: Remove Query.Centre(Polyline)
`Query.Centre(Polyline)` is wrong and should be removed (deprecated). Any references to be changed to `Centroid`.

LMarkowski:
### NOTE: Depends on 

Pull requests removing references to `Centre()`:
- [XML_Toolkit](https://github.com/BHoM/XML_Toolkit/pull/519)
- [IES_Toolkit](https://github.com/BHoM/IES_Toolkit/pull/168)
 
  
   
Closes #944 

Deprecating `Query.Centre(Polyline)` as it is generally incorrect. All references are being changed to `Centroid()`.

### Test files

Not needed as it doesn't affect a logic of any method, however [here](https://burohappold.sharepoint.com/:u:/s/BHoM/ETgDDrmqBoZNvApme34KNDcBOT7zjI5DWfZlWrTE3oOcAA?e=Jr2EtM) is the script I used to check if any reference has been missed.

### Changelog

- Deprecating `Query.Centre(Polyline)` method in the `Geometry_Engine`.
- Updating methods in `Environment_Engine` to call for an average non-duplicated control point.

### DIFF
--- a/Environment_Engine/Query/BottomLeft.cs
+++ b/Environment_Engine/Query/BottomLeft.cs
@@ -60,13 +60,12 @@ namespace BH.Engine.Environment
             Vector normal = polyline.Normal();
             if (!polyline.NormalAwayFromSpace(panelsAsSpace))
                 normal = polyline.Flip().Normal();
+            if (normal == null)
+                return null;
 
             Point centre = polyline.Centroid();
             if (centre == null)
                 centre = polyline.ControlPoints.CullDuplicates().Average();
-
-            if (normal == null)
-                return null;
 
             Line line = new Line
             {
--- a/Environment_Engine/Query/BottomLeft.cs
+++ b/Environment_Engine/Query/BottomLeft.cs
@@ -63,9 +63,9 @@ namespace BH.Engine.Environment
 
             Point centre = polyline.Centroid();
             if (centre == null)
-                centre = polyline.Centre();
+                centre = polyline.ControlPoints.CullDuplicates().Average();
 
-            if(normal == null)
+            if (normal == null)
                 return null;
 
             Line line = new Line
--- a/Environment_Engine/Query/BottomRight.cs
+++ b/Environment_Engine/Query/BottomRight.cs
@@ -60,13 +60,12 @@ namespace BH.Engine.Environment
             Vector normal = polyline.Normal();
             if (!polyline.NormalAwayFromSpace(panelsAsSpace))
                 normal = polyline.Flip().Normal();
+            if (normal == null)
+                return null;
 
             Point centre = polyline.Centroid();
             if (centre == null)
                 centre = polyline.ControlPoints.CullDuplicates().Average();
-
-            if (normal == null)
-                return null;
 
             Line line = new Line
             {
--- a/Environment_Engine/Query/BottomRight.cs
+++ b/Environment_Engine/Query/BottomRight.cs
@@ -63,9 +63,9 @@ namespace BH.Engine.Environment
 
             Point centre = polyline.Centroid();
             if (centre == null)
-                centre = polyline.Centre();
+                centre = polyline.ControlPoints.CullDuplicates().Average();
 
-            if(normal == null)
+            if (normal == null)
                 return null;
 
             Line line = new Line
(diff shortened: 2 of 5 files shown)


LMarkowski:
@BHoMBot check compliance

LMarkowski:
@BHoMBot check compliance

Geometry_Engine/Query/Centre.cs
```diff
@@ -40,6 +39,7 @@ public static Point Centre(this Arc arc, double tolerance = Tolerance.Distance)
 
         /***************************************************/
 
+        [ToBeRemoved("4.1", "To be removed as it is generally incorrect. Advising using Centroid instead.")]
```
FraserGreenroyd, line 42: Could we do some versioning so that scripts automatically change from `centre` to `centroid`? @IsakNaslundBh what do you think in terms of UX on that?

Environment_Engine/Query/BottomLeft.cs
```diff
@@ -63,9 +63,9 @@ public static Point BottomLeft(this Polyline polyline, List<Panel> panelsAsSpace
 
             Point centre = polyline.Centroid();
             if (centre == null)
-                centre = polyline.Centre();
+                centre = polyline.ControlPoints.CullDuplicates().Average();
 
-            if(normal == null)
+            if (normal == null)
```
FraserGreenroyd, line 68: While we're here... could we move this to happen after line 62? Seems redundant to get the vector and then do a computation before returning `null` if the vector was `null`.

Environment_Engine/Query/BottomRight.cs
```diff
@@ -63,9 +63,9 @@ public static Point BottomRight(this Polyline polyline, List<Panel> panelsAsSpac
 
             Point centre = polyline.Centroid();
             if (centre == null)
-                centre = polyline.Centre();
+                centre = polyline.ControlPoints.CullDuplicates().Average();
 
-            if(normal == null)
+            if (normal == null)
```
FraserGreenroyd, line 68: Same here can we move this to be after line 62

Environment_Engine/Query/TopRight.cs
```diff
@@ -63,7 +63,7 @@ public static Point TopRight(this Polyline polyline, List<Panel> panelsAsSpace)
 
             Point centre = polyline.Centroid();
             if (centre == null)
-                centre = polyline.Centre();
+                centre = polyline.ControlPoints.CullDuplicates().Average();
 
             if(normal == null)
```
FraserGreenroyd, line 68: Same here, can we move this to be after line 62

FraserGreenroyd (changes requested):
Very minor changes @LMarkowski - otherwise happy with this PR and would be happy to approve/merge pending discussion on the versioning mentioned :smile:

FraserGreenroyd:

Geometry_Engine/Query/Centre.cs
LMarkowski: Wouldn't `ReplacedAttribute` do the trick here?

FraserGreenroyd: I think it would, do you agree we need it? I'm happy with a `ToBeRemoved` but I'm wondering if changing to `centroid` or `average` is a better UX is all. Maybe `average` as that's closest to what the original method was doing? Though `centroid` might be more related to what people _thought_ it was doing?

IsakNaslundBh: I actually think keeping it as is, with the `ToBeRemoved`, makes most sense. Given, as you stated above, that this functionally could be replaced with average of controlpoints, and semantically with the centroid, I think simply deprecating it (with ToBeRemoved) seems like the right thing to do.

Adding versioning so that Centroid was called instead of centre would ofc be trivial from the code side, but would mean a functional change that I am not sure users would want.

TL;DR I think what Lukasz has done makes sense.

FraserGreenroyd: Ok, SGTM, marking as resolved, approving and merging :smile:

FraserGreenroyd (approved):

FraserGreenroyd:

FraserGreenroyd:

PR #2273, opened 2021-01-22
Versioning_Engine: Add support for custom message on items without an upgrade

adecler:
Supports this PR in the Versioning Toolkit: https://github.com/BHoM/Versioning_Toolkit/pull/134

See that PR for more details

### DIFF
--- a/Versioning_Engine/Convert/ToNewVersion.cs
+++ b/Versioning_Engine/Convert/ToNewVersion.cs
@@ -91,7 +91,12 @@ namespace BH.Engine.Versioning
                 // Get the new version back
                 BsonDocument result = ReadDocument(pipe);
                 if (result != null)
-                    document = result;
+                {
+                    if (result.Contains("_t") && result["_t"] == "NoUpdate" && result.Contains("Message"))
+                        Engine.Reflection.Compute.RecordError(result["Message"].ToString());
+                    else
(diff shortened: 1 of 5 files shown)


Versioning_Engine/Convert/ToNewVersion.cs
```diff
@@ -91,7 +91,12 @@ public static BsonDocument ToNewVersion(this BsonDocument document, string versi
                 // Get the new version back
                 BsonDocument result = ReadDocument(pipe);
                 if (result != null)
-                    document = result;
+                {
+                    if (result.Contains("_t") && result["_t"] == "NoUpdate" && result.Contains("Message"))
```
FraserGreenroyd, line 95: Maybe a dumb question from not having read in detail everything on the Versioning_TK PR, but if the result contains `_t` and that type is of `NoUpdate` but the result does not contain a message, do we still want to set the `document = result;`?

adecler: Thanks @FraserGreenroyd, Good catch. That should never happen since we are responsible for both end of the pipeline but better safe than sorry 😄 . Code updated

adecler:
@BHoMBot check copyright

adecler:
@BHoMBot check installer

PR #2279, opened 2021-01-25
Geometry_Engine: Fix Create.Grid(origin, direction, length) not projecting properly

### LINKED ISSUE #2278: Geometry_Engine: Create.Grid(origin, direction, length) only partially projects to XY plane
#### Description:

image

When using the (origin, direction, length) create methods for Grids, the origin point will be projected to the XY plane when used as the start point, but not when the direction vector is added to calculate the end point, resulting in the behaviour you see here.

#### Expected behaviour:

Resulting grid line should be in the XY plane.

IsakLarbornBH:

Closes #2278

Fixes `Create.Grid(origin, direction, length)` so that grid lines will now be properly projected to the XY plane even when `Point origin` is not. Previously only the start point would be projected.

I also took the liberty of deprecating the methods without the `String name` parameter, since I didn't see a need for them to exist when you can just default name to `""`. Versioning for this works perfectly and I don't see any issues, but we can roll that commit back if that was a bit too much.

### Test files

Here's a simple test file that also tests versioning since it contains methods that lack the name parameter. 
[Grid test script.zip](https://github.com/BHoM/BHoM_Engine/files/5866653/Grid.test.script.zip)
I've also added unit tests which are more thorough.

### Changelog
* Fixed `Create.Grid()` vector methods so that they now properly project to the XY plane.
* Removed `Create.Grid()` methods without a name parameter and added a default to name.


IsakLarbornBH:
My first attempt at unit tests, here we go:

@BHoMBot check unit tests

IsakNaslundBh:
@BHoMBot check unit tests

IsakNaslundBh:
@BHoMBot check compliance

IsakNaslundBh:
Cant see why the compliance check is complaining about the previous attribute methods in here. They do look fine to me. Will have another look Monday. Think the code looks fine!

IsakLarbornBH:
> Cant see why the compliance check is complaining about the previous attribute methods in here. They do look fine to me. Will have another look Monday. Think the code looks fine!

Great, thanks! I can see it's also complaining about the lack of descriptions, which makes sense - maybe it wouldn't be a terrible idea to add some of those since this PR is changing it up quite a bit anyway.

IsakLarbornBH:
@BHoMBot check unit tests

IsakLarbornBH:
@BHoMBot check compliance

IsakNaslundBh (approved):
All looks fine by me! Good work on the descs as well!

PR #2283, opened 2021-01-26
Added Create method to specify Domain extremes inclusion.

### LINKED ISSUE #2282: Data_Engine: Add Create method for Domain to align to new feature "extremesInclusion"
#### Description:

As title, see https://github.com/BHoM/BHoM/pull/1165

alelom:
### NOTE: Depends on 

https://github.com/BHoM/BHoM/pull/1165
   
Closes #2282

### Test files

Check if versioning upgrades correctly the `Create.Domain` components here:
https://burohappold.sharepoint.com/:f:/s/BHoM/ElNd6u9leC9Pn5My-bl1R7YBsgG5jOcEvbb95CzLUm5CCA?e=vjmYuV

### DIFF
--- a/Data_Engine/Create/Domain.cs
+++ b/Data_Engine/Create/Domain.cs
@@ -41,17 +41,9 @@ namespace BH.Engine.Data
         [Description("Creates a numerical domain around the two values.")]
         [Input("min", "The lowest value in the domain, will be set as max if larger than max. Defaults to -infinite (double.MinValue).")]
         [Input("max", "The highest value in the domain, will be set as min if larger than min. Defaults to +infinite (double.MaxValue).")]
-        [Output("domain", "A numerical domain with its extreme values at min and max.")]
-        public static Domain Domain(double min = double.MinValue, double max = double.MaxValue)
-        {
-            return new Domain(Math.Min(min, max), Math.Max(min, max));
-        }
-
-        [Description("Creates a numerical domain around the two values.")]
-        [Input("min", "The lowest value in the domain, will be set as max if larger than max. Defaults to -infinite (double.MinValue).")]
-        [Input("max", "The highest value in the domain, will be set as min if larger than min. Defaults to +infinite (double.MaxValue).")]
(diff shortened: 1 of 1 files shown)


Data_Engine/Create/Domain.cs
```diff
@@ -39,14 +39,24 @@ public static partial class Create
...
         {
             return new Domain(Math.Min(min, max), Math.Max(min, max));
         }
 
+        [Description("Creates a numerical domain around the two values.")]
+        [Input("min", "The lowest value in the domain, will be set as max if larger than max. Defaults to -infinite (double.MinValue).")]
+        [Input("max", "The highest value in the domain, will be set as min if larger than min. Defaults to +infinite (double.MaxValue).")]
+        [Input("extremesInclusion", "Whether the Min and Max extremes are included in the Domain or not. `IncludeBoth` by default.")]
+        [Output("domain", "A numerical domain with its extreme values at min and max.")]
+        public static Domain Domain(double min = double.MinValue, double max = double.MaxValue, BH.oM.Data.DomainExtremesInclusion extremesInclusion = BH.oM.Data.DomainExtremesInclusion.IncludeBoth)
```
FraserGreenroyd, line 55: Given that both have default options across the board, what is the benefit of both create methods? Seems like it'd be better to just have one?

FraserGreenroyd (commented):
> Test files
> Not needed.

I disagree :wink: Especially given I don't think you want two methods here, which means you'll want to do some versioning which will inherently need a test file :smile:

Data_Engine/Create/Domain.cs
adecler: That and the fact that the system will not be able to decide which method to call for less than three arguments.
Agreed that the first one should not exist.

alelom: @adecler , @FraserGreenroyd thanks and agreed - removed and added versioning, plus test script.

alelom:
Parked for now.

PR #2298, opened 2021-02-04
Data_Engine: Methods to flatten and split the trees made of nested ILogicalRequests added

### LINKED ISSUE #2296: Data_Engine: Add method to flatten the tree of ILogicalRequests
#### Description:

The `IRequests` can be joined into logical and/or queries using `ILogicalRequests`. `ILogicalRequests` can stack up creating trees, that in turn can be degenerate (e.g. a `LogicalAndRequest` consisting of one `IRequest` is in practice the latter). It would be nice to have a method to flatten the trees, i.e. merge `LogicalAndRequest` embedded into `LogicalAndRequest` of the same type or squash the ones with only one item.

### LINKED ISSUE #2297: Data_Engine: Add a method to split the tree of ILogicalRequests into a set of disjoint trees with same total meaning, but with max. one item of a given type
#### Description:

The `IRequests` can be joined into logical and/or queries using `ILogicalRequests`. `ILogicalRequests` can stack up creating trees with multiple levels. It may happen that `IRequests` of given type would not be allowed to exist in a single tree because each of them needs to have its _own_ tree to evaluate (e.g. in Revit_Toolkit that will be the requirement for `LinkRequests`, where each link document will need to have a separate request tree). Therefore, we need a method that would decompose a tree with multiple items of given type into a set of trees with only one item, which would in total represent the same request as the original tree.

Naturally, this will only work if the items of given type are linked with `LogicalOrRequest` (`LogicalAndRequest` imposes a direct conflict - the tree cannot be disjoint).

pawelbaran:

Closes #2296
Closes #2297

### Test files

On [SharePoint](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?RootFolder=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FIssue%2FBHoM%2FBHoM%5FEngine%2FData%5FEngine&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4). I had hard times coming up with automated way of generating and testing the data, therefore hand-made trees are provided for visual inspection. Happy to improve the tests if a smarter way is suggested.

### Changelog

- added method to simplify trees made of nested `ILogicalRequests`
- added method to split trees made of nested `ILogicalRequests` in a way that ensures that each resultant tree contains max. one `IRequest` of given type and all resultant trees combined constitute same request as the original tree
- added a method to check whether a tree created by a set of nested `ILogicalRequests` contains two or more `IRequests` of given type that can conflict (fall into a logical and statement)

### Additional comments

The functionality can look a bit abstract without a use case - this will follow in Revit_Toolkit, so please focus on the correctness of the methods for now 😃 

I was convinced the problem will be rather easy to solve (rather simple tree traversal), but it turned to be more tricky than it looked, mainly due to the existence of `LogicalNotRequest` that stores single `IRequest`, not a list. The code got a bit convoluted, but it works (I hope 🙈), happy to accept suggestions on potential simplifications.

### DIFF
--- a/Data_Engine/Compute/SplitRequestByType.cs
+++ b/Data_Engine/Compute/SplitRequestByType.cs
@@ -1,0 +1,152 @@
+﻿/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2021, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 
+ * GNU Lesser General Public License for more details.                          
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using BH.Engine.Base;
+using BH.oM.Base;
+using BH.oM.Data.Requests;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace BH.Engine.Data
+{
+    public static partial class Compute
+    {
+        /***************************************************/
+        /****              Public methods               ****/
+        /***************************************************/
+
+        public static List<IRequest> SplitRequestTreeByType(IRequest request, Type splittingType)
+        {
+            if (!typeof(IRequest).IsAssignableFrom(splittingType))
+            {
+                BH.Engine.Reflection.Compute.RecordError($"Type {splittingType} does not implement {nameof(IRequest)} interface.");
+                return null;
+            }
+
+            if (typeof(ILogicalRequest).IsAssignableFrom(splittingType))
+            {
+                BH.Engine.Reflection.Compute.RecordError($"It is not allowed to split by types that implement {nameof(ILogicalRequest)} interface.");
+                return null;
+            }
+
+            if (request.IsPotentialOverlap(splittingType))
+            {
+                BH.Engine.Reflection.Compute.RecordError($"The request could not be split by type {splittingType} because there is a potential logical AND overlap between two requests of the given type.");
+                return null;
+            }
+
+            IRequest flattened = request.FlattenRequestTree();
+            List<IRequest> extracted = new List<IRequest>();
+            flattened.ExtractTrees(splittingType, extracted, new List<IRequest>());
+
+            extracted = extracted.Select(x => x.FlattenRequestTree()).ToList();
+            extracted.Add(flattened.FlattenRequestTree());
+
+            return extracted.Where(x => x != null).ToList();
+        }
+
+
+        /***************************************************/
+        /****              Private methods              ****/
+        /***************************************************/
+
+        private static void ExtractTrees(this IRequest request, Type typeToExtract, List<IRequest> extracted, List<IRequest> history)
+        {
+            List<IRequest> newHistory = new List<IRequest>(history);
+
+            Type type = request.GetType();
+            if (type == typeToExtract)
+            {
+                extracted.Add(request.Extract(newHistory));
+
+                if (history.Count != 0)
+                    ((ILogicalRequest)history.Last()).IRequests().Remove(request);
+
+                return;
+            }
+
+            newHistory.Add(request);
+
+            if (request is LogicalAndRequest)
+            {
+                List<IRequest> subRequests = ((LogicalAndRequest)request).Requests;
+                IRequest found = subRequests.FirstOrDefault(x => x.GetType() == typeToExtract);
+                if (found != null)
+                {
+                    extracted.Add(found.Extract(newHistory));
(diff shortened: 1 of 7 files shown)


alelom (dismissed):
The code looks good and works. There is some overlap with some work we are doing in the combining of Conditions in CIH, and we discussed some refactoring of the Requests to be done in the future, and some work of this PR may be part of it then.

pawelbaran:
@BHoMBot check copyright

IsakNaslundBh:
@BHoMBot check compliance

pawelbaran:
@FraserGreenroyd could you have a look at this?
- branch name is just fine (did @BHoMBot go mad?)
- I would opt for keeping the private `enum` inside the class - alternatively can change it to an integer (in the end the values of the enumeration represent 0, 1 and more occurences), but this will make the code less readable

FraserGreenroyd:
> * branch name is just fine (did @BHoMBot go mad?)

See previous reply on the linked PR regarding this.

> * I would opt for keeping the private `enum` inside the class - alternatively can change it to an integer (in the end the values of the enumeration represent 0, 1 and more occurences), but this will make the code less readable

I disagree, this is heavily against BHoM guidelines of one class per file and making objects/methods easy to find when reflected from a UI. If there is something which does not need to be within a UI, but is required to make an engine method work, that is where the `Objects` folder comes in within the Engine. I believe Structure_Engine has an example of this in use that could be replicated here and have the enum moved to its own file there :smile:

FraserGreenroyd (changes requested):
PR title needs to include engine changed.

Compliance needs resolving, see previous comment.

pawelbaran:
> > * I would opt for keeping the private `enum` inside the class - alternatively can change it to an integer (in the end the values of the enumeration represent 0, 1 and more occurences), but this will make the code less readable
> 
> I disagree, this is heavily against BHoM guidelines of one class per file and making objects/methods easy to find when reflected from a UI. If there is something which does not need to be within a UI, but is required to make an engine method work, that is where the `Objects` folder comes in within the Engine. I believe Structure_Engine has an example of this in use that could be replicated here and have the enum moved to its own file there 😄

But `enum` is not a class? On top of that, it is `private` and makes absolutely zero sense to be reused anywhere else, being very problem-specific, so putting it in _Objects_ is pointless too. Therefore, I see two options here:
a) accept it as an edge case and update the compliance rules to accept it, when time allows (IMO that would be fully valid)
b) change the enum into an `int`, make the code less readable but keep the compliance happy

As usual in such cases, my opinion is that compliance checks should be for developers, not developers for compliance checks @FraserGreenroyd.

FraserGreenroyd:
> But `enum` is not a class? On top of that, it is `private` and makes absolutely zero sense to be reused anywhere else, being very problem-specific, so putting it in _Objects_ is pointless too.

`enum` is a special class representing constants. In these terms, it is treated in the same way as `interface`. If it is not intended to be reused anywhere, then the `Objects` folder is ideal for it. It allows the class to have its own file which avoids hiding it/obfuscating it for future maintainers, as is BHoM's philosophy.

> a) accept it as an edge case and update the compliance rules to accept it, when time allows (IMO that would be fully valid)

This is not an acceptable edge case so I would not be prepared to update the rules that check it at this stage unless there is additional evidence to this being necessary and beneficial.

> b) change the enum into an `int`, make the code less readable but keep the compliance happy

I think the move to `Objects` is cleaner, but changing to an `int` is acceptable but that is just a box-ticking at that point. The wider philosophy should be considered for the decision.

> As usual in such cases, my opinion is that compliance checks should be for developers, not developers for compliance checks @FraserGreenroyd.

It's important to remember the guidelines we have that BHoMBot enforces are to help keep the code true to BHoM's core principles of being accessible and maintainable by those without heavy coding backgrounds. Hiding classes (enums, interfaces) within classes begins to break down that ease of access philosophy. There are concepts of software development that BHoM has actively chosen to avoid for various reasons. That said, even in traditional software development, I don't know of many places that would enjoy the practice of hiding things within things, even in software houses most style guides will say to keep classes separate unless completely unavoidable :smile:

pawelbaran:
> > But `enum` is not a class? On top of that, it is `private` and makes absolutely zero sense to be reused anywhere else, being very problem-specific, so putting it in _Objects_ is pointless too.
> 
> `enum` is a special class representing constants. In these terms, it is treated in the same way as `interface`. If it is not intended to be reused anywhere, then the `Objects` folder is ideal for it. It allows the class to have its own file which avoids hiding it/obfuscating it for future maintainers, as is BHoM's philosophy.

That is your opinion I would argue with. First of all, `enum` is a value type, so rather not too close to types declared using standard classes or interfaces. As far as I know, _Objects_ folder is used to store Engine-specific objects, not enums - to me, a private `enum` would look there at least weird and would not improve clarity at all. So again, please do consider and respect the fact that others may have different opinions on things that were never discussed in detail and verified. We have been there with other compliance issues before, @IsakNaslundBh remembers for sure.

> 
> > a) accept it as an edge case and update the compliance rules to accept it, when time allows (IMO that would be fully valid)
> 
> This is not an acceptable edge case so I would not be prepared to update the rules that check it at this stage unless there is additional evidence to this being necessary and beneficial.

Please see the reply above.

> 
> > b) change the enum into an `int`, make the code less readable but keep the compliance happy
> 
> I think the move to `Objects` is cleaner, but changing to an `int` is acceptable but that is just a box-ticking at that point. The wider philosophy should be considered for the decision.

That is what will happen, credits to you 😉

pawelbaran:
I have pushed the change, not too much of a disaster after renaming the method param @FraserGreenroyd - a well-named `int` looks like a fair compromise in this case.

@alelom could you please re-review? Please note the scope of changes made in my commit, likely not much re-testing should be needed. Thanks!

pawelbaran:
@BHoMBot check compliance

pawelbaran:
@BHoMBot check installer

alelom (approved):
re-reviewing, LGTM.

FraserGreenroyd (approved):
All happy now, thanks @pawelbaran :smile:

PR #2311, opened 2021-02-15
Data_Engine: Handling of nested request issues fixed

### LINKED ISSUE #2310: Data_Engine: nested request handling issues
#### Description:

A few issues related to handling of requests nested in `ILogicalRequests` have been found when working on https://github.com/BHoM/Revit_Toolkit/issues/396:
- running `AllRequestsOfType` on `LogicalNotRequests` with null child
- running `SimplifyRequestTree` on `LogicalOrRequest`/`LogicalAndRequest` with a single child nested inside `LogicalNotRequest`
- running `SplitRequestTreeByType` on `LogicalNotRequest` with a child `IRequest` of the same type as splitting type

pawelbaran:

Closes #2310

### Test files

[This](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4&id=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FIssue%2FBHoM%2FBHoM%5FEngine%2FData%5FEngine%2F%232310%2DNestedRequestIssues%2Egh&parent=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FIssue%2FBHoM%2FBHoM%5FEngine%2FData%5FEngine) is an issue-specific file. However, all other files from [this folder](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?RootFolder=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FIssue%2FBHoM%2FBHoM%5FEngine%2FData%5FEngine&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4) should still work correctly.

#### Description:

- issues related to handling of nested requests have been fixed:
    - running `AllRequestsOfType` on `LogicalNotRequests` with null child
    - running `SimplifyRequestTree` on `LogicalOrRequest`/`LogicalAndRequest` with a single child nested inside `LogicalNotRequest`
    - running `SplitRequestTreeByType` on `LogicalNotRequest` with a child `IRequest` of the same type as splitting type

### DIFF
--- a/Data_Engine/Compute/SplitRequestTreeByType.cs
+++ b/Data_Engine/Compute/SplitRequestTreeByType.cs
@@ -67,8 +67,11 @@ namespace BH.Engine.Data
                 simplified.ExtractTrees(splittingType, extracted, new List<IRequest>());
                 extracted = extracted.Select(x => x.SimplifyRequestTree()).ToList();
 
-                if (simplified.AllRequestsOfType(splittingType).Count == 0)
-                    extracted.Add(simplified.SimplifyRequestTree());
+                if (!(simplified is LogicalNotRequest) || ((LogicalNotRequest)simplified).Request != null)
+                {
+                    if (simplified.AllRequestsOfType(splittingType).Count == 0)
+                        extracted.Add(simplified.SimplifyRequestTree());
+                }
             }
             else
                 extracted.Add(simplified);
--- a/Data_Engine/Compute/SplitRequestTreeByType.cs
+++ b/Data_Engine/Compute/SplitRequestTreeByType.cs
@@ -69,7 +69,7 @@ namespace BH.Engine.Data
 
                 simplified = simplified.SimplifyRequestTree();
                 if (simplified != null)
-                    extracted.Add(simplified.SimplifyRequestTree());
+                    extracted.Add(simplified);
             }
             else
                 extracted.Add(simplified);
--- a/Data_Engine/Compute/SplitRequestTreeByType.cs
+++ b/Data_Engine/Compute/SplitRequestTreeByType.cs
@@ -67,11 +67,9 @@ namespace BH.Engine.Data
                 simplified.ExtractTrees(splittingType, extracted, new List<IRequest>());
                 extracted = extracted.Select(x => x.SimplifyRequestTree()).ToList();
 
-                if (!(simplified is LogicalNotRequest) || ((LogicalNotRequest)simplified).Request != null)
-                {
-                    if (simplified.AllRequestsOfType(splittingType).Count == 0)
-                        extracted.Add(simplified.SimplifyRequestTree());
-                }
+                simplified = simplified.SimplifyRequestTree();
+                if (simplified != null)
+                    extracted.Add(simplified.SimplifyRequestTree());
             }
             else
                 extracted.Add(simplified);
--- a/Data_Engine/Modify/SimplifyRequestTree.cs
+++ b/Data_Engine/Modify/SimplifyRequestTree.cs
@@ -93,12 +93,17 @@ namespace BH.Engine.Data
                                 subRequests.RemoveAt(i);
                             else if (subSub.Count == 1 && !(logical is LogicalNotRequest))
                             {
-                                subRequests.RemoveAt(i);
-
-                                if (subSub[0].GetType() == request.GetType())
-                                    subRequests.InsertRange(i, ((ILogicalRequest)subSub[0]).IRequests());
+                                if (request is LogicalNotRequest)
+                                    ((LogicalNotRequest)request).Request = subSub[0];
                                 else
-                                    subRequests.Insert(i, subSub[0]);
+                                {
+                                    subRequests.RemoveAt(i);
+
+                                    if (subSub[0].GetType() == request.GetType())
+                                        subRequests.InsertRange(i, ((ILogicalRequest)subSub[0]).IRequests());
+                                    else
+                                        subRequests.Insert(i, subSub[0]);
+                                }
                             }
                         }
                     }
--- a/Data_Engine/Query/AllRequestsOfType.cs
+++ b/Data_Engine/Query/AllRequestsOfType.cs
@@ -58,7 +58,7 @@ namespace BH.Engine.Data
 
         private static void RequestsOfType(this IRequest request, Type requestType, List<IRequest> instances)
         {
-            Type type = request.GetType();
+            Type type = request?.GetType();
             if (type == requestType)
                 instances.Add(request);
 
--- a/Data_Engine/Query/Requests.cs
+++ b/Data_Engine/Query/Requests.cs
@@ -59,7 +59,11 @@ namespace BH.Engine.Data
 
         public static List<IRequest> Requests(this LogicalNotRequest request)
         {
-            return new List<IRequest> { request.Request };
+            List<IRequest> result = new List<IRequest>();
+            if (request.Request != null)
+                result.Add(request.Request);
+
+            return result;
         }
 
         /***************************************************/


adecler (changes requested):
I am not a big fan of the changes made in this PR. The issue you are trying to fixed is cause by `AllRequestsOfType` not by `SplitRequestTreeByType`. 
I got the test file you provided working by making the following changes on the master instead:

image

pawelbaran:
> I am not a big fan of the changes made in this PR. The issue you are trying to fixed is cause by `AllRequestsOfType` not by `SplitRequestTreeByType`.

Not really @adecler: the issue consists of 3 sub-issues, your suggested change seems to be fixing only the 1st from what I can see. The other two still require intervention (and therefore I am changing three files instead of one).

From what I can deduct from reading the code, the change proposed by you in `AllRequestsOfType` will provide the same behaviour as the change introduced in this PR (you moved the null check before the recursive call).

By the way, I agree that the code looks pretty convoluted as for a relatively simple tree handling problem, but I could not make it any better due to the fact that some `ILogicalRequests` store children in a list, while `LogicalNotRequest` has a single child property - that is the origin of the problem. I will be more than happy to simplify the code, so further improvement suggestions are more than welcome 👍

adecler:
> Not really @adecler: the issue consists of 3 sub-issues, your suggested change seems to be fixing only the 1st from what I can see. The other two still require intervention (and therefore I am changing three files instead of one).
>
> From what I can deduct from reading the code, the change proposed by you in `AllRequestsOfType` will provide the same behaviour as the change introduced in this PR (you moved the null check before the recursive call).

It solved the two red components in the example file for me so didn't look further (difficult to debug something that is already working). Then consider this change request limited to the `SplitRequestTreeByType` file. Fixes should be made inside the method causing problem, not in calling methods. Otherwise, the problem will reappear every time you have a new method calling it.

> By the way, I agree that the code looks pretty convoluted as for a relatively simple tree handling problem, but I could not make it any better due to the fact that some `ILogicalRequests` store children in a list, while `LogicalNotRequest` has a single child property - that is the origin of the problem. I will be more than happy to simplify the code, so further improvement suggestions are more than welcome 👍 

Yes, I am not a big fan of having to constantly check if an `Irequest` is an `ILogicalRequest`. What if you add more types of requests ? Will you have to constantly add more checks everywhere `IRequests` are being used ? This is why it is important to keep the needs for those check as local as possible. `AllRequestsOfType` is playing the role of the getter so you could have a similar method that would play the role of the setter. So `SimplifyRequestTree` wouldn't have to do those checks itself.

pawelbaran:
> > Not really @adecler: the issue consists of 3 sub-issues, your suggested change seems to be fixing only the 1st from what I can see. The other two still require intervention (and therefore I am changing three files instead of one).
> > From what I can deduct from reading the code, the change proposed by you in `AllRequestsOfType` will provide the same behaviour as the change introduced in this PR (you moved the null check before the recursive call).
> 
> It solved the two red components in the example file for me so didn't look further (difficult to debug something that is already working). Then consider this change request limited to the `SplitRequestTreeByType` file. Fixes should be made inside the method causing problem, not in calling methods. Otherwise, the problem will reappear every time you have a new method calling it.

Yeah, and the fixes are made inside the methods causing problems, nowhere else - there is simply 3 methods causing different problems, not one (see the issue). And the problem is not only the component turning red, it is also about the output being wrong, that is why there is a few scripts, not one.

> 
> > By the way, I agree that the code looks pretty convoluted as for a relatively simple tree handling problem, but I could not make it any better due to the fact that some `ILogicalRequests` store children in a list, while `LogicalNotRequest` has a single child property - that is the origin of the problem. I will be more than happy to simplify the code, so further improvement suggestions are more than welcome 👍
> 
> Yes, I am not a big fan of having to constantly check if an `Irequest` is an `ILogicalRequest`. What if you add more types of requests ? Will you have to constantly add more checks everywhere `IRequests` are being used ? This is why it is important to keep the needs for those check as local as possible. `AllRequestsOfType` is playing the role of the getter so you could have a similar method that would play the role of the setter. So `SimplifyRequestTree` wouldn't have to do those checks itself.

Actually, the problem is not only the existence of `ILogicalRequest`, but also `LogicalNotRequest` that stores children in a different structure. This mismatch makes it really hard to make the code more simple (and I am afraid a setter method will not help much here).

To push this PR forward, I would opt for treating it as a bugfix and no more than that, not to keep it hanging. Afterwards, I am happy to raise an issue to discuss simplification and possible wider refactoring of the `IRequests`, which I believe is beyond the scope of this PR.

adecler:
> To push this PR forward, I would opt for treating it as a bugfix and no more than that, not to keep it hanging. Afterwards, I am happy to raise an issue to discuss simplification and possible wider refactoring of the IRequests, which I believe is beyond the scope of this PR.

Agreed, I just provided a suggestion since you were asking for one but don't expect it to be part of this PR.

I would sill like to see like to see the fix on `SplitRequestTreeByType` to be done in `AllRequestsOfType` instead though as it is currently made in the incorrect location

pawelbaran:
> I would still like to see the fix on `SplitRequestTreeByType` to be done in `AllRequestsOfType` instead though as it is currently made in the incorrect location

`SplitRequestTreeByType` requires a fix that is non-related to `AllRequestsOfType`, it is about handling of `LogicalNotRequest` a bit more explicitly than before 😞

pawelbaran:
I have pushed the discussed fixes @adecler - not sure if this can be made look much better without major refactoring 🙈

Data_Engine/Compute/SplitRequestTreeByType.cs
```diff
@@ -67,7 +67,8 @@ public static List<IRequest> SplitRequestTreeByType(this IRequest request, Type
                 simplified.ExtractTrees(splittingType, extracted, new List<IRequest>());
                 extracted = extracted.Select(x => x.SimplifyRequestTree()).ToList();
 
-                if (simplified.AllRequestsOfType(splittingType).Count == 0)
+                simplified = simplified.SimplifyRequestTree();
+                if (simplified != null)
```
adecler, line 71: This is getting closer but I am confused as to why you need to simplify the tree so many times. Simplifying an already simplified tree should return the exact same result, no ?

image

I guess that's kind of fine for now but you should at least have 

```c#
extracted.Add(simplified);
```

and not 

```c#
extracted.Add(simplified.SimplifyRequestTree());
```

since the whole point was to avoid adding a null into extracted, right ?

Data_Engine/Modify/SimplifyRequestTree.cs
```diff
@@ -93,12 +93,17 @@ private static void Flatten(this IRequest request)
...
                                 else
-                                    subRequests.Insert(i, subSub[0]);
+                                {
+                                    subRequests.RemoveAt(i);
+
+                                    if (subSub[0].GetType() == request.GetType())
+                                        subRequests.InsertRange(i, ((ILogicalRequest)subSub[0]).IRequests());
+                                    else
+                                        subRequests.Insert(i, subSub[0]);
+                                }
```
adecler, line 106: I'll be honest, this whole method is confusing me a lot. I get the purpose but I just don't get why it is so complex. Since refactoring this method is not the purpose of this PR, I'll ignore it but I cannot guarantee that this part of the code is bug-free. So I'll have to trust you on this one (with the assumption that this will be reworked soon anyways as discussed).

adecler (changes requested):
Getting closer. Just a final change needed on `SplitRequestTreeByType.cs`

Data_Engine/Compute/SplitRequestTreeByType.cs
pawelbaran: Of course, that call is entirely redundant. I have just pushed a fix.

Data_Engine/Modify/SimplifyRequestTree.cs
pawelbaran: Yeah, 100% agreed about everything above - happy to reiterate together with @alelom in the next milestone 👍

pawelbaran:
@BHoMBot check compliance

pawelbaran:
@BHoMBot check installer

FraserGreenroyd:
@BHoMBot this is a CI/CD instruction. I am authorising dispensation to be granted on check ref. 1954410922

This is following a discussion with @pawelbaran and the agreement to raise #2332 to resolve this as it is outside the scope of this Pull Request which is already tackling an in depth issue.

FraserGreenroyd:
@BHoMBot check ready-to-merge

FraserGreenroyd:
@BHoMBot check serialisation

FraserGreenroyd:
@BHoMBot check core

adecler (approved):
I am now happy with the changes. 
I guess what is left is to deal with the bot 😋

pawelbaran:
@BHoMBot check ready-to-merge

PR #2325, opened 2021-02-19
Analytical_Engine: Add null checks to Geometry methods

### LINKED ISSUE #2324: Auto-constructor component for structural bar turns red when created
#### Description:

image

To be investigated why the geometry extension method is not working properly anymore.

adecler:

Closes #2324

Any object with a nul geometry property would cause the component in the UI to turn red due to missing null checks in the Geometry methods. This PR fixes that by adding `?` everywhere it is relevant.

### Test files
Just drop an auto-constructor component for a Bar for example.


adecler:
@BHoMBot check copyright

rolyhudson (approved):
LGTM. Dropped an auto constructor to call each of the methods changed. All working as expected.

PR #2321, opened 2021-02-18
Spatial_Engine, Geometry_Engine: Rigid body transformation method added

pawelbaran:
### Issues addressed by this PR

Closes #2320

### Test files

- [Geometry_Engine side](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?RootFolder=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F02%5FPull%20Request%2FBHoM%2FBHoM%5FEngine%2FGeometry%5FEngine%2F%232320%2DIsRigidTransformation&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4)
- [Spatial_Engine side](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?RootFolder=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F02%5FPull%20Request%2FBHoM%2FBHoM%5FEngine%2FSpatial%5FEngine%2F%232320%2DRigidBodyTransform&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4)

### Changelog

- rigid body transformation method added for `IElement0D`, `IElement1D`, `IElement2D`
- `IsRigidTransformation`, `IsValid` and `IsEqual` methods added for `TransformMatrix`

### Additional comments

This PR will require an immediate follow-up that will cover special cases that cannot be resolved with simple `GetGeometry` and `SetGeometry`. The requirements for that are logged in #2322, please feel free to contribute to the list. Thanks!

PR #2327, opened 2021-02-19
Reflection_Engine: SortExtensionMethods improved to return the most relevant method

### LINKED ISSUE #2326: Reflection_Engine: improve SortExtensionMethods to return the most relevant method when the type-specific one does not exist
#### Description:

As stated in the code:
https://github.com/BHoM/BHoM_Engine/blob/86ce04660b3b02ba92f1f6508e30169c2f4b5816/Reflection_Engine/Modify/SortExtensionMethods.cs#L45-L49

pawelbaran:

Closes #2326

### Test files

On [SharePoint](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?RootFolder=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F02%5FPull%20Request%2FBHoM%2FBHoM%5FEngine%2FReflection%5FEngine&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4) - the file is a bit lame, but I could not come up with a better example 🙈. I hope the code change will be self-explanatory.

### Changelog

- `SortExtensionMethods` improved to return the most relevant method when the type-specific one does not exist

### Additional comments

`type` parameter is now used to filter out the methods that are not relevant at all - possibly it could be removed, based on assumption that the methods are already prefiltered.

I was a bit concerned about the performance of `GetParameters` - I have tested it explicitly and there is less than 1ms difference between 1000 and 10000 calls, ergo it should not be even noticeable.

### DIFF
--- a/Reflection_Engine/Modify/SortExtensionMethods.cs
+++ b/Reflection_Engine/Modify/SortExtensionMethods.cs
@@ -42,11 +42,7 @@ namespace BH.Engine.Reflection
         [Output("metods", "Sorted methods")]
         public static List<MethodInfo> SortExtensionMethods(this IEnumerable<MethodInfo> methods, Type type)
         {
-            //TODO: Sort methods based on closeness to the type, not just exact vs non exact.
-            //Example A : B and B : C
-            //Then if the type is A and list of methods contain one method with first parameter matching each type, then the list should be
-            //Sorted so that the method with A comes first followed by B and lastly C.
-            return methods.OrderBy(x => x.GetParameters().FirstOrDefault()?.ParameterType == type ? 0 : 1).ToList();
+            return methods.OrderBy(x => methods.Count(y => x.GetParameters()[0].ParameterType.IsAssignableFrom(y.GetParameters()[0].ParameterType))).ToList();
         }
 
         /***************************************************/
--- a/Reflection_Engine/Modify/SortExtensionMethods.cs
+++ b/Reflection_Engine/Modify/SortExtensionMethods.cs
@@ -42,8 +42,72 @@ namespace BH.Engine.Reflection
         [Output("metods", "Sorted methods")]
         public static List<MethodInfo> SortExtensionMethods(this IEnumerable<MethodInfo> methods, Type type)
         {
-            IEnumerable<MethodInfo> typeMethods = methods.Where(x => x.GetParameters()[0].ParameterType.IsAssignableFrom(type));
-            return methods.OrderBy(x => methods.Count(y => x.GetParameters()[0].ParameterType.IsAssignableFrom(y.GetParameters()[0].ParameterType))).ToList();
+            List<List<Type>> hierarchy = type.InheritanceHierarchy();
+            IEnumerable<int> levels = methods.Select(x => hierarchy.InheritanceLevel(x.GetParameters()[0].ParameterType));
+            return methods.Zip(levels, (m, l) => new { m, l }).Where(x => x.l != -1).OrderBy(x => x.l).Select(x => x.m).ToList();
+        }
+
+
+        /***************************************************/
+        /**** Private Methods                           ****/
+        /***************************************************/
+
+        private static List<List<Type>> InheritanceHierarchy(this Type type)
+        {
+            List<Type> typeAncestry = new List<Type>();
+            Type ancestor = type;
+            while (ancestor != null && ancestor != typeof(object))
+            {
+                typeAncestry.Add(ancestor);
+                ancestor = ancestor.BaseType;
+            }
+
+            List<List<Type>> result = new List<List<Type>>();
+
+            for (int i = typeAncestry.Count - 1; i >= 0; i--)
+            {
+                Type child = typeAncestry[i];
+                result.Insert(0, new List<Type> { child });
+
+                IEnumerable<Type> alreadyMapped = result.SelectMany(x => x);
+                List<List<Type>> interfaces = child.InterfaceHierarchy();
+                for (int j = 0; j < interfaces.Count; j++)
+                {
+                    IEnumerable<Type> notMapped = interfaces[j].Except(alreadyMapped);
+                    if (notMapped.Any())
+                    {
+                        if (j + 1 < result.Count)
+                            result[j + 1].AddRange(notMapped);
+                        else
+                            result.Add(notMapped.ToList());
+                    }
+                    else
+                        break;
+                }
+            }
+
+            return result;
+        }
+
+        /***************************************************/
+
+        private static List<List<Type>> InterfaceHierarchy(this Type type)
+        {
+            Type[] interfaces = type.GetInterfaces();
+            return interfaces.GroupBy(x => interfaces.Count(y => x.IsAssignableFrom(y))).OrderBy(x => x.Key).Select(x => x.ToList()).ToList();
+        }
+
+        /***************************************************/
+
+        private static int InheritanceLevel(this List<List<Type>> hierarchy, Type type)
+        {
+            for (int i = 0; i < hierarchy.Count; i++)
+            {
+                if (hierarchy[i].Contains(type))
+                    return i;
+            }
+
+            return -1;
         }
 
         /***************************************************/
--- a/Reflection_Engine/Modify/SortExtensionMethods.cs
+++ b/Reflection_Engine/Modify/SortExtensionMethods.cs
@@ -42,6 +42,7 @@ namespace BH.Engine.Reflection
         [Output("metods", "Sorted methods")]
         public static List<MethodInfo> SortExtensionMethods(this IEnumerable<MethodInfo> methods, Type type)
         {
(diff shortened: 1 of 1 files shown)


Reflection_Engine/Modify/SortExtensionMethods.cs
```diff
@@ -42,11 +42,8 @@ public static partial class Modify
         [Output("metods", "Sorted methods")]
         public static List<MethodInfo> SortExtensionMethods(this IEnumerable<MethodInfo> methods, Type type)
         {
-            //TODO: Sort methods based on closeness to the type, not just exact vs non exact.
-            //Example A : B and B : C
-            //Then if the type is A and list of methods contain one method with first parameter matching each type, then the list should be
-            //Sorted so that the method with A comes first followed by B and lastly C.
-            return methods.OrderBy(x => x.GetParameters().FirstOrDefault()?.ParameterType == type ? 0 : 1).ToList();
+            IEnumerable<MethodInfo> typeMethods = methods.Where(x => x.GetParameters()[0].ParameterType.IsAssignableFrom(type));
+            return methods.OrderBy(x => methods.Count(y => x.GetParameters()[0].ParameterType.IsAssignableFrom(y.GetParameters()[0].ParameterType))).ToList();
```
IsakNaslundBh, line 46: Not sure if I agree with this change. The purpose of this method is to say that say `Geometry(Bar bar)` should be prioritised over `Geometry(IElement1d element)`. That is, exact match before any interface level.

By using `IsAssignableFrom` it feels like both of those cases will be returning the same match...

pawelbaran: Not really, the methods are ordered by the number of target types their target type is assignable from, so for example:
```
MyMethod(Line)
MyMethod(ICurve)
MyMethod(IGeometry)
```

`MyMethod(Line)` will score 1 (`typeof(Line)` is assignable from `typeof(Line)` only), the others 2 and 3 respectively.

In your case, `Geometry(Bar bar)` should score 1, `Geometry(IElement1d element)` - 2. So the former will be prioritised.

Does that make sense or am I missing a point?

pawelbaran:
After a conversation with @IsakNaslundBh I felt like I will let someone hold my beer and gave it another try to build an actual inheritance hierarchy, based on which one can pick the most suitable method to call.

Looks like I have succeeded to some extent, apart from possible idiosyncrasies that may appear in case of class inheritance - this, however, seems to be safe in the context of BHoM where it barely exists (usually the only cases are types that inherit from `BHoMObject`, which then is rather safe because there is literally zero extension methods on `BHoMObject` in the entire BHoM_Engine solution).

Performance wise, it still runs in tens of ms at max, so should be fine, taken into account the method is called only once per method and type set.

For now I made all methods private due to their complex return types, happy to change that if you find it desirable @adecler.

Reflection_Engine/Modify/SortExtensionMethods.cs
```diff
@@ -42,11 +42,72 @@ public static partial class Modify
...
         public static List<MethodInfo> SortExtensionMethods(this IEnumerable<MethodInfo> methods, Type type)
         {
-            //TODO: Sort methods based on closeness to the type, not just exact vs non exact.
-            //Example A : B and B : C
-            //Then if the type is A and list of methods contain one method with first parameter matching each type, then the list should be
-            //Sorted so that the method with A comes first followed by B and lastly C.
-            return methods.OrderBy(x => x.GetParameters().FirstOrDefault()?.ParameterType == type ? 0 : 1).ToList();
+            List<List<Type>> hierarchy = type.InheritanceHierarchy();
+            IEnumerable<int> levels = methods.Select(x => hierarchy.InheritanceLevel(x.GetParameters()[0].ParameterType));
+            return methods.Zip(levels, (m, l) => new { m, l }).Where(x => x.l != -1).OrderBy(x => x.l).Select(x => x.m).ToList();
```
adecler, line 47: You should look into what the Weighting system in the BHoM_UI does since this is a very similar problem:
https://github.com/BHoM/BHoM_UI/blob/631e784772f456dd881492262d7f10c5c2fb7322/UI_Engine/Query/Weight.cs#L190

I am not convinced it needs to be as complex as that though. Speed is quite an important factor here and I think we should cut everything that is not that relevant. I wonder if this wouldn't be enough:
- exact type match : comes first
- type is assignable from: comes second
- type not assignable from: comes last

I am not aware of the details of the situation you are trying to solve by this seems like a good compromise to me.

pawelbaran: The proposed solution works as follows:
- exact type match : comes first
- type is assignable from, closer to the top of inheritance tree: comes second
- type is assignable from, further from the top of inheritance tree: comes thirs
- type not assignable from: gets filtered out (it will throw an error at some point anyways)

The reason for such complex approach can be explained on the `Transform` method developed in https://github.com/BHoM/BHoM_Engine/pull/2321 (WIP) for `BH.oM.Physical.Elements.Beam`:
- there will be an interface method `ITransform` taking `BH.oM.Dimensional.IElement1D` that will call `Transform`
- there will be a general `Transform` method taking `BH.oM.Dimensional.IElement1D` that will only transform the driving curve
- there will be a more specific `Transform` method taking `BH.oM.Physical.Elements.IFramingElement` that besides driving curve transformation will change the rotation property
- there will be no type-specific `Transfrom` method for `BH.oM.Physical.Elements.Beam`

It is important to call the method taking `BH.oM.Physical.Elements.IFramingElement` not to lose the rotation part. Naturally, one could add `Transform` method for each type implementing `BH.oM.Physical.Elements.IFramingElement` instead, but this sounds like quite a bit of boilerplate, plus what if one adds another type being not aware that this method needs to be added?

pawelbaran: When it comes to performance, I have tested this and it runs in <10ms, plus it it is ran only once per type/method name combination (then stored in a public field). So I think we should be good with that.

adecler: I see what you mean. I am glad you are keeping those methods private though as their result is not exactly matching what a human would produce. Take for example the output of `InheritanceHierarchy` for a structural bar:

image

For example, it doesn't really make sense for `BHoMObject` to have such a big priority compared to other interfaces. Same thing for `IBHoMObject` being prioritised over `IElement`. Not the fault of you algorithm though as interfaces in the BHoM are mostly used as a tag and no so much as a strict hierarchy. On the other hand, trying to sort base types and interfaces on the same list is always going to be a bit iffy. 

That's why I was suggesting a simpler 3-levels system in order to avoid a complex sorting system that has many imperfect cases. But, again, I understand why you needed something more.

I would like to have @IsakNaslundBh final opinion as well on this one before agreeing to merge.

pawelbaran: I fully agree with your comment @adecler, the system is not perfect. However, I approached it based on the assumption that an approximate solution is better than quasi-random:
- on `master`, the exact type goes first, all others are random
- on this branch, the exact type goes first, then the order is reflecting the inheritance tree, which has the faults listed by you

On top of that, from the problem-specific point of view it should be fine to have `BHoMObject` high in the hierarchy because it will be simply ignored due to the lack of extension methods. This does not mean the method is any more correct than it actually is, but it should not fail in the discussed application of `RunExtensionMethod` 🙈 

Happy to hear @IsakNaslundBh's opinion too.

adecler: I'll approve this so you're not necessarily stuck waiting for @IsakNaslundBh if this needs merging. Personally, I am ok with this being merged.

adecler (approved):

Reflection_Engine/Modify/SortExtensionMethods.cs
IsakNaslundBh: Sorry. Been struggeling a bit with time!

Just reading through the comments here, I think this looks like an improvement! Agree that the speed is not _that_ big of an issue for the reasons @pawelbaran gave, rather have this behave as good as it can. Ideally, we would have this do the exact same thing that C# itself does when it comes to method overloading/choice, which I guess this is one step closer to doing compared to where it was before.

Have not tested this myself, but given that @adecler has, you do not have to wait for me. I approve on the concept 😄

pawelbaran: > Ideally, we would have this do the exact same thing that C# itself does when it comes to method overloading/choice, which I guess this is one step closer to doing compared to where it was before.

I am afraid we do not want that @IsakNaslundBh, the compiler would throw an error about method ambiguity when pushed to choose between an interface method and the parent type method:

image

So there might be no perfect solution I am afraid.

pawelbaran:
@BHoMBot check compliance

pawelbaran:
@BHoMBot check core

pawelbaran:
@BHoMBot check installer

PR #2343, opened 2021-02-24
Geometry_Engine: IGeometry and IGeometry3D to check if ExtensionMethod is found before calling it to avoid exposing error when not relevant

### LINKED ISSUE #2340: Geometry_Engine: the new RunExtensionMethod "not found" error is returned by the Goos preview every time a BHoMObject does not own one
#### Description:

Due to https://github.com/BHoM/BHoM_Engine/pull/2331, when an object's Goo tries to preview, if the Geometry() method is not found for it, an error is exposed:

image

The error is correctly returned by the newly introduced `RunExtensionMethod(object obj, MethodInfo method)`, because the method for IGeometry provided to it is actually `null`. The search for an IGeometry method is invoked by the Goo. See: <details> 
MicrosoftTeams-image
</details>

**In general**, the change introduced by https://github.com/BHoM/BHoM_Engine/pull/2331 brings a slight change in the usage of the `RunExtensionMethod()` pattern.

In the case of **_common_ extension methods**, like `IGeometry()` – i.e. _method called from a variety of different places_ – **we need to be careful when using `RunExtensionMethod()` _alone_**. You may be exposing irrelevant errors.

We need to make sure that `ExtensionMethodToCall()` is first used to retrieve the method, then a case `if (method != null)` should be introduced before using `RunExtensionMethod(object obj, MethodInfo method)`. An appropriate error can be then be returned (or not, depending on the usage) if no Method was found.

### Suggested fix

Introduce the check in the IGeometry fallback, from:

```cs
 private static IGeometry Geometry(this IBHoMObject obj)
{
    return Reflection.Compute.RunExtensionMethod(obj, "Geometry") as IGeometry;
}
```

To:

```cs
 private static IGeometry Geometry(this IBHoMObject obj)
{
      MethodInfo method = BH.Engine.Reflection.Query.ExtensionMethodToCall(obj, "Geometry");
      if (method != null)
          value = BH.Engine.Reflection.Compute.RunExtensionMethod(obj, method);
      else
          return null; // returns null and no Error appears in UI, restoring previous behaviour.
}
```
@pawelbaran @adecler @al-fisher @IsakNaslundBh @FraserGreenroyd @rolyhudson

alelom:

Closes #2340

See the text in #2340.

### Test files

Try to create a Physical Construction or Layer object. It should not return any error. See issue text for more info.

### DIFF
--- a/BHoM_Engine/Query/Geometry.cs
+++ b/BHoM_Engine/Query/Geometry.cs
@@ -94,7 +94,11 @@ namespace BH.Engine.Base
 
         private static IGeometry Geometry(this IBHoMObject obj)
         {
-            return Reflection.Compute.RunExtensionMethod(obj, "Geometry") as IGeometry;
+            System.Reflection.MethodInfo mi = Reflection.Query.ExtensionMethodToCall(obj, "Geometry");
+            if (mi != null)
+                return Reflection.Compute.RunExtensionMethod(obj, "Geometry") as IGeometry;
+            else
+                return null;
         }
 
         /***************************************************/
--- a/BHoM_Engine/Query/Geometry3D.cs
+++ b/BHoM_Engine/Query/Geometry3D.cs
@@ -101,7 +101,11 @@ namespace BH.Engine.Base
 
         private static IGeometry Geometry3D(this IObject obj)
         {
-            return Reflection.Compute.RunExtensionMethod(obj, "Geometry3D") as IGeometry;
+            System.Reflection.MethodInfo mi = Reflection.Query.ExtensionMethodToCall(obj, "Geometry3D");
+            if (mi != null)
+                return Reflection.Compute.RunExtensionMethod(obj, "Geometry3D") as IGeometry;
+            else
+                return null;
         }
 
         /***************************************************/


pawelbaran (approved):
LGTM 👍

pawelbaran:
@BHoMBot check compliance

pawelbaran:
@BHoMBot check core

pawelbaran:
@BHoMBot check installer

al-fisher (commented):
This makes sense.

As discussed with @alelom - I think this pattern of 
`QueryExtensionMethodToCall` 
`if( != null)`
`Compute.RunExtensionMethod)`
will be used a lot - so wonder about implementing a `TryRunExtensionMethod` method in Reflection Engine?

pawelbaran:
> As discussed with @alelom - I think this pattern of
> `QueryExtensionMethodToCall`
> `if( != null)`
> `Compute.RunExtensionMethod)`
> will be used a lot - so wonder about implementing a `TryRunExtensionMethod` method in Reflection Engine?

Actually I was thinking about too @al-fisher. My main concern was not to get stuck with the `out` params vs `Outputs`. But in general I like the idea!

FraserGreenroyd:
@pawelbaran has shown me the scope of problem caused without this PR which was brought in following a merge to a PR this morning. In the interest of improving the alpha tomorrow, following discussion with @pawelbaran I am merging this, with a discussion issue to be raised by @pawelbaran to discuss the comment by @al-fisher 😄 :

BHoM_Engine/Query/Geometry.cs
```diff
@@ -94,7 +94,11 @@ private static IGeometry Geometry(this object obj)
 
         private static IGeometry Geometry(this IBHoMObject obj)
         {
-            return Reflection.Compute.RunExtensionMethod(obj, "Geometry") as IGeometry;
+            System.Reflection.MethodInfo mi = Reflection.Query.ExtensionMethodToCall(obj, "Geometry");
+            if (mi != null)
+                return Reflection.Compute.RunExtensionMethod(obj, "Geometry") as IGeometry;
+            else
+                return null;
```
adecler, line 101: Even if this is already merged, it is worth noting that `Reflection.Compute.RunExtensionMethod(obj, mi)` should be called instead to avoid calling `ExtensionMethodToCall(obj, "Geometry")` twice

Not a massive deal breaker but something to keep in mind if this pattern is to be repeated elsewhere.

I also agree with Al that a `TryRunExtensionMethod` would keep things tidier.

alelom: You're right @adecler , I actually thought I did that - will correct in separate PR.

Happy to push the TryRunExtensionMethod.

alelom:
Raised https://github.com/BHoM/BHoM_Engine/issues/2348

PR #2347, opened 2021-02-25
CI/CD: Add test to make sure all methods can work with null inputs

### LINKED ISSUE #2335: CI/CD: Provide testing to make sure that all methods can be executed with null inputs without throwing an exception
#### Definition of the test :
For each engine method: invoke with only null inputs and make sure that no `NullReferenceException` exception is being thrown.

adecler:

Closes #2335

Note that I have considered, for now, that it is acceptable to assume that strings and Lists will be at worst empty, never null.
This is because the UI can be used as a safe-guard on those two. That doesn't obviously protect against internal calls that could provide a null for those,. But I think we can start by focussing on the null objects given how many methods are failing even that right now (1729).

### Test files
Run `./TestRunner NullChecks`. 
The full result is exceeding the content of the command window right now. So I have added a few lines to write to an temp file that you can uncomment.

### Additional comments
- There was a method in `LifeCycleAssessment` that was causing a stack overflow. Since we cannot catch those, I have just added the method to the list of exception for now until it is fixed.
- This is the kind of test where I feel CI/CD should call `NullChecks(MethodInfo method)` and not `NullChecks()` since this is purely a local impact so might then be quite fast to run on each PR since there are only going to be a few methods tested. Otherwise, the test takes a fairly long time.

### DIFF
--- a/.ci/code/Engine_Test/Engine_Test.csproj
+++ b/.ci/code/Engine_Test/Engine_Test.csproj
@@ -1,0 +1,73 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
+    <ProjectGuid>{216E8DF8-9650-4B3B-8416-959B94BC527C}</ProjectGuid>
+    <OutputType>Library</OutputType>
+    <AppDesignerFolder>Properties</AppDesignerFolder>
+    <RootNamespace>BH.Test.Engine</RootNamespace>
+    <AssemblyName>Engine_Test</AssemblyName>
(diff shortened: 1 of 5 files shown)


FraserGreenroyd (changes requested):
Minor change just because of how long it took to get to this PR 🙈

FraserGreenroyd (approved):
Have run the test runner, all looks good to me. Will want to implement on BHoMBot this milestone but that's a separate kettle of fish that I can tackle later. Happy to merge this for now though as everything looks to be working adequately.

@michaelhoehn it would be worth looking into the LCA method though, would be good to remove that as an exception ASAP when you're next doing a 5-minute compliance update to the toolkit :wink:

FraserGreenroyd:
@BHoMBot check core

FraserGreenroyd:
@BHoMBot check compliance

FraserGreenroyd:
@BHoMBot check installer

FraserGreenroyd:
@BHoMBot check versioning

FraserGreenroyd:
@BHoMBot check serialisation

FraserGreenroyd:
@adecler leave this PR with me, I'm waiting for the other checks to finish then I intend to just provide dispensation to the failing checks because they're clearly not related to this PR and appear to be caused by changes around this PR on the oM/Engine which are just not relevant in my view. I just CBA to wait for them to finish right now given the time, so I'll come back to them in the morning.

FraserGreenroyd:
@BHoMBot this is a CI/CD instruction. I am authorising dispensation to be granted on check ref. 2250106700

FraserGreenroyd:
@BHoMBot this is a CI/CD instruction. I am authorising dispensation to be granted on check ref. 2250107755

FraserGreenroyd:
@BHoMBot check ready-to-merge

PR #2353, opened 2021-02-25
Structure_Engine: Add conversion methods for FEMesh to Panel and vice versa for a limit use case

### LINKED ISSUE #2338: Structure_Engine: Methods from MidasCivil_Toolkit to Structure_Engine for PanelToFEMesh and FEMeshToPanel
#### Description:

Port over method for PanelToFEMesh and FEMeshToPanel, see https://github.com/BHoM/MidasCivil_Toolkit/pull/291 for initial comments.

This should sit under `Compute` I would think.

This is also a good opportunity to create two sets of unit tests for the methods.

Note: Be clear on description that it is not a meshing method, just a like-for-like convert.

StephennipBH:
Closes #2338 

### Test files

https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?cid=506a082e%2Daf0c%2D43ad%2D8fb5%2D4505bb1e26bb&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4&sortField=Modified&isAscending=false&viewid=a5a93cba%2Dfcca%2D46cc%2Da31a%2D80b8d5cbfd7b&id=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F01%5FIssue%2FBHoM%2FBHoM%5FEngine%2FStructure%5FEngine%2F%232338%2D%20PanelToFEMesh%20and%20FEMeshToPanel

### Changelog
- Added `PanelToFEMesh` which converts a `Panel` with 3 or 4 `Nodes` to an `FEMesh`
- Added `FEMeshToPanel` with converts an `FEMesh` object in to a `Panel`


StephennipBH:
@BHoMBot  check unit tests

Structure_Engine/Compute/FEMeshToPanel.cs
```diff
@@ -0,0 +1,80 @@
...
+                return null;
+            }
+                List<Polyline> polylines = new List<Polyline>();
+
+            List<Point> points = new List<Point>();
+
+            foreach (Node node in mesh.Nodes)
+            {
+                points.Add(node.Position);
+            }
```
peterjamesnugent, line 56: This should use the `MeshFace` and a `for` loop. One `feMesh` can own multiple `MeshFaces`, so you should loop over the `MeshFaces` using them to get the indexes in the `List<Node>.` Something like this:
```suggestion
            foreach(FEMeshFace feMeshFace in feMesh.Faces)
            {
                List<Point> points = new List<Point>();
                foreach (int nodeIndex in feMeshFace.NodeListIndices)
                    points.Add(mesh.Nodes[nodeIndex].Position);
                polylines.Add(BH.Engine.Geometry.Create.Polyline(points));
            }
```

Maybe a `for` loop would be better so you can add the final first index at the end to close the `Polyline`.

Structure_Engine/Compute/FEMeshToPanel.cs
```diff
@@ -0,0 +1,80 @@
...
+                panels = BH.Engine.Structure.Create.Panel(polylines.Cast<ICurve>().ToList(), mesh.Property);
+                return panels[0];
+            }
+            else
+            {
+                panels = BH.Engine.Structure.Create.Panel(polylines.Cast<ICurve>().ToList());
+                Engine.Reflection.Compute.RecordWarning("Meshs don't have any Section Property input");
+                return panels[0];
+            }
+
```
peterjamesnugent, line 73: You need to account for the multiple `Panel` from each `FEMeshFace` and you can simplify the `if` statement as well as adding `Fragments`. I've added `CustomData` even though it should be used as we have migrated that information to `Fragments`.

```suggestion
            List<Panel> panels = new List<Panel>();
            foreach(Polyline polyline in polylines)
            {
                Panel panel = Create.Panel(polyline, null, null, mesh.Name);
                if (mesh.Property != null)
                    panel.Property = mesh.Property;
                if (mesh.Fragments != null)
                    panel.Fragments = mesh.Fragments;
                panel.CustomData = mesh.CustomData;
            }
            return panels;
```

Structure_Engine/Compute/FEMeshToPanel.cs
```diff
@@ -0,0 +1,80 @@
...
+    public static partial class Compute
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+        [Description("Converting FEmesh to a panel")]
+        [Input("FEMesh", "FEMesh to be converted to an Panel")]
+        [Output("Panel", "Panel converted from a FEMesh.")]
+
+        public static Panel FEMeshToPanel(FEMesh mesh)
```
peterjamesnugent, line 42: This needs to return `List<Panel>` see below.

Structure_Engine/Compute/PanelToFEMesh.cs
```diff
@@ -0,0 +1,105 @@
...
+namespace BH.Engine.Structure
+{
+    public static partial class Compute
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+        [Description("Converting panel with 3 or 4 control points to FEmesh ")]
+        [Input("panel", "Panel to be converted to an FEMesh.")]
+        [Output("FEMesh", "FEMesh converted from a Panel.")]
```
peterjamesnugent, line 41:
```suggestion
        [Description("Converts a  Panel with three or four control points to an FEMesh with a single Face. This is not a method to discretise a Panel, it simply converts a simple Panel to an identical FEMesh.")]
        [Input("panel", "Panel to be converted to a FEMesh.")]
        [Output("FEMesh", "FEMesh converted from a Panel.")]
```

Structure_Engine/Compute/PanelToFEMesh.cs
```diff
@@ -0,0 +1,105 @@
...
+
+        public static FEMesh PanelToFEMesh(Panel panel)
+        {
+            List<Point> points = new List<Point>();
+            List<Edge> edges = panel.ExternalEdges;
+            List<Face> faces = new List<Face>();
+            if (panel.Openings.Count()>0)
```
peterjamesnugent, line 48: Just check if it `!= null`.

Structure_Engine/Compute/PanelToFEMesh.cs
```diff
@@ -0,0 +1,105 @@
...
+            {
+                Engine.Reflection.Compute.RecordWarning("PanelToFEMesh does not support panel with opening");
```
peterjamesnugent, line 50:
```suggestion
                Engine.Reflection.Compute.RecordWarning("This method does not support Panels with Openings.");
```

Structure_Engine/Compute/PanelToFEMesh.cs
```diff
@@ -0,0 +1,105 @@
...
+                return null;
+            }
+            if (panel.ExternalEdges.Count>4)
+            {
+                Engine.Reflection.Compute.RecordWarning("Panel contains more than 4 control points");
+                return null;
+            }
```
peterjamesnugent, line 57: You need to add some additional checks here, because you could have an `Edge` that is defined by a single `Polyline` with multiple control points. Whereas your error message refers to control points.

Structure_Engine/Compute/PanelToFEMesh.cs
```diff
@@ -0,0 +1,105 @@
...
+            if (panel.ExternalEdges == null)
+                {
+                Engine.Reflection.Compute.RecordWarning("Checks identify no External edges");
+                return null;
+            }
+            if (panel == null)
+            {
+                Engine.Reflection.Compute.RecordWarning("Checks identify null panel");
+                return null;
+            }
```
peterjamesnugent, line 67: The `Panel` `null` check should come first, no point doing the rest of the checks otherwise, order of checks should be:
- `Panel == null`
- `Panel.ExternalEdges == null` and then check if any if any are `null` by looping over the `List<Edge>`
- If any of the `Curve` owned by the `Edge` is `null`
- Then if the `Curves` are Linear 
- Then if any of the points on the `Curve` is `null`

Then I would do all the circumstantial checks about number of edges etc.

Structure_Engine/Compute/PanelToFEMesh.cs
```diff
@@ -0,0 +1,105 @@
...
+            foreach (Edge edge in edges)
+            {
+                ICurve curve = BH.Engine.Analytical.Query.Geometry(edge);
+
+
+                points.AddRange(BH.Engine.Geometry.Convert.IToPolyline(curve).ControlPoints);
```
peterjamesnugent, line 73: Delete the whitespace.

Structure_Engine/Compute/PanelToFEMesh.cs
```diff
@@ -0,0 +1,105 @@
...
+            int Count = points.Distinct().Count();
+            Face face = new Face();
+            if (Count == 4)
+            {
+                face = Geometry.Create.Face(0, 1, 2, 3);
+            }
+            else if (Count == 3)
+            {
+                face = Geometry.Create.Face(0, 1, 2);
+            }
```
peterjamesnugent, line 84: What if the count is > 4? You need a check once you get the points out.

Structure_Engine/Compute/PanelToFEMesh.cs
```diff
@@ -0,0 +1,105 @@
...
+                return null;
+            }
+            foreach (Edge edge in edges)
+            {
+                ICurve curve = BH.Engine.Analytical.Query.Geometry(edge);
+
+
+                points.AddRange(BH.Engine.Geometry.Convert.IToPolyline(curve).ControlPoints);
+            }
+            int Count = points.Distinct().Count();
```
peterjamesnugent, line 75: Does this maintain the order of the `Points`? I think it might as the duplicates appear later in the list but worth confirming.

Also, store the `List<Point>` that you get a distinct list of, because you use it further down on L86.

Structure_Engine/Compute/PanelToFEMesh.cs
```diff
@@ -0,0 +1,105 @@
...
+            {
+                fEMesh = BH.Engine.Structure.Create.FEMesh(mesh, panel.Property, null, panel.Name);
+                return fEMesh;
+            }
+            else
+            {
+                fEMesh = BH.Engine.Structure.Create.FEMesh(mesh, null, null, panel.Name);
+                Engine.Reflection.Compute.RecordWarning("Panels don't have any Section Property input");
+                return fEMesh;
+            }
```
peterjamesnugent, line 98: See my example in the `FEMeshToPanel` method, the create should be outside of the `if` statement, and then a series of `null` checks before assigning `SurfaceProperty`, `Fragments` and `CustomData`.

peterjamesnugent (changes requested):
Some changes to the checks, bolstering of descriptions and slight reordering of code.

Can you also upload the script that you used to generate the UnitTest so more cases can be added if needed.

Also, you need to update from master, that might be why the CI is failing.

Structure_Engine/Compute/PanelToFEMesh.cs
IsakNaslundBh: Would have a look into the nullcheck methods added by @IsakLarbornBH here. For example: `NullCheck(this Panel panel, string methodName = "Method")` that should be checking through all the geometrical aspects of the object. SImilar methods exists for the other elements as well.

IsakLarbornBH: That method purposely doesn't do quite as thorough a check as it sounds like you're going for here, but it will check if the panel itself is null, has a non-zero amount of edges, and that each of those edges has a curve, so basically the first three steps Peter lists here.

StephennipBH: Thanks both, I will adapt some of the null checks into the code.

StephennipBH:
@BHoMBot Check unit tests

Structure_Engine/Compute/FEMeshToPanel.cs
```diff
@@ -0,0 +1,112 @@
...
+            {
+                if (feMesh.Property != null)
+                {
+                    panel = Create.Panel(polyline, null, feMesh.Property, feMesh.Name);
+                }
+                else
+                {
+                    panel = Create.Panel(polyline, null, null, feMesh.Name);
+                    Reflection.Compute.RecordWarning("FEMeshes don't have any Section Property input");
+                }
```
peterjamesnugent, line 99: You can create the `Panel` without a `SurfaceProperty` and then check `if (feMesh.Property != null)` whether you assign it or not - then you don't have the `Panel` create method duplicated.

Structure_Engine/Compute/FEMeshToPanel.cs
```diff
@@ -0,0 +1,112 @@
...
+                if (feMesh.Fragments.Count > 0)
+                {
+                    panel.Fragments = feMesh.Fragments;
+                }
+                panels.Add(panel);
+            }
```
peterjamesnugent, line 105: We need to make sure the `tags` get carried across as well.

Structure_Engine/Compute/PanelToFEMesh.cs
```diff
@@ -0,0 +1,106 @@
...
+using BH.Engine.Geometry;
+using BH.oM.Reflection.Attributes;
+using System.Collections.Generic;
+using System.Linq;
+using System.ComponentModel;
+using BH.Engine.Analytical;
+
+namespace BH.Engine.Structure
+{
+    public static partial class Compute
```
peterjamesnugent, line 35: Convert

Structure_Engine/Compute/PanelToFEMesh.cs
peterjamesnugent: `count` camelCase

Structure_Engine/Compute/PanelToFEMesh.cs
```diff
@@ -0,0 +1,106 @@
...
+            foreach (Edge edge in edges)
+            {
+                ICurve curve = Analytical.Query.Geometry(edge);
+                points.AddRange(Geometry.Convert.IToPolyline(curve).ControlPoints);
+            }
+            int Count = points.Distinct().Count();
+            Face face = new Face();
+            if (Count>4)
+            {
+                Reflection.Compute.RecordError("Panel contains more than 4 Control Points");
```
peterjamesnugent, line 76: Panel contains more than four control points.

Structure_Engine/Compute/PanelToFEMesh.cs
```diff
@@ -0,0 +1,106 @@
...
+            {
+                face = Geometry.Create.Face(0, 1, 2, 3);
+            }
+            else if (Count == 3)
+            {
+                face = Geometry.Create.Face(0, 1, 2);
+            }
+            faces.Add(face);
+            Mesh mesh = Geometry.Create.Mesh(points.Distinct(), faces);
+            FEMesh fEMesh = new FEMesh();
```
peterjamesnugent, line 89: feMesh

Structure_Engine/Compute/PanelToFEMesh.cs
```diff
@@ -0,0 +1,106 @@
...
+            if (panel.Property != null)
+            {
+                fEMesh = Create.FEMesh(mesh, panel.Property, null, panel.Name);
+                return fEMesh;
+            }
+            else
+            {
+                fEMesh = Create.FEMesh(mesh, null, null, panel.Name);
+                Reflection.Compute.RecordWarning("Panels don't have any Section Property input");
```
peterjamesnugent, line 98: Remove warning - if we raised a warning for every property that is `null` we would have way too many.

Structure_Engine/Compute/PanelToFEMesh.cs
```diff
@@ -0,0 +1,106 @@
...
+            {
+                face = Geometry.Create.Face(0, 1, 2);
+            }
+            faces.Add(face);
+            Mesh mesh = Geometry.Create.Mesh(points.Distinct(), faces);
+            FEMesh fEMesh = new FEMesh();
+            if (panel.Property != null)
+            {
+                fEMesh = Create.FEMesh(mesh, panel.Property, null, panel.Name);
+                return fEMesh;
```
peterjamesnugent, line 93: As above on the `FEMeshToPanel` - create the `FEMesh` first, then check if the `SurfaceProperty` is `null`, then you can assign it or skip.

Structure_Engine/Compute/PanelToFEMesh.cs
```diff
@@ -0,0 +1,106 @@
...
+            }
+            else
+            {
+                fEMesh = Create.FEMesh(mesh, null, null, panel.Name);
+                Reflection.Compute.RecordWarning("Panels don't have any Section Property input");
+                return fEMesh;
```
peterjamesnugent, line 99: You need to make sure the `tags` are ported too.

peterjamesnugent (changes requested):
Some minor changes to descriptions, error messages and ordering of code.

Functionality looks good, just need to make sure it's water tight and the limitations are properly communicated.

Structure_Engine/Convert/FEMeshToPanel.cs
```diff
@@ -0,0 +1,112 @@
...
+            List<Polyline> polylines = new List<Polyline>();
+           
+            foreach (FEMeshFace feMeshFace in feMesh.Faces)
+            {
+                List<Point> points = new List<Point>();
+                foreach (int nodeIndex in feMeshFace.NodeListIndices)
+                {
+                    points.Add(feMesh.Nodes[nodeIndex].Position);    
+                }
+                points.Add(feMesh.Nodes.First().Position);
```
peterjamesnugent, line 84:
```suggestion
                points.Add(feMesh.Nodes[feMeshFace.NodeListIndices.First()].Position);
```

peterjamesnugent (changes requested):
See comment for error when using `FEMeshToPanel`.

I've added an `FEMeshFace` with multiple `FEMeshFaces` as one of the tests - please make sure that works.

For the unit tests I think we should have:
`PanelToFEMesh`
1. 3 noded Panel made up of 4 lines
2. 4 noded Panel made up of 4 lines
3. 4 noded Panel made up from a polyline i.e. 1 `Edge`

`FEMeshToPanel`
1. 3 noded `FEMesh` with 1 `FEMeshFace`
2. 4 noded `FEMesh` with 1 `FEMeshFace`
3. `FEMesh` with several `FEMeshFaces` see test script for example

Each of the above should have a `SurfaceProperty`, `Tags` and a `Name` to demonstrate the information being carried over when the convert occurs.

peterjamesnugent:
@BHoMBot check compliance

Structure_Engine/Convert/FEMeshToPanel.cs
```diff
@@ -0,0 +1,112 @@
...
+        [Output("panel", "Panel converted from a FEMesh.")]
+
+        public static List<Panel> FEMeshToPanel(FEMesh feMesh)
+        {
+            if (feMesh == null)
+            {
+                Engine.Reflection.Compute.RecordError("FEMesh is null, please check inputs.");
+                return null;
+            }
+            if (feMesh.Nodes == null)
```
peterjamesnugent, line 49:
```suggestion
            if (feMesh.Nodes == null || feMesh.Nodes.Count < 3) 
```

Structure_Engine/Convert/FEMeshToPanel.cs
```diff
@@ -0,0 +1,112 @@
...
+            {
+                Engine.Reflection.Compute.RecordError("FEMesh Nodes are null, please check inputs");
+                return null;
+            }
+            if (feMesh.Faces == null)
```
peterjamesnugent, line 54:
```suggestion
            if (feMesh.Faces == null || feMesh.Faces.Count < 1)
```

peterjamesnugent (changes requested):
- Need to add check that `fe.Mesh.Nodes` are not < 3, and that `fe.Mesh.Faces` are not < 1 - this test case in the test script:
image
- Needs to return a null for one of the `Panel` checks
- I've updated the test script to essentially check all the different null checks and different `Panel` and `FEMesh` configuations, this is reflected in the unit test script - please push the new unit test json 

Once these are done, happy to approve.

peterjamesnugent (changes requested):
I've just updated the Unit Testing script to include the date/time and some names for `Dataset` and `Source` objects. Can you delete the .json first before you push - when I read the JSON I get multiple datasets - I think we should only have one.

peterjamesnugent (dismissed):
LGTM.

peterjamesnugent:
@BHoMBot check copyright

peterjamesnugent:
@BHoMBot check compliance

peterjamesnugent:
@BHoMBot check compliance

peterjamesnugent (dismissed):

peterjamesnugent:
@BHoMBot check null-handling

peterjamesnugent:
@BHoMBot check serialisation

peterjamesnugent:
@BHoMBot check versioning

peterjamesnugent:
@BHoMBot check core

peterjamesnugent:
@BHoMBot run installer

peterjamesnugent:
@BHoMBot check installer

Structure_Engine/Convert/PanelToFEMesh.cs
```diff
@@ -0,0 +1,108 @@
...
+
+        public static FEMesh PanelToFEMesh(this Panel panel)
+        {
+            List<Point> points = new List<Point>();
+            List<Edge> edges = panel.ExternalEdges;
+            List<Face> faces = new List<Face>();
+            if (BH.Engine.Structure.Compute.NullCheck(panel) == false)
+            {
+                return null;
+            }
```
IsakNaslundBh, line 53: Need to make null checks before accessing any property of the object. Also, small coding style tweak

```suggestion
            if (!BH.Engine.Structure.Compute.NullCheck(panel))
            {
                return null;
            }
            List<Point> points = new List<Point>();
            List<Edge> edges = panel.ExternalEdges;
            List<Face> faces = new List<Face>();

```

IsakNaslundBh (dismissed):
External edges are accessed before the panel is checked for null. Should be the reason the failing null check. See suggestion below.

IsakNaslundBh:
@BHoMBot check required

peterjamesnugent:
@BHoMBot check serialisation

IsakNaslundBh:
@BHoMBot check versioning

IsakNaslundBh:
@BHoMBot check installer

IsakNaslundBh:

IsakNaslundBh:

peterjamesnugent (dismissed):

IsakNaslundBh:
@BHoMBot check required

IsakNaslundBh:
Due to github complaining about conflicts, I had to rebase this branch. Just got the bot re-running all the checks. @peterjamesnugent can you re-review and check that things still work your end?

IsakNaslundBh:

IsakNaslundBh:

peterjamesnugent (approved):
Tested with test script and still behaves as expected.

PR #2357, opened 2021-02-26
Reflection_Engine: Added TryRunExtensionMethod

### LINKED ISSUE #2348: Reflection_Engine: introduce TryRunExtensionMethod
#### Description:

As discussed in https://github.com/BHoM/BHoM_Engine/pull/2343#pullrequestreview-597742860.

alelom:

Closes #2348

Adds a TryRunExtensionMethod. If no method is found, returns false. Otherwise, invokes the method and passes the output in the `out` parameter, then returns true.

### Test files

Test with any choice of extension methods.

### DIFF
--- a/Reflection_Engine/Compute/TryRunExtensionMethod.cs
+++ b/Reflection_Engine/Compute/TryRunExtensionMethod.cs
@@ -1,0 +1,67 @@
+/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2021, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 
+ * GNU Lesser General Public License for more details.                          
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using BH.oM.Reflection.Debugging;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Runtime.CompilerServices;
+using System.Reflection;
+using System.Text;
+using System.Threading.Tasks;
+using System.ComponentModel;
+using BH.oM.Reflection.Attributes;
+
+namespace BH.Engine.Reflection
+{
+    public static partial class Compute
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Looks for an extension method applicable to the input object with the provided `methodName` and, if found, invokes it.\n" +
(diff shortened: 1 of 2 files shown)


Reflection_Engine/Compute/TryRunExtensionMethod.cs
```diff
@@ -0,0 +1,67 @@
...
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Looks for an extension method applicable to the input object with the provided `methodName` and, if found, invokes it.\n" +
+            "Extension methods are searched using Reflection through all BHoM assemblies.\n" +
+            "If no method is found, this returns `false`, and the `result` is null.")]
+        [Input("obj", "Object whose extension method is to be found, and to which the method will be applied in order to obtain the result.")]
+        [Input("methodName", "Name of the extension method defined for the input object that is to be found in any of the BHoM assemblies.")]
+        [Input("result", "Result of the method invokation, if any. If no method is found this is null.")]
+        [Output("True if a method was found and successfully invoked. False otherwise.")]
```
pawelbaran, line 50: Successful invocation is not a condition driving the output - the output is equal only to whether the method was found or not.

Reflection_Engine/Compute/TryRunExtensionMethod.cs
```diff
@@ -0,0 +1,67 @@
...
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Looks for an extension method applicable to the input object with the provided `methodName` and, if found, invokes it.\n" +
+            "Extension methods are searched using Reflection through all BHoM assemblies.\n" +
+            "If no method is found, this returns `false`, and the `result` is null.")]
+        [Input("obj", "Object whose extension method is to be found, and to which the method will be applied in order to obtain the result.")]
+        [Input("methodName", "Name of the extension method defined for the input object that is to be found in any of the BHoM assemblies.")]
+        [Input("result", "Result of the method invokation, if any. If no method is found this is null.")]
```
pawelbaran, line 49:
```suggestion
        [Input("result", "Result of the method invocation, if the method had been invoked. If no method is found this is null.")]
```

pawelbaran (commented):
I have a found minor misconceptions in the descriptions - please see the code comments (plus possibly consider copy-pasting stuff from `RunExtensionMethod`).

On top of that, I would consider two changes in the code itself:
- Adding a similar method taking allowing to call an extension method with other parameters than the target (similar as in `RunExtensionMethod`).
- Refactoring the method to return `Output<bool, object>` instead of `bool` and `out` param - this would allow using the method from the UI level, which may be a nice to have. However, in this case I would lean on @adecler to have the final say.

Reflection_Engine/Compute/TryRunExtensionMethod.cs
pawelbaran: Ah, BTW the output is missing name.

alelom:
> Refactoring the method to return Output<bool, object> instead of bool and out param - this would allow using the method from the UI level, which may be a nice to have. However, in this case I would lean on @adecler to have the final say.

My preference is to keep the `out` parameter instead, to follow C#'s convention. In particular, I expect this method to be used 95% of the time from the code, rather than from the UI. Also, I believe that the ideal solution for scenarios like these would be for the framework to support the correct reflection of `out` parameters in the outputs of a component (as far as I understand, currently `out` parameters end up among the inputs instead).

pawelbaran:
Agreed with @alelom, the standard C# convention with `out` parameter makes sense to me in this particular case, so the code LGTM now.

However, due to the fact that the PR introduces a UI glitch (nonharmful, the UI is not crashing, it just looks weird), I would like @adecler to have a final say before we approve/merge.

adecler (approved):
I am fine with the `out` parameter. It better aligns with the general C# conventions on the `Try`  methods and we will eventually support `out` parameters in the UI anyways. In the meantime, people can just use the `RunExtensionMethod` component in the UI if they ever need that for some reason.

pawelbaran:
@BHoMBot check all

pawelbaran (approved):
Confirming LGTM, happy to see this merged.

Issue #2322, opened 2021-02-18
Add type-specific rigid body transforms where needed


pawelbaran:
#### Description:

As discussed with @IsakNaslundBh, some object types will require a dedicated rigid body transform method that goes beyond simple `GetGeometry` and `SetGeometry`:
- [ ] `BH.oM.Structure.Elements.Node` - orientation
- [ ] `BH.oM.Structure.Elements.Bar` - orientation angle and nodes
- [ ] `BH.oM.Structure.Elements.FEMesh` - nodes
- [ ] `BH.oM.Structure.Elements.RigidLink` - nodes
- [ ] `BH.oM.Structure.Elements.Panel`- orientation angle
- [ ] `BH.oM.Physical.Elements.IFramingElement` (all implementations) - orientation angle
- [ ] `BH.oM.MEP.System.CableTray` - orientation angle
- [ ] `BH.oM.MEP.System.Duct` - orientation angle
- [ ] `BH.oM.Lighting.Elements.Luminaire` - direction
- [ ] `BH.oM.Security.Elements.CameraDevice` - target position
- [ ] `BH.oM.Architecture.Elements.Ceiling` - tiles
- [ ] `BH.oM.Architecture.BuildersWork.Opening` - it is neither `IElement0D` nor `IElement1D`, but will be nice to unlock it

Uff... the list got suspiciously long... I will be thankful if others could check if I am not missing anything here @al-fisher @IsakNaslundBh @FraserGreenroyd @rolyhudson @LMarkowski @alelom @adecler - happy to do the coding part by myself.

Issue #2339, opened 2021-02-24
Incorrect rigid transformation check in Geometry_Engine transforms


pawelbaran:
#### Description:

As proven in the test file for #2333, the determinant of the entire transform matrix equal to 1 is not equivalent to the transform being rigid. Instead, the rotation part of the matrix needs to be checked for two requirements ([source](https://en.wikipedia.org/wiki/Rigid_transformation)):
- R<sup>T</sup> == R<sup>-1</sup>
- det(R) == 1

This means that `Transform` methods in the Geometry_Engine need revisiting and updating the checks.

PR #2359, opened 2021-02-26
Physical_Engine: Transform method added for IFramingElement

pawelbaran:

Contributes to resolving #2322

### Test files

On [SharePoint](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4&id=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F02%5FPull%20Request%2FBHoM%2FBHoM%5FEngine%2FPhysical%5FEngine%2F%232322%2DTransform%2Egh&parent=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F02%5FPull%20Request%2FBHoM%2FBHoM%5FEngine%2FPhysical%5FEngine).

### Changelog

- `Transform` method added for `IFramingElement`


Physical_Engine/Modify/Transform.cs
```diff
@@ -0,0 +1,82 @@
...
+        [Input("framingElement", "IFramingElement to transform.")]
+        [Input("transform", "Transform matrix.")]
+        [Output("transformed", "Modified IFramingElement with unchanged properties, but transformed location and orientation angle.")]
+        public static IFramingElement Transform(this IFramingElement framingElement, TransformMatrix transform)
+        {
+            if (!transform.IsRigidTransformation())
+            {
+                BH.Engine.Reflection.Compute.RecordError("Transformation failed: only rigid body transformations are currently supported.");
+                return null;
+            }
```
LMarkowski, line 52: Do we want to include this check in higher level methods as well? In #2339 we plan to put this check in the geometry level transform methods. It'll become redundant.
On the other hand the results of a failing check are different on those levels. Here it returns null (which is fair and logical) and in Geo_Eng it converts the input curve into `Nurbs`.

PS. tolerance passing.. 👀

Physical_Engine/Modify/Transform.cs
```diff
@@ -0,0 +1,82 @@
...
+            result.Location = result.Location.ITransform(transform);
+
+            ConstantFramingProperty property = result.Property as ConstantFramingProperty;
+            if (property == null)
+                BH.Engine.Reflection.Compute.RecordWarning($"Orientation angle of the IFramingElement has not been transformed because its property is not ConstantFramingProperty. BHoM_Guid: {framingElement.BHoM_Guid}");
+            else
+            {
+                Line locationBefore = framingElement.Location as Line;
+                if (locationBefore == null)
+                    BH.Engine.Reflection.Compute.RecordWarning($"Orientation angle of the IFramingElement has not been transformed because the element is not linear. BHoM_Guid: {framingElement.BHoM_Guid}");
```
LMarkowski, line 64: It's probably my lack of experience with `IFramingElement` interface but why don't we transform the orientation angle of non-linear elements?

LMarkowski (commented):
In general everything looks and works neat! I've raised two questions in the code however.

Also.. <sub><sup>\*ahem, ahem\* unit test... \*ahem, ahem\*</sup></sub>

Physical_Engine/Modify/Transform.cs
pawelbaran: I am not sure how would that be done tbh (not even sure how is orientation defined in case of nonlinear elements 🙈). Possibly I could do that by reproducing the linear approach on the line spanning between start and end of the location curve, but what if the curve is closed? Etc. etc.

I would rather not add anything than add something I am not too sure about. As long as anyone can tell me how to do that properly.

pawelbaran:
Tolerance and UT added 🐒

Physical_Engine/Modify/Transform.cs
IsakNaslundBh: Can maybe help out a bit here..

https://github.com/BHoM/documentation/wiki/BHoM-Structural-Conventions#coordinate-system gives the conventions.

Basically, for a planar non-linear element, the starting normal is defined by the plane of the curve, and then the infinite numer of normals along the curve is rotated around the curve axis with the orientation angle.

Unless I am completely mistaken, I think this mean that this actually gets simpler... Just transforming the curve itself, should basically bring the normal with it, meaning you are done there. Good if one of you guys can have a think about that as well just making sure I am not completely wrong.

For non-planar curves we actually do not have any convention yet as no usecases requiring it has popped up yet (as Revit does not allow them, it is far less common). Not 100% sure what to do about them yet either, but might for the purpose of this PR just transform the curves as well, as I am pretty sure any convention we put in will be dependant on the Location curve, alternatively, just skip non-planar curves for now and raise an error.

pawelbaran: Big thanks for help @IsakNaslundBh! The above is very informative and makes perfect sense (as confirmed during our discussion) 👍 I have updated the code the reflect the following convention:
- linear - same as structural bars
- nonlinear planar - only transform the curve, rotation being driven by its plane
- nonlinear nonplanar - only transform the curve + a warning that correctness could not be ensured 100%

I think this is maximum we can get for now. I have also updated UT and GH script, so the PR should be ready for re-review.

Physical_Engine/Modify/Transform.cs
pawelbaran: From what I understand, the checks will serve different purpose: failed check here disallows any further actions because of the discipline-specific constraints that we cannot handle, while in the Geometry_Engine it will simply change the return type (at least in the ideal world, when we have all geometry scaling/mirroring/skewing operations implemented). Therefore we need the checks on both levels I am afraid.

LMarkowski:
@BHoMBot check unit tests

LMarkowski:
@BHoMBot check unit tests

LMarkowski (approved):
Code, test file, and UT all lgtm.

RTM as soon as @BHoMBot will be up and provide a pass on UTs.

pawelbaran:
> RTM as soon as @BHoMBot will be up and provide a pass on UTs.

Will that ever happen? 🐒

FraserGreenroyd:
> > RTM as soon as @BHoMBot will be up and provide a pass on UTs.
> 
> Will that ever happen? 🐒

https://media.giphy.com/media/aQGqcObSxfixy/giphy.gif

pawelbaran:
@BHoMBot check all

LMarkowski:
@BHoMBot check all

pawelbaran:
@BHoMBot check all

FraserGreenroyd:

pawelbaran:
@BHoMBot check versioning

pawelbaran:
@BHoMBot check ready-to-merge

FraserGreenroyd:
@BHoMBot check installer

FraserGreenroyd:
@BHoMBot check ready-to-merge

PR #2370, opened 2021-03-04
Geometry_Engine: Cuboid IsContaining method.

### LINKED ISSUE #2369: Geometry_Engine: IsContaining implementation that works for Cuboid
#### Description:

Currently, the `IsContaining()` method has its input limited to `BoundingBox` only.

As per its definition, a BoundingBox is aligned to the global coodinate axes, as it only stores a `Point Min` and `Point Max` properties.

However there is the need, for a variety of different workflows – in particular the Digital Configurator – to have a "containment check" for generic objects. For such cases, the user wants to check whether an object, however aligned, is contained or not within a certain location in space.  For example, the user wants to check whether a Beam – generally not aligned to global axes – is within a certain location in space, defined for example by a cube.

This means that we need another method that works with a geometrical entity not constrained to global axes, as Bounding Box is.

There are several ways to implement this:
- the most powerful and general way of checking containment would be to consider Mesh => containment in closed mesh. Complex, requires checking if Mesh is closed (tight) and manifold. This approach may be simplified by generating the container Mesh ourselves and ensuring its tightness; however, we discussed that this should be part of a separate set of issues where we also reorganise the methods now in TriangleNet.
- more fitting for the immediate needs, we can introduce a IsContaining method that works with `Cuboid`.

This last method may work by using the OrientationMatrix methods. What we can do is orientating the input geometry from the orientation system of the given cuboid to the global orientation system. From there, we can check whether the dimensions of the given geometry are within `±Cuboid.Depth/2`, `±Cuboid.Length/2`, `±Cuboid.Height/2`.

rolyhudson:

Closes #2369 
Added method to test IGeometry containment in Cuboid.

### Test files
[Test file with validation against grasshopper.](https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/02_Pull%20Request/BHoM/BHoM_Engine/Geometry_Engine/%232370-CuboidIsContaining/%232370-CuboidIsContaining.gh?csf=1&web=1&e=pfoXye)


Geometry_Engine/Query/IsContaining.cs
```diff
@@ -441,6 +441,40 @@ public static bool IsContaining(this PolyCurve curve1, ICurve curve2, bool accep
...
+                return ( Math.Abs(min1.X) <= cuboid.Length / 2 + tolerance &&
+                         Math.Abs(max1.X) <= cuboid.Length / 2 + tolerance &&
+                         Math.Abs(min1.Y) <= cuboid.Depth / 2 + tolerance &&
+                         Math.Abs(max1.Y) <= cuboid.Depth / 2 + tolerance &&
+                         Math.Abs(min1.Z) <= cuboid.Height / 2 + tolerance &&
+                         Math.Abs(max1.Z) <= cuboid.Height / 2 + tolerance);
+            }
+            else
+            {
+                return ( Math.Abs(min1.X) < cuboid.Length / 2 + tolerance &&
```
IsakNaslundBh, line 469: Wonder if this should be 
```suggestion
                return ( Math.Abs(min1.X) < cuboid.Length / 2 - tolerance &&
```

instead (if so ofc same for the other rows below if we agree)

Reason being, we are saying we do _not_ accept on edge, and then within a tolerance.

Looking at it like this:

image

with the black lines being the tolerances. Then the black lines would be the edge-zone, basically anything within that zone would be "On edge", hence accept on edge should be true for anything within it, but not accepting on edge should exclude it.

Remeber having similar chat with @pawelbaran on some other is containing methods (where I was thinking it should be done as you have done on this PR, but was convinced by pawel that it should be a -tolerance).

Does this make sense? @pawelbaran not sure if you remember the discussion, and if so can confirm?

pawelbaran: I do not remember the discussion @IsakNaslundBh, but I agree with the suggested changes. I would actually consider going a step further and building a `BoundingBox` based on the input `Cuboid`, then calling `IsContaining` on the former. That could lead to a pattern that could be easily reproduced for other methods implemented for `BoundingBox`.

rolyhudson: @IsakNaslundBh yes makes sense regarding the subtraction of tolerance when not accepting on edge.
@pawelbaran sorry I don't understand the suggestion of the `BoundingBox`. Can you explain?
Just thinking though... If we create a bounding box on an arbitrarily oriented input cuboid it will be globalCS aligned. Where green is the `BoundingBox` of the red input cuboid below:
Screenshot 2021-03-04 180024
I can see how that would work if you mean:
1. create a `BoundingBox` using the cuboid properties around the globalCS
2. then transform the input testGeometry from cuboidCS toGlobal
3. then test box.IsContaining( testGeometry)

pawelbaran: > I can see how that would work if you mean:
> 
> 1. create a `BoundingBox` using the cuboid properties around the globalCS
> 2. then transform the input testGeometry from cuboidCS toGlobal
> 3. then test box.IsContaining( testGeometry)

Yeah, that is exactly what I had in mind. Not sure how helpful would that be, I can imagine the method becoming a few lines thanks to that:

```
TransformMatrix transform = Create.OrientationMatrixLocalToGlobal(cuboid.CoordinateSystem);
IGeometry globalGeo = geometry.ITransform(transform);
BoundingBox geoBox = globalGeo.IBounds();
BoundingBox cuboidBox = cuboid.BoundingBoxFromProperties();
return cuboidBox.IsContaining(geoBox, tolerance);
```

Then for example `IsInRange` would be limited to a mere copy-paste and change of the last line. Not sure if worth it tbh, it was just an idea that came to my mind.

rolyhudson: Added in that approach @pawelbaran - much cleaner code. 
I wonder where `BoundingBoxFromProperties` belongs? Any ideas on that? 
Right now as temporary measure I have left it as a private method in the IsContaining file. 
It probably should be public but where logically? It seems like it could be viewed as a `Query`, `Create` or `Convert`? There is risk of confusion if wrongly located / poorly described. 

`IBounds` does not implement a method for cuboids which could add to the confusion. If there was `IBounds` for cuboids it could (and should) return different results from `BoundingBoxFromProperties` when the cuboid is not globally aligned.

rolyhudson: Based on conversation with @alelom and @al-fisher `BoundingBoxFromProperties` has been refactored as a `Create.BoundingBox` method with appropriate description and parameter names.

pawelbaran: Hmm, just trying to organize things in my head... How will we calculate the bounding box of a non-aligned `Cuboid` in this case? Through `Bounds` query?

It tempts me to question the above and propose using `Convert` instead of `Create` etc., but it seems that any convention we apply will lead to a certain degree of ambiguity. So makes sense to me, happy with the code now 😉

alelom: We agree there is some overlap with the Convert - the reason we eventually went for `Create.BoundingBox(Cuboid globallyAlignedCuboid)` is that `Convert` normally expects a more 1:1 conversion. Instead, what this method is doing is creating a new BoundingBox given some properties of a Cuboid; for this new BoundingBox to make sense, you need the Cuboid to be aligned to the global axes, hence why we named the input **`globallyAlignedCuboid`**.

Geometry_Engine/Create/BoundingBox.cs
```diff
@@ -51,6 +53,19 @@ public static BoundingBox BoundingBox(Point centre, Vector extent)
             };
         }
 
+        /***************************************************/
+
+        [Description("Create a BoundingBox using the properties of a cuboid.")]
+        [Input("globallyAlignedCuboid", "Cuboid that is assumed to be globally aligned.")]
```
pawelbaran, line 59: It could be worth mentioning that globally aligned means centre == origin && aligned axes (particularly the former is not clear from the current description).

Geometry_Engine/Create/BoundingBox.cs
```diff
@@ -51,6 +53,19 @@ public static BoundingBox BoundingBox(Point centre, Vector extent)
             };
         }
 
+        /***************************************************/
+
+        [Description("Create a BoundingBox using the properties of a cuboid.")]
```
pawelbaran, line 58: Maybe worth mentioning that the bounding box will have its centre in the origin of the global coordinate system?

pawelbaran (changes requested):
The code itself LGTM 👍 However, it would be great if the descriptions of `Create.BoundingBox` could be a bit more informative - the behaviour is a bit counterintuitive, so a nice explanation what should be expected will be super helpful.

On top of that, I would consider spending some more effort on the test script and creating a UT based on it, but of course I am aware of our time constraints so I assume @rolyhudson knows best what is the acceptable level as the discipline lead.

rolyhudson:
Cheers @pawelbaran updated the comment - hopefully now clearer what is happening there. Also as suggested adding the UT for this new IsContaining method. UT creation file adds to the other IsContaining dataset see [here](https://burohappold.sharepoint.com/:f:/r/sites/BHoM/02_Current/12_Scripts/02_Pull%20Request/BHoM/BHoM_Engine/Geometry_Engine/%232370-CuboidIsContaining?csf=1&web=1&e=13oqdO) for the GH script and old IsContianing UT dataset.

rolyhudson:
@BHoMBot check unit-tests

pawelbaran (approved):
LGTM now 👍

pawelbaran:
@BHoMBot check compliance

pawelbaran:
@BHoMBot check installer

pawelbaran:
@BHoMBot check core

pawelbaran:
@BHoMBot check serialisation

pawelbaran:
@BHoMBot check versioning

PR #2395, opened 2021-03-15
Geometry_Engine: Adding proper rigidness check to the Transform methods

### LINKED ISSUE #2339: Incorrect rigid transformation check in Geometry_Engine transforms

LMarkowski:

Closes #2339 

Replacing former matrix determinant check with new `IsRigidTransformation` query.

### Test files

[Here](https://burohappold.sharepoint.com/:f:/s/BHoM/En2Ibg00v25HmGzTJBBz76cBJaFHnTPrNFJoBsnZQo7SVQ?e=aRLQKc) sits the general test for transfom. We can also just run the unit test ☺️ 

### Changelog

- Added `IsRigidTransformation` check to the `Transform` methods in the `Geometry_Engine`.

### DIFF
--- a/Geometry_Engine/Modify/Transform.cs
+++ b/Geometry_Engine/Modify/Transform.cs
@@ -105,7 +105,7 @@ namespace BH.Engine.Geometry
                 };
             else
             {
-                Reflection.Compute.RecordNote("Transformation is not rigid or uniform. Converting into NurbsCurve. May occur change in shape");
+                Reflection.Compute.RecordNote("Transformation is not rigid or uniform. Converting into NurbsCurve. Change in shape may occur.");
                 return curve.ToNurbsCurve().Transform(transform);
             }
         }
@@ -123,7 +123,7 @@ namespace BH.Engine.Geometry
                 };
             else
             {
-                Reflection.Compute.RecordNote("Transformation is not rigid or uniform. Converting into NurbsCurve. May occur change in shape");
+                Reflection.Compute.RecordNote("Transformation is not rigid or uniform. Converting into NurbsCurve. Change in shape may occur.");
                 return curve.ToNurbsCurve().Transform(transform);
             }
         }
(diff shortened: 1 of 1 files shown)


LMarkowski:
@BHoMBot check unit-tests

Geometry_Engine/Modify/Transform.cs
```diff
@@ -124,7 +123,7 @@ public static ICurve Transform(this Circle curve, TransformMatrix transform)
                 };
             else
             {
-                Reflection.Compute.RecordNote("Transformation is not rigid. Converting into NurbsCurve. May occure change in shape");
+                Reflection.Compute.RecordNote("Transformation is not rigid or uniform. Converting into NurbsCurve. May occur change in shape");
```
pawelbaran, line 126:
```suggestion
                Reflection.Compute.RecordNote("Transformation is not rigid or uniform. Converting into NurbsCurve. Change in shape may occur.");
```

pawelbaran (commented):
In general no rocket science, LGTM, I have only suggested a change on messaging (3x).

I also have one question: wouldn't it be worth adding `tolerance` parameter? This will align well with all other `Transform` methods in the engine.

LMarkowski:
@pawelbaran, I've applied the suggestions to the messages. When it comes to passing the tolerance I'd delay it for two reasons:
1. It is en existing and quite a core feature. I'd rather not start doing the versioning for it during the testing sprint.
2. It is not exactly the distance tolerance anyway. Eventually it needs to be approached during the global tolerance rework.

pawelbaran (approved):
LGTM, agreed it may not be the best moment to introduce the tolerance as it will require versioning and extra testing. However, it would be good to have the one merging this PR to have a final look and confirm he/she is fine with chosen strategy.

pawelbaran:
@BHoMBot check compliance

pawelbaran:
@BHoMBot check core

pawelbaran:
@BHoMBot check installer

pawelbaran:
@BHoMBot check serialisation

pawelbaran:
@BHoMBot check versioning

Issue #2393, opened 2021-03-12
Spatial_Engine: GeneralisedFabricatedBoxSection cannot be created with zero corbel widths


peterjamesnugent:
#### Description:

If the corbel width is set to zero the section curves cannot be produced. 

#### Steps to reproduce:

Some checks need to be added here:
https://github.com/BHoM/BHoM_Engine/blob/69d1a93e6a53fccdc7613a023b6a171955f7d25e/Spatial_Engine/Create/ShapeProfiles/GeneralisedFabricatedBoxProfile.cs#L93-L105

#### Expected behaviour:

Allow `GeneralisedFabricatedBoxProfile` to be created using zero corbel widths.

PR #2410, opened 2021-03-19
Geometry_Engine: Do not check zero length segments in IsSelfIntersecting

### LINKED ISSUE #2409: Geometry_Engine: IsSelfIntersecting to avoid zero-length lines
#### Description:

A polycurve containing zero-length segements is currently flagged as self-intersecting. Suggesting that any zero-length curve is filtered out before the check to avoid these polycurves to be flagged.

Currently causing issues with Profile creates, rasied in https://github.com/BHoM/BHoM_Engine/issues/2393 .

That issue should be resolved separately, by making sure there are no zero-length segements in any curves, but that has turned out to be a bit more complicated than first anticipated, and fixing the `IsSelfIntersecting` method seem to be a blanket fix.

#### Steps to reproduce:

Check a curve with a zero-length line

#### Expected behaviour:

For curves containing zero-length segments, but not self intersecting in any other way to not be flagged as self-intersecting.

IsakNaslundBh:

Closes #2409

This seem to fix the urgent issue in #2393 , but leaving that issue open as a review of the profile creates will be needed to avoid any zero-length segments to be created in the first place.

### Test files

https://burohappold.sharepoint.com/:f:/s/BHoM/EqqGnXiRrSVPkYN2v1QyjoIB51GZm6ShUdu-f_E62Hn-YA?e=X5oILq

### Additional comments

@LMarkowski @pawelbaran would be good if this could be checked quite thoroughly, to ensure this is not causing any issues elsewhere, and to make sure I have not missed any critical reason why this would not be ok!


IsakNaslundBh:
@BHoMBot check unit tests

IsakNaslundBh:
@BHoMBot Check Unit Tests

IsakNaslundBh:
@BHoMBot check unit-tests

rolyhudson (changes requested):
Given that these changes are now known not to be the cause of the failing unit tests check. 
I think this is a good solution. 

We should add a unit test for this method - which should not be based on the profile creates but direct calls from all the Curve inputs.

Other methods that call `IsSelfIntersecting`  directly are:
`Normal, Centroid, Offset` 
All of which have unit tests and these are passing. 
`OffsetVariable` does not yet have a unit test but I cannot see a problem.

Would also be good to get thoughts from others on other methods that may be indirectly calling `IsSelfIntersecting` and possible impact.

rolyhudson:
In setting up a batch of unit tests for `IsSelfInteresting` I note an inconsistency when passing a `Polyline` with zero length segments to the changed `PolyCurve` method compared with the `Polyline` method.

`PolyCurve` method returns false
`Polyline` method return true

Should the Polyline method also remove the short(zero) length segments?
`List<Line> curves = curve.SubParts().Where(x => x.ILength() > tolerance).ToList();`

Also a possible confusion for users is that in Rhino those Curves with zero length segments are marked as Invalid Curves.

rolyhudson:
Unit test creation script [here](https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/02_Pull%20Request/BHoM/BHoM_Engine/Geometry_Engine/%232410-AvoidShortCurvesForIsSelfIntersecting/IsSelfIntersecting.gh?csf=1&web=1&e=zp0jYl)

rolyhudson:
@BHoMBot check unit-tests

rolyhudson:
@BHoMBot check unit-tests

IsakNaslundBh:
@BHoMBot check compliance

IsakNaslundBh:
@BHoMBot check installer

IsakNaslundBh:
@BHoMBot check core

IsakNaslundBh:
@BHoMBot check versioning

IsakNaslundBh:
@BHoMBot check serialisation

rolyhudson (approved):
Happy with this now.

FraserGreenroyd:
@BHoMBot check core

PR #2420, opened 2021-03-25
BHoM_Engine: Adding null checks on geometry methods

### LINKED ISSUE #2417: Reflection_Engine: error message inside try/catch in RunExtensionMethod causes unwanted errors to appear in the UI
#### Description:

As in the title, currently creating e.g. an empty `BH.oM.Physical.Elements.Floor` results in a weird error showing up as in the snip below:

image

After the discussion with @IsakNaslundBh we came to a conclusion that the try/catch in `RunExtensionMethod` seems to be redundant or even harmful because:
- it prohibits the actual errors from being propagated outside `RunExtentionMethod`, catching them and generating a vague message instead
- it logs a preview error (see the snip above), even though preview is meant to fail silently

@IsakNaslundBh please feel free to add/correct.

adecler:

Closes #2417 

Adding the necessary null checks on `Query.Geometry` methods.

the `Query.Geometry(IFace)` method actually records errors on its own. Not sure if we should keep them or not. They make sense for that very specific case so I'm happy either way.

### Test files
https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/02_Pull%20Request/BHoM/BHoM_Engine/Geometry_Engine/%232417-NullChecksForGeometryMethods/GeometryNullChecks.gh?csf=1&web=1&e=ZXKtpw

### Additional comments
@FraserGreenroyd , those changes are purely adding null checks so I assume this is safe to merge (after thorough review of course). But happy to wait if you disagree.

### DIFF
--- a/Analytical_Engine/Query/Geometry.cs
+++ b/Analytical_Engine/Query/Geometry.cs
@@ -155,7 +155,7 @@ namespace BH.Engine.Analytical
         [Output("face", "The geometry of the IFace as geometrical Mesh Face.")]
         public static Face Geometry(this IFace face)
         {
-            if (face == null || face.NodeListIndices == null)
+            if (face?.NodeListIndices == null)
                 return null;
 
             if (face.NodeListIndices.Count < 3)
@@ -232,4 +232,3 @@ namespace BH.Engine.Analytical
         
     }
 }
-
--- a/Analytical_Engine/Query/Geometry.cs
+++ b/Analytical_Engine/Query/Geometry.cs
@@ -155,7 +155,7 @@ namespace BH.Engine.Analytical
         [Output("face", "The geometry of the IFace as geometrical Mesh Face.")]
         public static Face Geometry(this IFace face)
         {
-            if (face == null)
+            if (face == null || face.NodeListIndices == null)
                 return null;
 
             if (face.NodeListIndices.Count < 3)
--- a/MEP_Engine/Query/Geometry.cs
+++ b/MEP_Engine/Query/Geometry.cs
@@ -39,7 +39,10 @@ namespace BH.Engine.MEP
         
         public static IGeometry Geometry(this IFlow flowObj)
         {
-            return new Line { Start = flowObj?.StartPoint, End = flowObj?.EndPoint};
+            if (flowObj?.StartPoint == null || flowObj?.EndPoint == null)
+                return null;
+            else
+                return new Line { Start = flowObj.StartPoint, End = flowObj.EndPoint};
         }
 
         /***************************************************/
--- a/MEP_Engine/Query/Geometry.cs
+++ b/MEP_Engine/Query/Geometry.cs
@@ -39,7 +39,7 @@ namespace BH.Engine.MEP
         
         public static IGeometry Geometry(this IFlow flowObj)
         {
-            return new Line { Start = flowObj.StartPoint, End = flowObj.EndPoint};
+            return new Line { Start = flowObj?.StartPoint, End = flowObj?.EndPoint};
         }
 
         /***************************************************/
--- a/Physical_Engine/Query/Geometry.cs
+++ b/Physical_Engine/Query/Geometry.cs
@@ -49,12 +49,11 @@ namespace BH.Engine.Physical
         [Output("surface", "The defining location surface geometry of the ISurface with its openings represented")]
         public static oM.Geometry.ISurface Geometry(this oM.Physical.Elements.ISurface surface)
         {
-            PlanarSurface planarSurface = surface?.Location as PlanarSurface;
-            if (planarSurface == null)
+            ICurve exBound = (surface?.Location as PlanarSurface)?.ExternalBoundary;
+            if (exBound == null)
                 return null;
 
-            ICurve exBound = planarSurface.ExternalBoundary;
-            List<ICurve> inBound = surface?.Openings?.Select(o => (o?.Location as PlanarSurface)?.ExternalBoundary).ToList();
+            List<ICurve> inBound = surface?.Openings?.Select(o => (o?.Location as PlanarSurface)?.ExternalBoundary).Where(x => x != null).ToList();
             return Engine.Geometry.Create.PlanarSurface(exBound, inBound);
         }
 
@@ -63,5 +62,4 @@ namespace BH.Engine.Physical
 
     }
 }
-
 
--- a/Physical_Engine/Query/Geometry.cs
+++ b/Physical_Engine/Query/Geometry.cs
@@ -40,7 +40,7 @@ namespace BH.Engine.Physical
         [Output("CL", "The centre line curve of the framing element")]
         public static ICurve Geometry(this IFramingElement framingElement)
         {
-            return framingElement.Location;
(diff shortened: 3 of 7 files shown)


MEP_Engine/Query/Geometry.cs
```diff
@@ -39,7 +39,7 @@ public static partial class Query
         
         public static IGeometry Geometry(this IFlow flowObj)
         {
-            return new Line { Start = flowObj.StartPoint, End = flowObj.EndPoint};
+            return new Line { Start = flowObj?.StartPoint, End = flowObj?.EndPoint};
```
pawelbaran, line 42: I returning a line with `null` endpoints safe? Wouldn't it be better to simply return `null` if any of the endpoints of the `IFlow` is `null`?

Analytical_Engine/Query/Geometry.cs
```diff
@@ -155,7 +155,7 @@ public static IEnumerable<Face> Geometry<TFace>(this IEnumerable<TFace> faces)
         [Output("face", "The geometry of the IFace as geometrical Mesh Face.")]
         public static Face Geometry(this IFace face)
         {
-            if (face == null)
+            if (face == null || face.NodeListIndices == null)
```
pawelbaran, line 158: This would possibly be more consistent with other changes
```suggestion
            if (face?.NodeListIndices == null)
```

Physical_Engine/Query/Geometry.cs
```diff
@@ -49,8 +49,12 @@ public static ICurve Geometry(this IFramingElement framingElement)
...
         public static oM.Geometry.ISurface Geometry(this oM.Physical.Elements.ISurface surface)
         {
-            ICurve exBound = (surface.Location as PlanarSurface).ExternalBoundary;
-            List<ICurve> inBound = surface.Openings.Select(o => (o.Location as PlanarSurface).ExternalBoundary).ToList();
+            PlanarSurface planarSurface = surface?.Location as PlanarSurface;
+            if (planarSurface == null)
+                return null;
+
+            ICurve exBound = planarSurface.ExternalBoundary;
+            List<ICurve> inBound = surface?.Openings?.Select(o => (o?.Location as PlanarSurface)?.ExternalBoundary).ToList();
```
pawelbaran, line 57: I gave it a shot to flood it with even more null checks 🙈 
```suggestion
            ICurve exBound = (surface?.Location as PlanarSurface)?.ExternalBoundary;
            if (exBound == null)
                return null;

            List<ICurve> inBound = surface?.Openings?.Select(o => (o?.Location as PlanarSurface)?.ExternalBoundary).Where(x => x != null).ToList();
```

Structure_Engine/Query/Geometry.cs
```diff
@@ -72,10 +78,15 @@ public static IGeometry Geometry(this RigidLink link)
         {
             List<IGeometry> lines = new List<IGeometry>();
 
-            foreach (Node sn in link.SecondaryNodes)
+            if (link?.SecondaryNodes != null && link?.PrimaryNode != null)
```
pawelbaran, line 81:
```suggestion
            if (link?.SecondaryNodes != null && link?.PrimaryNode?.Position != null)
```

pawelbaran (commented):
In general LGTM (nothing to go wrong here), but I tried an exercise to improve the checks even further - please see the code comments.

MEP_Engine/Query/Geometry.cs
adecler: I am fine either way. I implemented your change though

adecler:
@pawelbaran , I have pushed your changes after checking each of them separately.

Structure_Engine/Query/Geometry.cs
```diff
@@ -60,7 +63,10 @@ public static CompositeGeometry Geometry(this ConcreteSection section)
         [Output("outlines", "The geometry of the GeometricalSection as its outline in the global XY plane.")]
         public static IGeometry Geometry(this IGeometricalSection section)
         {
-            return new CompositeGeometry { Elements = section.SectionProfile.Edges.ToList<IGeometry>() };
+            if (section?.SectionProfile?.Edges == null)
+                return new CompositeGeometry();
```
pawelbaran, line 67: Possibly this could return `null`, to distinguish between empty geometries and `null` ones.

Structure_Engine/Query/Geometry.cs
```diff
@@ -72,10 +78,15 @@ public static IGeometry Geometry(this RigidLink link)
         {
             List<IGeometry> lines = new List<IGeometry>();
 
-            foreach (Node sn in link.SecondaryNodes)
+            if (link?.SecondaryNodes != null && link?.PrimaryNode?.Position != null)
```
pawelbaran, line 81: Possibly we could have an `else` statement here returning `null`, to make sure that we distinguish between ampty and corrupt geometry?

Structure_Engine/Query/Geometry.cs
```diff
@@ -108,7 +119,10 @@ private static IGeometry Geometry(this object section)
         [Deprecated("3.1", "Replaced with method for base interface IGeometricalSection.", typeof(Query), "Geometry(this IGeometricalSection section).")]
         public static IGeometry Geometry(this SteelSection section)
         {
-            return new CompositeGeometry { Elements = section.SectionProfile.Edges.ToList<IGeometry>() };
+            if (section?.SectionProfile?.Edges == null)
+                return new CompositeGeometry();
```
pawelbaran, line 123: Same as above, this could possibly return `null`

pawelbaran (commented):
The code LGTM, however while having a thorough look I started wondering how important is it to distinguish between empty geometries (say, empty list) and corrupt ones (nulls). This is reflected in the code comments - I think you will have a better feel for the above @adecler, so it would be great if you could have a look and decide whether it is worth changing or not. I am ready to approve the PR as is if you decide too keep it this way.

Structure_Engine/Query/Geometry.cs
adecler: Personally, I prefer to return an empty one. It is generally advised to return null only when an empty object is not a valid match to the inputs. In this case, I feel an empty `CompositeGeometry` makes sense.

Structure_Engine/Query/Geometry.cs
adecler: Same as above. Given that we are returning a `CompositeGeometry`, its good to be consistent.

Structure_Engine/Query/Geometry.cs
adecler: Same answer too 😄

adecler:

adecler:
@BHoMBot check core

adecler:
@BHoMBot check compliance

adecler:
@BHoMBot check versioning

adecler:
@BHoMBot check serialisation

adecler:
@BHoMBot check null-handling

adecler:
@BHoMBot check installer

pawelbaran (approved):
Happy to approve, LGTM

PR #2437, opened 2021-03-31
Base_Engine: Create a useable DateTime for RandomObject

### LINKED ISSUE #2408: BHoM_Engine: RandomObject creation creates unusable random DateTime dtypes
#### Description:

When generating a random object containing a DateTime dtype, the values generated default to `0001-01-01 00:00:00:00`. While a perfectly valid date, computers don't like this (specifically SQL servers, which need dates to be after 1753 AD (ish). I guess that as BHoM doesn't account for this it decomposes the DateTime dtype into its composite ints/doubles?

#### Steps to reproduce:

image

#### Expected behaviour:

The default DateTime generated to be epoch (`1970-01-01 00:00:00:00`) (or earlier if we really want)

I think I've found [where this needs to change](https://github.com/BHoM/BHoM_Engine/blob/4cc47ca26b4046ce921d4a9aded77de6e6d05dd1/BHoM_Engine/Create/IObject/RandomObject.cs#L167-L198):
```C#
    ...
    case "DateTime":
        return new DateTime(1970, 1, 1).AddDays(rnd.Next((DateTime.Today - start).Days));
    ...

```

adecler:
Closes #2408 

Added code suggestion made in the issue with a few minor tweaks. 

### Test files
Create random objects of any type that has a `DateTime` property (I used Issue).

### DIFF
--- a/BHoM_Engine/Create/IObject/RandomObject.cs
+++ b/BHoM_Engine/Create/IObject/RandomObject.cs
@@ -195,6 +195,9 @@ namespace BH.Engine.Base
                     return new Point { X = rnd.NextDouble(), Y = rnd.NextDouble(), Z = rnd.NextDouble() };
                 case "Type":
                     return typeof(BHoMObject);
+                case "DateTime":
+                    DateTime startDate = new DateTime(1970, 1, 1);
+                    return startDate.AddDays(rnd.Next((DateTime.Today - startDate).Days)).AddSeconds(rnd.Next(86400));
             }
 
             if (type.IsPrimitive)


FraserGreenroyd (approved):
Couldn't find any test scripts but replicated what the PR suggested with an `Inspection.Issue`. Date related fields had some sufficiently `random` dates which resolves the scope of the issue raised.

tg359 (approved):
I checked using a similar script to the one posted in the issue raised, and dates created are all after 1970 so it looks like it's working.

Thanks @adecler!

FraserGreenroyd:
@BHoMBot check compliance

FraserGreenroyd:
@BHoMBot check core

FraserGreenroyd:
@BHoMBot check null-handling

FraserGreenroyd:
@BHoMBot check serialisation

FraserGreenroyd:
@BHoMBot check installer

FraserGreenroyd:
@BHoMBot check versioning

FraserGreenroyd:
@BHoMBot check serialisation

FraserGreenroyd:
@BHoMBot check installer

FraserGreenroyd:
@BHoMBot check versioning

PR #2439, opened 2021-04-01
Architecture_Engine: Add distance method for Builders Work openings

### LINKED ISSUE #2302: Add Distance(BH.oM.Architecture.BuildersWork.Opening, BH.oM.Architecture.BuildersWork.Opening) method
#### Description:

As in the title.

michal-pekacki:

Closes #2302 

Adding `IsValidForDistanceCheck` and `Distance` methods to calculate planar distance between two openings.

### Test files

On [SharePoint](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?RootFolder=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F02%5FPull%20Request%2FBHoM%2FBHoM%5FEngine%2FArchitecture%5FEngine%2F%232439%2DDistance&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4)


pawelbaran:
@BHoMBot check core

pawelbaran:
@BHoMBot check core

Architecture_Engine/Query/Distance.cs
```diff
@@ -0,0 +1,90 @@
...
+
+namespace BH.Engine.Architecture
+{
+    public static partial class Query
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        public static double Distance(this Opening opening1, Opening opening2, double maxExpectedDistance = double.NaN, double distanceTolerance = Tolerance.Distance, double angleTolerance = Tolerance.Angle)
```
pawelbaran, line 40: Actually, I would consider renaming the method to make it a bit more specific, i.e. `DistanceInPlane` instead of `Distance`, to indicate that the method calculates the distance between two openings that lie in a common plane (that is what the method actually does).

Besides that, all newly added public methods should have descriptions, see the example below:
https://github.com/BHoM/BHoM_Engine/blob/e454bc241841a94c8d44f4725a7f4229f586aa57/Data_Engine/Create/Dataset.cs#L39-L45

Architecture_Engine/Query/IsValidForDistanceCheck.cs
```diff
@@ -0,0 +1,100 @@
...
+
+namespace BH.Engine.Architecture
+{
+    public static partial class Query
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        public static bool IsValidForDistanceCheck(this Opening opening1, Opening opening2, double maxDistance = double.NaN, double distanceTolerance = Tolerance.Distance, double angleTolerance = Tolerance.Angle)
```
pawelbaran, line 41: Following the suggestion made in my previous comment, maybe we could make the method name even more explicit, e.g. `IsValidForInPlaneDistanceCheck` instead of `IsValidForDistanceCheck`.

On top of that, please be so kind and add descriptions to the method, inputs and outputs as requested in the previous comment.

Architecture_Engine/Query/Distance.cs
```diff
@@ -0,0 +1,90 @@
...
+        {
+
```
pawelbaran, line 42: This blank line is redundant, can be deleted (it is good to split different sections of the method with blank lines, but is does not make much sense to start a method with a blank line - @FraserGreenroyd, do we have any aesthetics guide at hand?)

Architecture_Engine/Query/Distance.cs
```diff
@@ -0,0 +1,90 @@
...
+            if (opening1 == null || opening2 == null)
+            {
+                BH.Engine.Reflection.Compute.RecordError("Opening is null");
```
pawelbaran, line 45: Please try being decently informative, particularly in case of error messages.
```suggestion
                BH.Engine.Reflection.Compute.RecordError("One of the openings is null, the distance could not be calculated.");
```

Architecture_Engine/Query/Distance.cs
```diff
@@ -0,0 +1,90 @@
...
+            double minDistance = double.MaxValue;
+
+            Point centerPoint2 = opening2.CoordinateSystem.Origin;
+            Plane plane2 = new Plane { Origin = centerPoint2, Normal = opening2.CoordinateSystem.Z };
+
+            if (!opening1.IsValidForDistanceCheck(opening2, maxExpectedDistance, distanceTolerance, angleTolerance)) 
+            {
+                BH.Engine.Reflection.Compute.RecordError("Distance can't be calculated");
+                return double.NaN;
+            }
```
pawelbaran, line 58: I would move this `if` statement before line 49 and take the code below outside the `else` statement. By doing that you will first have method applicability checks, with the actual query following afterwards.

So it will be:
1. check if the openings are not null
2. check if the openings are valid for distance check
3. if both above are OK then start declaring variables and calculating stuff

Besides that, as above: please make the error message slightly more informative (why can't the distance be calculated?) - using full stops at the end of sentences will earn you an extra CI/CD credit (@FraserGreenroyd 😃)

Architecture_Engine/Query/Distance.cs
```diff
@@ -0,0 +1,90 @@
...
+
+            else
+            {               
+                foreach (ICurve edge1 in opening1.Profile.Edges)
+                {
+                    foreach (ICurve edge2 in opening2.Profile.Edges)
```
pawelbaran, line 64: This is very inefficient because you need to extract the edges of `opening2` and orient them for every edge of `opening1`. Wouldn't it be better to it once before entering any `for` loop?

```
List<ICurve> edges2 = opening2.Profile.Edges.Select(x => x.Orient(global, opening2.CoordinateSystem).IProject(plane1)).ToList();
foreach (ICurve edge1 in opening1.Profile.Edges)
{
    //do your checks here
    //could also orient edges of opening1 before entering for loop for better code clarity
}
```

Architecture_Engine/Query/Distance.cs
```diff
@@ -0,0 +1,90 @@
...
+                        if (dist < minDistance)
+                        {
+                            minDistance = dist;
+                        }
+                    }
+                }
+            }
+
+            return minDistance;
+
```
pawelbaran, line 81: as above, blank line here is redundant

Architecture_Engine/Query/Distance.cs
```diff
@@ -0,0 +1,90 @@
...
+                }
+            }
+
+            return minDistance;
+            
+        }
+
+        /***************************************************/
+
+
```
pawelbaran, line 86: same as above

Architecture_Engine/Query/IsValidForDistanceCheck.cs
```diff
@@ -0,0 +1,100 @@
...
+            {
+                return false;
+            }
+
+            // Check if the openings are closer than maxDistance
+            if (!double.IsNaN(maxDistance))
+            {
+                double diag1 = opening1.MaxDiagonal();
+                double diag2 = opening2.MaxDiagonal();
+
```
pawelbaran, line 67: I would add a check if neither `diag1` nor `diag2` is `double.NaN` (that is what is going to be returned in case of an opening with unusual shape being fed into `MaxDiagonal`). If the above is true, we need to calculate the distance because the pair cannot be excluded.

```
if (double.IsNaN(diag1) || double.IsNaN(diag2))
    return true;
```

Architecture_Engine/Query/IsValidForDistanceCheck.cs
```diff
@@ -0,0 +1,100 @@
...
+                Point point1onPlane2 = centerPoint1.Project(plane2);
+
+                double c = centerPoint2.Distance(point1onPlane2);
```
pawelbaran, line 70: this variable is never used , so I guess it is redundant

Architecture_Engine/Query/IsValidForDistanceCheck.cs
```diff
@@ -0,0 +1,100 @@
...
+
+                if (maxDistance + (diag1/2 + diag2/2) + distanceTolerance < centerPoint2.Distance(point1onPlane2)) 
+                {
+                    return false;
+                }
+            }
+
+            return true;
+
```
pawelbaran, line 79: redundant blank

Architecture_Engine/Query/IsValidForDistanceCheck.cs
```diff
@@ -0,0 +1,100 @@
...
+        }
+
```
pawelbaran, line 81: In BHoM we try to keep the code readable by separating the methods with decorators, on top of that each method type group (e.g. `private` vs `public`) should have its dedicated header - please feel free to explore other files for inspiration, in the meantime I post a suggestion how would I do it in this case.

```suggestion

        /***************************************************/
        /**** Private Methods                           ****/
        /***************************************************/
```

Architecture_Engine/Query/Distance.cs
```diff
@@ -0,0 +1,90 @@
...
+ */
+
+using BH.oM.Architecture.BuildersWork;
+using BH.oM.Architecture.Elements;
+using BH.oM.Geometry;
+using System.Collections.Generic;
+using System.Linq;
+using BH.oM.Spatial.ShapeProfiles;
+using BH.Engine.Geometry;
+using BH.oM.Geometry.CoordinateSystem;
```
pawelbaran, line 30: You have some unused usings here - you can clean them up easily by right clicking in VS => Organise Usings => Remove and Sort Usings (you can also use keyboard shortcuts ofc).

pawelbaran (changes requested):
I went through the code thoroughly and complained about every single issue I could find - sometimes possibly going over the top, but wanted to make sure that the good practice is cultivated. Not much to complain functionality wise, apart from slight performance issues as well as potential errors in case of unusual shapes. Everything is explained in the code comments.

Regarding the test files, they are really nice as for the 1st attempt ever, particularly `IsValidForDistanceCheck` - could you please randomise the inputs to `Distance` in a similar way, so that there was more test data than only single opening vs single opening?

Good stuff!

Architecture_Engine/Query/Distance.cs
FraserGreenroyd: Remind me next week, we do have something but I don't think it's up on the wiki and it really should be, so remind me and I'll make sure it gets up there :smile:

Architecture_Engine/Query/Distance.cs
```diff
@@ -0,0 +1,90 @@
...
+                        Cartesian global = new Cartesian();
+                        ICurve edge1O = edge1.Orient(global, opening1.CoordinateSystem);
+                        ICurve edge1Onew = edge1O.IProject(plane2);
+                        ICurve edge2O = edge2.Orient(global, opening2.CoordinateSystem);
+
+                        double dist = edge1Onew.Distance(edge2O);
+                        if (dist < minDistance)
+                        {
+                            minDistance = dist;
+                        }
```
FraserGreenroyd, line 75:
```suggestion
                        minDistance = Math.Min(minDistance, dist);
```

Not sure how this fits in with @pawelbaran above comment but this is a slightly neater way of ensuring the minimum distance is always the smaller of the two variables that is quite common place in various parts of the code. As this isn't my engine, this is not a mandatory suggestion if you prefer to keep the `if` statement that's fine too :smile;

Architecture_Engine/Query/IsValidForDistanceCheck.cs
```diff
@@ -0,0 +1,100 @@
...
+    public static partial class Query
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        public static bool IsValidForDistanceCheck(this Opening opening1, Opening opening2, double maxDistance = double.NaN, double distanceTolerance = Tolerance.Distance, double angleTolerance = Tolerance.Angle)
+        {
+            // Check if the openings are parallel
+            if (opening1.CoordinateSystem.Z.IsParallel(opening2.CoordinateSystem.Z, angleTolerance) == 0)
```
FraserGreenroyd, line 44: I haven't checked but why is `IsParallel` returning a number? Shouldn't it be returning a `true`/`false`? @pawelbaran maybe you know about this? Just seems odd to me that we would need to `== 0`?

Equally, if we must do `== 0` because `IsParallel` is being weird and returning a number rather than a `bool`, then it is probably good to add a tolerance, in case the return is `0.00000000000000000001` which is `!= 0` so checking for a range of acceptable values on a tight tolerance is probably preferable?

pawelbaran: This method returns an `int` as follows:
- -1 when the vectors have same direction, but different turn
- 0 when the vectors are not parallel
- 1 when the vectors have same direction and turn 

It would be faster if you looked up the method instead of writing lengthy comments @FraserGreenroyd 🙈 😉

FraserGreenroyd:
@BHoMBot check core

FraserGreenroyd:
@BHoMBot check compliance

FraserGreenroyd:
@BHoMBot check serialisation

FraserGreenroyd:
@BHoMBot check null-handling

FraserGreenroyd:
@BHoMBot check installer

FraserGreenroyd:
@BHoMBot check versioning

michal-pekacki:
Thanks for the comments, all remarks have been implemented in the code. Grasshopper files have been updated on SharePoint.

Architecture_Engine/Query/IsValidForDistanceInPlaneCheck.cs
```diff
@@ -0,0 +1,97 @@
...
+
+namespace BH.Engine.Architecture
+{
+    public static partial class Query
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        public static bool IsValidForDistanceInPlaneCheck(this Opening opening1, Opening opening2, double maxDistance = double.NaN, double distanceTolerance = Tolerance.Distance, double angleTolerance = Tolerance.Angle)
```
pawelbaran, line 37: I would reorder the words to make it sound a bit more English - feel free to disagree though, I am not a native speaker either 😉 
```suggestion
        public static bool IsValidForInPlaneDistanceCheck(this Opening opening1, Opening opening2, double maxDistance = double.NaN, double distanceTolerance = Tolerance.Distance, double angleTolerance = Tolerance.Angle)
```

In case you accept the suggestion, please remember to rename the .cs file too.

pawelbaran (changes requested):
Looks really good, both the code and GH is really neat! Two minor changes suggested - please see the code comments.

Architecture_Engine/Query/IsValidForDistanceInPlaneCheck.cs
michal-pekacki: I agree and I will correct it 👍

pawelbaran (approved):
LGTM now, thank you @michal-pekacki!

pawelbaran:
@BHoMBot check required

Architecture_Engine/Query/IsValidForInPlaneDistanceCheck.cs
```diff
@@ -36,6 +36,12 @@ public static partial class Query
 
         public static bool IsValidForInPlaneDistanceCheck(this Opening opening1, Opening opening2, double maxDistance = double.NaN, double distanceTolerance = Tolerance.Distance, double angleTolerance = Tolerance.Angle)
         {
+            // Check if openings aren't nulls
+            if (opening1 == null || opening2 == null)
+            {
+                return true;
```
pawelbaran, line 42: You sure this should return `true`? This indicates there is a point in checking distance between `null` openings 🙈

Btw, I would also add an error saying that it is impossible to verify the query because the opening is `null`.

pawelbaran (changes requested):

Architecture_Engine/Query/IsValidForInPlaneDistanceCheck.cs
michal-pekacki: You are right, it should return `false`. I will add an error notification 👍

pawelbaran (approved):
Again, LGTM 👍

michal-pekacki:
@BHoMBot check required

michal-pekacki:
@FraserGreenroyd can you check why @BHomBot shows a null-handling issue?

FraserGreenroyd:
It's due to the age of the PR. Lemme know when this is ready to merge and we'll dispensate on this occasion :smile:

pawelbaran:
@BHoMBot dataset-compliance

pawelbaran:

pawelbaran:

michal-pekacki:
@FraserGreenroyd PR is ready to merge! 😃

FraserGreenroyd:
@BHoMBot this is a CI/CD instruction. I am authorising dispensation to be granted on check ref. 2468292055

PR #2453, opened 2021-04-07
Serialiser_Engine: RegisterClassMaps added for entire assemblies

pawelbaran:
### Issues addressed by this PR

Closes #2451

### Test files

Very minimalistic test on [SharePoint](http://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?RootFolder=%2Fsites%2FBHoM%2F02_Current%2F12_Scripts%2F02_Pull%20Request%2FBHoM%2FBHoM_Engine%2FSerialiser_Engine%2F%232453-RegisterClassMaps&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4).

PR #2456, opened 2021-04-13
Serialisation_Engine: Fix Serialisation of Generics

### LINKED ISSUE #2455: Copied BHoM components become magic
#### Description:

alt-copying-bug
As in header. It happens with both alt and crtl+c,v copying. I'm using today's (210412) daily alpha.

#### Steps to reproduce:

Copy working component.

#### Expected behaviour:

Pasted component shouldn't be magic.

#### Test file(s):

[SP](https://burohappold.sharepoint.com/:f:/s/BHoM/EgnY_Hzl49BIn5FI5tz6QtYBra3_DqC2GDhVAnn-7rfHKw?e=TOr75m)

adecler:

Closes #2455 

The changes made in [this PR](https://github.com/BHoM/BHoM_Engine/pull/2453) broke the serialisation of components. When moving the code around, one behaviour was changed: Generic types were trying to add their discriminator conventions every time they were serialised instead of just the first time (since the code for registering types for the first time and on the fly was merged into one). Unfortunately, Mongo throws an exception when you try to do that.

So here's the changes that I have made:
- Only skip the registration of the class map if the type is a generic type definition (e.g. `List<T>`), not just a generic type (e.g. `List<string>`. This should have been done this way from the very beginning so this is an improvement on the original code too. The reason why class maps cannot be created for generic type definitions is because objects cannot be created from them.
- Bring back the try-catch inside `RegisterClassMap` to cover all cases.  After testing, this never gets hit but better safe than sorry.  The obvious way to avoid an exception altogether would have been to check if a discriminator convention is already registered before trying to register a new one. Sadly, There is no way to that. `IsTypeDiscriminated` is for checking if a discriminator (not a discriminator convention) was already registered and `LookupDiscriminatorConvention` adds an arbitrary convention when called if one doesn't exist already.

### Test files
I have run the unit tests on Serialisation and UI. I have also ran the same tests from the UI, just to make sure and all seems to be working fine. The one on serialisation can be run from the PR or the GH version can be found [here](https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/03_Alpha/BHoM/BHoM_Engine/Serialiser_Engine/Serialisation_Engine.gh?csf=1&web=1&e=RAAhT1). The one on the UI is not available yet but the GH can be found [here](https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/03_Alpha/BHoM/BHoM_UI/BHoM_UI_Test.gh?csf=1&web=1&e=6PnBiD). I have also reopened a series of older files from Grasshopper, Excel, and Dynamo to make sure all is back to normal.

It would be great if you could do the same. The most important part though is to open older files to make sure the components are deserialising correctly. Note that Any file you have save yesterday might be broken if you were on the latest installer. 

### Additional comments
I think it is critical to understand how such a critical bug managed to find its way to the master without any of us catching it. As we need to take the necessary action for this to not happen again.   
- the unit tests on serialisation didn't catch the issue because this is only affecting the serialisation of UI component ( a bit long to explain why here) and those are tested in the unit tests for UI. So I would say, two actions are needed here:
   - @FraserGreenroyd , can we have the units test for `ComponentInstantiation` and `ComponentReadWriteCycle` available through the bot ? Might be worth to have them as required tests on critical repos like the BHoM Engine.
   - Add serialisation unit tests to cover the cases that make the serialization of components unique. I'll do that straight away.
- Opening any existing file in any UI would have made the bug obvious. This mean that neither the PR creator nor any reviewer did this during the review process. I find this really hard to believe but I cannot think of any explanation. As I was one of the reviewer, here's what I did during my review process:
   - Initial check by looking at the file changes directly on GitHub to get a better idea of the scope of changes and identify any obvious problem
   - Switch to the branch on my local machine and review the code from Visual studio as the new code is always too hard to go through on GitHub.
   - Run the following tests
       - TestRunner on both Serialisation and UI
       - Equivalent of the TestRunner tests from my Grasshopper test scripts 
       - Re-opened a series of older files from different UIs

The only way that procedure could have failed to catch the bug is if I forgot to recompile the BHoM_Engine or didn't notice that a compilation failed due to a UI being opened. I am pretty OCD with my procedures so I find this very surprising but I cannot find any other explanation so :man_facepalming: me on this one.

### DIFF
--- a/Serialiser_Engine/Compute/RegisterClassMap.cs
+++ b/Serialiser_Engine/Compute/RegisterClassMap.cs
@@ -41,15 +41,15 @@ namespace BH.Engine.Serialiser
         {
             if (!BsonClassMap.IsClassMapRegistered(type))
             {
-                if (type.IsEnum)
+                try
                 {
-                    MethodInfo generic = m_CreateEnumSerializer.MakeGenericMethod(type);
-                    generic.Invoke(null, null);
-                }
-                else if (!type.IsGenericType)
-                {
-                    try
+                    if (type.IsEnum)
                     {
+                        MethodInfo generic = m_CreateEnumSerializer.MakeGenericMethod(type);
+                        generic.Invoke(null, null);
+                    }
+                    else if (!type.IsGenericTypeDefinition)
+                    { 
                         BsonClassMap cm = new BsonClassMap(type);
                         cm.AutoMap();
                         cm.SetDiscriminator(type.FullName);
@@ -59,15 +59,15 @@ namespace BH.Engine.Serialiser
 
                         BsonClassMap.RegisterClassMap(cm);
 
+                        BsonSerializer.RegisterDiscriminatorConvention(type, new GenericDiscriminatorConvention()); 
+                    }
+                    else
                         BsonSerializer.RegisterDiscriminatorConvention(type, new GenericDiscriminatorConvention());
-                    }
-                    catch (Exception e)
-                    {
-                        Debug.WriteLine(e.ToString());
-                    }
                 }
-                else
-                    BsonSerializer.RegisterDiscriminatorConvention(type, new GenericDiscriminatorConvention());
+                catch (Exception e)
+                {
+                    Debug.WriteLine(e.ToString());
+                }
             }
         }
 


pawelbaran (approved):
Changes LGTM, seem to fix the issue - I have debugged the failing case, confirmed that it is a problem of `List<ParamInfo>` skipping a try/catch safety net on class map registration because of `(!type.IsGenericType)` requirement, and going into `RegisterDiscriminatorConvention` instead, crashing there. Now, when the safety net is widened to encompass the entire method, it should not crash any more.

Following @adecler's long PR description, I share a large part of 🤦‍♂️ (probably most of it) as the originator of the change that introduced this bug. To give some perspective from my end:
- I have opened the PR, refactored it based on suggestions of @adecler 
- I have created the [minimalistic test file](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?RootFolder=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F02%5FPull%20Request%2FBHoM%2FBHoM%5FEngine%2FSerialiser%5FEngine%2F%232453%2DRegisterClassMaps&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4), checked that the components and objects get created correctly and `ToJson` and `FromJson` work - what I have not done is to close the file and reopen it
- I ran the serialisation test successfully
- it works, so I switched BHoM_Engine to another branch straight away and did not do any more manual testing
- the PR got approved by @adecler 
- merged, yay!

So the major gap I see here is me not really understanding the exact scope that is being affected by my changes and thus not testing properly, plus no one opening the test file (which has been broken from the very beginning!). On top of that, I 100% agree with all suggestions made by @adecler regarding the improvements of serialisation tests - I think we both trusted the bot a bit too much this time.

Luckily, it looks like the fix is easy and works. Thanks @LMarkowski for spotting this and @adecler for fixing 👍

pawelbaran:
@BHoMBot check serialisation

pawelbaran:
@BHoMBot check versioning

pawelbaran:
@BHoMBot check installer

pawelbaran:
@BHoMBot check core

pawelbaran:
@BHoMBot check compliance

pawelbaran:
@BHoMBot check null-handling

LMarkowski (approved):
I can't give much input on the code itself, but tested funcionailty-wise and everything works correct.
I've done copying, saving and clustering (which also caused issues) in both Rhino 5 & 6, and everything seems to be serialised an de-serialised flawlessly.

PR #2457, opened 2021-04-15
Reflection_Engine: Add support for setting an enum property from a string

adecler:
### NOTE: Supports
https://github.com/BHoM/Excel_Toolkit/pull/283

   
Provides support for https://github.com/BHoM/Excel_Toolkit/pull/283 although this can be merged separately.
Otherwise as the title said, this enables setting an enum property by providing a string directly instead of an enum of the correct type.

### Test files
https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/02_Pull%20Request/BHoM/BHoM_Engine/Reflection_Engine/ExcelToolkit%23158-SetEnumPropertyWithString.gh?csf=1&web=1&e=lFCcQN

image

Although I find testing in Grasshopper easier, you can obviously test this in Excel VBA using something like this:

```vba
Dim bhom As New Excel_UI.Server

Dim obj As New Excel_UI.Object
Call obj.SetType("BH.oM.Acoustic.Rasti")
Call obj.SetProperty("Frequency", "BH.oM.Acoustic.Frequency.Hz125")

Dim objJson As String
objJson = ToJson(obj)
```

### DIFF
--- a/Reflection_Engine/Modify/SetPropertyValue.cs
+++ b/Reflection_Engine/Modify/SetPropertyValue.cs
@@ -90,9 +90,16 @@ namespace BH.Engine.Reflection
                 if (propType.IsEnum && value is string)
                 {
                     string enumName = (value as string).Split('.').Last();
-                    object enumValue = Enum.Parse(propType, enumName);
-                    if (enumValue != null)
-                        value = enumValue;
+                    try
+                    {
+                        object enumValue = Enum.Parse(propType, enumName);
+                        if (enumValue != null)
+                            value = enumValue;
+                    }
+                    catch
+                    {
+                        Engine.Reflection.Compute.RecordError($"An enum of type {propType.ToText(true)} does not have a value of {enumName}");
+                    }   
                 }
 
                 if (propType == typeof(Type) && value is string)
--- a/Reflection_Engine/Modify/SetPropertyValue.cs
+++ b/Reflection_Engine/Modify/SetPropertyValue.cs
@@ -102,6 +102,9 @@ namespace BH.Engine.Reflection
                     }
                 }
 
+                if (propType == typeof(Type) && value is string)
+                    value = Create.Type(value as string);
+
                 if (value != null)
                 {
                     if (value.GetType() != propType && value.GetType().GenericTypeArguments.Length > 0 && propType.GenericTypeArguments.Length > 0)
--- a/Reflection_Engine/Modify/SetPropertyValue.cs
+++ b/Reflection_Engine/Modify/SetPropertyValue.cs
@@ -26,6 +26,7 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.ComponentModel;
+using System.Linq;
 using System.Reflection;
 
 namespace BH.Engine.Reflection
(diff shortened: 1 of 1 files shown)


Reflection_Engine/Modify/SetPropertyValue.cs
```diff
@@ -86,6 +86,25 @@ public static object SetPropertyValue(this object obj, string propName, object v
...
+                    if (index > 0)
+                    {
+                        Type enumType = Create.Type(fullName.Substring(0, index));
+                        if (enumType != null)
+                        {
+                            object enumValue = Enum.Parse(enumType, fullName.Substring(index + 1));
+                            if (enumValue != null)
+                                value = enumValue;
+                        }
+                    }
```
IsakNaslundBh, line 102: Any reason to not just use the property type, instead of trying to create a type from the providede value?

Have not tested this, just looked on github, but should nto something like this work?

```suggestion
                    string enumName= (value as string).Split('.').Last();
                    object enumValue = Enum.Parse(propType, enumName);
                    if (enumValue != null)
                        value = enumValue;
```

adecler:
@BHoMBot check required

adecler:
@BHoMBot check required

FraserGreenroyd:
@BHoMBot check required

adecler:

adecler:

IsakNaslundBh (changes requested):
Have tested this now, and enums when input correctly seem to work just fine.

Two things found through:

1. The error given when inputting a faulty enum type is not that clear. I can understand what it means, but not sure the general user would. Might be worth wrapping it up in a try catch and return a more sensible message if `Enum.Parse()` fails:
image
Something stating that the provided string is not part of the enum or similar I feel would be much clearer.

2. Setting type from string does not seem to work from within GH. Suspect this might be a UI issue, and not really related to this PR. If that is the case, happy for this to be merged without this being fixed first, but good then if a issue in GH/BHoM_UI can be raised as well to fix.

I think the first point should be fixed before we merge this PR though.

Have updated the testfile to include the cases mentioned above.

adecler:
@IsakNaslundBh ,

Regarding the conversion of strings to types, this is indeed a Grasshopper issue on enforcing the type `Type` on the input once the type of object to modify has been identified. It is easy to go around it for testing purposes:

image

Not sure I agree on needed to have that as a working feature in Grasshopper though. It was not working before, nobody ever asked for it, and it is still not working now. The reason for this PR is to compensate for the limited functionality in VBA by providing extra ways to create a type, not to add new functionality to all UIs.

Not sure I agree with your first point either to be honest. `adasd` cannot be found as a valid value for this enum. All makes sense to me, especially for a advanced feature only really has its place within the context of a text-based usage of the BHoM. I have added a clearer error message anyways though.

adecler:
@BHoMBot check required

IsakNaslundBh:
> Not sure I agree with your first point either to be honest. adasd cannot be found as a valid value for this enum. All makes sense to me

As mentioned, it makes sense to me as well, but not sure that is the case for the full userbase 😄

adecler:
@IsakNaslundBh , are you happy with the change on the error message ? Can you do a final review, please ?

IsakNaslundBh (approved):
Happy now!

adecler:

adecler:

FraserGreenroyd:
@BHoMBot this is a CI/CD instruction. I am authorising dispensation to be granted on check ref. 2398912341

PR #2497, opened 2021-05-04
Environment_Engine: Skip IInternalEdges when set to null

### LINKED ISSUE #2496: Environment_Engine: OpeningsFromPhysical to account for null internal boundaries
Line 64 of the `OpeningsFromPhysical.cs` file tries to get the internal boundaries of a physical opening, but if the internal boundaries are `null`, then this fails and crashes out the method, which if used in conjunction with the `PanelsFromPhysical` method crashes that out to, returning no data to the user.

This should be resolved to not do this, and handle the instance of `null` internal boundaries more effectively.

At the same time, we should also fix line 85 of the `Panel` Create method, which suffers from a similar problem.

JohannaOlin1:

Closes #2496 

Fixes the bug where PanelsFromPhysical and Panel components are not running for panels with openings without internal edges.

### Test files

[TestFiles.zip](https://github.com/BHoM/BHoM_Engine/files/6420173/TestFiles.zip)

### DIFF
--- a/Environment_Engine/Create/Panel.cs
+++ b/Environment_Engine/Create/Panel.cs
@@ -94,8 +94,11 @@ namespace BH.Engine.Environment
             if (connectedSpaceName == null)
                 connectedSpaceName = Guid.NewGuid().ToString();
 
-            List<Polyline> openingLines = surface.IInternalEdges().Select(x => x.ICollapseToPolyline(angleTolerance)).ToList();
-            List<Opening> openings = new List<Opening>();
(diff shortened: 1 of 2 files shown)


FraserGreenroyd:
@BHoMBot check required

FraserGreenroyd:

FraserGreenroyd (approved):
This works well, fixes the bug found by @svaz00 and does not appear to have any side effects :smile:

Thanks @JohannaOlin1 for the quick fix!

PR #2507, opened 2021-05-26
Structure_Engine: Added methods for ReinforcementDensity, BarRequiredArea and MeshRequiredArea classes

peterjamesnugent:
### NOTE: Depends on 

https://github.com/BHoM/BHoM/pull/1247
   
Closes https://github.com/BHoM/BHoM/issues/1239

### Changelog

- Added `Compute` methods to summate the required area in `BarRequiredArea` and `MeshRequiredArea`
- Added `Compute` method to evaluate a `ReinforcementDensity` object based on a `BarRequiredArea`
- Made the `FindIdentifier` method public and migrated to the BHoM_Engine
- Added descriptions for all `Min` and `Max` methods in the Results_Engine
- Added `MaxEnvelope`, `MaxBarRequiredArea` and `MaxMeshRequiredArea` methods 
- Added `MinEnvelope`, `MinBarRequiredArea` and `MinMeshRequiredArea` methods 
- Added `null` checks to the `Min` and `Max` methods in the Results_Engine

### Additional comments

~~We need to have some form of `Region` implemented for `IAreaElement` to rationalise them if we want to move freely between the different representations. For a `Bar` we can easily work out the length and extrapolate the required area along it, but it's not so easy with an `IAreaElement` - so some form of `Region` that makes use of the `Orientation` to determine the principle direction would be useful. I think this will set us up nicely for the next step which is the `RebarIntent` representation.~~

There is future scope to make use of the `MarchingSquares` methods to allow users to produce contour lines from the `MeshRequiredArea`. This could then be used with a section cut or a region to then calculate the maximum reinforcement in a region.


peterjamesnugent:
@BHoMBot check compliance

IsakNaslundBh:
@BHoMBot check required

peterjamesnugent:
@BHoMBot check required

Structure_Engine/Compute/ReinforcementDensity.cs
```diff
@@ -0,0 +1,199 @@
...
+                Reflection.Compute.RecordError("The Materials are null, these are required to calculate the density of the reinforcing objects.");
+                return null;
+            }
+            else if (bars == null || bars.All(x => x == null))
+            {
+                Reflection.Compute.RecordError("The Bars are null, these are required to calculate the density of the reinforcing objects.");
+                return null;
+            }
+
+            Dictionary<string, IMaterialFragment> materialsDict = materials.ToDictionary(x => x.Name.ToString());
```
pawelbaran, line 68: Why `x.Name.ToString()` instead of simply `x.Name`?

Structure_Engine/Compute/ReinforcementDensity.cs
```diff
@@ -0,0 +1,199 @@
...
+        [Description("Calculates the ReinforcementDensity of a MeshRequiredArea from a given Bar and Material.")]
+        [Input("meshRequiredArea", "The BarRequiredArea containing the reinforced areas and ids for the Bar and Material.")]
+        [Input("elements", "The elements to search for the relevant AreaElement associated with the MeshRequiredArea.")]
+        [Input("materials", "The Materials to search for the relevant Material associated with the MeshRequiredArea.")]
+        [Output("reinforcementDensity", "The ReinforcementDensity calculated using the inputs provided.")]
+        public static ReinforcementDensity ReinforcementDensity(MeshRequiredArea meshRequiredArea, List<IAreaElement> elements, List<IMaterialFragment> materials)
+        {
+            Reflection.Compute.RecordError("Requires an implementation of Regions before the required area can be extrapolated.");
+            return null;
+            //Add to IsNull when PR is merged
```
pawelbaran, line 127: I guess this is meant to stay here for the future, but just to double check - please make sure the code below is needed and should not be deleted.

pawelbaran (commented):
I went through the code and I am afraid this should be reviewed by @IsakNaslundBh due to all the magic that is happening around the bar results - the way of juggling the ids and materials is not intuitive to me and I have zero experience with it, so I would rather not be the one to approve it 🙈 

However, naturally I can give it a deep dive if really needed (i.e. @IsakNaslundBh is too busy in the foreseeable future).

Structure_Engine/Compute/ReinforcementDensity.cs
```diff
@@ -115,81 +115,15 @@ public static ReinforcementDensity ReinforcementDensity(BarRequiredArea barRequi
...
+        [NotImplemented]
         [Description("Calculates the ReinforcementDensity of a MeshRequiredArea from a given Bar and Material.")]
         [Input("meshRequiredArea", "The BarRequiredArea containing the reinforced areas and ids for the Bar and Material.")]
         [Input("elements", "The elements to search for the relevant AreaElement associated with the MeshRequiredArea.")]
         [Input("materials", "The Materials to search for the relevant Material associated with the MeshRequiredArea.")]
         [Output("reinforcementDensity", "The ReinforcementDensity calculated using the inputs provided.")]
         public static ReinforcementDensity ReinforcementDensity(MeshRequiredArea meshRequiredArea, List<IAreaElement> elements, List<IMaterialFragment> materials)
         {
-            Reflection.Compute.RecordError("Requires an implementation of Regions before the required area can be extrapolated.");
             return null;
```
pawelbaran, line 107: I believe there should be some sort of error here to inform the consumer why `null` is returned

pawelbaran (changes requested):
Just a quick issue related to one of the recent commits.

Structure_Engine/Compute/ReinforcementDensity.cs
peterjamesnugent: It has a `[Not Implemented]` attribute at the top so should not even appear in the UI - it was more a placeholder so we could develop the method later. I can remove it if you'd like.

pawelbaran: I guess one could still try to call this method from the code. Of cource not too likely but better be on the safe side IMO. Anyways, it is a minor one, I would not bother about it now, rather add it to the batch in case other issues pop up during the review.

Structure_Engine/Compute/ReinforcementDensity.cs
```diff
@@ -0,0 +1,114 @@
...
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+        [Description("Calculates the ReinforcementDensity of a BarRequiredArea from a given Bar and Material.")]
+        [Input("barRequiredArea", "The BarRequiredArea containing the reinforced areas and ids for the Bar and Material.")]
+        [Input("bars", "The Bars to search for the relevant Bar associated with the BarRequiredArea.")]
+        [Input("materials", "The Materials to search for the relevant Material associated with the BarRequiredArea.")]
+        [Output("reinforcementDensity", "The ReinforcementDensity calculated using the inputs provided.")]
+        public static ReinforcementDensity ReinforcementDensity(BarRequiredArea barRequiredArea, List<Bar> bars, List<IMaterialFragment> materials)
```
IsakNaslundBh, line 49: As a first quick comment, would make this list take  list of `BarRequiredArea` as input and output a list of `ReinforcementDensity` with the same length.

This is mainly for efficiency reasons, as the current way the method is set up, the dictionaries etc have to be generated once for each `BarRequiredArea`.

peterjamesnugent:
@BHoMBot check compliance

peterjamesnugent:
@BHoMBot check compliance

peterjamesnugent:
@BHoMBot check null-handling

Results_Engine/Query/Max.cs
```diff
@@ -22,22 +22,27 @@
...
+
 namespace BH.Engine.Results
 {
     public static partial class Query
     {
         /***************************************************/
         /**** Public Methods                            ****/
         /***************************************************/
-
+        [Description("Determines the maximum force for each component of the BarForce and returns a new BarForce with its concurrent forces.")]
```
IsakNaslundBh, line 41: minor comment on the description here. This is not returning a "new" Bar force, but is just giving the ones with maximum values for all different component. Also, it is always returning a list of length six, not a single new one.

Maybe something like (feel free to reword):

```suggestion
        [Description("Returns a list of the six BarForces that show the maximum value for each force or moment component respectively. The forces returned will have correct concurrent forces for the other component.")]
```

**EDIT**: Something like what you have in the output description I think is clearer.

Results_Engine/Query/Max.cs
```diff
@@ -50,21 +55,62 @@ public static List<BarForce> MaxForces(this IEnumerable<BarForce> forces)
...
+        [Input("forces", "The BarForces to be considered.")]
+        [Output("maxResults", "A list of BarForces, one for each id, enveloped to produce the maximum forces for that id.")]
         public static List<BarForce> MaxEnvelopeByObject(this IEnumerable<BarForce> forces)
         {
             return forces.GroupByObjectId().Select(x => x.MaxEnvelope(true, false)).ToList();
         }
 
         /***************************************************/
-
+        [Description("Determines the maximum force in each component of the BarForce and returns a single BarForce object with the enveloped forces.")]
```
IsakNaslundBh, line 118: Would note here that the resulting force will not be concurrent. Same for all methods where this is true.

IsakNaslundBh (changes requested):
Some comments on the description. Mainly making sure it is clear what the method is doing and not. Mainly with regard to if the forces are concurrent or not.

peterjamesnugent:
@BHoMBot check null-handling

peterjamesnugent:
@BHoMBot check compliance

BHoM_Engine/Query/FindIdentifier.cs
```diff
@@ -37,26 +37,32 @@ public static partial class Query
...
         /***************************************************/
 
-        [Description("Tries to find a IAdapterId on the object.")]
+        [Description("Tries to find an IAdapterId on the object.")]
+        [Input("o", "The object to search for an IAdapterId.")]
         [Output("identifier", "First plausible identifier present on the object.")]
         public static Type FindIdentifier(this IBHoMObject o)
         {
-            return o.Fragments.FirstOrDefault(fr => fr is IAdapterId)?.GetType();
+            return o?.Fragments.FirstOrDefault(fr => fr is IAdapterId)?.GetType();
```
IsakNaslundBh, line 45: Some suggestions for centralising messaging

```suggestion
            if (o == null)
            {
                Engine.Reflection.Compute.RecordError("Provided object is null. Cannot extract identifier.");
                return null;
            }
            Type adapterIdType = o.Fragments.FirstOrDefault(fr => fr is IAdapterId)?.GetType();

            if (adapterIdType == null)
            {
                Engine.Reflection.Compute.RecordError("No Identifier found");
                return null;
            }
            else
            {
                Engine.Reflection.Compute.RecordNote("Auto-generated Identifier as " + adapterIdType.Name);
                return adapterIdType;
            }
```

BHoM_Engine/Query/FindIdentifier.cs
```diff
@@ -0,0 +1,83 @@
...
+                else
+                    Reflection.Compute.RecordNote("Auto-generated Identifier as " + adapterIdType.Name);
+            }
+            else if (!typeof(IAdapterId).IsAssignableFrom(adapterIdType))
+            {
+                Reflection.Compute.RecordError("The provided adapterIdType need to be a type of IAdapterId.");
+                return null;
+            }
+
+            return adapterIdType;
```
IsakNaslundBh, line 76: With the above, this can be simplified down to:

```suggestion
            if (adapterIdType == null)
            {
               return o.FindIdentifier();
            }
            else if (!typeof(IAdapterId).IsAssignableFrom(adapterIdType))
            {
                Engine.Reflection.Compute.RecordError("The provided adapterIdType need to be a type of IAdapterId.");
                return null;
            }
            return adapterIdType;
```

IsakNaslundBh (changes requested):
Some tweaks to checks and messaging of FIndIdentifier

peterjamesnugent:
@BHoMBot check required

peterjamesnugent:

Results_Engine/Results_Engine.csproj
```diff
@@ -1,113 +1,117 @@
...
+    </Reference>
+    <Reference Include="Graphics_oM">
+      <HintPath>C:\ProgramData\BHoM\Assemblies\Graphics_oM.dll</HintPath>
+      <Private>False</Private>
+    </Reference>
+    <Reference Include="Reflection_oM">
+      <HintPath>C:\ProgramData\BHoM\Assemblies\Reflection_oM.dll</HintPath>
+      <Private>False</Private>
+    </Reference>
+    <Reference Include="Structure_Engine">
```
IsakNaslundBh, line 57: Do not think this dependency can be allowed actually. Result_Engine should sit under Structure_Engine.

Assume it is only there for the sake of the IsNull methods? Might have to live without them for this case I think.

IsakNaslundBh (changes requested):
Dependency chain issue, see bellow.

Suggesting either to add a IsNull method in Result_Engine accepting IResult or living without that method, doing explicit nullchecks and custom messages.

peterjamesnugent:
@BHoMBot check compliance

peterjamesnugent:
@BHoMBot check compliance

IsakNaslundBh:
@BHoMBot check required

IsakNaslundBh:
Will be ready to dispensate the null handling for this PR as it is only failing due to problems with the test unable to generate test data for a generic method:

image

IsakNaslundBh (approved):
Looks good to me now!

IsakNaslundBh:
@BHoMBot this is a CI/CD instruction. I am authorising dispensation to be granted on check ref. 2959754649

IsakNaslundBh:
@BHoMBot check ready-to-merge

IsakNaslundBh:
@BHoMBot check ready-to-merge

IsakNaslundBh:
@BHoMBot check ready-to-merge

Issue #2487, opened 2021-04-19
Geometry_Engine - handle nulls


FraserGreenroyd:
Note, where the `method` begins with `BH.oM` - this means it is a `Create` method.

The following methods need appropriate `null` handling of their inputs. You can use `@BHoMBot check null-handling` on a draft PR to check progress if needed - you're ultimately looking for a reduction in errors from the `master`/`main` branch. I can run more localised tests to confirm against if needed to confirm specific methods.

 - [ ] BH.Engine.Geometry.Compute.CurveProximity(Line curve1, Polyline curve2, Double tolerance)
 - [ ] BH.Engine.Geometry.Compute.CurveProximity(Arc curve1, Line curve2, Double tolerance)
 - [ ] BH.Engine.Geometry.Compute.CurveProximity(Arc curve1, Arc curve2, Double tolerance)
 - [ ] BH.Engine.Geometry.Compute.CurveProximity(Arc curve1, Circle curve2, Double tolerance)
 - [ ] BH.Engine.Geometry.Compute.CurveProximity(Arc curve1, PolyCurve curve2, Double tolerance)
 - [ ] BH.Engine.Geometry.Compute.CurveProximity(Arc curve1, Polyline curve2, Double tolerance)
 - [ ] BH.Engine.Geometry.Compute.CurveProximity(Circle curve1, Line curve2, Double tolerance)
 - [ ] BH.Engine.Geometry.Compute.CurveProximity(Circle curve1, Arc curve2, Double tolerance)
 - [ ] BH.Engine.Geometry.Compute.CurveProximity(Circle curve1, Circle curve2, Double tolerance)
 - [ ] BH.Engine.Geometry.Compute.CurveProximity(Circle curve1, PolyCurve curve2, Double tolerance)
 - [ ] BH.Engine.Geometry.Compute.CurveProximity(Circle curve1, Polyline curve2, Double tolerance)
 - [ ] BH.Engine.Geometry.Compute.CurveProximity(PolyCurve curve1, Line curve2, Double tolerance)
 - [ ] BH.Engine.Geometry.Compute.CurveProximity(PolyCurve curve1, Arc curve2, Double tolerance)
 - [ ] BH.Engine.Geometry.Compute.CurveProximity(PolyCurve curve1, Circle curve2, Double tolerance)
 - [ ] BH.Engine.Geometry.Compute.CurveProximity(PolyCurve curve1, Polyline curve2, Double tolerance)
[...]

PR #2516, opened 2021-06-14
Geometry_Engine: Null handling

rolyhudson:

Closes #

### DIFF
--- a/Geometry_Engine/Compute/BooleanDifference.cs
+++ b/Geometry_Engine/Compute/BooleanDifference.cs
@@ -37,6 +37,9 @@ namespace BH.Engine.Geometry
 
         public static List<Line> BooleanDifference(this Line line, Line refLine, double tolerance = Tolerance.Distance)
         {
+            if (line.IsNull() || refLine.IsNull())
+                return null;
+
             if (refLine.Length() <= tolerance)
                 return new List<Line> { line.DeepClone() };
 
@@ -127,6 +130,9 @@ namespace BH.Engine.Geometry
         
         public static List<Polyline> BooleanDifference(this Polyline region, List<Polyline> refRegions, double tolerance = Tolerance.Distance)
         {
+            if (region.IsNull() || refRegions.Any(x => x.IsNull()))
+                return null;
+
             if (!region.IsClosed(tolerance) || refRegions.Any(x => !x.IsClosed()))
             {
                 Reflection.Compute.RecordError("Boolean Difference works on closed regions.");
--- a/Geometry_Engine/Compute/BooleanIntersection.cs
+++ b/Geometry_Engine/Compute/BooleanIntersection.cs
@@ -96,6 +96,9 @@ namespace BH.Engine.Geometry
 
         public static List<Line> BooleanIntersection(this Line line, List<Line> refLines, double tolerance = Tolerance.Distance)
         {
+            if (line.IsNull() || refLines.Any(x => x.IsNull()))
+                return null;
+
             List<Line> result = new List<Line>();
             if (line.Length() <= tolerance)
                 return result;
@@ -133,6 +136,9 @@ namespace BH.Engine.Geometry
 
         public static List<Polyline> BooleanIntersection(this Polyline region, Polyline refRegion, double tolerance = Tolerance.Distance, double angleTolerance = Tolerance.Angle)
         {
+            if (region.IsNull() || refRegion.IsNull())
+                return null;
+
             if (!region.IsClosed(tolerance) || !refRegion.IsClosed(tolerance))
             {
                 Reflection.Compute.RecordError("Boolean Union works on closed regions.");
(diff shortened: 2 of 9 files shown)


rolyhudson:
@BHoMBot check null-handling

Geometry_Engine/Query/IsNullOrContainsNulls.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Threading.Tasks;
+
+namespace BH.Engine.Geometry
+{
+    /***************************************************/
+    /**** Public Methods                            ****/
+    /***************************************************/
+    public static partial class Query
+    {
+        [Description("Checks if a List of Geometry is null and if any element in the List is null and outputs relevant error message.")]
```
peterjamesnugent, line 18: @rolyhudson I added an `IsNullOrEmpty` in the BHoM_Engine here: https://github.com/BHoM/BHoM_Engine/blob/master/BHoM_Engine/Query/IsNullOrEmpty.cs

The thinking was it will check if the `IEnumerable<T>` was null or it was empty, and then within the actual code you would make use of the `IsNull` method for the object itself . `IEnumberable<T>.Any(x => x.IsNull())`

This way the error message returned is not generic and tells you what object is null rather than an object in the list is null.

rolyhudson: @peterjamesnugent cheers for looking at this appreciate your thoughts. 
I did see that method and the reasons I did not implement as you suggest are multiple...

First up I'm conscious of avoiding adding excessive lines of code with this checking of null handling. The intention of the  `IsNullOrContainsNulls` is to try and manage what could be come a maintenance headache by providing a single method to check any collection of IGeometries.

I am aware of the need for decent message reporting for users. These checks could run quite deep so I am not sure tracking the only object that is null is necessarily helpful. 
For example if a null Point is found in the ControlPoints of one of a an input collection of Polylines, is it useful to report a Point was found to be null?

To be useful would, I think, require providing a stack of messages tracking back to the offending input, which in some situations could be inefficient as the string processing itself could slow things down in some situations.

My thinking (yet to be committed) is just to report the name of the method input parameter where a null was found and nothing more.

It is still a WIP and really needs scrutiny from others before rolling out across the 600+ methods. Maybe worth a conversation with  you very soon.

peterjamesnugent: Understand your point about excessive lines of null checks - one thing that did come up on the Structure_Engine PR would be an option for a "deep check", especially for geometry objects that may be worth considering on this PR: https://github.com/BHoM/BHoM_Engine/issues/2487#issuecomment-853314761

> For example if a null Point is found in the ControlPoints of one of a an input collection of Polylines, is it useful to report a Point was found to be null?

With the `[CallerMemberName]` attribute it would feedback that "Cannot evaluate ControlPoints because a Point is null" - but I see your point about traceability, especially given that so many other Engine methods make use of Geometry_Engine methods. The user might not even know the `ControlPoints` method is being used and leads to even more confusion in the error message!

rolyhudson: @peterjamesnugent - updated this following the conversation. The method `ContainsNulls` cannot move to the base as it is specific to `IEnumerable<IGeometry>` as it call `Geometry.Query.IIsNull()` on each element in the collection. 

This is looking better but I think still needs further tuning in how the messages are handled.

I have added a basic test script  [here](https://burohappold.sharepoint.com/:f:/r/sites/BHoM/02_Current/12_Scripts/02_Pull%20Request/BHoM/BHoM_Engine/Geometry_Engine/%232516-NullHandling?csf=1&web=1&e=rDCZ9p). Quite simple for now - just working with Polylines and the BooleanDifference method.

rolyhudson:
First attempt at diagramming what is going on with this PR in its current state.
Is this insane? Would be good to get a wider consensus on the approach @peterjamesnugent @IsakNaslundBh @FraserGreenroyd @al-fisher.
Screenshot 2021-06-28 121550

peterjamesnugent:
@rolyhudson the diagram looks great- I think it aligns with what we discussed. The challenge is to either make the method generic enough that it can just accept an IGeometry and loop through as you've illustrated with the deep checks etc. or just having `IsNull` methods for the various objects in the Geometry_oM.

The only thing it doesn't capture is how the messaging is propagated and how much detail that we want to feed back up the chain. Last time we concluded that nested methods shouldn't be a problem because the null will get picked up early, with the deep check at least so I guess there's a decision to be made on how much you want to feedback to the user that it's useful.

FraserGreenroyd:
This PR has gone stale per our [guidelines](https://bhom.xyz/documentation/Contributing/Pull-Requests/#stale-pull-requests) and following discussion with @IsakNaslundBh it has been decided that this PR would not be suitable given the conflicts, and changes to the Geometry Engine since this PR has been raised. Due to the introduction of `check-null-handling` it is also difficult to make the situation worse going forward, so although this PR aimed to make things better, we will tackle it as smaller and more discrete tasks going forward. Thanks for the help though @rolyhudson 😄

PR #2522, opened 2021-06-25
Description query added for IRequest

pawelbaran:

Closes #

### Test files

Not there yet.


adecler (commented):
I would prefer to see the existing `IDescription` method in the reflection engine updated to include support to `TryRunExtensionMethod` instead. I don't see the benefit of adding a version that only offers that feature for `IRequest`.

pawelbaran:
100% agreed @adecler - I will close this thread and reopen a new one following the chain of other PRs 👍

PR #2530, opened 2021-07-06
Verification: Make sure NullCheck method can call Generic methods

### LINKED ISSUE #2529: Verification: NullChecks does not handle generic methods
#### Description:

Generic methods crash out when being invoked as they have not been given generic parameters. A simple call to MakeFromGenerics should resort this issue

IsakNaslundBh:

Closes #2529 

Adding check if the method is a Generic method, and if so, making use of the MakeFromGeneric method to make sure it can be called.

Code without this change simply crashes for all generic methods.

### Test files

Testfile for a couple of methods here:
https://burohappold.sharepoint.com/:f:/s/BHoM/EjNwbhC-4KFMkZBRlmhKV0gBNiFkDp-F0fKFZXf2tBeuuQ?e=pWaxCa

Note that all of the methods in the testfile are not passing, but that they now throw more appropriate errors and warnings, not failing to cast errors.

### DIFF
--- a/.ci/code/Engine_Test/Verify/NullChecks.cs
+++ b/.ci/code/Engine_Test/Verify/NullChecks.cs
@@ -87,16 +87,13 @@ namespace BH.Test.Engine
             {
                 if (method.IsGenericMethodDefinition)
                     method = method.MakeFromGeneric();
+
+                if (method == null)
+                    return GenericsFailedResult(methodDescription);
             }
             catch (Exception e)
             {
-                return new TestResult
-                {
-                    Description = methodDescription,
-                    Status = TestStatus.Warning,
-                    Message = $"Warning: Failed to make method {methodDescription} into a generic method. It will not be tested.",
-                    Information = new List<ITestInformation> { new EventMessage { Message = e.Message, StackTrace = e.StackTrace } }
-                };
+                return GenericsFailedResult(methodDescription, e);
             }
 
             // Collect the inputs setting themm to null when relevant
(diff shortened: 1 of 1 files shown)


IsakNaslundBh:
@BHoMBot check null-handling

adecler (dismissed):
Looks good to me 👍

FraserGreenroyd (approved):
@IsakNaslundBh a test file might be beneficial just because BHoMBot does not generally check verification code against a PR. The code LGTM though so I'm happy with this being merged in as my local test was happy. Just thinking a test file for future use may be beneficial :smile:

IsakNaslundBh:
after chat with @adecler realised we needed to actually null check the method after trying to turn it into a generic type. Moved the return TestResult with message to its own method as wanted to return the same thing in the catch and for the null check

adecler:
I would still add a null check to make sure the input method itself is not null. It is a method focused on null checks after all so better if respects its own rules.

IsakNaslundBh:
> I would still add a null check to make sure the input method itself is not null. It is a method focused on null checks after all so better if respects its own rules.

Fair point. Check added

IsakNaslundBh:
> @IsakNaslundBh a test file might be beneficial just because BHoMBot does not generally check verification code against a PR. The code LGTM though so I'm happy with this being merged in as my local test was happy. Just thinking a test file for future use may be beneficial 😄

Have added a test file now! Link in description :)

adecler (approved):
looks good to me now

IsakNaslundBh:
@BHoMBot check required

IsakNaslundBh:
@BHoMBot check complaince

IsakNaslundBh:
@BHoMBot check compliance

IsakNaslundBh:
@BHoMBot check ready-to-merge

IsakNaslundBh:
Requesting dispensation for this PR as the methods changed are not part of the main system

IsakNaslundBh:
@BHoMBot this is a CI/CD instruction. I am authorising dispensation to be granted on check ref. 3008948011

PR #2535, opened 2021-07-09
Reflection_Engine: Provide better support for setting a DateTime property with a string input

### LINKED ISSUE #2534: Reflection_Engine: Provide better support for setting a DateTime property with a string input
#### Description:
- Try to convert the string into a date
- Provide a proper error message if it fails

adecler:

Closes #2534

This is as good as it can get without providing the format of the string containing the date. The C# method `dateTime.TryParse` does its best to figure out the format (e.g. even if only the date is provided without the time) but there will always be cases that are unsolvable. For example the American `mm-dd-yyyy` vs the British `dd-mm-yyyy`.  Better than what we had before without breaking anything so I am happy with the state of this PR.

### Test files
This was mainly created within the context of the [Building Performance Excel form](https://github.com/BuroHappoldEngineering/BuildingPerformance_Toolkit/pull/6). So I recommend to make sure invalid dates are now properly handled within that context.

### DIFF
--- a/Reflection_Engine/Modify/SetPropertyValue.cs
+++ b/Reflection_Engine/Modify/SetPropertyValue.cs
@@ -102,6 +102,18 @@ namespace BH.Engine.Reflection
                     }   
                 }
 
+                if (propType == typeof(DateTime) && value is string)
+                {
+                    DateTime date;
+                    if (DateTime.TryParse(value as string, out date))
+                        value = date;
+                    else
+                    {
+                        Engine.Reflection.Compute.RecordError($"The value provided for {propName} is not a valid DateTime.");
+                        value = DateTime.MinValue;
+                    }
+                }
+
                 if (propType == typeof(Type) && value is string)
                     value = Create.Type(value as string);
 


tg359 (approved):
Tested via the BuildingPerformance_Toolkit input form where this error is reflected up to the user when given something silly like "2021-02-200".

adecler:
@BHoMBot check required

Reflection_Engine/Modify/SetPropertyValue.cs
```diff
@@ -102,6 +102,18 @@ public static object SetPropertyValue(this object obj, string propName, object v
                     }   
                 }
 
+                if (propType == typeof(DateTime) && value is string)
```
FraserGreenroyd, line 105: What if someone tries to set an int as the DateTime using a number of seconds since 1970 rather than a string? Is that a use case we want to consider?

adecler: I was sure someone would come in and ask for more than this PR wants to cover 😋 

I purposefully kept the scope of this PR as small as possible because this can become a fairly complex problem. I actually did an initial version of this same change a few weeks ago and threw it away because it was getting increasingly complicated. 

So the point of this PR is to do a very small improvement on the existing code without trying to solve the whole thing. As I said in the original comment, the problem doesn't really need to be solved for anything but Excel VBA. So  I don't think there is value in investing too much time in it until proven otherwise.

adecler:
@BHoMBot check ready-to-merge

PR #2586, opened 2021-07-30
Geometry_Engine: Added tolerance to Create PlanarSurface and methods it relies on

### LINKED ISSUE #2575: Geometry_Engine: Tolerance added to PlanarSurface create method
#### Description:
Tolerance needs to be added as a parameter to the PlanarSurface Create method for use when creating PlanarSurfaces with varying levels of tolerance of their external/internal curves.

enarhi:

Closes #2575 

Tolerance param has been added to the create planar surface method and the `IArea` method that it relies on by way of the `DistributeOutlines` method. Take note of the change this has on the `IArea` method, as it now invokes tolerance but only a couple of the overloads actually use this tolerance (eg if `IIsClosed` is called before getting the area).

### Test files
https://burohappold.sharepoint.com/:f:/r/sites/BHoM/02_Current/12_Scripts/02_Pull%20Request/BHoM/BHoM_Engine/Geometry_Engine/%232575-ToleranceForPlanarSurfaceCreate?csf=1&web=1&e=Y0M3jI

Note that the test file shows a specific example of tolerance behaving as desired, but does not cover all existing unit tests and the like. @pawelbaran @rolyhudson please add a link here to any relevant test files we already have for this create method, as I could not find them if they do already exist.

### DIFF
--- a/Geometry_Engine/Compute/DistributeOutlines.cs
+++ b/Geometry_Engine/Compute/DistributeOutlines.cs
@@ -43,6 +43,18 @@ namespace BH.Engine.Geometry
         [Output("distributedOutlines", "Lists representing each set of outlines. For each list, the first outline is the containing outline and the remainder are the outlines it contains.")]
         public static List<List<Polyline>> DistributeOutlines(this List<Polyline> outlines, double tolerance = Tolerance.Distance)
         {
+            if (outlines == null || outlines.Where(x => x != null).Count() == 0)
+            {
+                BH.Engine.Reflection.Compute.RecordError("Cannot distribute a null list of outlines.");
+                return new List<List<Polyline>>();
+            }
+
+            if (tolerance == double.NaN)
+            {
+                BH.Engine.Reflection.Compute.RecordError("Cannot distribute outlines as tolerance is null.");
+                return new List<List<Polyline>>();
+            }
+
             foreach (Polyline p in outlines)
             {
                 if (!p.IsClosed(tolerance))
@@ -86,6 +98,18 @@ namespace BH.Engine.Geometry
         [Output("distributedOutlines", "Lists representing each set of outlines. For each list, the first outline is the containing outline and the remainder are the outlines it contains.")]
         public static List<List<ICurve>> DistributeOutlines(this List<ICurve> outlines, double tolerance = Tolerance.Distance)
         {
+            if (outlines == null || outlines.Where(x => x != null).Count() == 0)
+            {
+                BH.Engine.Reflection.Compute.RecordError("Cannot distribute a null list of outlines.");
+                return new List<List<ICurve>>();
+            }
+
+            if (tolerance == double.NaN)
+            {
+                BH.Engine.Reflection.Compute.RecordError("Cannot distribute outlines as tolerance is null.");
+                return new List<List<ICurve>>();
+            }
+
             foreach (ICurve p in outlines)
             {
                 if (!p.IIsClosed(tolerance))
@@ -132,6 +156,18 @@ namespace BH.Engine.Geometry
         [Output("distributedOutlines", "Lists representing each set of outlines. For each list, the first outline is the containing outline and the remainder are the outlines it contains.")]
         private static List<List<Polyline>> DistributeOpenings(this List<Polyline> panels, List<Polyline> openings, double tolerance = Tolerance.Distance)
         {
+            if (panels == null || panels.Where(x => x != null).Count() == 0 || openings == null || openings.Where(x => x != null).Count() == 0)
+            {
+                BH.Engine.Reflection.Compute.RecordError("Cannot distribute a null list of panels/openings.");
+                return new List<List<Polyline>>();
+            }
+
+            if (tolerance == double.NaN)
+            {
+                BH.Engine.Reflection.Compute.RecordError("Cannot distribute outlines as tolerance is null.");
+                return new List<List<Polyline>>();
+            }
+
             List<List<Polyline>> result = new List<List<Polyline>>();
             foreach (Polyline p in panels)
             {
@@ -163,6 +199,18 @@ namespace BH.Engine.Geometry
         [Output("distributedOutlines", "Lists representing each set of outlines. For each list, the first outline is the containing outline and the remainder are the outlines it contains.")]
         private static List<List<PolyCurve>> DistributeOpenings(this List<PolyCurve> panels, List<PolyCurve> openings, double tolerance = Tolerance.Distance)
         {
+            if (panels == null || panels.Where(x => x != null).Count() == 0 || openings == null || openings.Where(x => x != null).Count() == 0)
+            {
+                BH.Engine.Reflection.Compute.RecordError("Cannot distribute a null list of panels/openings.");
+                return new List<List<PolyCurve>>();
+            }
+
+            if (tolerance == double.NaN)
+            {
+                BH.Engine.Reflection.Compute.RecordError("Cannot distribute outlines as tolerance is null.");
+                return new List<List<PolyCurve>>();
+            }
+
             List<List<PolyCurve>> result = new List<List<PolyCurve>>();
             foreach (PolyCurve p in panels)
             {
@@ -194,6 +242,12 @@ namespace BH.Engine.Geometry
         [Output("distributedOutlines", "Lists representing each set of outlines. For each list, the first outline is the containing outline and the remainder are the outlines it contains.")]
         private static List<List<ICurve>> DistributeOpenings(this List<ICurve> panels, List<ICurve> openings, double tolerance = Tolerance.Distance)
         {
+            if (panels == null || panels.Where(x => x != null).Count() == 0 || openings == null || openings.Where(x => x != null).Count() == 0)
+            {
+                BH.Engine.Reflection.Compute.RecordError("Cannot distribute a null list of panels/openings.");
+                return new List<List<ICurve>>();
+            }
+
             List<List<ICurve>> result = new List<List<ICurve>>();
             foreach (ICurve p in panels)
             {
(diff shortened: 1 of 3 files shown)


FraserGreenroyd (changes requested):
You will need to provide `PreviousVersion` attributes for versioning of the methods that have changed.

Geometry_Engine/Create/PlanarSurface.cs
```diff
@@ -126,7 +126,7 @@ public static PlanarSurface PlanarSurface(ICurve externalBoundary, List<ICurve>
                 if (intCurve.ISubParts().Any(x => x is NurbsCurve || x is Ellipse))
                     continue;
 
-                if (externalBoundary.ICurveIntersections(intCurve).Count != 0)
+                if (externalBoundary.ICurveIntersections(intCurve, tolerance).Count != 0)
                 {
                     externalBoundary = externalBoundary.BooleanDifference(new List<ICurve>() { intCurve }).Single();
                     internalBoundaries.RemoveAt(i);
```
pawelbaran, line 132: You are missing tolerance pass in line 142

pawelbaran (commented):
I found one missing pass of `tolerance` within the method - see the code comment.

Regarding the `PreviousVersion` attribute, will it work in case of change of parameter count @FraserGreenroyd? I thought it only works if the name/namespace changes but the parameters stay intact.

Sticking to the attributes `DeprecatedAttributes` on `DistributeOutlines` look like legacy - shouldn't we get rid of them @FraserGreenroyd @rolyhudson?

FraserGreenroyd:
> Regarding the `PreviousVersion` attribute, will it work in case of change of parameter count @FraserGreenroyd? I thought it only works if the name/namespace changes but the parameters stay intact.

`PreviousVersion` is for methods who's parameters have changed as well as name/namespace.

> Sticking to the attributes `DeprecatedAttributes` on `DistributeOutlines` look like legacy - shouldn't we get rid of them @FraserGreenroyd @rolyhudson?

Agreed, they are probably methods to be deleted, so we should remove them and use a `PreviousVersion` attribute to point to the new ones to keep versioning happy.

pawelbaran:
> > Regarding the `PreviousVersion` attribute, will it work in case of change of parameter count @FraserGreenroyd? I thought it only works if the name/namespace changes but the parameters stay intact.
> 
> `PreviousVersion` is for methods who's parameters have changed as well as name/namespace.

Sorry but in the case of increase in the parameter count, how does versioning know what to assign to the newly added parameters? I can imagine this being OK when these are defaulted (as in here), but what if not?

> > Sticking to the attributes `DeprecatedAttributes` on `DistributeOutlines` look like legacy - shouldn't we get rid of them @FraserGreenroyd @rolyhudson?
> 
> Agreed, they are probably methods to be deleted, so we should remove them and use a `PreviousVersion` attribute to point to the new ones to keep versioning happy.

I think the methods are OK, but the attributes were dragged around and could be deleted (worth checking though).

FraserGreenroyd:
> Sorry but in the case of increase in the parameter count, how does versioning know what to assign to the newly added parameters? I can imagine this being OK when these are defaulted (as in here), but what if not?

Nulls - hence why `null-handling` is important.

> I think the methods are OK, but the attributes were dragged around and could be deleted (worth checking though).

They were deprecated for a reason, presumably a long time ago, so we should either remove them, or the attributes. If we remove them, we need to version that correctly.

pawelbaran:
> They were deprecated for a reason, presumably a long time ago, so we should either remove them, or the attributes. If we remove them, we need to version that correctly.

Oke, ja, this will be more tricky than it looks - there is a sister method in Spatial_Engine, but it takes collections of `IElement1D` - they could be merged but from what I vaguely remember they are not, for valid reason. Closer look needed.

enarhi:
@BHoMBot check compliance

enarhi:
@BHoMBot check versioning

enarhi:
@FraserGreenroyd versioning and descriptions have been added per our conversation, and the deprecated attribute removed as the Geometry DistributedOutlines methods are unique from the Spatial ones and were referring to a CommonEngine deprecation that no longer exists.

pawelbaran (approved):
LGTM 👍

pawelbaran:
@BHoMBot check null-handling

pawelbaran:
Ha, as usual, all goes alright until you trigger `null-handling` 😃

enarhi:
@BHoMBot check null-handling

FraserGreenroyd:
@BHoMBot check null-handling

FraserGreenroyd (dismissed):
I'm happy from a DevOps perspective if @pawelbaran is still happy from a functionality perspective.

FraserGreenroyd:
@BHoMBot check required

FraserGreenroyd:
@BHoMBot check required

Geometry_Engine/Query/Area.cs
```diff
@@ -43,6 +43,12 @@ public static partial class Query
         [PreviousVersion("4.3", "BH.Engine.Geometry.Query.IArea(BH.oM.Geometry.IGeometry)")]
         public static double IArea(this IGeometry geometry, double tolerance = Tolerance.Distance)
         {
+            if (geometry == null)
+            {
+                BH.Engine.Reflection.Compute.RecordError("Cannot query the area of null geometry.");
+                return double.NaN;
```
pawelbaran, line 49: Shouldn't we return 0 in such cases, @FraserGreenroyd? I do not remember what was the final decision about it?

FraserGreenroyd: 0 could be a valid number for downstream operations, whereas `double.NaN` prevents downstream operations carrying on as if everything was ok with just the error recorded (particularly if used in code calling the method rather than just UI components). Final decision is it needs to make sense at the time of being done, some instances 0 or -1 are acceptable, others are `double.NaN`. My preference is on `double.NaN` but they obviously then need checking for in code calling them, but that's just good practice to garbage check anything coming back.

Geometry_Engine/Query/Area.cs
```diff
@@ -43,9 +43,9 @@ public static partial class Query
         [PreviousVersion("4.3", "BH.Engine.Geometry.Query.IArea(BH.oM.Geometry.IGeometry)")]
         public static double IArea(this IGeometry geometry, double tolerance = Tolerance.Distance)
         {
-            if (geometry == null)
+            if (geometry == null || tolerance == double.NaN)
```
pawelbaran, line 46: I do not think such construct will work for `double`, `double.IsNaN(tolerance)` is recommended, following [StackOverflow](https://stackoverflow.com/questions/558267/equality-with-double-nan):

> The [IEEE 754 floating point standard](https://en.wikipedia.org/wiki/IEEE_754) states that comparing NaN with NaN will always return false. If you must do this, use `Double.IsNaN()`.

But TBH, I would consider ignoring the check on `tolerance`, otherwise it would be good to rule out zero too etc. I think null checks are enough, good to hear others' opinion too though.

Geometry_Engine/Compute/DistributeOutlines.cs
```diff
@@ -194,6 +242,12 @@ private static List<List<PolyCurve>> DistributeOpenings(this List<PolyCurve> pan
         [Output("distributedOutlines", "Lists representing each set of outlines. For each list, the first outline is the containing outline and the remainder are the outlines it contains.")]
         private static List<List<ICurve>> DistributeOpenings(this List<ICurve> panels, List<ICurve> openings, double tolerance = Tolerance.Distance)
         {
+            if (panels == null || panels.Where(x => x != null).Count() == 0 || openings == null || openings.Where(x => x != null).Count() == 0)
+            {
+                BH.Engine.Reflection.Compute.RecordError("Cannot distribute a null list of panels/openings.");
+                return new List<List<ICurve>>();
```
pawelbaran, line 248: Same as my previous comment, shall we return a new collection or `null` in case of null inputs? Is it formalised somewhere @FraserGreenroyd? This would be really helpful to turn this exercise into a nobrainer.

pawelbaran (changes requested):
There is one serious issue with the `double.NaN` checks, plus I am not 100% if the return values after failed null checks are correct (see the comments).

Geometry_Engine/Compute/DistributeOutlines.cs
FraserGreenroyd: https://github.com/BHoM/documentation/wiki/Null-Handling

enarhi:
@BHoMBot check null-handling

enarhi:
Updated to remove the tolerance checks for now as they are not done elsewhere. Figuring if we want to add true validity checks of the tolerance value it can be done separately as a wider effort.

Geometry_Engine/Compute/DistributeOutlines.cs
```diff
@@ -35,9 +37,18 @@ public static partial class Compute
...
         /***************************************************/
 
-        [DeprecatedAttribute("Moved to Common_Engine")]
+        [Description("Distributes outlines into lists representing containing outlines and the outlines that they each contain.")]
+        [Input("outlines", "The outlines to sort.")]
+        [Input("tolerance", "The tolerance to apply when detecting if outlines are closed and contained.")]
+        [Output("distributedOutlines", "Lists representing each set of outlines. For each list, the first outline is the containing outline and the remainder are the outlines it contains.")]
         public static List<List<Polyline>> DistributeOutlines(this List<Polyline> outlines, double tolerance = Tolerance.Distance)
         {
+            if (outlines == null || outlines.Where(x => x != null).Count() == 0)
```
pawelbaran, line 46: Personally I would use different syntax, but it is rather a matter of preference (negligible difference in performance in favour of my proposal I think, one iteration over the collection avoided).
```suggestion
            if (outlines == null || outlines.All(x => x == null))
```

pawelbaran (approved):
LGTM now, one comment left but feel free to ignore and go for merge if you prefer 👍

enarhi:
@BHoMBot check required

Geometry_Engine/Compute/DistributeOutlines.cs
enarhi: will leave as is as I'm with you, its an improvement but a negligible one.

FraserGreenroyd:
@BHoMBot check ready-to-merge

PR #2617, opened 2021-09-12
Physical_Engine: Added associated Engine methods for 3D reinforcement representation

peterjamesnugent:
### NOTE: Depends on 

https://github.com/BHoM/BHoM/pull/1286
   
Closes https://github.com/BHoM/BHoM/issues/1282

### Test files

https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/02_Pull%20Request/BHoM/BHoM/Physical_oM/%231282-AddReinforcement3DRepresentation.gh?csf=1&web=1&e=VqcrUR

### Changelog

- Added constructor for `Reinforcement` method which verifies the `IShapeCode` is compliant with BS 8666:2020
- Added methods to calculate `SchedulingRadius`, `HookDiameter`, `GeneralEndProjection` and `LinksEndProjection` based on BS 8666:2020 Table 2#
- Added `Centreline` method to produce a centreline for the `IShapeCode` provided

### Additional comments

@pawelbaran and @IsakNaslundBh this is a WIP but I wanted to get something up to review before it goes too far - happy to discuss the approach if needed.


Physical_Engine/Query/SchedulingRadius.cs
```diff
@@ -0,0 +1,59 @@
...
+    public static partial class Query
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Gets the minimum scheduling radius based on the diameter of the reinforcement bar using the values given in BS 8666:2020 Table 2.")]
+        [Input("diameter", "The diameter of the reinforcement bar to determine the minimum scheduling radius.")]
+        [Output("radius", "The minimum scheduling radius based on the diameter of the reinforcement bar", typeof(Length))]
+        public static double SchedulingRadius(this double diameter)
```
pawelbaran, line 41: I think this method should take an input representing the code it complies to. This means we would need an enum containing the codes that we support, plus the whole system would need to be tweaked to pass this value around seamlessly. Not sure how to marry this with my comment on https://github.com/BHoM/BHoM/pull/1286 regarding the shape code namespaces.

What do you think @peterjamesnugent?

Physical_Engine/Create/Reinforcement.cs
```diff
@@ -0,0 +1,84 @@
...
+        /***************************************************/
+
+        [Description("Creates a Reinforcement object storing the ShapeCode, diameter, bend radius and coordinate system.")]
+        [Input("diameter", "The diameter of the reinforcement.")]
+        [Input("coordinateSystem", "The name of the beam, default empty string")]
+        [Input("shapeCode", "The name of the beam, default empty string")]
+        [Input("bendRadius", "The bend radius of the reinforcement. This will be calculated based on the diameter if the provided value " +
+            "is less than the the minimum scheduling radius defined in BS 8666:2020.")]
+        [Output("reinforcement", "The reinforcement object with a compliant shape code in accordance with BS 8666:2020.")]
+        public static Reinforcement Reinforcement(double diameter, Cartesian coordinateSystem, IShapeCode shapeCode, double bendRadius = 0)
```
pawelbaran, line 51: Following the previous comment of mine, here the applicable code could be extracted from the `shapeCode` argument and then passed downstream.

```
ReinforcementCode reinforcementCode = shapeCode.ReinforcementCode();

(...)

// Then the reinforcement code could be passed on to other methods
diameter.SchedulingRadius(reinforcementCode)
```

Physical_Engine/Query/CompliantShapeCode.cs
```diff
@@ -0,0 +1,1090 @@
...
+    public static partial class Query
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Verifies the dimensions of the ShapeCode to verify their compliance with BS 8666:2020.")]
+        [Input("reinforcement", "The reinforcement containing the ShapeCode, reinforcement and bending radius to be verified.")]
+        [Output("bool", "True if the shape code is compliant with BS 8666:2020.")]
+        public static bool CompliantShapeCode(this Reinforcement reinforcement)
```
pawelbaran, line 43: I would call this method to start with _Is_ prefix to follow the general convention: `IsShapeCodeCompliant` or similar.

Physical_Engine/Query/GeneralEndProjection.cs
```diff
@@ -0,0 +1,57 @@
...
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Gets the minimum end projection for general bends (bobs) or links where the bend is greater than or equal to 150 degrees." +
+            "This is based on the diameter of the reinforcement bar using the values given in BS 8666:2020 Table 2.")]
+        [Input("diameter", "The diameter of the reinforcement bar to determine the minimum scheduling radius.")]
+        [Output("endProjection", "The minimum end projection based on the diameter of the reinforcement bar", typeof(Length))]
+        public static double GeneralEndProjection(this double diameter)
```
pawelbaran, line 42: This method could also accept the `ReinforcementCode` enum.

Physical_Engine/Query/HookDiameter.cs
```diff
@@ -0,0 +1,58 @@
...
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Gets the hook diameter based on the diameter of the reinforcement bar using the anticipated values given in BS 8666:2020 Table 2.")]
+        [Input("diameter", "The diameter of the reinforcement bar to determine the hook diameter.")]
+        [Input("bendingRadius", "The bending radius of the bar used to override the SchedulingRadius method that looks up the value in BS 8666:2020 Table 2.")]
+        [Output("diameter", "The anticipated hook dianeter based on the diameter of the reinforcement bar", typeof(Length))]
+        public static double HookDiameter(this double diameter, double bendingRadius = 0)
```
pawelbaran, line 43: This method could also accept the `ReinforcementCode` enum.

Physical_Engine/Query/IsNull.cs
```diff
@@ -0,0 +1,90 @@
...
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Checks if a Reinforcement is null and outputs relevant error message.")]
+        [Input("reinforcement", "The Reinforcement to test for null.")]
+        [Input("methodName", "The name of the method to reference in the error message.")]
+        [Input("msg", "Optional message to be returned in addition to the generated error message.")]
+        [Output("isNull", "True if the Reinforcement or its defining properties are null.")]
+        public static bool IsNull(this Reinforcement reinforcement, string msg = "", [CallerMemberName] string methodName = "Method")
```
pawelbaran, line 44: I like the idea of using `CallerMemberName` attribute here! Really curious how will it prove applicable - if well, I would consider widening this approach to other parts of BHoM.

Physical_Engine/Query/LinksEndProjection.cs
```diff
@@ -0,0 +1,57 @@
...
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Gets the minimum end projection for links where the bend is less than 150 degrees." +
+            "This is based on the diameter of the reinforcement bar using the values given in BS 8666:2020 Table 2.")]
+        [Input("diameter", "The diameter of the reinforcement bar to determine the minimum scheduling radius.")]
+        [Output("endProjection", "The minimum end projection based on the diameter of the reinforcement bar", typeof(Length))]
+        public static double LinksEndProjection(this double diameter)
```
pawelbaran, line 42: This method could also accept the `ReinforcementCode` enum.

pawelbaran (changes requested):
I have not checked the methods in detail, but looked at the general code structure, which makes sense to me.

The one bit I would consider changing is the introduction of an enum (or string) correspondent to the relevant code (BS 8666:2020 etc.). This value could be extracted from the `IShapeCode` and passed around to the methods such as `SchedulingRadius` etc. - see the code comments.

Physical_Engine/Query/IsNull.cs
peterjamesnugent: We've used it for the `IsNull` checks on the Structure_Engine and I think it's being included across the others. From the tests carried out it look promising.

Physical_Engine/Query/SchedulingRadius.cs
peterjamesnugent: As discussed the input will be a `Reinforcement` and the `ShapeCode` used to determine the standard used.

peterjamesnugent:
@BHoMBot check compliance

Physical_Engine/Query/MaximumRadius.cs
```diff
@@ -0,0 +1,122 @@
...
+            int position = shapeCode.GetType().Namespace.LastIndexOf(".") + 1;
+            string standard = fullNamespace.Substring(position, fullNamespace.Length - position);
+
+            switch (standard)
+            {
+                case "BS8666":
+                    List<double> diameters = new List<double>() { 6, 8, 10, 12, 16, 20, 25, 32, 40 }.Select(x => x / 1000).ToList();
+                    List<double> radii = new List<double>() { 2.5, 2.75, 3.5, 4.25, 7.5, 14, 30, 43, 58 };
+
+                    Dictionary<double, double> maximumBendingRadius = diameters.Zip(radii, (k, v) => new { k, v }).ToDictionary(x => x.k, x => x.v);
```
pawelbaran, line 70: This dictionary seems to be constant - I would consider declaring a private readonly field. It will avoid creating the dictionary from scratch per each method call, plus it would be more readable, almost like a table when each pair declared in a separate line.

Physical_Engine/Query/MaximumRadius.cs
```diff
@@ -0,0 +1,122 @@
...
+                    Reflection.Compute.RecordError("Standard not recognised or supported, the scheduling radius could not be calculated.");
+                    return 0;
+            }
+        }
+
+        /***************************************************/
+        /**** Private Methods                            ****/
+        /***************************************************/
+
+        private static double LinearInterpolate(this Dictionary<double, double> reference, double x )
```
pawelbaran, line 83: It tempts me to ask for making it public and putting in a more central engine - not sure where, however, and not sure if the UI would lift it. So OK with keeping it as is.

pawelbaran (changes requested):
_MASSIVE_ amount of code and work in this PR. I had another run, looks good in general, only a few minor comments left. Still need to dig deeper in the details of functionality, but this may take time, so a leaving a code (only code) review for now.

peterjamesnugent:
@BHoMBot check compliance

IsakNaslundBh (changes requested):
Some initial comments.

As discussed on Teams, will let you look through the script with the dimensions and check before I do a full review :)

Physical_Engine/Query/MaximumRadius.cs
peterjamesnugent: I assume it could be used elsewhere, but we can always promote it up later.

Physical_Engine/Create/ShapeCode/BS8666/ShapeCode00.cs
```diff
@@ -0,0 +1,56 @@
...
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Creates a ShapeCode object using the parameters provided. Refer to the object descriptions for alignment.")]
+        [Output("shapeCode", "A ShapeCode to be used with Reinforcement objects.")]
+        public static ShapeCode00 ShapeCode00(double a)
+        {
+            if(a < Tolerance.Distance)
```
IsakNaslundBh, line 43:
```suggestion
            if(a <= 0)
```
Would just check these values against 0.

checking against tolerance makes 0.9e-6 fail and 1.1e-6 pass, which to me feels quite arbitrary

Physical_Engine/Create/ShapeCode/BS8666/ShapeCode14.cs
```diff
@@ -0,0 +1,53 @@
...
+{
+    public static partial class Create
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Creates a ShapeCode object using the parameters provided. Refer to the object descriptions for alignment.")]
+        [Output("shapeCode", "A ShapeCode to be used with Reinforcement objects.")]
+        public static ShapeCode14 ShapeCode14(double a, double b, double c, double d)
```
IsakNaslundBh, line 41: This create method is allowing you to create an invalid object (where a^2 != b^2+d^2).

To me at least, the sole reason of making the shape codes immutable is to avoid this.

Would not keep this create method, but instead have some methods that calculates what needs to be calculated. For example (not sure if these are the exactly right ones):

- take b,c,d -> calculate a
- take a,c and angle -> calculate a and d
- take a,b,c -> calculate d

not sure we need all of them, but just to exemplify.

When thinking of it, I guess you kind of need the diameter and bend radius to do this. Wonder then if it is worth moving those properties to the shape code objects as well.

That would also actually help simplify some bits of your code. For example the `IsValid` and `Centreline` methods that currently requires the shape code object as well as those two parameter to be able to check the object/generate the curve. The more I think of it, I think it is worth promoting up those properties just for that reason.

Physical_Engine/Query/IsShapeCodeCompliant.cs
```diff
@@ -0,0 +1,983 @@
...
+                Reflection.Compute.RecordError("The parameters A and C of ShapeCode13 shall not be less than B/2 + the greater of 5d or 0.09.");
+            }
+
+
+            return true;
+        }
+
+        /***************************************************/
+
+        private static bool IsShapeCodeCompliant(this ShapeCode14 shapeCode, double diameter)
```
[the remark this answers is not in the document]
peterjamesnugent: image

peterjamesnugent:
@IsakNaslundBh - the three checks are:
- Discontinuities: Zero discontinuities between start and end points
- Length check: <1% difference between calculated length (based on BS 8666 which is not precise) compared against the drawn length
- Dimensions: Dimensions match those input in to the object (against `centreline()`)

### Three confirmed correct and ready to merge ((Discontinuities ✔️ , Length ✔️ , dimensions match ✔️ ):
- [ ] Shape Code 00 
- [ ] Shape Code 11
- [ ] Shape Code 12
- [ ] Shape Code 13
- [ ] Shape Code 21
- [ ] Shape Code 22
- [ ] Shape Code 23
- [ ] Shape Code 31
- [ ] Shape Code 32
- [ ] Shape Code 33
- [ ] Shape Code 41
- [ ] Shape Code 44
- [ ] Shape Code 47
- [ ] Shape Code 48 (length is 3.43% but I believe that's because the radius is deliverable large `10d` and the calculation is simple)
- [ ] Shape Code 51 (there is a discontinuity where the bar overlaps in the Z plane)
- [ ] Shape Code 52 (there is a discontinuity where the bar overlaps in the Z plane)
- [ ] Shape Code 63 (there is a discontinuity where the bar overlaps in the Z plane)
- [ ] Shape Code 64 (there is a discontinuity where the bar overlaps in the Z plane)
- [ ] Shape Code 67 (the dimensions in Rhino are slightly off (by a diameter vertically and two diameters horizontally)
- [ ] Shape Code 75
- [ ] Shape Code 77
- [ ] Shape Code 98 (just within 1%)

### Minor difference in dimensions (10s of mm) but proposed to merge (Discontinuities ✔️ , Length ✔️ , dimensions match ❌):
- [ ] Shape Code 14
- [ ] Shape Code 15
- [ ] Shape Code 25
- [ ] Shape Code 27
- [ ] Shape Code 28
- [ ] Shape Code 34 (bottom of the visualisations)
- [ ] Shape Code 35
- [ ] Shape Code 36


### Not implemented (2/3: Discontinuities ❌ , Length ❌  , dimensions match ❌) - @pawelbaran these can be the ones we give attention tomorrow.
- [ ] Shape Code 24 (bottom left arc is not correct and shown as discontinuity) 
- [ ] Shape Code 26 (bottom left arc is not correct and shown as discontinuity) .
- [ ] Shape Code 29 (angled section is 100s mm out)
- [ ] Shape Code 46 (1.16% difference in length and difference in dimensions)
- [ ] Shape Code 56 (angled section is 100s mm out)

Shape Code 35 has a similar shape as 24/26 (an angled segment between two horizontal segments, so could try to reuse the code for 35 on 24/26).

Physical_Engine/Query/IsShapeCodeCompliant.cs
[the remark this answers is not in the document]
peterjamesnugent: Ended up removing this because of the ratios of A,B and D do not form a triangle.

peterjamesnugent:
@BHoMBot check required

peterjamesnugent:
@BHoMBot check required

peterjamesnugent:
@BHoMBot check required

FraserGreenroyd:
@BHoMBot check unit-tests

FraserGreenroyd:

peterjamesnugent:

FraserGreenroyd:
@BHoMBot check installer

FraserGreenroyd:
@BHoMBot this is a DevOps instruction. I am requesting neutral checks on: versioning

PR #2622, opened 2021-09-15
Geometry_Engine: planar surface creation and DistributeOutlines checks/messages refined

### LINKED ISSUE #2621: Geometry_Engine: input checks on Create.PlanarSurface(List<ICurve>) and Compute.DistributeOutlines are not correct
#### Description:

As in the title, the method creating `PlanarSurfaces` based on a collection of `ICurves` contains some input curve checks, but these are not sufficient, and are duplicated in `DistributeOutlines`. `DistributeOutlines`, on the other hand, contains another set of checks, insufficient either. So the plan would be:
- remove all checks from `Create.PlanarSurface`
- make sure all required checks exist in `DistributeOutlines`:
    - null checks
    - unsupported type check (nurbs/ellipse)
    - planarity check
    - closed curve check
    - self-intersection check

pawelbaran:

Closes #2621

### Test files

Problem-specific test file has been uploaded on [SharePoint](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?RootFolder=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F02%5FPull%20Request%2FBHoM%2FBHoM%5FEngine%2FGeometry%5FEngine%2F%232622%2DDistributeOutlinesChecks&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4)
On top of that, worth running it in parallel with https://github.com/BHoM/Revit_Toolkit/pull/1082 and check the test files there (this was the original source of the issue actually).

### Additional comments

This PR is as minimalistic as possible, it only sorts out the checks and messages, but does not change the return values etc. to avoid breaking changes if possible - to merge it safely before the Beta release. Possible future improvements in the next milestone:
- ignore invalid/unsupported curves in `Create.PlanarSurface(List<ICurve>)` instead of crashing the entire method, to align to the behaviour of `Create.PlanarSurface(ICurve, List<ICurve>)`
- add a check whether the input curves to `Create.PlanarSurface(List<ICurve>)` are coplanar, if not then raise warning
- get rid of `DistributeOutlines` for polylines and leave only one version taking `IEnumerable<ICurve>`
- get rid of `DistributeOpenings` overloads that do not work on `ICurves`
- get rid of checks inside `DistributeOpenings` (they are redundant, the method is `private` and the checks are done by the caller)

The above to be picked up after the Beta.

### DIFF
--- a/Geometry_Engine/Compute/DistributeOutlines.cs
+++ b/Geometry_Engine/Compute/DistributeOutlines.cs
@@ -67,12 +67,6 @@ namespace BH.Engine.Geometry
                 return new List<List<Polyline>>();
             }
 
-            if (outlines.Any(p => p.IsSelfIntersecting(tolerance)))
-            {
-                BH.Engine.Reflection.Compute.RecordError("At least one of the input outlines is self-intersecting.");
-                return new List<List<Polyline>>();
-            }
-
             outlines.Sort(delegate (Polyline p1, Polyline p2)
             {
                 return p1.Area().CompareTo(p2.Area());
@@ -137,12 +131,6 @@ namespace BH.Engine.Geometry
             if (outlines.Any(p => !p.IIsPlanar(tolerance)))
             {
                 BH.Engine.Reflection.Compute.RecordError("At least one of the input outlines is not planar within the input tolerance. Please try changing the tolerance if this behaviour is not expected.");
-                return new List<List<ICurve>>();
-            }
-
-            if (outlines.Any(p => p.IIsSelfIntersecting(tolerance)))
-            {
-                BH.Engine.Reflection.Compute.RecordError("At least one of the input outlines is self-intersecting.");
                 return new List<List<ICurve>>();
             }
 
--- a/Geometry_Engine/Compute/DistributeOutlines.cs
+++ b/Geometry_Engine/Compute/DistributeOutlines.cs
@@ -55,20 +55,19 @@ namespace BH.Engine.Geometry
                 return new List<List<Polyline>>();
             }
 
-            List<Point> controlPoints = outlines.SelectMany(x => x.IControlPoints()).ToList();
-            if (!controlPoints.IsCoplanar(tolerance))
-            {
-                BH.Engine.Reflection.Compute.RecordError("The input outline curves are not coplanar, therefore they could not be distributed. Please try changing the tolerance if this behaviour is not expected.");
-                return new List<List<Polyline>>();
-            }
-
             if (outlines.Any(p => !p.IsClosed(tolerance)))
             {
                 BH.Engine.Reflection.Compute.RecordError("At least one of the input outlines is not closed within the input tolerance. Please try changing the tolerance if this behaviour is not expected.");
                 return new List<List<Polyline>>();
             }
 
-            if (outlines.Any(p => p.IsSelfIntersecting(tolerance)))
+            if (outlines.Any(p => !p.IsPlanar(tolerance)))
+            {
+                BH.Engine.Reflection.Compute.RecordError("At least one of the input outlines is not planar within the input tolerance. Please try changing the tolerance if this behaviour is not expected.");
+                return new List<List<Polyline>>();
+            }
+
+            if (outlines.Any(p => !p.IsSelfIntersecting(tolerance)))
             {
                 BH.Engine.Reflection.Compute.RecordError("At least one of the input outlines is self-intersecting.");
                 return new List<List<Polyline>>();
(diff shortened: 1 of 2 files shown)


pawelbaran:
@BHoMBot check compliance

pawelbaran:
@BHoMBot check null-handling
@BHoMBot check core
@BHoMBot check installer
@BHoMBot check serialisation
@BHoMBot check versioning

Geometry_Engine/Compute/DistributeOutlines.cs
```diff
@@ -43,15 +43,33 @@ public static partial class Compute
...
+                return new List<List<Polyline>>();
+            }
+
+            if (outlines.Any(p => !p.IsPlanar(tolerance)))
+            {
+                BH.Engine.Reflection.Compute.RecordError("At least one of the input outlines is not planar within the input tolerance. Please try changing the tolerance if this behaviour is not expected.");
+                return new List<List<Polyline>>();
+            }
+
+            if (outlines.Any(p => !p.IsSelfIntersecting(tolerance)))
```
enarhi, line 70: This causes issues as self intersecting does not use the same tolerance, in that a higher tolerance here means a curve can be self intersecting when it is actually further from self intersecting when the tolerance is higher. This is inherently problematic as a higher tolerance is often needed for the closed checks and the like with very small geometry, where the geometry is much more likely to come closer to itself as well (for example, my test file from /BHoM/Revit_Toolkit/pull/1082 fails for this reason). I suggest we either leave the tolerance on this check at the default value, or don't run this check as part of distribute outlines as before.

Geometry_Engine/Compute/DistributeOutlines.cs
```diff
@@ -92,15 +110,39 @@ public static List<List<Polyline>> DistributeOutlines(this List<Polyline> outlin
...
+                return new List<List<ICurve>>();
+            }
+
+            if (outlines.Any(p => !p.IIsPlanar(tolerance)))
+            {
+                BH.Engine.Reflection.Compute.RecordError("At least one of the input outlines is not planar within the input tolerance. Please try changing the tolerance if this behaviour is not expected.");
+                return new List<List<ICurve>>();
+            }
+
+            if (outlines.Any(p => p.IIsSelfIntersecting(tolerance)))
```
enarhi, line 143: See previous comment

enarhi (changes requested):
Just a couple changes as the self intersecting check is causing issues.

Geometry_Engine/Compute/DistributeOutlines.cs
pawelbaran: Very valid point @enarhi! Actually, this highlights a wider issue of how the tolerance affects the results of different checks. I see two behaviours, depending on the tolerance gradient:
- higher tolerance => easier to pass (`IsClosed`, `IsPlanar`)
- higher tolerance => harder to pass (`IsSelfIntersecting`)

At the first sight, the trends above should be consistent across all preliminary checks, so could possibly be captured in a form of a table, what do you think @al-fisher @IsakNaslundBh? Following that, we could consider prototyping `IsValidInput` method or similar, with a range of Boolean parameters or or enum flags, in order to centralise the checking procedure for geometry inputs.

Back to the origin: I see no safe/smart solution with regard to the problem highlighted by @enarhi, therefore I simply removed the self-intersection checks from this PR - better safe than sorry in the last sprint 😉

pawelbaran:
Following the offline discussion we had with @IsakNaslundBh, the possibly breaking changes were excluded from this PR, leaving it with only 2 modifications:
- meaningful warnings and errors added to `Create.PlanarSurface`
- `null` curves as well as `NurbsCurves` and `Ellipses` are now filtered out instead of crashing `Create.PlanarSurface`

The work on `DistributeOutlines` to be done in the new milestone. I hope you like it @IsakNaslundBh 🙈

Geometry_Engine/Create/PlanarSurface.cs
```diff
@@ -165,17 +165,45 @@ public static PlanarSurface PlanarSurface(ICurve externalBoundary, List<ICurve>
...
+
+            //--------------Closed-Boundary-Curves-----------------------//
+            count = checkedCurves.Count;
+            checkedCurves = checkedCurves.Where(x => x.IIsClosed(tolerance)).ToList();
+            if (checkedCurves.Count != count)
+                Reflection.Compute.RecordWarning("Some of the input boundary curves were not closed within the tolerance and have been ignored on planar surface creation. Please make sure if the output is correct and tweak the input tolerance if needed.");
+
+            if (checkedCurves.Count == 0)
+            {
+                Reflection.Compute.RecordError("Planar surface could not be created because all input boundary curves turned to be invalid (null, unsupported type, non-planar, not closed within the tolerance).");
```
enarhi, line 200: Wording of this error is a bit weird, suggest we change it to:
"Planar surface could not be created because all input boundary curves are invalid (null, unsupported type, non-planar, or not closed within the tolerance)."

enarhi (approved):
LGTM. One comment regarding the error message for all curves invalid, but just down to wording.

enarhi (approved):
LGTM!

pawelbaran:
@BHoMBot check required

IsakNaslundBh (approved):
Happy about changed made in here. Does not change general behaviour, but greatly increases warning/error messages produced by the method. Feels safe to be merged and go into the 4.3 beta release

pawelbaran:
@BHoMBot check ready-to-merge

PR #2648, opened 2021-10-07
Library_Engine: Enable user defined paths for libraries

### LINKED ISSUE #1153: Library_Engine: Expose source path and allow appending source paths
#### Description:

Expose the source path for libraries to users to allow them to switch between sources for datasets/append more sources. Ideally this would work with SharePoint/URL's too, but that's most likely for later/will need SharePoint_Toolkit.

IsakNaslundBh:
### NOTE: Depends on 

https://github.com/BHoM/BHoM/pull/1301
   
Closes #1153

Adding ability for the library_Engine to be able to handle custom user paths to be considered. The custom paths are stored in the added LibrarySettings object.

### Test files

Testfiles to test different aspects of the change in here:

https://burohappold.sharepoint.com/:f:/s/BHoM/Erd1GBSDJGVJlYMO8FYYhDgBxoBJaYu5Xx5xzZ2x2fT57A?e=FTY0qo


IsakNaslundBh:
@BHoMBot check compliance

IsakNaslundBh:
@BHoMBot check compliance

Library_Engine/Query/IsValidUserPath.cs
```diff
@@ -0,0 +1,68 @@
...
+    public static partial class Query
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Checks if a path is valid to add to the library settings. Will return false if the path is empty, or if it is a parent or subfolder of the default source folder.")]
+        [Input("userPath", "The userpath to check.")]
+        [Output("isValid", "Returns true if the user path is valid to be added.")]
+        public static bool IsValidUserPath(this string userPath)
```
JosefTaylor, line 41: This doesn't check if a path exists - should it?

Library_Engine/Query/Library.cs
```diff
@@ -170,24 +170,60 @@ private static string[] LibraryString(string name)
...
+            foreach (string path in UserPaths())
+            {
+                GetPathsAndLoadLibraries(path, "", "");
+            }
+
+        }
+
+        /***************************************************/
+
+        [Description("Loop through all subfolders if provided source folder and extract all json files contained within it or any subfolder.")]
```
JosefTaylor, line 201:
```suggestion
        [Description("Loop through all subfolders of provided source folder and extract all json files contained within.")]
```

Library_Engine/Query/Library.cs
```diff
@@ -170,24 +170,60 @@ private static string[] LibraryString(string name)
...
+            m_libraryStrings = new Dictionary<string, string[]>();
+            m_datasets = new Dictionary<string, Dataset>();
+            m_deserialisationEvents = new Dictionary<string, List<Tuple<oM.Reflection.Debugging.EventType, string>>>();
+            m_dbTree = new Tree<string>();
+            GetPathsAndLoadLibraries();
+        }
+
+        /***************************************************/
+
+        [Description("Loops through all subfolders of default library folder and any additional userpaths and reads all json files contained in those folders.")]
```
JosefTaylor, line 185:
```suggestion
        [Description("Loops through all subfolders of default library folder and any additional userpaths and reads all json files contained within.")]
```

Library_Engine/Query/IsValidUserPath.cs
```diff
@@ -0,0 +1,68 @@
...
+            {
+                Reflection.Compute.RecordWarning("Null or empty userpaths are not valid.");
+                return false;
+            }
+
+            string path = System.IO.Path.GetFullPath(userPath);
+
+            if (m_sourceFolder.ToLower().Contains(path.ToLower()))
+            {
+                Reflection.Compute.RecordWarning($"Not allowed to add a parent folder of the default source folder {m_sourceFolder} to the user libraries.");
```
JosefTaylor, line 53:
```suggestion
                Reflection.Compute.RecordWarning($"Not allowed to add a parent folder of the default dataset folder {m_sourceFolder} to the user libraries.");
```

Library_Engine/Compute/SaveLibrarySettings.cs
```diff
@@ -0,0 +1,83 @@
...
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Updates the LibrarySetting used for extraction of datasets to the one provided and stores it in the settings folder.")]
+        [Input("settings", "The LibrarySettings object to store on the system.")]
+        [Input("replacePreexisting", "If true, any preexisting settings file will be replaced. If false, LibrarySettings will only be updated if no settings already exists.")]
+        [Input("refreshLibraries", "If true, all loaded libraries will be refreshed and reloaded, making use of the provided LibrarySettings object.")]
+        [Output("sucess", "Returns true of the settings was successfully updated.")]
```
JosefTaylor, line 44:
```suggestion
        [Output("sucess", "Returns true if the settings were successfully updated.")]
```

Library_Engine/Compute/SaveLibrarySettings.cs
```diff
@@ -0,0 +1,83 @@
...
+    public static partial class Compute
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Updates the LibrarySetting used for extraction of datasets to the one provided and stores it in the settings folder.")]
+        [Input("settings", "The LibrarySettings object to store on the system.")]
+        [Input("replacePreexisting", "If true, any preexisting settings file will be replaced. If false, LibrarySettings will only be updated if no settings already exists.")]
+        [Input("refreshLibraries", "If true, all loaded libraries will be refreshed and reloaded, making use of the provided LibrarySettings object.")]
```
JosefTaylor, line 43:
```suggestion
        [Input("refreshLibraries", "If true, all loaded libraries will be refreshed and reloaded, making use of the new library settings.")]
```

Library_Engine/Compute/SaveLibrarySettings.cs
```diff
@@ -0,0 +1,83 @@
...
+{
+    public static partial class Compute
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Updates the LibrarySetting used for extraction of datasets to the one provided and stores it in the settings folder.")]
+        [Input("settings", "The LibrarySettings object to store on the system.")]
+        [Input("replacePreexisting", "If true, any preexisting settings file will be replaced. If false, LibrarySettings will only be updated if no settings already exists.")]
```
JosefTaylor, line 42:
```suggestion
        [Input("replacePreexisting", "If true, any preexisting settings file will be replaced. If false, settings will only be updated if none exists.")]
```

Library_Engine/Compute/SaveLibrarySettings.cs
```diff
@@ -0,0 +1,83 @@
...
+namespace BH.Engine.Library
+{
+    public static partial class Compute
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Updates the LibrarySetting used for extraction of datasets to the one provided and stores it in the settings folder.")]
+        [Input("settings", "The LibrarySettings object to store on the system.")]
```
JosefTaylor, line 41:
```suggestion
        [Input("settings", "The new LibrarySettings object.")]
```

Library_Engine/Compute/SaveLibrarySettings.cs
```diff
@@ -0,0 +1,83 @@
...
+
+namespace BH.Engine.Library
+{
+    public static partial class Compute
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Updates the LibrarySetting used for extraction of datasets to the one provided and stores it in the settings folder.")]
```
JosefTaylor, line 40:
```suggestion
        [Description("Updates the settings used for accessing library datasets. These settings are stored in the Settings folder of your BHoM installation")]
```

Library_Engine/Compute/AddUserPath.cs
```diff
@@ -0,0 +1,55 @@
...
+namespace BH.Engine.Library
+{
+    public static partial class Compute
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Adds a custom folderpath to the User libraries accessed by the Library_Engine.")]
+        [Input("customPath", "Path to folder with custom libraries to be extracted from the Library_Engine.", typeof(FolderPathAttribute))]
```
JosefTaylor, line 39:
```suggestion
        [Input("customPath", "Path to folder containing user libraries to be added.", typeof(FolderPathAttribute))]
```

Library_Engine/Compute/AddUserPath.cs
```diff
@@ -0,0 +1,55 @@
...
+        [Input("refreshLibraries", "If true, all loaded libraries will be refreshed and reloaded, making use of the provided LibrarySettings object.")]
```
JosefTaylor, line 40:
```suggestion
        [Input("refreshLibraries", "If true, all loaded libraries will be refreshed and reloaded, making use of the added libraries.")]
```

Library_Engine/Compute/AddUserPath.cs
```diff
@@ -0,0 +1,55 @@
...
+        [Output("sucess", "Returns true of the settings was successfully updated.")]
```
JosefTaylor, line 41:
```suggestion
        [Output("sucess", "Returns true if the settings was successfully updated.")]
```

JosefTaylor (approved):
Looks good as-is, and tests properly. I've added some suggestions on the descriptions and error messages. Also, the checkValidPath method doesn't check if the path exists- should it?

Library_Engine/Query/IsValidUserPath.cs
IsakNaslundBh: First added a check for that, but then realised that there might be a usecase where you want to get all the settings up before you have added a folder. Like, if you want some specific datasets for a project or something, and someone sets up a script to add to settings and then get the data in order afterwards or similar.

Not really a common usecase I guess, but thought it might be viable, hence the lack of checking for this.

If you think my reasoning is overthinking it, I am happy to add this check as well!

IsakNaslundBh:
@BHoMBot check installer
@BHoMBot check serialisation

JosefTaylor (approved):

IsakNaslundBh:
@BHoMBot check required

IsakNaslundBh:

FraserGreenroyd:
@BHoMBot check serialisation
@BHoMBot check null-handling

pawelbaran (commented):
I went through the code, both oM and Engine, and have 2 questions:
- Couldn't `RefreshLibraries` be moved to `Compute` class? This looks very similar to one of the recent Reflection_Engine PR issues.
- Will the settings be saved on shutdown and loaded on startup? There is such a mechanism in BHoM_UI, but I cannot find a trigger for it here.

Functionality wise, I do not have a strong opinion as I barely use any datasets - would lean towards @JosefTaylor's opinion.

IsakNaslundBh:
> I went through the code, both oM and Engine, and have 2 questions:
> 
> * Couldn't `RefreshLibraries` be moved to `Compute` class? This looks very similar to one of the recent Reflection_Engine PR issues.

Agreed, it could be. There are a lot fewer moving parts in this case though compared to reflection engine, why I kind of think it makes sense to keep it where it is.
What I mean by this is that we do not have multiple different corelated collections spanning multiple query methods the way we did in reflection engine, but all can sit nicely in one file. Hence personally think it is fine for this case to keep everything in the query class, as moving it only would complicate things more I think.

> * Will the settings be saved on shutdown and loaded on startup? There is such a mechanism in BHoM_UI, but I cannot find a trigger for it here.
> 

Settings will be loaded on startup (the trigger is via the call to UserPaths when initialising the libraries.)
Settings will not be saved on shutdown, but will be saved directly when you modify them. I am aware of the save functionality in BHoM_UI, but can not access those methods from the Engine so, currently it is being done explicitly in the SaveLibrarySettings method.

> Functionality wise, I do not have a strong opinion as I barely use any datasets - would lean towards @JosefTaylor's opinion.

pawelbaran:
> Agreed, it could be. There are a lot fewer moving parts in this case though compared to reflection engine, why I kind of think it makes sense to keep it where it is.
What I mean by this is that we do not have multiple different corelated collections spanning multiple query methods the way we did in reflection engine, but all can sit nicely in one file. Hence personally think it is fine for this case to keep everything in the query class, as moving it only would complicate things more I think.

I _would prefer_ to use the private fields only within the scope of the file in which they are declared. This, however, would most likely lead to making these fields `internal`, ergo the Reflection_Engine solution. But this is just my preference, so will not request changes because of it - happy to get this merged.

One more note on the general concept of this PR: settings manipulation is a process that is (or could be) applicable to multiple engines and toolkits. Wouldn't it be worth to formalise the work you are doing here a bit more, to make it easy to pick by others on one hand and to integrate it with BHoM_UI on the other?

IsakNaslundBh:
@BHoMBot check ready-to-merge

pawelbaran:
Thanks @IsakNaslundBh, I captured my ideas in https://github.com/BHoM/BHoM_Engine/issues/2709 - please feel free to jump in if something is missing.

PR #2649, opened 2021-10-10
Reflection_Engine: assembly loading improved

### LINKED ISSUE #2620: Revisit the assembly loading and type/method querying mechanism
#### Description:

It would be good to revisit and refine the existing methods used to load the assemblies and query methods and types. There is a few issues identified:
- currently `LoadAllAssemblies` executes only once, even if called a few times with different inputs, so cannot dynamically load assemblies
- all methods are very rigid in terms of naming and namespace conventions, which would be worth loosening a bit, e.g. by adding optional file name suffixes or namespace conventions

### LINKED ISSUE #2638: Reflection_Engine: BHoMTypeDictionary overfilled when there are no adapters available
#### Description:

Edge case bug is happening when loading up all types when you have 0 adapters built.

Issue is that ExtractAllTypes get called multiple types as https://github.com/BHoM/BHoM_Engine/blob/0cd98f0a0b930af1daccbdab817ed884c5370958/Reflection_Engine/Query/TypeList.cs#L75

will return false when no adapter has been extracted (which is valid for the edge case of no adapters built and included).

This together with the fact that the BHoMTypeDictionary is _not_ rest by the `ExtractAllTypes` method

https://github.com/BHoM/BHoM_Engine/blob/0cd98f0a0b930af1daccbdab817ed884c5370958/Reflection_Engine/Query/TypeList.cs#L124-L130

leads to the BHoMTypeDictionary being filled up with more and more of the same type, which in turn causes issues in the CreateType methods depending on the dictionary. This is in turn an issue for the serialisation TK, and makes it fail trying to deserialise BHoM types.

#### Steps to reproduce:

- Clean BHoM folder in Program data
- Build BHoM, BHoM_Engine, BHoM_Adapter, BHoM_UI and Grasshopper_Toolkit. Do _not_ build any adapter
- Try using BHoM in grasshopper, where all kinds of issues occur with things from creating component to deserialising certain objects 

#### Expected behaviour:

For either the ExtractAllTypes to also make sure to clean up the BHoMTypeDictionary and/or for the ExtractAllTypes to not be called multiple times for the edge case of having exactly 0 adapter installed.

pawelbaran:

Closes #2620
Closes #2638

### Test files

#2620 tests on [SharePoint](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?RootFolder=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F02%5FPull%20Request%2FBHoM%2FBHoM%5FEngine%2FReflection%5FEngine%2F%232649%2DAssemblyLoadingImproved&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4):
- _ExistingFunctionality.gh_ - compare `main` with this branch - minor differences, I believe nonharmful
- _AssemblyFilters.gh_ - self-explanatory
- _DynamicAssemblyLoad.gh_ - move some assemblies away from _C:\ProgramData\BHoM\Assemblies_ to a custom folder, then use this folder to load from in the script, see if the assemblies, types and methods become available

#2638 should be tested as explained in https://github.com/BHoM/BHoM_Engine/issues/2638.

### Changelog

- dynamic loading of assemblies enabled, incl. extra suffix filter
- assembly, type and method extraction mechanism refined to allow the above
- as a result of the above, private fields used by the assembly/type/method mechanism are only called from within the files they are declared in

~~The major trade-off introduced in this PR is related to the fact that the extracted assemblies are not stored locally, but they are parsed from the current domain on each query call - luckily this seems to be pretty fast (a matter of 20-25ms on my machine). Massive slowdown starts on extraction of types/methods from the assemblies: therefore, they are stored locally and only the newly loaded assemblies are processed when the types/methods are requested.~~

The current iteration of the idea is similar to the original mechanism (just a few tweaks and cleanups added), with one major new feature: on dynamic load of the assemblies, `Query.RefreshAssemblyList` is called, which subsequently calls `ExtractAllAssemblies`, `ExtractAllTypes` and `ExtractAllMethods` in one go.

I decided to take this approach instead of optional bools suggested by @FraserGreenroyd because the latter would require each query method to call the entire chain of resets to make sure we are working with the latest set of assemblies/types/methods, e.g. to extract the BHoM methods with `reloadAssemblies == true`, one would need to force reload the assemblies and types as it is not known if they were reset before. This sounds like adding extra complexity and redundant relationships, which can be avoided if all resets are centralised in a single method.

### Additional comments

As in the header - the PR still requires some work (mainly to make the compliance happy plus I would like to get rid of the `LoadAllAssemblies` method in favour of  `LoadAssemblies`, which will require versioning and a chain of other PRs - probs better to do in a separate push). But should be ready for the review as the core functionality is there.

Extra notes for the reviewers:
- I have ruled out __Test_ assembly name suffix as it is not used any more - could someone confirm?
- I left support for _External_ namespace untouched, but I am not aware of any use of it - could someone better informed tell me if it could be removed?
- BHoM assembly names are required to end with _oM_, __UI_, __Engine_ or __Adapter_ or to contain _\_oM_\_,  _\_UI_\_,  _\_Engine_\_ or  _\_Adapter_\_, with following exceptions:
    - _BH.UI._ file name prefix enabled to pass Grasshopper UI (that was not the case on `main`)
    - namespace requirements for oM and Engine have been loosened to allow a middle word between `BH` and `oM/Engine`


pawelbaran:
@BHoMBot check compliance
@BHoMBot check null-handling

pawelbaran:
> > If this was an issue, and if we could see loading extra assemblies as an "edge case" that you do very rarely (as in maybe max 1/2 times per session/load time) an alternative approach could be to simply null out the cached type and method lists/dictionaries, and just let the reflection engine populate them again when queried.
> 
> This would be my preferred solution too I think - have a `bool` flag for the method for `reset cache` or something similar, that allows the existing functionality to work as current, but developers (and advanced users) could call the method(/component) and reset the lists as necessary.

Yeah, I was thinking about it as well, but was looking for a solution without any extra bools or resetting of private fields across files/classes. But taken @IsakNaslundBh's point about the methods being called over and over by the in some cases, I am afraid my approach will not be sustainable. I will will test the performance a bit more and think about alternative solutions.

pawelbaran:
@BHoMBot check compliance

pawelbaran:
@BHoMBot check required

pawelbaran:
Replaced with #2681

PR #2652, opened 2021-10-15
Diffing_Engine: Toolkit-specific Diffing to rely on TryRunExtensionMethod; restructured Diffing result and related refactorings

### LINKED ISSUE #2651: Diffing_Engine: improve Toolkit-specific support backend functionality
#### Description:

After https://github.com/BHoM/BHoM_Engine/pull/2647, @pawelbaran and me identified some points that could be improved in the backend system.

In particular, we could leverage more the TryRunExtensionMethods().

Further, if we add mechanisms based on TryRunExtensionMethods(), we should be able to add a more reliable support for Toolkit-specific objects PropertiesToConsider (and Exceptions, etc.) as input separately, instead of putting all together in the `DiffingConfig.PropertiesToConsider`. Although this latter option should remain available to users (i.e. any toConsider "thing" should be insertable in the `DiffingConfig.PropertiesToConsider`, and the Diffing should be intelligent enough to dispatch as appropriate), we should also be able to have more ToolkitSpecific Diffing methods that take inputs separately.

### Proposed changes

- See how much we can ditch the delegate approach for a TryRunExtensionMethods approach instead
- Minimize the needed number of extension methods needed.
- See if we can simplify the RevitDiffing method with this.

### LINKED ISSUE #2684: BHoM_Engine: reference to System.Management.Automation is to be removed
#### Description:

`System.Management.Automation` and related nuget were added for the [WildcardPattern ](https://docs.microsoft.com/en-us/dotnet/api/system.management.automation.wildcardpattern?view=powershellsdk-7.0.0 ) feature. Here is [an example use case](https://stackoverflow.com/a/35655453/3873799).

However, the dll [is not meant to be referenced/used directly](https://stackoverflow.com/questions/66607550/system-management-automation-is-incompatible-with-net-core-3-1/66607812#66607812), and its direct support is already fading (.NET Core apps do not support it).

### Proposed changes
Remove the reference.

Add a Wildcard matching mechanism for all instances where it is needed.

alelom:
### NOTE: Depends on 

https://github.com/BHoM/BHoM/pull/1306/

   
Closes #2651
Closes https://github.com/BHoM/BHoM_Engine/issues/2664
Closes #2684

#### Toolkit-specific `*Diffing()` methods
Toolkit-specific Diffing now rely on TryRunExtensionMethod and a discovery of `*Diffing()` methods that can be defined in specific namespaces. The IDiffing calls those automatically. This replaces the previous temporary solution that used delegates.

#### Toolkit-specific `ComparisonInclusion()` methods
When analysing object properties, Toolkit-specific `Query.ComparisonInclusion(SomeObject object)` methods are discovered and called by the Hash/Diffing functions. These methods evaluate if/how specific properties should be included in the comparison. Implementing a `Query.ComparisonInclusion(SomeObject object)` in your specific toolkit allows for options that are specific to your Toolkit. 

See https://github.com/BHoM/Revit_Toolkit/pull/1109 for [an example of this](https://github.com/BHoM/Revit_Toolkit/pull/1109/files#diff-a6a0b416b094c093ce1500364a67d5d5d5e779c495ff09f6ff2b1807d759af18). More details below.

<details>

A `Query.ComparisonInclusion()` function must have the following signature:

```cs 
ComparisonInclusion ComparisonInclusion(SomeObject object, string propertyFullName, BaseComparisonConfig comparisonConfig)
```

where: 
- `object`: the object whose difference/hash is being evaluated.
-  `propertyFullName`: the full name (path) of the object being evaluated.
- `comparisonConfig`: additional configurations.

It must return a `ComparisonInclusion` object populated with:
- `Include`: true if the property is to be included.
- `DisplayName`: if we want to display the name of the Difference found in this comparison with some custom name, this can be specified here. 

</details>

#### Alignments to the new `Diff` object structure.
The new `Diff` object has a simplified structure for the `ModifiedPropertiesPerObject` property (no more `Dictionary<Dictionary<Tuple<etc>>>`) and therefore adjustments were made for that.

#### Conclusion

This set of PRs is the final wave of big changes done to the Diffing and related objects/functions (e.g. Hash) in this Beta, and I expect that it will more or less settle here.
Once these are approved (i.e. we like the solution), I will document the entire structure of the code, add a proper wiki, diagrams and many tutorial resources.

### Test files

See https://github.com/BHoM/Revit_Toolkit/pull/1109

### Changelog

- Toolkit-specific Diffing now rely on TryRunExtensionMethod and a discovery of `*Diffing` methods that can be defined in specific namespaces.
- Toolkit-specific `Query.ComparisonInclusion(SomeObject object)` methods are discovered and called by the Hash/Diffing functions when analysing object properties. The `ComparisonInclusion` evaluates if/how specific properties should be included in the comparison; it can be implemented in specific Toolkits to allow for options that are specific to that Toolkit.
- Alignments to the new `Diff` object structure.
- General tidy up and quality of life improvements.


pawelbaran (changes requested):
I have reviewed the diffing code line by line (with an exception of `BH.Engine.Base.Query.Hash`). Before I move on to the actual code, a few remarks on the PR itself:
- the PR description requires improvement to capture the changes for reference
- we need proper tests to verify the diffing engine in general - so far we have been using a few variations of the same file with Revit objects, which I believe is far from enough, therefore more test content would be more than welcome before we merge this
- the mechanics of diffing should be captured/documented/diagrammed - it is not a trivial thing and it would be a shame not to have it explained (not necessarily now, more of a general comment)

Now to the code: it is not allowed to leave comments on unchanged bits (which I have also reviewed), so I am listing out all my thoughts/questions below, file by file. Sorry if that is not ideal, but I could not come up with any smarter way of communicating such a variety of remarks 😉

- `ListModifiedProperties` and `ListDifferentProperties` have not been updated to work with the new objects?
- `HashComparer` object sits in wrong namespace
- what is the use of `ValueFromSource`, if in the code you are still using `BH.Engine.Reflection.Query.PropertyValue`? The file itself looks a bit messy btw
- `IDiffing`:
    - lines 60-61 objects per namespace not needed (never referenced)
    - line 76 you could simply return the diff as there is nothing to combine?
    - line 238 `WithCommonPersistentAdapterId` could accept a collection of `IBHoMObjects` instead of objects
- `DiffWithFragmentId`:
    - lines 71-75 are redundant
    - lines 85-86 dangerous to use as `IEnumerable` due to deferred execution!
    - lines 85-86 isn't it better to check using `string.IsNullOrWhiteSpace(s)` than `s.IsNullOrEmpty()`?
    - line 88 `missingIds` variable seems not to be needed, just return `null` instead of assigning a variable and continuing execution?
    - line 124 will crash on null `idFragments`
- `DiffRevisions`:
    - line 94 you can check if `prevObjs_nonBHoM` and `currObjs_nonBHoM` are empty instead of the range of checks you are doing
- `AdaptersDiffingMethods`:
    - line 52 will `x.Name == "Diffing"` catch Revit diffing? shouldn't it be `EndsWith("Diffing")`?
- `ObjectDifferences`:
    - line 118 `propertyFullName_noIndexes.EndsWith(pe)` can be dangerous in case if `pe` is a substring of another property name? wouldn't it be safer to call `EndsWith($".{pe}"`?
- `DiffWithCustomDataKeyId`:
    - line 73 the out parameters cannot be of type `HashSet` because the order of Ids matters!

Finally, I would be really careful with using `IEnumerables` where they are passed on as method parameters - due to the deferred execution mechanism, they will be evaluated separately per each call, which can cause serious performance issues (I noticed a few of such cases in your code @alelom).

alelom:
### Answers to general comments

> the PR description requires improvement to capture the changes for reference
Done

> we need proper tests to verify the diffing engine in general 
Added

> the mechanics of diffing should be captured/documented/diagrammed 
Agreed. As discussed, once we get this current mechanism approved, I will diagram out the solution and the rest of the diffing appropriately for the Wiki, so I effectively "demistify" it. I was simply waiting to see if we liked this solution.

### Answers to code comments

> ListModifiedProperties and ListDifferentProperties have not been updated to work with the new objects?
- `ListDifferentProperties` is still relevant. We still have a `DifferentProperties` method that, unlike Diffing methods, returns the differences without a time-based perspective (change control - e.g. past/following-added/deleted). It is used elsewhere and I think it's useful as it has a complimentary function to diffing methods.
- `ListModifiedProperties` per se is not useful on the new Diff object structure, but it may be used to "explode" possible old serialized Diff objects stored somewhere. I added the deprecated attribute now.


> what is the use of ValueFromSource, if in the code you are still using BH.Engine.Reflection.Query.PropertyValue? The file itself looks a bit messy btw

True, removed, was there when prototyping, not really implemented for now.

alelom:
@BHoMBot check required

alelom:
@BHoMBot check compliance

Diffing_Engine/Properties/launchSettings.json
```diff
@@ -0,0 +1,9 @@
+{
```
FraserGreenroyd, line 1: Can we not commit this file to the repo please

Reflection_Engine/Query/IsAssignableFromIncludeGenerics.cs
```diff
@@ -56,6 +56,26 @@ public static bool IsAssignableFromIncludeGenerics(this Type assignableTo, Type
...
+
+        [Description("Checks if a type is assignable from another type by first checking the system IsAssignableFrom." +
+            "\nIf this is false, checks if the assignable is generic and tests if it can be assigned as a generics version." +
+            "\nIf this is false, checks if the types are the same type, only a byRef version (&).")]
+        [Input("assignableTo", "The type to check if it can be assigned to.")]
+        [Input("assignableFrom", "The type to check if it can be assigned from.")]
+        [Output("result", "Returns true if 'assignableTo' is assignable from 'assignableFrom'.")]
+        public static bool IsAssignableFromIncludeGenericsAndRefTypes(this Type assignableTo, Type assignableFrom)
+        {
+            if (IsAssignableFromIncludeGenerics(assignableTo, assignableFrom))
```
FraserGreenroyd, line 68: This method looks new and probably needs some `null-handling`

Diffing_Engine/Query/ListModifiedProperties.cs
```diff
@@ -40,6 +40,7 @@ public static partial class Query
         [MultiOutput(1, "propNames", "List of properties changed per each object.")]
         [MultiOutput(2, "value_current", "List of current values of the properties.")]
         [MultiOutput(3, "value_past", "List of past values of the properties.")]
+        [Deprecated("5.0", "Deprecated due to the new Diff object structure.")]
```
FraserGreenroyd, line 43: This attribute has been deprecated - `ToBeRemoved` should be used instead

Diffing_Engine/Query/ObjectDifferences.cs
```diff
@@ -0,0 +1,170 @@
...
+        [Input("pastObject", "Past version of the object (created before `followingObject`).")]
+        [Input("followingObject", "Following version of the object (modified or created after `pastObject`).")]
+        [Input("comparisonConfig", "Additional configurations to be used for the comparison.")]
+        [Output("Returns an `ObjectDifferences` object storing all the found differences between `previousObject` and `followingObject`." +
+            "\nIf no difference was found, returns null.")]
+        [PreviousVersion("5.0", "BH.Engine.Diffing.DifferentProperties(System.Object, System.Object, BH.oM.Base.ComparisonConfig)")]
+        public static ObjectDifferences ObjectDifferences(this object pastObject, object followingObject, BaseComparisonConfig comparisonConfig = null)
+        {
+            // Set ComparisonConfig if null. Clone it for immutability in the UI.
+            BaseComparisonConfig cc = comparisonConfig == null ? new ComparisonConfig() : comparisonConfig.DeepClone();
```
FraserGreenroyd, line 50: Might need some specific null handling on `pastObject` and `followingObject` here

FraserGreenroyd (changes requested):

pawelbaran:
> > `DiffWithFragmentId`:
> > lines 71-75 are redundant
> 
> Why? Genuinely cant' see why, I think I'm missing something obvious..

From what I can read in the code, `fragmentIdProperty` will never be `null` nor whitespace - check in line 62 will always overwrite it in such case.

> > lines 85-86 dangerous to use as IEnumerable due to deferred execution!
> 
> Sorry, how? Asking sincerely, I might need some knowledge here

This is a [random link from Google that seems to explain the issue in simple words](https://www.tutorialsteacher.com/linq/linq-deferred-execution). To map it onto your case, variable `IEnumerable<string> pastObjsIds` declared in line 85 **will not get computed at the moment of declaration**. Instead, it will contain an expression that will be evaluated every time you use the variable, in this case, in line 89 (`pastObjsIds.Count()`) and line 104 (`Diffing(pastObjects, pastObjsIds, followingObjs, follObjsIds, diffConfigCopy)` - in this case you are not passing a collection as a parameter, you are passing _expression_ that still needs to be evaluated).

The simplest way to avoid the above issue is to cast the Linq output to list - then, naturally, the expression needs to be evaluated to populate the collection.


> > line 88 missingIds variable seems not to be needed, just return null instead of assigning a variable and continuing > execution?
> 
> As I mentioned before, I prefer to gather all problems and return them all to the user. Returning when 1 condition is not satisfied forces user to do trial-error if multiple problems are present.

Right, I had a blackout - now when I looked at the code it made full sense to me 😃 

> You keep mentioning this but I keep having trouble understanding what cases you refer to/how it is an issue in specific cases. Might need to learn something here. If you put specific comments where you concerns sit I'm happy to go through anything, or even if we want to have a specific call.

Please see my reply a few lines above ⬆️

Diffing_Engine/Properties/launchSettings.json
alelom: committed by mistake, thanks for noticing. Removed.

Reflection_Engine/Query/IsAssignableFromIncludeGenerics.cs
alelom: Thanks, done 6a77ecb488d2af5ca5561d062af5c28b09c94293

Diffing_Engine/Query/DifferentProperties.cs
```diff
@@ -40,131 +40,36 @@ public static partial class Query
...
-        [Output("Dictionary whose key is the name of the property, and value is a tuple with the different values found in obj1 and obj2.\n" +
-            "This dictionary can be 'exploded' in the UI by using `ListDifferentProperties` method.")]
-        public static Dictionary<string, Tuple<object, object>> DifferentProperties(this object obj1, object obj2, DiffingConfig diffingConfig = null)
+        [Input("comparisonConfig", "Additional configurations to be used for the comparison.")]
+        [Output("Dictionary whose Key is the Full Name of the property that is different among the two objects, " +
+            "and whose Value is a Tuple: Item1 is the value of the property in obj1, Item2 the value in obj2." +
+            "\nThis dictionary can be 'exploded' in the UI by using the `ListDifferentProperties()` method." +
+            "\nIf no difference was found, returns null.")]
+        [PreviousVersion("5.0", "BH.Engine.Diffing.DifferentProperties(System.Object, System.Object, BH.oM.Diffing.DiffingConfig)")]
```
FraserGreenroyd, line 48:
```suggestion
        [PreviousVersion("5.0", "BH.Engine.Diffing.Query.DifferentProperties(System.Object, System.Object, BH.oM.Diffing.DiffingConfig)")]
```

Think this will fix the versioning issue @alelom

Diffing_Engine/Query/ListModifiedProperties.cs
alelom: Thanks, done 6a77ecb488d2af5ca5561d062af5c28b09c94293

FraserGreenroyd (changes requested):

Diffing_Engine/Query/ObjectDifferences.cs
alelom: Added one for both objects `null`. db23c5ff2972acec88405153ed6d154148938a20

If only one is null, we still need to defer to Kellerman library method below in order to get the appropriate result.

alelom:
Thanks @pawelbaran ,

> > > `DiffWithFragmentId`:
> > > lines 71-75 are redundant
> >
> > Why? Genuinely cant' see why, I think I'm missing something obvious..
>
> From what I can read in the code, fragmentIdProperty will never be null nor whitespace - check in line 62 will always overwrite it in such case.

Absolutely right, I was missing something obvious 😄! Fixed in c6531c2a652f54c21b0de4eb02103902e84dd12b

>  variable IEnumerable<string> pastObjsIds declared in line 85 will not get computed at the moment of declaration. Instead, it will contain an expression that will be evaluated every time you use the variable, in this case, in line 89 (pastObjsIds.Count()) and line 104 (Diffing(pastObjects, pastObjsIds, followingObjs, follObjsIds, diffConfigCopy) - in this case you are not passing a collection as a parameter, you are passing expression that still needs to be evaluated).
> 
> The simplest way to avoid the above issue is to cast the Linq output to list - then, naturally, the expression needs to be evaluated to populate the collection.

Thanks! Done in 05d3e4b60cfd3c7ea4fe3ea8bd346f66b913ddce

pawelbaran (dismissed):
Happy to approve based on the code and scarce tests provided - looking forward to running more tests tomorrow (any issues can be resolved in separate PRs then)

alelom:
@BHoMBot check required

alelom:
@BHoMBot check required

alelom:
@BHoMBot check required

FraserGreenroyd:
@BHoMBot check null-handling

pawelbaran:
@BHoMBot check core

pawelbaran (dismissed):
This does not compile @alelom 🙈

alelom:
@BHoMBot check required

alelom:
@BHoMBot check required

alelom:
> This does not compile @alelom 🙈

@pawelbaran sorry, you must have arrived in a moment when I was doing changes to comply with some check requirements or testing.

alelom:
@BHoMBot check null-handling

alelom:
@BHoMBot check required

FraserGreenroyd:
@BHoMBot check serialisation

alelom:
@BHoMBot check required

FraserGreenroyd:

FraserGreenroyd:
@BHoMBot this is a CI/CD instruction. I am authorising dispensation to be granted on check ref. 4188450018

FraserGreenroyd (dismissed):
I am approving for the sole purpose of checking mergeability and will be dismissing this review regardless of the outcome without merge. This review is not a formal acceptance or agreement with the changes proposed by this PR and still requires formal review.

FraserGreenroyd:
@BHoMBot check ready-to-merge

FraserGreenroyd:
@BHoMBot this is a CI/CD instruction. I am authorising dispensation to be granted on check ref. 4191879619

FraserGreenroyd:
image
As we suspected @alelom.

alelom:
Closing to reopen in a new, clean branch after noticing PR Merge issues.

alelom:
Replaced with https://github.com/BHoM/BHoM_Engine/pull/2698

PR #2672, opened 2021-10-28
Environment_Engine Updated Method for isExternal.cs

### LINKED ISSUE #2525: Environment_Engine: Update to Panels IsExternal method
#### Description:

Need to add additional approaches and rigor to the _[IsExternal](https://github.com/BHoM/BHoM_Engine/blob/master/Environment_Engine/Query/IsExternal.cs)_ method for Environment Panels to support geometry build workflows and automated model data population.

Options for identifying whether a Panel is 'external':

1. Based on the Panel Type enum - Quick and easy to implement but this can be subjective or simply incorrectly assigned

2. Based on overlaps with other panels; panels with overlaps must be internal, no overlaps = external panel. - Quite thorough but geometrically heavy to compute. Will also fail if _[TidyPanels](https://github.com/BHoM/BHoM_Engine/blob/master/Environment_Engine/Compute/TidyPanels.cs)_ has been applied as there will be no overlaps

3. Based on number of ConnectedSpaces; if ConnectedSpaces <= 1 then external panel, else internal panel. - Relatively quick and easy to implement and workflow can reliably depend on Panels [_ToSpace_](https://github.com/BHoM/BHoM_Engine/blob/master/Environment_Engine/Query/ToSpace.cs) method.

4. Based on geometrical proximity or intersection check with an additional guide geometry or object e.g. floorplate/space perimeter line intersection with the Panel edges. @linaeriksson - I believe this is what you are already starting to develop?

Potentially we could use all of these options as different methods or some in conjunction with others etc. My preference would be to have four distinct methods accessible to allow users freedom to work which the best suited to the task at hand but I am open to suggestions.

emilsvedjer:

Closes #2525 

Added new method to determine whether a panel is external or internal. The new method sorts the panels using the isContaining method instead of the panel type.

### Additional comments

The component is working by own testing

### DIFF
--- a/Environment_Engine/Query/IsExternal.cs
+++ b/Environment_Engine/Query/IsExternal.cs
@@ -58,18 +58,27 @@ namespace BH.Engine.Environment
                  Point panelCentre = p.Polyline().Centroid();
                  Point normalPt = panelCentre + (0.01 * panelNormal);
                  Point negNormalPt = panelCentre - (0.01 * panelNormal);
-                 List<bool> boolContains = new List<bool>();
-
+                int insideCounter = 0;
                 foreach (List<Panel> space in definedSpaces)
-                 {
-                     boolContains.Add(space.IsContaining(normalPt));
-                     boolContains.Add(space.IsContaining(negNormalPt));
-                 }
-                 if (boolContains.Where(x => x).Count() > 1)
-                     internalPanels.Add(p);
-                 else 
-                     externalPanels.Add(p);
-             }
+                {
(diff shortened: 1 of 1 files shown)


FraserGreenroyd:
@BHoMBot check required

FraserGreenroyd:
@BHoMBot check ready-to-merge

FraserGreenroyd (approved):

FraserGreenroyd:
@BHoMBot check ready-to-merge

PR #2679, opened 2021-11-04
Serialiser_Engine: Serialisation of properties of type MethodInfo enabled

### LINKED ISSUE #2678: Serialiser_Engine: Serialisation of properties of type MethodInfo does not work
#### Description:

As in the title, if the object has a property of type `MethodInfo`, on serialisation its value is turned into `System,Runtime.MethodInfo`, missing such details as the method name and declaring type. The snip below shows the difference between `MethodInfo` serialised as a standalone object vs wrapped into a property:

image

pawelbaran:

Closes #2678

### Test files

On [SharePoint](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?RootFolder=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F02%5FPull%20Request%2FBHoM%2FBHoM%5FEngine%2FSerialiser%5FEngine%2F%232679%2DMethodInfoPropertyBug&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4).

### Additional comments

As discovered together with @IsakNaslundBh, registering `MethodBaseSerializer` for `MethodInfo` seems to fix the issue. However, when we wanted to push it forward and do the same for other types such as `Type.GetType("System.RuntimeMethodInfo")` or `Type.GetType("System.RuntimeConstructorInfo")`, we broke the entire serialisation system. Therefore, the intervention I am proposing in this PR is as small as possible, with any further changes to be done in a separate, dedicated thread (if ever needed).

### DIFF
--- a/Serialiser_Engine/Convert/Bson.cs
+++ b/Serialiser_Engine/Convert/Bson.cs
@@ -177,6 +177,7 @@ namespace BH.Engine.Serialiser
                 BsonSerializer.RegisterSerializer(typeof(object), new BH_ObjectSerializer());
                 BsonSerializer.RegisterSerializer(typeof(System.Drawing.Color), new ColourSerializer());
                 BsonSerializer.RegisterSerializer(typeof(MethodBase), new MethodBaseSerializer());
+                BsonSerializer.RegisterSerializer(typeof(MethodInfo), new MethodBaseSerializer());
                 BsonSerializer.RegisterSerializer(typeof(Guid), new GuidSerializer(BsonType.String));
                 BsonSerializer.RegisterSerializer(typeof(CustomObject), new CustomObjectSerializer());
                 BsonSerializer.RegisterSerializer(typeof(FragmentSet), new BHoMCollectionSerializer<FragmentSet, IFragment>());
--- a/Serialiser_Engine/Convert/Bson.cs
+++ b/Serialiser_Engine/Convert/Bson.cs
@@ -176,8 +176,6 @@ namespace BH.Engine.Serialiser
             {
                 BsonSerializer.RegisterSerializer(typeof(object), new BH_ObjectSerializer());
                 BsonSerializer.RegisterSerializer(typeof(System.Drawing.Color), new ColourSerializer());
-                BsonSerializer.RegisterSerializer(typeof(MethodBase), new MethodBaseSerializer());
-                BsonSerializer.RegisterSerializer(typeof(MethodInfo), new MethodBaseSerializer());
                 BsonSerializer.RegisterSerializer(typeof(Guid), new GuidSerializer(BsonType.String));
                 BsonSerializer.RegisterSerializer(typeof(CustomObject), new CustomObjectSerializer());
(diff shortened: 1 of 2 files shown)


pawelbaran:
@BHoMBot check compliance
@BHoMBot check null-handling

pawelbaran:
@BHoMBot check serialisation

pawelbaran:
@BHoMBot check serialisation

pawelbaran:
@BHoMBot check serialisation

pawelbaran:
@BHoMBot check compliance
@BHoMBot check null-handling
@BHoMBot check core
@BHoMBot check installer
@BHoMBot check versioning

IsakNaslundBh (changes requested):
Small tweak on serialiser registration.

Then before approving, think it is good to get @adecler s view on this before merging.

Serialiser_Engine/Convert/Bson.cs
```diff
@@ -177,6 +177,7 @@ private static void RegisterSerializers()
                 BsonSerializer.RegisterSerializer(typeof(object), new BH_ObjectSerializer());
                 BsonSerializer.RegisterSerializer(typeof(System.Drawing.Color), new ColourSerializer());
                 BsonSerializer.RegisterSerializer(typeof(MethodBase), new MethodBaseSerializer());
+                BsonSerializer.RegisterSerializer(typeof(MethodInfo), new MethodBaseSerializer());
```
IsakNaslundBh, line 180: My comment here did not make it to the review for whatever reason.

Had commented here that I think we should initialising this the same way as the TypeSerialiser is being registred. That is, create the `MethodBaseSerializer` once, and then assign it to both MethodBase and MethodInfo, rather than creating two separate isntances of the `MethodBaseSerializer` class.

pawelbaran: Done 👍

pawelbaran:
@BHoMBot check required

adecler (approved):
I am ok with this change. 

Two comments though:
- Technically, it would be better to add the support for `ConstructorInfo` as well while you're at it since MethodBase is covering both `MethodInfo` and `ConstructorInfo`. Support for constructors i very fringe so I am ok for it to be added when it becomes needed if you don't have the time now to do the proper testing.
- `RunTimeX` should not be added as those types are only defined for current execution (as name implies). I am actually surprised it was added for types at all. @IsakNaslundBh , do you remember why you added it ?

adecler:
>But, probably better to not touch it in this PR and raise separately if we want to remove it, as this quick test is not really a full one.

Agreed.

FraserGreenroyd (approved):

FraserGreenroyd:
@BHoMBot check ready-to-merge

FraserGreenroyd:
@BHoMBot this is a CI/CD instruction. I am authorising dispensation to be granted on check ref. 4116115523

FraserGreenroyd:
@BHoMBot this is a CI/CD instruction. I am authorising dispensation to be granted on check ref. 4116115316

PR #2681, opened 2021-11-05
Reflection_Engine: assembly loading improved (2nd try)

### LINKED ISSUE #2620: Revisit the assembly loading and type/method querying mechanism

### LINKED ISSUE #2638: Reflection_Engine: BHoMTypeDictionary overfilled when there are no adapters available

pawelbaran:

Closes #2620
Closes #2638

### Test files

#2620 tests on [SharePoint](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?RootFolder=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F02%5FPull%20Request%2FBHoM%2FBHoM%5FEngine%2FReflection%5FEngine%2F%232649%2DAssemblyLoadingImproved&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4):
- _ExistingFunctionality.gh_ - compare `main` with this branch - I think the changes are only on plus
- _AssemblyFilters.gh_ - self-explanatory
- _DynamicAssemblyLoad.gh_ - move some assemblies away from _C:\ProgramData\BHoM\Assemblies_ to a custom folder, then use this folder to load from in the script, see if the assemblies, types and methods become available

#2638 should be tested as explained in https://github.com/BHoM/BHoM_Engine/issues/2638.

### Changelog

- internal `BH.Engine.Reflection.Global` class introduced to store the reflected assemblies, types and methods **as well as to monitor the current domain for any newly loaded assemblies**
- `BH.Engine.Reflection.Compute.ReflectAssembly` method added to extract all relevant information from an assembly in one place
- `BH.Engine.Reflection.Compute.LoadAssembly` method added to allow for loading single assemblies
- `BH.Engine.Reflection.Compute.LoadAllAssemblies` refactored to allow for multiple calls in lifetime of the `Compute` class
- other methods refactored to leverage the above concept

### Additional comments

This PR differs from what we have discussed last week @adecler @alelom @IsakNaslundBh @FraserGreenroyd. I have split the explanation into 3 subchapters to map the evolution process - for TL;DR; I recommend the last chapter 😉 

#### Current solution
The key problem now is the fact that LoadAllAssemblies can be loaded only once, then `m_AssemblyAlreadyLoaded` becomes `true` and that's it.

However, this is not the only issue. `LoadAllAssemblies` is called on UI startup, then each of the collection query methods (assembly/type/method list) performs extraction of relevant items from loaded assemblies on the 1st call, for example:

https://github.com/BHoM/BHoM_Engine/blob/1a1cd5a03dc721df3c8d79a59ee1c3b6ca7138f5/Reflection_Engine/Query/MethodList.cs#L40-L45

Since the type and method lists are queried on UI startup, this means that they are scraping only the assemblies that have been loaded before BHoM UI starts up - anything that gets loaded after BHoM (incl. other libraries loaded by GH) is not reflected _ever_. On the other hand, the assemblies are not collected on startup, therefore they are scraped when the user calls the method, so when GH is fully armed, leading to a different source set than in case of types methods. This is illustrated by the diagram:

current

#### Discussed last week
The solution we came up with was based on two new concepts:
- `Global` class that would store all relevant assemblies/types/methods in `internal` properties
- System that would eagerly reflect the assemblies loaded on 2 occasions:
    - startup
    - assemblies being dynamically loaded on user's request

This solution has a flaw that we did not take into account. Namely, it would not reflect any assemblies that have been loaded silently (not as a result of BHoM method call) after BHoM startup. These are for example:
- assemblies loaded by GH after BHoM starts up
- dependencies loaded as a result of other assemblies being loaded

The diagram below shows the problem:
next

#### Final solution
`AppDomain.CurrentDomain` has an event `AssemblyLoad` that triggers every time a new assembly is loaded to the domain. Thanks to that, it is possible to reflect and parse _any_ assembly that is loaded from _anywhere_, _anytime_. I leveraged this and extended the `Global` class with a static constructor that does 2 things:
- subscribes to `AppDomain.CurrentDomain.AssemblyLoad` to reflect any newly loaded assembly
- reflects all assemblies loaded up to that moment

The above means that the assemblies start being parsed using BHoM reflection system as soon as the latter is loaded, plus the collections stored by `Global` are _always_ up to date. I think this is pretty helpful.

final

Naturally, this solution is a fundamental change compared to what we were using so far, but I think it is worth it. Happy to hear the opinion of others 👍


pawelbaran:
@BHoMBot check required

pawelbaran:
@BHoMBot check versioning

pawelbaran:
@FraserGreenroyd failure of the versioning check is interesting - I have just tried to deserialise the failing methods from the .ci datasets and it works just fine! Do you have any idea what could go wrong here?

FraserGreenroyd:
I'd have said it was caused by the branches being behind but looking at the network graph it looks ok. One to investigate next week I think.

adecler:
I still have to do more checks but here's a few initial comments:

**Global**
- I really like the processing of the assembly being triggered by ` AppDomain.CurrentDomain.AssemblyLoad` 👍 
- Quite a few of the properties could be merged

**UsedAssemblies**
- The assembly dictionary is still replaced with a list ? 

**AssemblyList**
- Line 50: why is this line there ?

**LoadAllAssemblies**
- `LoadAssembly` return the assembly regardless of it already being loaded or not while `LoadAllAssemblies` only return those that are newly loaded. Not a big deal but might be worth thinking if it is worth being consistent.
- Not sure I like that assembly is added to `m_AlreadyLoaded` even if not really loaded (e.g. folder doesn't exist). I understand it's done for efficiency reasons but I wonder if it might just be a bug waiting to happen if we later want to leverage the list of loaded assemblies. Not a big deal though.

IsakNaslundBh:
I like the solution as well.

Some initial comments:

### LoadAllAssemblies

- m_ALreadyLoaded can get into race conditions on line 58 if two threads simultaneously tries to add the same key to the hashset (if both slip by the check on line 55 before any of the threads have reached line 58)
- Sort of related to this, I wonder if it is worth locking this whole method, as a case could occur where two separate threads both call LoadAllAssemblies with default inputs and then expects all standard things to be loaded. If the second thread hits line 55, and it then proclaims "Yes, this is already loaded" then the thread might continue for example trying to de-serialise some objects/loading up menus/something else believing all assemblies to be loaded, when in fact, they are in the process of being loaded by another thread. This could then ofc lead to failure if the Type/Methodlists are assumed to be correctly filled when in fact they are not. Think the safest would be to put a lock around the method from Line 53 down to the end of the method to ensure threads calling this method wait until it can be made sure that it is done before proceeding.
- Need to ensure that all entry points to the BHoM (for example Analytics) makes an explicit call to the LoadAllAssemblies() as it was previously done implicitly via ExtractAllMethods.


Could also be worth checking if Concurrent collections are worth to be used for the Type/method/assembly storages, but that might not be required as I think the standard collections are thread safe for reading, and the only method currently modifying them is the extraction method, which is already put under a pad lock.

But, like the solution :)

pawelbaran:
Thanks @adecler for the review, please see my replies below:

> **Global**
> 
> * I really like the processing of the assembly being triggered by ` AppDomain.CurrentDomain.AssemblyLoad` 👍
> * Quite a few of the properties could be merged

Yeah I also noticed that, but decided not to put the extra risk on this PR (particularly that I am not the originator of the code and may miss some caveats relevant to each of the collections).

> **UsedAssemblies**
> 
> * The assembly dictionary is still replaced with a list ?

Yep, my mistake - fixed.

> **AssemblyList**
> 
> * Line 50: why is this line there ?

Legacy from debugging - removed.

> **LoadAllAssemblies**
> 
> * `LoadAssembly` return the assembly regardless of it already being loaded or not while `LoadAllAssemblies` only return those that are newly loaded. Not a big deal but might be worth thinking if it is worth being consistent.

I have refactored `m_AlreadyLoaded` into a dictionary that stores assemblies loaded (at any point) that are relevant to the given combination of folder and suffix.

> * Not sure I like that assembly is added to `m_AlreadyLoaded` even if not really loaded (e.g. folder doesn't exist). I understand it's done for efficiency reasons but I wonder if it might just be a bug waiting to happen if we later want to leverage the list of loaded assemblies. Not a big deal though.

Yeah I was concerned about it too. I made two improvements to address this:
- I moved the folder existence check to the very beginning
- I added `forceParseFolder` method parameter to force re-parsing the folder when required

I hope the above answers your concerns @adecler 👍

pawelbaran:
Thanks @IsakNaslundBh, thank you for the deep dive and spotting a few potential issues!

> * m_ALreadyLoaded can get into race conditions on line 58 if two threads simultaneously tries to add the same key to the hashset (if both slip by the check on line 55 before any of the threads have reached line 58)
> * Sort of related to this, I wonder if it is worth locking this whole method, as a case could occur where two separate threads both call LoadAllAssemblies with default inputs and then expects all standard things to be loaded. If the second thread hits line 55, and it then proclaims "Yes, this is already loaded" then the thread might continue for example trying to de-serialise some objects/loading up menus/something else believing all assemblies to be loaded, when in fact, they are in the process of being loaded by another thread. This could then ofc lead to failure if the Type/Methodlists are assumed to be correctly filled when in fact they are not. Think the safest would be to put a lock around the method from Line 53 down to the end of the method to ensure threads calling this method wait until it can be made sure that it is done before proceeding.

Both points make sense to me, can imagine both issues occurring with my original implementation. To embrace them, I put a dedicated lock on `LoadAllAssemblies` - as discussed online, I was trying to have as many locks as possible, but this one is indeed needed.

> * Need to ensure that all entry points to the BHoM (for example Analytics) makes an explicit call to the LoadAllAssemblies() as it was previously done implicitly via ExtractAllMethods.

Yeah - I will investigate it today, then anything I miss should be caught during the UI tests 👍 

> Could also be worth checking if Concurrent collections are worth to be used for the Type/method/assembly storages, but that might not be required as I think the standard collections are thread safe for reading, and the only method currently modifying them is the extraction method, which is already put under a pad lock.

Yeah, should be safe now, plus changing into concurrent collections should be easy - happy to change it if requested, otherwise I will gamble on the current solution 😃 
 
> But, like the solution :)

🚀 🚀 🚀

pawelbaran:
@BHoMBot check required

adecler (dismissed):
Looks ok to me now. I would prefer if Isak could also do a final review before merging though.

There is one thing that I would like to see changed though. But this can be done in a separate PR: The handling of suffix needs to be improved. Right now, you are purely relying on the date suffix to load all revit assemblies while ignoring the `Revit` prefix. This means that, if another set of assemblies  are created for another software following the same naming conventions you used, both set of dlls will be loaded as soon as either of them is trying to load its assemblies. I can think of two options for this:
- either change the naming conventions of the assemblies so the `Revit` part is at the end (e.g. `_Engine_Revit2018` instead of `Engine_2018`),
- or use regex style pattern instead of suffix as input of `LoadAllAssemblies`.

pawelbaran:
@BHoMBot check versioning

pawelbaran:
@BHoMBot check required

pawelbaran:
@adecler I have introduced a few tweaks to the code:
- [40801f0f](https://github.com/BHoM/BHoM_Engine/pull/2681/commits/40801f0fd97518fa2e32b8c6d34b3559b9522320) - this is a peculiar case, when running Versioning PostBuild, it is loading assemblies from its dedicated folder, not _BHoM\Assemblies_, which then resulted in the same assembly (with the same full name!) being loaded twice, as shown in the snip from debug of PostBuild:
    image

    The above caused type mismatches that broke Versioning. We spent almost half a day getting to the bottom of this problem with @IsakNaslundBh, but finally cracked it 😃 

- [016c5823](https://github.com/BHoM/BHoM_Engine/pull/2681/commits/016c5823339bf3d136909b9125f774e6874c87e8) - the `Global` collection properties we made concurrent because together with @IsakNaslundBh we came up with a scenario that could cause to race conditions if they were not:
    - thread A is loading/reflecting the assemblies
    - thread B is attempting to access one of the collections, e.g. by calling `BHoMMethodList` etc.

- [50a51ce0](https://github.com/BHoM/BHoM_Engine/pull/2681/commits/50a51ce04862dc3fa55c0a97ce256309db514b94) and [98a4ccb0](https://github.com/BHoM/BHoM_Engine/pull/2681/commits/98a4ccb051f06670ef15b1e47c0088fe5c963de2) - performance tweaks that should not affect the overall mechanics (the types and methods are only captured from BH assemblies anyways due to the namespace requirements, so I simply ruled out parsing other assemblies)

pawelbaran:
@BHoMBot check required

adecler (dismissed):
The new changes make sense. Great catch, @IsakNaslundBh 👍

alelom:
I am leaving a comment here just to state that I really like this solution:

- I like how this solves the dynamic assembly loading using the event. 
- I also really like that this solution allows to get also assemblies that are not explicitly loaded by BHoM, which is a decisive advantage over last week's proposed approach and a fundamental feature to cover.

pawelbaran:
@BHoMBot check ready-to-merge

pawelbaran:
@BHoMBot check required

IsakNaslundBh (approved):
Now happy! Working as expected.

Some potential improvements for better filtering of allowable namespaces, as discussed offline, which I have raised an issue for https://github.com/BHoM/BHoM_Engine/issues/2686.

Still happy with current solution that I think improves the system as a whole and make it significantly cleaner, so good work!

IsakNaslundBh:
@BHoMBot check ready-to-merge

PR #2683, opened 2021-11-10
Physical_Engine: Methods to query dimensions from Openings added

### LINKED ISSUE #2675: Physical_Engine: Add method to query IOpening's dimensions
#### Description:
Add a simple and generic method to query dimensions from physical IOpenings.

tiagogrossi:

Closes #2675 

Following the example from `Environment`, a couple of dimension query methods have been added for openings, with specific implementation of Doors. Other opening objects will have to be implemented when required.

### Test files
Create a BHoM Door object and query its dimensions, like in this sample [file](https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/02_Pull%20Request/BHoM/BHoM_Engine/Physical_Engine/Physical_Engine-%232675-AddQueryOpeningDimensions.gh?csf=1&web=1&e=a75Jpt) 

### Changelog
Added `BH.Engine.Geometry.Query.Width`
Added `BH.Engine.Geometry.Query.Height`
Added `BH.Engine.Physical.Query.Height`
Added `BH.Engine.Physical.Query.Width`
Added `BH.Engine.Physical.Query.Bottom`

### Additional comments
@FraserGreenroyd these methods are a repurposing of existing public and private methods in `Environment`. Since other toolkit might take advantage of it I copied some of it to `Geometry_Engine`, and if this is deemed appropriate I reckon `Environment` could adjust its relevant code to use it instead.


Physical_Engine/Query/Height.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        /***************************************************/
+        /****             Fallback Methods              ****/
+        /***************************************************/
+        
+        private static double Height(this object opening)
+        {
+            BH.Engine.Reflection.Compute.RecordError(String.Format("Height query for {0} type is not implemented.",opening.GetType()));
+            return 0;
+        }
```
IsakNaslundBh, line 70: Think this gets cleaner if it is dealt with straight in the physical_Engine, without a roundtrip by extracting the geometry. Think something like the suggested below should solve it.

Have also added the Length quantity attribute that should be put on any output returning a length of something :).

```suggestion
using BH.oM.Geometry;
using BH.Engine.Geometry;
using BH.Engine.Spatial;
using BH.oM.Quantities.Attributes;

namespace BH.Engine.Physical
{
    public static partial class Query
    {
        /***************************************************/
        /****               Public Methods              ****/
        /***************************************************/

        [Description("Returns the height of a generic opening.")]
        [Input("opening", "A generic Opening object.")]
        [Output("height", "The total height of the generic opening.", typeof(Length))]
        public static double Height(this IOpening opening)
        {
            if (opening == null)
            {
                BH.Engine.Reflection.Compute.RecordError("Cannot query the height of a null opening.");
                return 0;
            }
            BoundingBox bBox = opening.IBounds();
            return (bBox.Max.Z - bBox.Min.Z);
        }
```

Physical_Engine/Query/Width.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        /***************************************************/
+        /****             Fallback Methods              ****/
+        /***************************************************/
+        
+        private static double Width(this object opening)
+        {
+            BH.Engine.Reflection.Compute.RecordError(String.Format("Width query for {0} type is not implemented.",opening.GetType()));
+            return 0;
+        }
```
IsakNaslundBh, line 70: Same as for the height method. Think this is better dealt with in the physical engine.

```suggestion
using BH.oM.Geometry;
using BH.Engine.Spatial;
using BH.oM.Quantities.Attributes;

namespace BH.Engine.Physical
{
    public static partial class Query
    {
        /***************************************************/
        /****               Public Methods              ****/
        /***************************************************/

        [Description("Returns the width of a generic opening.")]
        [Input("opening", "A generic Opening object.")]
        [Output("width", "The total width of the generic opening.", typeof(Length))]
        public static double Width(this IOpening opening)
        {
            if (opening == null)
            {
                BH.Engine.Reflection.Compute.RecordError("Cannot query the width of a null opening.");
                return 0;
            }

            BoundingBox bBox = opening.IBounds();

            double diffX = Math.Abs(bBox.Max.X - bBox.Min.X);
            double diffY = Math.Abs(bBox.Max.Y - bBox.Min.Y);

            return Math.Sqrt((diffX * diffX) + (diffY * diffY));
        }
```

IsakNaslundBh (changes requested):
Have put some initial reviews below.

Also, think the descriptions needs to be exapanded to tell more what they are doing. FOr example, this is giving heights in widths in global coordinates. An opening that was tilted for whatever reason would not give the dimension of it in its own plane, but just the global coordinate boundingbox distance (same before my suggested changes and after).

With that said, can see the use of the methods, but again, need to be described to clearly state what they are doing.

tiagogrossi:
@IsakNaslundBh thanks for the review! I agree with you that it seemed a bit awkward going to `Geometry` and then to `Physical`, I've done that because this whole thing exists already in `Environment`, but some methods are private, and therefore I thought we could just move it `Geometry` and any other toolkit could use it. Would you still be in favor of each toolkit that needs to handle it privately?

IsakNaslundBh:
> @IsakNaslundBh thanks for the review! I agree with you that it seemed a bit awkward going to `Geometry` and then to `Physical`, I've done that because this whole thing exists already in `Environment`, but some methods are private, and therefore I thought we could just move it `Geometry` and any other toolkit could use it. Would you still be in favor of each toolkit that needs to handle it privately?

Think my take on this is twofold, and why I suggested the changes I made:

1. To me it does not make sense to have a method in Geometry_Engine that measures the the `Width` or `Height` of a curve. My basic argument to that is that those are not well defined measurements for _any_ curve. To exemplify, what is the width of a line? (there are ofc exceptions, like if we added had a `Rectangle` class or similar, but then the current implementation would not work anyway). 
2. You do not need to measure the curve. as the curve method was working on the basis of the `BoundingBox` of the outline curve anyway. As we already have methods for getting out `BoundingBox` for any `IElement2d` (at least for planar versions) it feels more sensible to just get that out straight away and do the processing on it.

I could potentially see us adding a `Height` method for the `BoundingBox` class in geometry engine, as well as `HorizontalDiagonalLength` (or similar) for the current implementation of the `Width`, as those are querible properties that would be well defined for the class, in all cases.

tiagogrossi:
Good point @IsakNaslundBh , boundingbox does makes more sense. I've made the necessary changes and added more detail to the descriptions.

tiagogrossi:
@BHoMBot check compliance

Geometry_Engine/Query/Height.cs
```diff
@@ -0,0 +1,52 @@
...
+    public static partial class Query
+    {
+
+        /***************************************************/
+        /****       Public Methods - BoundingBox        ****/
+        /***************************************************/
+
+        [Description("Returns the vertical orthogonal height of a BHoM BoundingBox based on its Z minimum and maximum values.")]
+        [Input("boundingBox", "BHoM BoundingBox to query its height.")]
+        [Output("height", "The height of the BoundingBox based on the difference in Z minimum and maximum values.")]
```
IsakNaslundBh, line 38: The output here should have the Length quantity attribute added. Make sure you add the Using of `BH.oM.Quantities.Attributes` as well.

This should be updated for all methods.

```suggestion
        [Output("height", "The height of the BoundingBox based on the difference in Z minimum and maximum values.", typeof(Length))]
```

Geometry_Engine/Query/Width.cs
```diff
@@ -0,0 +1,56 @@
...
+    {
+
+        /***************************************************/
+        /****       Public Methods - BoundingBox        ****/
+        /***************************************************/
+
+        [Description("Returns the horizontal orthogonal width of a BHoM BoundingBox.")]
+        [Input("boundingBox", "BHoM BoundingBox to query its width.")]
+        [Output("width", "The horizontal width of the BoundingBox based on the difference in XY values for its bounding box")]
+        public static double Width(this BoundingBox boundingBox)
```
IsakNaslundBh, line 40: Don't think Width is the right name for this method. Would call it something like `HorizontalDiagonalLength` or similar. Does not need to be exactly that, but needs to be something that states what the method is returning.

Physical_Engine/Query/Height.cs
```diff
@@ -0,0 +1,75 @@
...
+        /***************************************************/
+        /****               Public Methods              ****/
+        /***************************************************/
+
+        [Description("Returns the vertical orthogonal height of a generic opening based on global coordinates of its BoundingBox.")]
+        [Input("opening", "A generic Opening object to query its height.")]
+        [Output("height", "The total height of the generic opening.")]
+        public static double IHeight(this IOpening opening)
+        {
+            return Height(opening as dynamic);
```
IsakNaslundBh, line 43: This needs a null check.

Physical_Engine/Query/Height.cs
```diff
@@ -0,0 +1,75 @@
...
+        [Output("height", "The total height of the door object.")]
+        public static double Height(this Door door)
+        {
+            if(door == null)
+            {
+                BH.Engine.Reflection.Compute.RecordError("Cannot query the height of a null door.");
+                return 0;
+            }
+
+            return BH.Engine.Geometry.Query.Height(door.Location.IBounds());
```
IsakNaslundBh, line 59: By adding the using of Spatial_Engine, you could query the bounds straight from the door. That is just details though, and I actually think what you have here might be quicker anyway.

Physical_Engine/Query/Width.cs
```diff
@@ -0,0 +1,75 @@
...
+        /***************************************************/
+        /****               Public Methods              ****/
+        /***************************************************/
+
+        [Description("Returns the horizontal orthogonal width of a generic opening based on global coordinates of its BoundingBox.")]
+        [Input("opening", "A generic Opening object to query its width.")]
+        [Output("width", "The total width of the generic opening.")]
+        public static double IWidth(this IOpening opening)
+        {
+            return Width(opening as dynamic);
```
IsakNaslundBh, line 43: again, null check needed

IsakNaslundBh (changes requested):
Some more comments below. Mainly name of the method currently called "Width" for the bounding box, and for correct quantity attributes to be assigned to all of the outputs.

IsakNaslundBh (approved):
Now lgtm

tiagogrossi:
@BHoMBot check required

tiagogrossi:
@BHoMBot check compliance

tiagogrossi:
@BHoMBot check required

tiagogrossi:
@BHoMBot check serialisation

tiagogrossi:
@BHoMBot check serialisation

IsakNaslundBh (approved):
Re-approving after compliance updates

IsakNaslundBh:
@BHoMBot check ready-to-merge

PR #2689, opened 2021-11-11
Reflection_Engine: reflection of non-BHoM types enabled

### LINKED ISSUE #2687: Reflection_Engine: add a feature to reflect non-BHoM types so that they showed up in AllTypeList query results
#### Description:

As in the title - the value of that would be enabling versioning (on demand) of non-BHoM types.

pawelbaran:

Closes #2687

### Test files

On [SharePoint](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?RootFolder=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F02%5FPull%20Request%2FBHoM%2FBHoM%5FEngine%2FReflection%5FEngine%2F%232689%2DReflectNonBHoMTypes&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4).

### Additional comments

I moved `m_ReflectAssemblyLock` to `Global` to become a central lock for all methods that attempt to modify the global collections. In the future it could be split into sub-locks per each collection, but for now there is no need for going so granular I think.


pawelbaran:
@BHoMBot check required

pawelbaran:
@BHoMBot check compliance
@BHoMBot check null-handling

pawelbaran:
@BHoMBot check core
@BHoMBot check installer
@BHoMBot check serialisation
@BHoMBot check versioning

adecler:
The main reason for this PR is that `ReflectAssembly` is ignoring all types with a namespace that doesn't start with `BH.` even when adding to `AllTypeList`. 

https://github.com/BHoM/BHoM_Engine/blob/28500ef755945ddbb263e0da0eae8cbae173e99b/Reflection_Engine/Compute/ReflectAssembly.cs#L120-L127

I vagely remember something around performances as the reason for filtering out non-BHoM types even in `AllTypeList` but would be good to properly assess/capture that now with the new version. If there is no valid reason to filter out those types, then it might be better to directly edit the `ExtractTypes` method.

Reflection_Engine/Compute/ReflectNonBHoMTypes.cs
```diff
@@ -0,0 +1,87 @@
...
+            lock (Global.AssemblyReflectionLock)
+            {
+                foreach (Type type in assembly.GetTypes())
+                {
+                    if (type.Namespace != null && !type.IsAutoGenerated())
+                        Global.AllTypeList.Add(type);
+                }
+
+                m_NonBHoMTypesReflected.Add(assembly.FullName);
+            }
```
IsakNaslundBh, line 62: Think the lock needs to hit before you check if the assembly has already been reflected. That is the critical part. So something like this could do it:

```suggestion
            lock (Global.AssemblyReflectionLock)
            {
                if (m_NonBHoMTypesReflected.Contains(assembly.FullName) || assembly.IsBHoM())
                    return;

                m_NonBHoMTypesReflected.Add(assembly.FullName);

                foreach (Type type in assembly.GetTypes())
                {
                    if (type.Namespace != null && !type.IsAutoGenerated())
                        Global.AllTypeList.Add(type);
                }
            }
```

Then, actually, the looping through the types in the assembly should be fine to keep outside the lock really, as the AllTypeList is a ConcurrentCollection, so could be changing it to the following really:

```suggestion
            lock (Global.AssemblyReflectionLock)
            {
                if (m_NonBHoMTypesReflected.Contains(assembly.FullName) || assembly.IsBHoM())
                    return;

                m_NonBHoMTypesReflected.Add(assembly.FullName);
            }
            
            foreach (Type type in assembly.GetTypes())
            {
                 if (type.Namespace != null && !type.IsAutoGenerated())
                   Global.AllTypeList.Add(type);
            }
```

But that is not how we have it in the other method, so might opt for option 1 to start with.

Might also be good to put this in a try-catch, as has been done with the other methods.

(Also, code suggestions made in the browser so indentation is probably a bit off 😄 )

IsakNaslundBh (changes requested):
Small tweak, see below.

pawelbaran:
> The main reason for this PR is that `ReflectAssembly` is ignoring all types with a namespace that doesn't start with `BH.` even when adding to `AllTypeList`.
> 
> https://github.com/BHoM/BHoM_Engine/blob/28500ef755945ddbb263e0da0eae8cbae173e99b/Reflection_Engine/Compute/ReflectAssembly.cs#L120-L127
> 
> I vagely remember something around performances as the reason for filtering out non-BHoM types even in `AllTypeList` but would be good to properly assess/capture that now with the new version. If there is no valid reason to filter out those types, then it might be better to directly edit the `ExtractTypes` method.

Actually, the current setup makes sense to me: as highlighted on a few occasions, we support only BHoM types ootb, while all other types cannot be guaranteed. So for performance reasons, it makes perfect sense to me to avoid loading all types from all assemblies. I did some basic tests to check the influence of switching:
- **from** loading only types from _BH.*_ namespaces from assemblies that pass `IsBHoM` query
- **to** loading all types from all assemblies

BHoM startup on my machine goes from 6.5s to 14.5s, plus the number of stored types goes up from less than 3000 to almost 50000. This sounds like a proper versioning overkill to me 😃 

image

Happy to be proven wrong though 👍

pawelbaran:
@BHoMBot check required

FraserGreenroyd (changes requested):
This is being blocked until the previous PRs problems have been resolved.

adecler:
Thanks @pawelbaran , I knew there was method in the madness! 
Always good to double-check though since I never remember what I did more than a few days ago 😋

FraserGreenroyd (commented):
This looks to be only adding things and should have no impact on bot related activity, so from DevOps perspective am happy for this to be merged following full testing.

IsakNaslundBh (approved):
Happy with this now

pawelbaran:
@BHoMBot check required

pawelbaran:
@BHoMBot check core

pawelbaran:
@BHoMBot check ready-to-merge

PR #2702, opened 2021-11-18
Geometry_Engine: Add Length and area method to ellipse

### LINKED ISSUE #2636: Geometry_Engine: Missing methods for Ellipse
#### Description:

Length and Area methods missing

IsakNaslundBh:

Closes #2636 

Adding Area and Length calculation methods for ellipse.

Area is trivial.

Length is based on infinite series calculation. 

Reference can be found here: https://en.wikipedia.org/wiki/Ellipse#Circumference (the second option by Ivory and Bessel)
And here: https://www.mathsisfun.com/geometry/ellipse-perimeter.html (Infinite series 2)

The number of terms used in the infinite series is 10 times the ratio between a and b (long and short radius) which seem to render very good results in general.

Limited the maximum number of terms to be calculated to 100 000, which will give close to exact results for ellipses with a ratio of up to 1:10 000 to 1:20 000. Calculating that number of terms takes about a minute on my machine (as a one-off calculation) that then will run quick for subsequent ellipses.

Only required number of coefficients are computed, which means even the first length calculation for more sane ellipses (rations of less than 1:100 (not sure if 1:100 should even qualify as a normal case)), will be quick to calculate even for the first pass, with subsequent calculations being even quicker. 

Can ofc increase this hard cutoff, but that seemed like a reasonable cutoff to me and as a balance for how long the computer should be allowed to be frozen. 

### Test files

Test file here:

https://burohappold.sharepoint.com/:f:/s/BHoM/EuQzVNKKlupGhuwpr3UhdFgBcLhH1RZGAu0xZyAqwPYyhA?e=xzdrsb

Some differences between the added calculation and what rhino gives, but pretty confident that the added method is more exact.


IsakNaslundBh:
@BHoMBot check serialisation

IsakNaslundBh:
@BHoMBot check core

IsakNaslundBh:
@BHoMBot check installer

IsakNaslundBh:
@BHoMBot check installer

Geometry_Engine/Query/Length.cs
```diff
@@ -51,41 +59,114 @@ public static double SquareLength(this Vector vector)
...
+            //Get out a as the long radius and b as the short radius
+            double a = Math.Max(curve.Radius1, curve.Radius2);
+            double b = Math.Min(curve.Radius1, curve.Radius2);
+
+            //Special case circle
+            if (a == b)
+                return 2 * Math.PI * a;
+
+            //Special case line
+            if (b == 0)
```
pawelbaran, line 97: Wouldn't it be worth applying tolerance here?

Geometry_Engine/Query/Length.cs
```diff
@@ -118,6 +202,81 @@ private static double Length(this ICurve curve)
...
+                    }
+                }
+            }
+            return m_ellipseCoeficients;
+        }
+
+        /***************************************************/
+
+        private static List<double> m_ellipseCoeficients = null;
+        private static object m_cooeficientLock = new object();
```
pawelbaran, line 278: Shouldn't the values be written in m_PascalCase?

Geometry_Engine/Query/Length.cs
```diff
@@ -118,6 +202,81 @@ private static double Length(this ICurve curve)
...
+
+        [Description("Calculates the square binomial cooefficients used in the infinite series of the ellipse length calculation.")]
+        private static List<double> SquareBinomialCoefficientsEllipse(int requiredCount)
+        {
+            //Only allow for one thread to modify the cache at once.
+            lock (m_cooeficientLock)
+            {
+                //Calcuation of coefficients only needs to be done once.
+                //Coefficients are being stored for next ellipse to be evaluated
+                if (m_ellipseCoeficients == null)
```
pawelbaran, line 217: Couldn't you declare the initial value of `m_ellipseCoeficients = new List<double>();`?

pawelbaran (changes requested):
LGTM functionality wise, just a few comments on the code itself.

Geometry_Engine/Query/Length.cs
IsakNaslundBh: true, will fix. Leftover from previous iteration of the code where the coefficient list was always fully populated.

Geometry_Engine/Query/Length.cs
```diff
@@ -51,41 +59,114 @@ public static double SquareLength(this Vector vector)
...
+        [Input("curve", "The Ellipse to calculate the length of.")]
+        [Output("length", "The length of the Ellipse.", typeof(Length))]
+        public static double Length(this Ellipse curve)
+        {
+            //Get out a as the long radius and b as the short radius
+            double a = Math.Max(curve.Radius1, curve.Radius2);
+            double b = Math.Min(curve.Radius1, curve.Radius2);
+
+            //Special case circle
+            if (a == b)
```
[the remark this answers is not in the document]
IsakNaslundBh: potentially. A bit tricky to know what tolerance to use though to ensure this works for any scale.

The check here is mainly for performance reasons and should not impact the result

Geometry_Engine/Query/Length.cs
IsakNaslundBh: Same as above. This is mainly here for performance. Guess I could check if b/a < 1e-9 or something here though

IsakNaslundBh:
@BHoMBot check required

pawelbaran (dismissed):
LGTM now 👍

IsakNaslundBh:
@BHoMBot check ready-to-merge

IsakNaslundBh:
@BHoMBot check required

IsakNaslundBh:
@pawelbaran I had to add nullchecks, do you mind re-approving?

pawelbaran (approved):

pawelbaran:
@BHoMBot check ready-to-merge

PR #2705, opened 2021-11-19
Architecture_Engine: Geometry() query for BuildersWork.Opening refactored to return Point

### LINKED ISSUE #2704: Architecture_Engine: Geometry() query for BuildersWork.Opening does not return a Point
#### Description:

...instead, it returns `PlanarSurface`, which is against the principle of all IElements0D to return a point.

pawelbaran:

Closes #2704

### Test files

It is a simple property return, so no tests should be needed.

### Additional comments

This is a breaking change (different return type compared to the original version of the method) - this cannot be avoided taken the nature of the bug (`Geometry()` _has to_ return `Point` for all types that implement `IElement0D` interface). Luckily, I do not think this type has been heavily used so far (if at all), taken it does not provide any interop features atm.

@IsakNaslundBh please shout if you are concerned about it and you have ideas how to lower the risk I introduce with this change.

### DIFF
--- a/Architecture_Engine/Query/Geometry.cs
+++ b/Architecture_Engine/Query/Geometry.cs
@@ -71,17 +71,11 @@ namespace BH.Engine.Architecture
         [Description("Extracts geometry of the given BuildersWork Opening.")]
         [Input("opening", "BuildersWork Opening to be queried for its geometry.")]
         [Output("Geometry of the input BuildersWork Opening.")]
-        public static PlanarSurface Geometry(this Opening opening)
+        public static Point Geometry(this Opening opening)
         {
             if (opening == null)
             {
                 BH.Engine.Reflection.Compute.RecordError("Cannot extract the geometry from a null opening.");
-                return null;
-            }
-
-            if (opening.Profile?.Edges == null || opening.Profile?.Edges.Count == 0)
-            {
-                BH.Engine.Reflection.Compute.RecordError("Cannot extract the geometry from an opening without profile edges.");
                 return null;
             }
 
@@ -91,7 +85,7 @@ namespace BH.Engine.Architecture
                 return null;
             }
 
-            return new PlanarSurface(new PolyCurve { Curves = opening.Profile.Edges.ToList() }, new List<ICurve>()).Orient(new Cartesian(), opening.CoordinateSystem);
+            return opening.CoordinateSystem.Origin;
         }
 
         /***************************************************/


IsakNaslundBh:
> @IsakNaslundBh please shout if you are concerned about it and you have ideas how to lower the risk I introduce with this change.

Not overly concerned by a change as long as you have checked through all use cases of the class. More wondering then about why an opening is an IElement0D in the first place?

IsakNaslundBh:
> > More wondering then about why an opening is an IElement0D in the first place?
> 
> Thanks @IsakNaslundBh. That is the question I was asking myself when working on the original oM PR: [BHoM/BHoM#1258](https://github.com/BHoM/BHoM/pull/1258). The main rationale behind implementing `IElement0D` interface was that `Opening` is based on `Cartesian` coordinate system, so the most simplistic (and sufficient for now) approach is to treat it as point-based.
> 
> One could argue that it could also be 1D (taken its depth into account) or 2D (as windows/doors etc.). However, in the cases I can foresee now, there is no need for more than a point location from the code perspective, while turning `Opening` into `IElement2D` would lead to issues related with extraction of the shape and dimensions (which is usually what you need on `Push`).

Yeah, thinking a bit more about it myself, it is quite analogous to the Bar class (a curve element with a swept profile) so can sort of understand the ifs and whys of it.

Then, I think the change is fine if you have checked through that it does not break anything (as mentioned before) and if it can be tested by someone who has used the class.

pawelbaran:
> Then, I think the change is fine if you have checked through that it does not break anything (as mentioned before) and if it can be tested by someone who has used the class.

Yeah, agreed - will make sure this gets verified before merging. This PR is a link in a larger chain of PRs to be raised today (still polishing the code), so more info, test files etc. will be added. Thanks!

michal-pekacki (approved):
LGTM 👍

pawelbaran:
@BHoMBot check compliance
@BHoMBot check null-handling

pawelbaran:
@BHoMBot check core
@BHoMBot check serialisation

FraserGreenroyd:
@BHoMBot check required

michal-pekacki (approved):
LGTM

pawelbaran:

pawelbaran:

FraserGreenroyd:
@BHoMBot check ready-to-merge

PR #2711, opened 2021-12-02
.ci: simple console app running the Engine checks locally and saving the results to path

### LINKED ISSUE #2710: Improve the ability to run and interpret the CI engine checks locally
#### Description:

As in the title, following the today's discussion between @al-fisher, @FraserGreenroyd, @IsakNaslundBh, @alelom and myself, it would be a good feature to streamline the local engine check process if possible. I will raise a PR with a prototype set up for my personal us in a second.

@adecler FYI

pawelbaran:

Closes #2710

### Test files

Just run it on your machine on 2 different sets of branches, then compare the outputs, e.g. in GH.

### Additional comments

This is a polished prototype that I have built when grappling with null-handling recently, feel free to say it is rubbish 😃 

BTW, while opening this PR, I just thought that we could possibly reflect the checks even in GH - wouldn't it that make running them locally even easier?

### DIFF
--- a/.ci/code/Engine_Test/Verify/NullChecks.cs
+++ b/.ci/code/Engine_Test/Verify/NullChecks.cs
@@ -46,8 +46,8 @@ namespace BH.Test.Engine
         public static TestResult NullChecks()
         {
             // Ignore because of stack overflow or need to install external libraries
-            List<string> toIgnore = new List<string> { "MachineLearning", "Python", "LifeCycleAssessment.Query.GetEvaluationValue" };
-
+            List<string> toIgnore = new List<string> { "MachineLearning", "Python", "LifeCycleAssessment.Query.GetEvaluationValue", "BH.Engine.Adapters.BIM360.Query.AuthorisationCode" };
+            
             // Test all the BHoM types available
             List<MethodInfo> methods = BH.Engine.Reflection.Query.BHoMMethodList()
                 .Where(x => !x.IsDeprecated())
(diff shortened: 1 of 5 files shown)


FraserGreenroyd (changes requested):
The console app belongs on Test Toolkit, not as part of BHoM_Engine

pawelbaran:
> The console app belongs on Test Toolkit, not as part of BHoM_Engine

Oke, and what do you think about the app itself? Does it make sense to:
a) have it as a separate project in Test_Toolkit
b) try to merge it into the TestRunner
c) not have it at all because it is rubbish 🙈 

Not sure whether worth pushing it forward, and if so, what direction to take - here 100% up to you @FraserGreenroyd, I do not know much about the CI/CD setup.

FraserGreenroyd:
Sorry @pawelbaran I was in a rush earlier so just wanted to get the key point down of this needs to be in Test Toolkit, not to dismiss the idea entirely 😄 

I would do it as a separate project to `TestRunner` - `TestRunner` is used by some DevOps processes so I'd be keen to avoid clashes at the moment. However, if we're doing this as a longer term development option, then I am happy for us to consider changing those processes to make use of a combined system. But on initial feedback, separation is best for this.

pawelbaran:
OK thanks @FraserGreenroyd - I will migrate it then and see how does it fit into the existing code base 👍

al-fisher:
Above all sounds good - great stuff!

FraserGreenroyd:
Following discussion with @pawelbaran this PR is being closed in favour of the roadmap towards the next versions of BHoMBot and additional tools via Test Toolkit to provide this functionality.

PR #2724, opened 2022-01-05
Serialiser_Engine: Fix issue with classmaps of inner generic types

### LINKED ISSUE #2722: Serialiser_Engine: Issue with classmaps of Generic types
#### Description:

Currently there are issues with class maps that are being registered not by themselves but by a top level object owning them for generic types. For example, if a `PointMatrix<object>` is being deserialised before `LocalData<object>`, then the BsonClassMap for the local data will be incorrectly registred, with incorrect Discriminator as well as `IgnoreExtraElements` set to false instead of true.

This only seem to be an issue for generic types that are owned by another type, if the owning type is Deserialised before the owned type is being serialised in the session. When this order is used the inner obejct (LocalData) will crash when it does not know how to handle the extra element "_bhomVersion" and return as a custom object.

Have a solution to the issue and will raise a PR for discussion on it.

#### Steps to reproduce:

Try deserialising a PointMatrix and then LocalData of the same generic type. If LocalData is deserialised first, all work fine in the rest of the session, if PointMatrix is deserialised first, LocalData will fail for the rest of the session.

#### Expected behaviour:

For class maps of generic objects to be correctly serialised with correct discriminator as well as IgnoreExtraElements set to true.

#### Test file(s):

will link in PR

IsakNaslundBh:

Closes #2722 

Adding ClassMapConvetion to correctly set properties of class maps registered outside the compute method.

See issue for more details as to why this is required.

Could also have a think if we should change how those settings are handled in general by removing all the settings in the compute method `RegisterClassMap` and change it so that all objects make use of this added ClassMapConvention instead, to ensure that all objects get these settings, independent on what registered their class map.

Would be a quite small change code wise, but tried to disrupt as little as possible with this PR.

@adecler if you have any other suggestion for how this can be handled better, I would be up to changing what ever is needed. 

**EDIT**: After chat with @adecler decided that updating the code so the convention is applied to every single class map gives a cleaner solution. By doing this, the same calls made in the compute method is now redundant and has hence been removed.

### Test files

https://burohappold.sharepoint.com/:f:/s/BHoM/EkaZJA6Qzr1Krl472i5JiWEB0wzyDhfPmrKhI200c0EFvQ?e=Jgm5LR

### Changelog

- Add convention for handling settings of all BSONClassMaps

### Additional comments

🤦‍♂️ forgot the issue number in the branch name. If this is a problem I will re-raise..


adecler (commented):
I am happy with the solution proposed here. Fixes the problem nicely without creating other issues (that I could find).

Only thing I would change is the name of the class from `BHoMDefaultClassMapConvention` to something like `GenericClassMapConvention` to make it clearer what it does.

IsakNaslundBh:
> I am happy with the solution proposed here. Fixes the problem nicely without creating other issues (that I could find).
> 
> Only thing I would change is the name of the class from `BHoMDefaultClassMapConvention` to something like `GenericClassMapConvention` to make it clearer what it does.

Yes, happy to change to that. Called it something like that first, but change the name when I realised that this really could be applied to all classes, but agree that with current setup that name is better. Will update!

IsakNaslundBh:
@BHoMBot check serialisation
@BHoMBot check versioning

IsakNaslundBh:
@BHoMBot check serialisation
@BHoMBot check versioning

IsakNaslundBh:
@adecler found some issues with some objects, namely the system DataTable class on the BHoM Table class when I applied the class map to everything. Hence added in the class map convention that it should only be applied to IObjects.

Let me know if you think this is not enough.

What we can do is to keep this new class map convention as is, but also keep the settings in the compute class. It will ofc be doubling up things, but might be a way out.

I am also happy to revert this PR to what it was initially, just fixing the inner generic types, if you think that is better.

IsakNaslundBh:
@BHoMBot check serialisation
@BHoMBot check versioning

adecler:
@IsakNaslundBh , let have another catch-up on this.

IsakNaslundBh:
@BHoMBot check serialisation
@BHoMBot check versioning

IsakNaslundBh:
@BHoMBot check serialisation
@BHoMBot check versioning

IsakNaslundBh:
@BHoMBot check required

FraserGreenroyd:

FraserGreenroyd (approved):
From my understanding of discussion with @IsakNaslundBh this is ok to merge today to unblock work on migrating stuff from Reflection_Engine to Base_Engine, following this mornings conversation between @IsakNaslundBh and @adecler .

The checks are happy with this PR on serialisation, so further eyes will be kept out on changes there. @adecler if you could check over the files changed from @IsakNaslundBh last commit, and if there are any issues with them, please raise new issues so they can be tackled ASAP.

Thanks both!

FraserGreenroyd:
@BHoMBot check ready-to-merge

FraserGreenroyd:
BHoMBot this is a DevOps instruction. I am authorising dispensation to be granted on check ref. 4760430000

FraserGreenroyd:
@BHoMBot this is a DevOps instruction. I am authorising dispensation to be granted on check ref. 4760429546

adecler:
>  @adecler if you could check over the files changed from @IsakNaslundBh last commit, and if there are any issues with them, please raise new issues so they can be tackled ASAP.

@FraserGreenroyd , I am happy with the last commit.

PR #2744, opened 2022-01-28
MEP_Engine: Update CompositeSolidVolumes to support lists of elements

### LINKED ISSUE #2743: MEP_Engine: Solid Volumes for List of Pipes and Ducts
#### Description:

Add the ability to query lists of pipes and ducts for their solid volumes.

michaelhoehn:

Closes #2743 

### Test files

@kayleighhoude @alexissantella could you please upload your current test files (specifically the excel file you were demonstrating with)?

### DIFF
--- a/MEP_Engine/Query/CompositeSolidVolumes.cs
+++ b/MEP_Engine/Query/CompositeSolidVolumes.cs
@@ -195,6 +195,66 @@ namespace BH.Engine.MEP
             };
             return output;
         }
+
+        /***************************************************/
+
+        [Description("Queries the solid volume of a Pipe by multiplying the section profile's solid area by the element's length.")]
+        [Input("pipe", "The Pipe to query solid volume.")]
+        [MultiOutput(0, "elementSolidVolume", "SolidVolume of the Element itself within a compiled SectionProfile.")]
+        [MultiOutput(1, "insulationSolidVolume", "The solid volume of the Pipe's exterior insulation.")]
+        [MultiOutput(2, "liningSolidVolume", "The solid volume of the Pipe's interior lining.")]
+        public static Output<double, double, double> CompositeSolidVolumes(this List<Pipe> pipe)
+        {
+            int pipeCount = pipe.Count();
+            List<double> lengthList = pipe.Select(x => x.Length()).ToList();
+            int lenghtListCount = lengthList.Count();
+
+            if (pipeCount != lenghtListCount)
+            {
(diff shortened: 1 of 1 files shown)


FraserGreenroyd (changes requested):
This is a lot of code duplication to enable list based operations.

It would be easier to just remove the old methods and then version them so any instances of those methods go to this method instead.

That though would then require changes in other codebases and would remove the capacity to easily call `Duct duct = new Duct(); duct.SolidVolume();` because it would need to be passed into a list of 1 to work.

Which gives another possibility to allow both options, which is for the existing methods which take a singular object to become something like:

```
public static Output<double, double, double> CompositeSolidVolumes(this Duct duct)
{
    List<Duct> ducts = new List<Duct>() { duct };
    return ducts.CompositeSolidVolume();
}
```

Ultimately though, both options are a workaround for the actual issue it seems from reading the linked issue, which is that Excel functionality does not match Grasshopper functionality. Which to me suggests the actual solution to the whole problem is to improve Excel_Toolkit rather than add more methods here. However, this one will take longer to achieve, so the quick fix into this engine may allow some prototyping to take place sooner. But ideally it needs to be done without the code duplication 😄

al-fisher:
> the actual solution to the whole problem is to improve Excel_Toolkit rather than add more methods here. However, this one will take longer to achieve, so the quick fix into this engine may allow some prototyping to take place sooner. 

Yes - agreed on this as an approach! 
Prototyping up the required functionality/workflow will also help drive the concrete need and demand to improve Excel. Sounds good! 👍 
Good to keep track on this work - referenced in Excel toolkit already say - so we don't loose track with these prototypes growing _too_ many legs before long term strategy found. 😃

michaelhoehn:
> It would be easier to just remove the old methods and then version them so any instances of those methods go to this method instead.

I was trying to be sneaky and not require any versioning 🦖 
But yes, I totally agree. 

> That though would then require changes in other codebases and would remove the capacity to easily call `Duct duct = new Duct(); duct.SolidVolume();` because it would need to be passed into a list of 1 to work.

Seems someone should just build lacing support into excel please. 🙏 
Your proposed solution would definitely work, but as you stated, the real problem is to improve excel in general. Which may take ages. 

> so the quick fix into this engine may allow some prototyping to take place sooner. But ideally it needs to be done without the code duplication 😄

I'll be closing this PR and moving the code to a prototype repo for further discussion.

michaelhoehn:
A new prototype toolkit has been created and code has been copied over. Closing this here on the core engine. 

We will continue this in the prototype code space. Thanks all 😃

PR #2750, opened 2022-02-11
BHoM_Engine: ThreadStatic attribute taken off the debug log and replaced by a lock

### LINKED ISSUE #2009: Reflection_Engine: Event log to be able to record events from multiple threads
#### Description:

The DebugLog is currently thread static, which is good from a stability perspective, but means that any use of linq 'AsParallel' for example mean events raised on a separate thread does not get passed to the UI.

Logging across multiple threads should all show up on a component in the UI for example. For this, the logging system needs to be made non-thread static and made thread safe.

### LINKED ISSUE #2745: Base_Engine: Remove ThreadStatic attribute from m_DebugLog field to enable recording events in multithreaded code
#### Description:

As in the title, I would like to enable recording events in multithreaded code. Currently it is impossible because of `m_DebugLog` field having `ThreadStatic` attribute, which results in the field value being set to default on each thread that attempts to access it. I guess the attribute has been added for a reason, so it would be great if someone could shed some light on that, @adecler @IsakNaslundBh @FraserGreenroyd?

I am considering whether it would be OK to make the access to log thread safe by putting it behind `lock` or, alternatively, by changing its `List` properties to `ConcurrentBag`? The 1st option sounds best to me, but curious what do others think.

Happy to action the decision made in here as it is pretty relevant to my current work 👍

pawelbaran:

Closes #2009
Closes #2745

### Test files

Not sure how to test it meaningfully - the code still works as previously and does not show any signs of issues. I think that the best way of tasting can be simply merging the change and watching the logs carefully for the coming 2 sprint to make sure that this does not break anything.

### Additional comments

The approach taken has been discussed with @IsakNaslundBh: it seems to deliver best value for price:
- the log sits safely behind the lock
- the lock is `Global`, so it can be used across the entire project
- `Query.DebugLog` has already been `internal` so there is no risk of the collection being modified from the outside

An alternative that we discussed with @IsakNaslundBh was moving `BH.oM.Base.Debugging.Log` to the engine's `Objects` and encapsulating it properly. However, this would require more effort that does not seem to be justified for now.

### DIFF
--- a/BHoM_Engine/Compute/ClearCurrentEvents.cs
+++ b/BHoM_Engine/Compute/ClearCurrentEvents.cs
@@ -38,9 +38,12 @@ namespace BH.Engine.Base
         [PreviousVersion("5.1", "BH.Engine.Reflection.Compute.ClearCurrentEvents()")]
         public static bool ClearCurrentEvents()
         {
-            Log log = Query.DebugLog();
-            log.CurrentEvents.Clear();
-            return true;
+            lock (Global.DebugLogLock)
+            {
+                Log log = Query.DebugLog();
+                log.CurrentEvents.Clear();
+                return true;
+            }
         }
 
         /***************************************************/
--- a/BHoM_Engine/Compute/ClearCurrentEvents.cs
+++ b/BHoM_Engine/Compute/ClearCurrentEvents.cs
@@ -25,6 +25,7 @@ using BH.oM.Base.Debugging;
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.Reflection;
 
 namespace BH.Engine.Base
@@ -36,6 +37,8 @@ namespace BH.Engine.Base
         /***************************************************/
 
         [PreviousVersion("5.1", "BH.Engine.Reflection.Compute.ClearCurrentEvents()")]
+        [Description("Clears the current event log buffer.")]
+        [Output("success", "True if the buffer is cleared successfully.")]
         public static bool ClearCurrentEvents()
         {
             lock (Global.DebugLogLock)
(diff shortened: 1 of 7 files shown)


pawelbaran:
@BHoMBot check compliance
@BHoMBot check null-handling

pawelbaran:
@BHoMBot check compliance

pawelbaran:
@BHoMBot check core
@BHoMBot check null-handling
@BHoMBot check serialisation
@BHoMBot check versioning
@BHoMBot check installer

adecler (approved):
I am happy with the solution.
Agreed that it is best to keep things simple and leave the log where it is for now.

pawelbaran:
@BHoMBot check ready-to-merge

IsakNaslundBh (approved):
Looks good to me. All instances using the `m_DebugLog` or calling the `DebugLog()` method are protected with the same lock.

Looks safe to me as long as the `DebugLog()` method stays internal as stated in the PR description by @pawelbaran .

In terms of testing, could be done with a dummy method with a Parallel.ForEach or similar simply raising tons of events and making sure no race conditions occurs. From the code side I can not see that being an issue, but ofc, good to test before merging.

pawelbaran:
Thanks @adecler @IsakNaslundBh. Following the request, I am pasting the code that I have used to (simply) test the feature, with positive results:

```C#
int errorCount = 10000;
BH.Engine.Base.Compute.ClearCurrentEvents();
Parallel.ForEach(Enumerable.Range(0, errorCount), i => { BH.Engine.Base.Compute.RecordError($"Error no. {i + 1}"); });

var errors = BH.Engine.Base.Query.CurrentEvents();
foreach (var e in errors)
{
    Console.WriteLine(e.Message);
}

Console.WriteLine($"\nTotal error count: {BH.Engine.Base.Query.CurrentEvents().Count}, test finished with result: {errors.Count == errorCount}");
BH.Engine.Base.Compute.ClearCurrentEvents();
```

I will merge the PR now, thanks all 👍

PR #2756, opened 2022-02-21
Geometry_Engine: Add empty bounds methods for Face, Basis and TransformMatrix

### LINKED ISSUE #2755: Geometry_Engine: Add Bounds method for Face and Basis
#### Broken rules:

The Basis, Face and TransformMatrix does not have a method for Bounds. This can lead to unwanted error being thrown in Grasshopper when the geometry is being displayed. 

Add Bounds methods for Basis, Face and TransformMatrix returning empty bounding boxes.

IsakNaslundBh:

Closes #2755 

Adding empty return methods for Bounds of Face, TransformMatrix and Basis. This is mostly done to help with the UI:s that are calling those methods implicitly through the display pipeline, and in some cases this leads to non-sensical errors being thrown.

### DIFF
--- a/Geometry_Engine/Query/Bounds.cs
+++ b/Geometry_Engine/Query/Bounds.cs
@@ -75,9 +75,23 @@ namespace BH.Engine.Geometry
 
         /***************************************************/
 
+        public static BoundingBox Bounds(this Basis basis)
+        {
+            return new BoundingBox();
+        }
+
+        /***************************************************/
+
         public static BoundingBox Bounds(this BoundingBox boundingBox)
         {
             return boundingBox == null ? null : boundingBox.DeepClone();
+        }
+
+        /***************************************************/
+
+        public static BoundingBox Bounds(this TransformMatrix matrix)
+        {
+            return new BoundingBox();
         }
 
         /***************************************************/
@@ -453,6 +467,13 @@ namespace BH.Engine.Geometry
         public static BoundingBox Bounds(this Mesh mesh)
         {
             return mesh?.Vertices?.Bounds();
+        }
+
+        /***************************************************/
+
+        public static BoundingBox Bounds(this Face face)
+        {
+            return new BoundingBox();
         }
 
         /***************************************************/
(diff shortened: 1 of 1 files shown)


IsakNaslundBh:
@BHoMBot check compliance

IsakNaslundBh:
@BHoMBot check null-handling

IsakNaslundBh:
@BHoMBot check required

pawelbaran (commented):
Just thinking about it... wouldn't it be possible to fix/suppress the errors on different level? The proposed changes look a bit suspicious, particularly the one for `Face` - it does not make much sense from the user's perspective to get empty bounds for a spatial elements. I get why does that happen, but the user would not, particularly taken the lack of notification (which I guess you are trying to avoid not to show them on the UI without a reason).

To me it looks like a design issue that we should try to solve instead of hiding. Happy to be proven wrong with my assumptions! 😉

FraserGreenroyd (approved):
Code changes look good, introduce the missing methods to avoid unnecessary errors to users. Improves UX as a result 👍

IsakNaslundBh:
@BHoMBot check required

> Just thinking about it... wouldn't it be possible to fix/suppress the errors on different level? The proposed changes look a bit suspicious, particularly the one for `Face` - it does not make much sense from the user's perspective to get empty bounds for a spatial element. I get why does that happen, but the user would not, particularly taken the lack of notification (which I guess you are trying to avoid not to show them on the UI without a reason).
> 
> To me it looks like a design issue that we should try to solve instead of hiding. Happy to be proven wrong with my assumptions! 😉

After discussion offline, will revert the empty return method for the Face class, and have raised issue in the oM here : https://github.com/BHoM/BHoM/issues/1365 considering removal of the IGeometry interface from the Face class as potentially better solution to the issue.

Will revert the addition of this method in this PR.

IsakNaslundBh:
@BHoMBot check compliance
@BHoMBot check required

pawelbaran (approved):

IsakNaslundBh:
@BHoMBot check compliance
@BHoMBot check required

IsakNaslundBh:
@BHoMBot check ready-to-merge

IsakNaslundBh:
@BHoMBot check ready-to-merge

PR #2768, opened 2022-03-04
Result_Engine: Result display improvements

### LINKED ISSUE #2740: Results_Engine: MapResults to allow for mapping by more than AdapterId
#### Description:

Currently the method relies on the `AdapterId` for mapping the results to objects, which is based on Structures general requirements. This should be generalised a bit more to also allow for mapping to be based on Name as well as BHoM_Guid.:

image

This will in turn improve the capabilities of the DispayMeshResults method, by allowing for the same mapping strategy to be used for faces as well as nodes:

image

### LINKED ISSUE #2757: Results_Engine: Add method for displaying results for IObjectResults
#### Description:

Method that takes a list of BHoMObejcts, a list of IObjectResults and a Gradient, maps the results to the objects and then returns a list of RenderGeometry using the Geometry from the BHoMObejct and colour based on the result

### LINKED ISSUE #2758: Graphics_oM: Add method for displaying the geometry of a gradient
#### Description:

Add method for generating a RenderMesh from a gradient and settings.

### LINKED ISSUE #2764: Results_Engine: Add Display results method for IElement1DResults
#### Description:

Method that takes a list of BHoMObejcts, a list of IElement1DResults and a Gradient, maps the results to the objects and then returns a list of RenderGeometry corresponding to the results along the element

### LINKED ISSUE #2765: Result_Engine: Result display improvements
#### Description:

Collection issue for various result display improvements

- #2764 
- #2757 
- #2758 
- #2740

### LINKED ISSUE #2772: Result_Engine: Improve filtering capabilities of MapResults
#### Description:

MapResults currently allows for filtering of results based on their ResultCase. This should be expanded to allow for filtering based on other properties as well, such as TimeStep and other relevant interfaces that might be added in the future.

This will in turn expand the capabilities of the display methods added and refactored in https://github.com/BHoM/BHoM_Engine/pull/2768/files

Plan to add a ResultFilter object that will be taken as an input for the methods instead of the current list of strings that will enable both filtering by result case, but also enable filtering by other properties.

IsakNaslundBh:
### NOTE: Depends on 

https://github.com/BHoM/BHoM/pull/1371
   
Closes #2765 
Closes  #2764 
Closes  #2757 
Closes  #2758 
Closes  #2740
Closes #2772 

- Expanding the functionality of MapResults to allow for a string pointing to a property to be used instead of just a IAdapterId type
- Adding method for fetching properties with potential to be displayed from result class
- Adding Display method for IObjectResult
- Adding Display method for IElement1DResult
- Adding methods for generating a legend from a GradientOptions object

### Test files

General test files:
https://burohappold.sharepoint.com/:f:/r/sites/BHoM/02_Current/12_Scripts/02_Pull%20Request/BHoM/BHoM_Engine/Results_Engine/%232768-ResultDisplayImprovements?csf=1&web=1&e=6fPLCE

Structural results extraction:
https://burohappold.sharepoint.com/:f:/r/sites/BHoM/Installers/Scripts/Burohappold_BHoM_v5.1/Structures%20-%20Results?csf=1&web=1&e=dQ3x2X


IsakNaslundBh:
@BHoMBot check null-handling

IsakNaslundBh:
@BHoMBot check compliance
@BHoMBot check null-handling

IsakNaslundBh:
@BHoMBot check compliance
@BHoMBot check null-handling

IsakNaslundBh:
@BHoMBot check compliance
@BHoMBot check null-handling

IsakNaslundBh:
@BHoMBot fix project file ref. 5445976307

IsakNaslundBh:
@BHoMBot check compliance
@BHoMBot check null-handling

IsakNaslundBh:
@BHoMBot check compliance
@BHoMBot check null-handling

IsakNaslundBh:
@BHoMBot check compliance
@BHoMBot check null-handling
@BHoMBot check serialisation

IsakNaslundBh:
@BHoMBot check versioning

IsakNaslundBh:
@BHoMBot check versioning

Graphics_Engine/Convert/ToSteppedGradient.cs
```diff
@@ -0,0 +1,75 @@
...
+namespace BH.Engine.Graphics
+{
+    public static partial class Convert
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Coverts a gradient to a stepped gradient by extracting colours in intervals.")]
+        [Input("gradient", "The gradient to covert. If already a stepped gradient, no action will be taken.")]
```
alelom, line 41:
```suggestion
        [Input("gradient", "The gradient to convert. If already a stepped gradient, no action will be taken.")]
```
```suggestion
        [Input("gradient", "The gradient to covert. If already a stepped gradient, no action will be taken.")]
```

alelom, line 41:
```suggestion
        [Input("gradient", "The gradient to convert. If already a stepped gradient, no action will be taken.")]
```
```suggestion
        [Input("gradient", "The gradient to covert. If already a stepped gradient, no action will be taken.")]
```

Graphics_Engine/Convert/ToSteppedGradient.cs
```diff
@@ -0,0 +1,75 @@
...
+        [Input("nbSteps", "Number of steps to be used. If -1, the markers of the provided gradient will be used.")]
```
alelom, line 42:
```suggestion
        [Input("steps", "(Optional, defaults to -1) Number of steps to be used. If -1, the markers of the provided gradient will be used.")]
```
```suggestion
        [Input("nbSteps", "Number of steps to be used. If -1, the markers of the provided gradient will be used.")]
```

alelom, line 42:
```suggestion
        [Input("steps", "(Optional, defaults to -1) Number of steps to be used. If -1, the markers of the provided gradient will be used.")]
```
```suggestion
        [Input("nbSteps", "Number of steps to be used. If -1, the markers of the provided gradient will be used.")]
```

Graphics_Engine/Modify/ApplyGradientOptions.cs
```diff
@@ -38,12 +38,14 @@ public static partial class Modify
...
         [Description("Sets up the properties of a GradientOptions object for usage.")]
         [Input("gradientOptions", "GradientOptions object to modify.")]
         [Input("allValues", "The values to set gradient auto range from. Optional if range is already set.")]
+        [Input("gradientBoundsWarning", "If true, a warning will be raised if the bounds have been manually set and any of the provided values are outside of the bounds domain.")]
         [Input("defaultGradient", "Sets which gradient to use as default if no gradient is already set. Defaults to BlueToRed.")]
         [Output("gradientOptions", "A GradientOptions object which is ready for usage.")]
-        public static GradientOptions ApplyGradientOptions(this GradientOptions gradientOptions, IEnumerable<double> allValues = null, string defaultGradient = "BlueToRed")
+        public static GradientOptions ApplyGradientOptions(this GradientOptions gradientOptions, IEnumerable<double> allValues = null, bool gradientBoundsWarning = true, string defaultGradient = "BlueToRed")
```
alelom, line 48: I think that the option `gradientBoundsWarning` could be better placed as the last input parameter, being only of contextual utility? At least that is what I do when I have similar cases

alelom, line 48: I think that the option `gradientBoundsWarning` could be better placed as the last input parameter, being only of contextual utility? At least that is what I do when I have similar cases

Graphics_Engine/Query/Bounds.cs
```diff
@@ -32,6 +35,7 @@ public static partial class Query
         /**** Public Methods                            ****/
         /***************************************************/
 
+        [Description("Computes and returns the BoundingBox the SVGObject which is the concatenated BoudingBox of all Shapes contained within the object..")]
```
alelom, line 38:
```suggestion
        [Description("Computes and returns the BoundingBox the SVGObject which is the concatenated BoudingBox of all Shapes contained within the object.")]
```
```suggestion
        [Description("Computes and returns the BoundingBox the SVGObject which is the concatenated BoudingBox of all Shapes contained within the object..")]
```

alelom, line 38:
```suggestion
        [Description("Computes and returns the BoundingBox the SVGObject which is the concatenated BoudingBox of all Shapes contained within the object.")]
```
```suggestion
        [Description("Computes and returns the BoundingBox the SVGObject which is the concatenated BoudingBox of all Shapes contained within the object..")]
```

Results_Engine/Query/DisplayElement1DResults.cs
```diff
@@ -0,0 +1,282 @@
...
+                    return results.GroupBy(x => new { ((ICasedResult)x).ResultCase, ((ITimeStepResult)x).TimeStep }).Select(x => x.ToList()).ToList();
+                else
+                    return results.GroupBy(x => ((ICasedResult)x).ResultCase).Select(x => x.ToList()).ToList();
+            }
+            else if (results.FirstOrDefault() is ITimeStepResult)
+                return results.GroupBy(x => ((ITimeStepResult)x).TimeStep).Select(x => x.ToList()).ToList();
+            else
+                return new List<List<T>> { results.ToList() };
+
+        }
```
alelom, line 278: I'm thinking we could expose these as public overloaded methods for different groupings. Could be very useful in other instances too.
E.g.
```cs
public List<List<T>> GroupByTimeStep<T>(this IEnumerable<T> results) where T : ITimeStepResult
{
  // do stuff
}

public List<List<T>> GroupByResultCase<T>(this IEnumerable<T> results) where T : ICasedResult
{
  // do stuff
}

public List<List<T>> GroupByResultCaseAndTimeStep<T>(this IEnumerable<T> results) where T : ICasedResult, ITimeStepResult
{
  // do stuff
}
```

alelom, line 278: I'm thinking we could expose these as public overloaded methods for different groupings. Could be very useful in other instances too.
E.g.
```cs
public List<List<T>> GroupByTimeStep<T>(this IEnumerable<T> results) where T : ITimeStepResult
{
  // do stuff
}

public List<List<T>> GroupByResultCase<T>(this IEnumerable<T> results) where T : ICasedResult
{
  // do stuff
}

public List<List<T>> GroupByResultCaseAndTimeStep<T>(this IEnumerable<T> results) where T : ICasedResult, ITimeStepResult
{
  // do stuff
}
```

Results_Engine/Query/DisplayMeshResults.cs
```diff
@@ -82,17 +85,16 @@ public static partial class Query
...
-            //Check if no identifier has been provided. If this is the case, identifiers are searched for in the objects
-            identifier = meshes.First().FindIdentifier(identifier);
 
             List<IMesh<TNode, TFace>> meshList = meshes.ToList();
 
             // Map the MeshResults to Meshes
-            List<List<IMeshResult<IMeshElementResult>>> mappedResults = meshList.MapResults(meshResults, "ObjectId", identifier, caseFilter);
-            Func<IMeshElementResult, double> propertyFuction = GetPropertFunc(meshResults.First().Results.First() as dynamic, meshResultDisplay);
+            List<List<IMeshResult<IMeshElementResult>>> mappedResults = meshList.MapResults(meshResults, "ObjectId", objectIdentifier, caseFilter);
+            //Get function for extracting property from results
```
alelom, line 93:
```suggestion
            //Get function for extracting property from results
```
```suggestion
            //Get tuple with result name, property selector function and quantity attribute
```

alelom, line 93:
```suggestion
            //Get function for extracting property from results
```
```suggestion
            //Get tuple with result name, property selector function and quantity attribute
```

Results_Engine/Query/DisplayMeshResults.cs
```diff
@@ -82,17 +85,16 @@ public static partial class Query
...
+            var resPropSelectorAndQuantity = meshResults.First().ResultItemValueProperty(meshResultDisplay);
```
alelom, line 94: To aid reading the code and remembering what is inside the Tuple, I generally like to use names with underscores e.g.
name_propSelector_quantity, but that's my personal preference

alelom, line 94: To aid reading the code and remembering what is inside the Tuple, I generally like to use names with underscores e.g.
name_propSelector_quantity, but that's my personal preference

Results_Engine/Query/ResultItemValueProperty.cs
```diff
@@ -0,0 +1,132 @@
...
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Returns a Fuction for extracting a value from a result item.")]
+        [Input("result", "The result used to extract the type to generate the Func for value extraction.")]
+        [Input("propertyName", "The name of the property to get the function for.")]
+        [MultiOutput(0, "propertyName", "The name of the property or method function.")]
+        [MultiOutput(1, "func", "The result extraction function.")]
+        [MultiOutput(2, "quantity", "The quantity corresponding to the function and property.")]
+        public static Output<string, Func<IResultItem, double>, QuantityAttribute> ResultItemValueProperty(this IResultItem result, string propertyName)
```
alelom, line 49: Maybe worth formalising the Output into a class, to make it easier to reuse this function and avoiding Tuples?

We could have a class `ResultItemValueProperty` with those three properties (PropertyName, ExtractorFunction, Quantity)

alelom, line 49: Maybe worth formalising the Output into a class, to make it easier to reuse this function and avoiding Tuples?

We could have a class `ResultItemValueProperty` with those three properties (PropertyName, ExtractorFunction, Quantity)

Results_Engine/Query/DisplayMeshResults.cs
alelom: Also see my comment below https://github.com/BHoM/BHoM_Engine/pull/2768/files#r822558773

alelom: Also see my comment below https://github.com/BHoM/BHoM_Engine/pull/2768/files#r822558773

alelom, line 94:
```suggestion
            //Get tuple with property name, function to extract its value and its quantity
            var resPropSelectorAndQuantity = meshResults.First().ResultItemValueProperty(meshResultDisplay);
```

alelom, line 94:
```suggestion
            //Get tuple with property name, function to extract its value and its quantity
            var resPropSelectorAndQuantity = meshResults.First().ResultItemValueProperty(meshResultDisplay);
```

Results_Engine/Query/MapResults.cs
```diff
@@ -72,22 +78,18 @@ public static partial class Query
...
 
-            Dictionary<string, IGrouping<string, TResult>> resGroups;
 
-            //Group results by Id and turn to dictionary
-            // Add null check for when the property of the name in whichId does not exist?
-            resGroups = filteredRes.GroupBy(x => Base.Query.PropertyValue(x, whichId).ToString()).ToDictionary(x => x.Key);
+            Func<TResult, string> resultIdFunction = GetResultIdentifier(results.First(), resultIdentifier);
+            //Turn to Lookup based on the result ID function
+            //Add null check for when the property of the name in whichId does not exist?
+            var resGroups = filteredRes.ToLookup(resultIdFunction);
```
alelom, line 85: These could be turned again in a set of public overloaded Query methods for grouping results, which I think could come in handy in other cases. E.g.

GroupByIdentifier(results, resultIdentifier)

alelom, line 85: These could be turned again in a set of public overloaded Query methods for grouping results, which I think could come in handy in other cases. E.g.

GroupByIdentifier(results, resultIdentifier)

Results_Engine/Query/MapResults.cs
```diff
@@ -111,6 +113,89 @@ private static string IdMatch(this IBHoMObject o, Type identifier)
...
+
+            if (idLower == nameof(IMeshElementResult.NodeId).ToLower() && obj is IMeshElementResult)
+                return x => ((IMeshElementResult)x).NodeId.ToString();
+
+            if (idLower == nameof(IMeshElementResult.MeshFaceId).ToLower() && obj is IMeshElementResult)
+                return x => ((IMeshElementResult)x).MeshFaceId.ToString();
+
+            //Fall back on PropertyValue
+            return x => Base.Query.PropertyValue(x, identifier).ToString();
+        }
```
alelom, line 195: Would be nice to make available publicly as per comment https://github.com/BHoM/BHoM_Engine/pull/2768/files#r822565874

alelom, line 195: Would be nice to make available publicly as per comment https://github.com/BHoM/BHoM_Engine/pull/2768/files#r822565874

Results_Engine/Query/DisplayObjectResult.cs
```diff
@@ -0,0 +1,150 @@
...
+        [Input("caseFilter", "Which cases to colour by, default is all.")]
+        [Input("displayProperty", "The name of the property on the result to colour by. If nothing is provided, the first available property will be used.")]
+        [Input("gradientOptions", "How to color the mesh, null defaults to `BlueToRed` with automatic range.")]
+        [MultiOutput(0, "results", "A List of Lists of RenderGeometry, where the outer list corresponds to the object and the inner list correspond to the matchis results..")]
+        [MultiOutput(1, "gradientOptions", "The gradientOptions that were used to colour the meshes.")]
+        public static Output<List<List<RenderGeometry>>, GradientOptions> DisplayObjectResult(this IEnumerable<IBHoMObject> objects, IEnumerable<IObjectResult> results, string displayProperty = "", object objectIdentifier = null, List<string> caseFilter = null, GradientOptions gradientOptions = null)
```
alelom, line 59: To allow for a CaseFilter input, shouldn't we place a type constraint on the `results` input to be `ICasedResults`?

I would probably have a separate overloaded method to allow for that CaseFilter input/result type combination

alelom, line 59: To allow for a CaseFilter input, shouldn't we place a type constraint on the `results` input to be `ICasedResults`?

I would probably have a separate overloaded method to allow for that CaseFilter input/result type combination

Results_Engine/Query/MapResults.cs
```diff
@@ -37,14 +38,17 @@ public static partial class Query
...
+        [Input("objectIdentifier", "Should either be a string specifying what property on the object that should be used to map the objects to the results, or a type of IAdapterId fragment to be used to extract the object identification, i.e. which fragment type to look for to find the identifier of the object. If no identifier is provided, the object will be scanned an IAdapterId to be used.")]
         [Input("caseFilter", "Optional filter for the case. If nothing is provided, all cases will be used.")]
         [Output("results", "Results as a List of List where each inner list corresponds to one BHoMObject based on the input order.")]
-        public static List<List<TResult>> MapResults<TResult, TObject>(this IEnumerable<TObject> objects, IEnumerable<TResult> results, string whichId = nameof(IObjectIdResult.ObjectId), Type identifier = null, List<string> caseFilter = null) 
+        public static List<List<TResult>> MapResults<TResult, TObject>(this IEnumerable<TObject> objects, IEnumerable<TResult> results, string resultIdentifier = nameof(IObjectIdResult.ObjectId), object objectIdentifier = null, List<string> caseFilter = null)
```
alelom, line 51: I'm considering the accuracy of this method name and inputs ordering. Because the "main subject" (the input that is specified with the `this` keyword) seems to be `objects`, just by considering that I would say that `MapObjects()` could be a more accurate name, especially because we are already inside the namespace `Results`.

However, `MapObjects()` does not seem to be what the method actually does. So if we keep the name `MapResults()` maybe we should consider changing the order of the inputs to be `(this IEnumerable<TResult> results, IEnumerable<TObject> objects, ...` instead?

Personally I would prefer an even more explicit solution, with swapped inputs:
`MapResultsToObjects(this IEnumerable<TResult> results, IEnumerable<TObject> objects, ...)`.

alelom, line 51: I'm considering the accuracy of this method name and inputs ordering. Because the "main subject" (the input that is specified with the `this` keyword) seems to be `objects`, just by considering that I would say that `MapObjects()` could be a more accurate name, especially because we are already inside the namespace `Results`.

However, `MapObjects()` does not seem to be what the method actually does. So if we keep the name `MapResults()` maybe we should consider changing the order of the inputs to be `(this IEnumerable<TResult> results, IEnumerable<TObject> objects, ...` instead?

Personally I would prefer an even more explicit solution, with swapped inputs:
`MapResultsToObjects(this IEnumerable<TResult> results, IEnumerable<TObject> objects, ...)`.

alelom, line 51: On the `caseFilter` input, see my comment https://github.com/BHoM/BHoM_Engine/pull/2768/files#r822570021

Results_Engine/Query/ResultItemValueProperties.cs
```diff
@@ -0,0 +1,133 @@
...
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Extract all potential result carrying property getters from a result class, i.e. properties of type double that is defined on the class. Properties defined on a base class are ignored.\n" +
+                     "Also extracts methods returning a double that has the result type as the only argument.\n" +         
+                     "Func<T,double> returned together with corresponding QuantityAttribute in a Dictionary with the property or method name as the Key.")]
+        [Input("result", "The result type to extract property functions from.")]
+        [Output("func", "Dictionary of string, Tuple<Func,Quantity> where the Key is the name of the property or method and the value is the compiled selector for extracting the value from a result of the type with matching QuantityAttribute.")]
+        public static Dictionary<string, Tuple<Func<T, double>, QuantityAttribute>> ResultItemValueProperties<T>(this T result) where T : IResultItem
```
alelom, line 48: Maybe the output here could be simplified too similarly to my comment https://github.com/BHoM/BHoM_Engine/pull/2768/files#r822558773?

alelom, line 48: Maybe the output here could be simplified too similarly to my comment https://github.com/BHoM/BHoM_Engine/pull/2768/files#r822558773?

Results_Engine/Query/ResultPropertyKeys.cs
```diff
@@ -0,0 +1,70 @@
...
+            {
+                IDictionary dict = ResultItemValueProperties(result as dynamic);
+                if (dict != null)
+                    return dict.Keys.Cast<string>().ToList();
+            }
+            else
+            {
+                Engine.Base.Compute.RecordError($"Unsupported result type. Currently only suporting {nameof(IResultItem)} and {nameof(IResultCollection<IResultItem>)} type results.");
+            }
+
```
alelom, line 64: Could we resolve this with multiple overloads and dynamic dispatch?

Results_Engine/Query/ResultPropertyKeys.cs
```diff
@@ -0,0 +1,70 @@
...
+    public static partial class Query
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Returns the result value carrying proeprties available for the result type provided. Currently only supported for IResultItem and IResultCollection<IResultItem> type results.")]
+        [Input("result", "The result to fetch applicable result properties from.")]
+        [Output("props", "Resulve value carrying properties for the result type.")]
+        public static List<string> ResultPropertyKeys(this IResult result)
```
alelom, line 43: Maybe this function could be simplified if we do something as per comment https://github.com/BHoM/BHoM_Engine/pull/2768/files#r822579307

Results_Engine/Query/ResultPropertyKeys.cs
```diff
@@ -0,0 +1,70 @@
...
+{
+    public static partial class Query
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Returns the result value carrying proeprties available for the result type provided. Currently only supported for IResultItem and IResultCollection<IResultItem> type results.")]
+        [Input("result", "The result to fetch applicable result properties from.")]
+        [Output("props", "Resulve value carrying properties for the result type.")]
```
alelom, line 42: Watch out for typos

alelom (changes requested):
Amazing work, well done!

I left a few comments here and there mainly on general tidiness and code structure. Functionality seems ok, I still have to test scripts though.

Graphics_Engine/Modify/ApplyGradientOptions.cs
IsakNaslundBh: I actually put it last when first adding the parameter, but then moved it in front of the default gradient. My reason for putting it there was that I thought it was more likely that you would want to change whether to display the warning or not than it was likely that you would want to change the default gradient, hence leaving the default gradient last.

Can move it to the last position if you disagree with this.

Results_Engine/Query/DisplayElement1DResults.cs
IsakNaslundBh: We actually have the first two methods already :)

For this one, I tried to think of a way to basically group by non-object related identifiers in a generic way, but then run a bit short of coming up with something at the spot.

Would like to have a method like that basically (GroupByNonObejctIdentifiers) or similar. Will have a bit of a think if I can think of something better. Agree it would be useful, and agree this method feels kind of a bit misplaced.

If I cant resolve it quickly, would you be fine with keeping this in this PR if I raise an issue for it to be fixed properly in a further iteration?

Results_Engine/Query/DisplayObjectResult.cs
IsakNaslundBh: Agree this needs to be expanded. Have some thinking around this, and have raised an issue here to capture my initial thoughts: https://github.com/BHoM/BHoM_Engine/issues/2772

This fundamentally comes down to a further expansion of the capabilities of MapResults filtering capabilities, and for sure is something that should be moved further. My initial plan was to do this in a later PR, but if you think it is better done for this PR, I can add the functionality.

Results_Engine/Query/MapResults.cs
IsakNaslundBh: see answer above 😄

IsakNaslundBh: Good comments, and interesting take as usual! :)

My personal preference would be to keep the objects as the first input, as they are the sort of controlling parameter. The outer list will for example be governed  by the objects input rather than the result input. Also, this method will discard all results that do not map to any of the provided objects. For example, inputting a single bar and results for 1000 bars this method will basically just filter the results down.

I have looked at it as "for _this_ set of objects, give me the corresponding results".

> Personally I would prefer an even more explicit solution, with swapped inputs:
MapResultsToObjects(this IEnumerable<TResult> results, IEnumerable<TObject> objects, ...).

Not against updating the method name though, cause I do agree that that is more explicit than the current name, and captures what the methods does even more, but would personally keep the input parameter order.

Results_Engine/Query/MapResults.cs
IsakNaslundBh: We do have GroupByObjectId already implemented, and GroupByProperty in the Base_Engine already handles the fallback case. Happy to add GroupByNodeId and GroupByMeshFaceId methods for IMeshElementResult types though!

Results_Engine/Query/ResultItemValueProperty.cs
IsakNaslundBh: Personally not that against keeping the output, as I am not sure how much use the object would be of being passed around. This way as well you do get all the information upfront if the method is called from the UI.

Lets have a quick call about it!

Results_Engine/Query/ResultPropertyKeys.cs
IsakNaslundBh: That is probably true. Will aim to fix that.

Results_Engine/Query/DisplayObjectResult.cs
IsakNaslundBh: Actually, after another think, I think this might be the PR to make this change, to avoid having to change the interface of the method again. Will get a first version in, that might not be perfect and cover all cases yet, but work as a start that can be expanded upon in a more isolated fashion.

Results_Engine/Query/MapResults.cs
alelom: Thanks I get your point, if you agree with just the renamign to `MapResultsToObjects` keeping the input order as is now and if that is not too much work then I would say let's go for it, otherwise no worries

Graphics_Engine/Modify/ApplyGradientOptions.cs
alelom: After our conversation, my understanding is that the proper solution would be:
- `defaultGradient` should be a default value of `GradientOptions` object
- the method should accept a null `GradientOptions` object or one with a null `DefaultGradient` property, and instantiate it if it is (therefore getting the default value populated)
- The first point is difficult to achieve due to the current mechanism of where the default values are pulled from (datasets)

I am not sure I summarised this correctly 😄 
Anyway, because this is not important, nothing needs to be done here in this PR.
I leave it to you if we want to raise an issue with the points above to be solved in the future or not.

Results_Engine/Query/DisplayElement1DResults.cs
alelom: To summarise in our discussion, this is more about having public methods to get a ILookup of results as you are doing here

IsakNaslundBh:
@BHoMBot check compliance

IsakNaslundBh:
@BHoMBot check null-handling

IsakNaslundBh:
@BHoMBot check compliance

IsakNaslundBh:
@BHoMBot check versioning

Graphics_Engine/Modify/ApplyGradientOptions.cs
IsakNaslundBh: Will leave it as it is for now!

IsakNaslundBh:
@BHoMBot check compliance
@BHoMBot check installer

Results_Engine/Query/ResultPropertyKeys.cs
IsakNaslundBh: WIll keep as is for now to keep IResultCollection as priority over the IResultItem in case a result implements both interfaces

Results_Engine/Query/ResultItemValueProperties.cs
IsakNaslundBh: a discussed in call, will keep as is for now as no real use outside passing code around for the additional object.

Results_Engine/Query/DisplayElement1DResults.cs
IsakNaslundBh: Done

Results_Engine/Query/ResultItemValueProperty.cs
IsakNaslundBh: a discussed in call, will keep as is for now as no real use outside passing code around for the additional object.

BHoM_Engine/Create/Type/EngineType.cs
```diff
@@ -0,0 +1,84 @@
...
+        /***************************************************/
+
+        [Description("Creates an Engine type that matches the given name.")]
+        [Input("name", "Name to be searched for among all Engine types.")]
+        [Input("silent", "If true, the error about no type found will be suppressed, otherwise it will be raised.")]
+        [Output("type", "BHoM Engine type that matches the given name.")]
+        [PreviousVersion("5.1", "BH.Engine.Reflection.Create.EngineType(System.String, System.Boolean)")]
+        public static Type EngineType(string name, bool silent = false)
+        {
+            if (name == null)
```
alelom, line 45:
```suggestion
            if (String.IsNullOrWhiteSpace(name))
```

BHoM_Engine/Create/Type/EngineType.cs
```diff
@@ -0,0 +1,84 @@
...
+                {
+                    if (types.Count == 0)
+                        Compute.RecordError($"A type corresponding to {name} cannot be found.");
+                    else
+                    {
+                        string message = "Ambiguous match: Multiple types correspond the the name provided: \n";
+                        foreach (Type t in types)
+                            message += "- " + t.FullName + "\n";
+
+                        message += "To get a Engine type from a specific Assembly, try adding ', NameOfTheAssmebly' at the end of the name string, or use the AllEngineTypes method to retreive all the types.";
```
alelom, line 71:
```suggestion
                        message += "To get a Engine type from a specific Assembly, try adding ', NameOfTheAssmebly' at the end of the name string, or use the AllEngineTypes method to retrieve all the types.";
```

BHoM_Engine/Create/Type/GenericType.cs
```diff
@@ -0,0 +1,111 @@
...
+        /***************************************************/
+
+        [Description("Creates a generic BHoM type that matches the given name.")]
+        [Input("name", "Name to be searched for among all BHoM generic types.")]
+        [Input("silent", "If true, the error about no type found will be suppressed, otherwise it will be raised.")]
+        [Output("type", "BHoM generic type that matches the given name.")]
+        [PreviousVersion("5.1", "BH.Engine.Reflection.Create.GenericType(System.String, System.Boolean)")]
+        public static Type GenericType(string name, bool silent = false)
+        {
+            if (name == null)
```
alelom, line 45:
```suggestion
            if (String.IsNullOrWhiteSpace(name))
```

Results_Engine/Query/ObjectIdentifier.cs
```diff
@@ -0,0 +1,125 @@
...
+
+        [Description("Gets the AdapterIdName from an object.")]
+        [Output("adapterIdName ", "The AdapterIdName of the specified identifier.")]
+        private static string IdMatch(this IBHoMObject o, Type identifier)
+        {
+            IFragment id;
+            if (!o.Fragments.TryGetValue(identifier, out id))
+                return "";
+
+            return ((IAdapterId)id).Id.ToString();
```
alelom, line 119:
```suggestion
            return ((IAdapterId)id)?.Id?.ToString();
```

alelom (changes requested):
Works for me. Minor changes requested, otherwise happy to approve

IsakNaslundBh:
@BHoMBot check compliance

IsakNaslundBh:
@BHoMBot check ready-to-merge

IsakNaslundBh:
@BHoMBot check ready-to-merge

PR #2819, opened 2022-05-11
Structure_Engine: Add metal deck surface property methods

JosefTaylor:
### NOTE: Depends on 

[BHoM PR#1386](https://github.com/BHoM/BHoM/pull/1388)
   
Closes #

### DIFF
--- a/Structure_Engine/Query/VolumePerArea.cs
+++ b/Structure_Engine/Query/VolumePerArea.cs
@@ -165,7 +165,7 @@ namespace BH.Engine.Structure
         /**** Public Methods - Interfaces               ****/
         /***************************************************/
 
-        [PreviousVersion("5.2","BH.Engine.Structure.Query.IAverageThickness(BH.oM.Structure.SurfaceProperties.ISurfaceProperty")]
+        [PreviousVersion("5.2","BH.Engine.Structure.Query.IAverageThickness(BH.oM.Structure.SurfaceProperties.ISurfaceProperty)")]
         [Description("Gets the average thickness of the property for the purpose of calculating solid volume.")]
         [Input("property", "The property to evaluate the average thickness of.")]
         [Output("volumePerArea", "The average thickness of the property for the purpose of calculating solid volume.", typeof(Length))]


IsakNaslundBh (changes requested):
Generally looks good.

Some changes are needed before this is merged:

- SolidVolume calculation need to be ensured to be correct for Panels with any of these types. It might be the case that the current solution of using AverageThickness is not enough any longer, ad it might need to be rewritten.
- MaterialComposition for IAreaElement need to be retwritten now that we have SurfaceProperties with more than one material.

Think both these two needs to be handled before this is merged.

JosefTaylor:
@IsakNaslundBh checking MaterialComposition and Mass(IElementM) these are both issues which also affect the Layered property. I'm going to have a try at re-writing those so they take account that these include multiple materials.

JosefTaylor:
Since this touched the method Geometry3D(Panel) method by renaming AverageThickness, I notice it's using the wrong thickness method (or, it's making an assumption about what it is being used for that may not be accurate, and is not communicated to the user). In the future, I would love for this to properly extrude ribbed, waffle, and slabondeck properties, but for now maybe it should just extrude TotalThickness and tell the user the geometry is not suitable for volume takeoff.

JosefTaylor:

IsakNaslundBh:
@BHoMBot check compliance

JosefTaylor:
@BHoMBot check compliance

JosefTaylor:
@BHoMBot check required

IsakNaslundBh:
@BHoMBot check versioning

IsakNaslundBh (dismissed):
Happy with this now. Good changes, and general improvements for the methods.

Happy for this to be merged as soon as all bot checks are passing

FraserGreenroyd:
@BHoMBot check versioning

FraserGreenroyd:
@BHoMBot check required

IsakNaslundBh:

IsakNaslundBh:
@BHoMBot check ready-to-merge

PR #2824, opened 2022-05-12
Serialiser_Engine: added null check in MethodBaseSerialiser

### LINKED ISSUE #2823: Serialiser_Engine: missing null check in MethodBaseSerialiser, throws exception on startup
#### Description

Just an annoyance, but it's slowing my debug of certain workflows, and leads to unnecessary reports of exceptions. It's an easy fix.

Here: https://github.com/BHoM/BHoM_Engine/blob/ae5f832e75f8f8a4d4f3b994e6253c4485a5e829/Serialiser_Engine/Objects/BsonSerializers/MethodBaseSerializer.cs#L169-L172

when `types` is a non-null `List<Type>` that contains only `null` items, the `types.ToArray()` method produces a `null` result.
Then, the method call `GetMethod()` or `GetConstructor()` do not admit `null` input for that parameter.

image

#### Steps to reproduce

It's easy to reproduce in a few lines of code:
```cs
List<Type> types = new List<Type>() { null };
Type[] typeArray = types.ToArray();

typeof(string).GetMethod("Count", typeArray);
```

The last line will throw an exception because we do not check if typeArray is null.

#### Expected behaviour:

We should first create `typeArray` and check if it is null.

#### Test file(s):

Suggest to reproduce with the lines of code above. On my machine, it happens due to some particular DLLs that are loaded at startup.

alelom:

Closes #2823

Added a null check in [MethodBaseSerializer](https://github.com/BHoM/BHoM_Engine/blob/ae5f832e75f8f8a4d4f3b994e6253c4485a5e829/Serialiser_Engine/Objects/BsonSerializers/MethodBaseSerializer.cs) to avoid it throwing an exception during startup (plugin loading) as explained in #2823.

### Test files

See #2823. I can provide an assembly that throws the error if deemed necessary.

### Changelog

- Added a null check in [MethodBaseSerializer](https://github.com/BHoM/BHoM_Engine/blob/ae5f832e75f8f8a4d4f3b994e6253c4485a5e829/Serialiser_Engine/Objects/BsonSerializers/MethodBaseSerializer.cs).


Serialiser_Engine/Objects/BsonSerializers/MethodBaseSerializer.cs
```diff
@@ -166,10 +166,12 @@ private MethodBase GetMethod(string methodName, string typeName, List<string> pa
...
-                            if (methodName == ".ctor")
-                                method = type.GetConstructor(types.ToArray());
-                            else
-                                method = type.GetMethod(methodName, types.ToArray());
+                            Type[] typesArray = types.ToArray();
+                            if (typesArray != null)
+                                if (methodName == ".ctor")
+                                    method = type.GetConstructor(typesArray);
+                                else
+                                    method = type.GetMethod(methodName, typesArray);
```
adecler, line 174: @alelom , can you add curly braces around the body of the `if` (line 170) for clarity ?

FraserGreenroyd:
@BHoMBot check versioning
@BHoMBot check compliance

alelom:
Because BHoMBot is fragile, I will close this PR and reopen with a new one *wink wink*

alelom:
Replaced with https://github.com/BHoM/BHoM_Engine/pull/2963

PR #2830, opened 2022-05-18
Structure_Engine: Add check if bar contains reinforcement

### LINKED ISSUE #2829: Structure_Engine: Bars with concrete sections with no BarRebarIntent set can not be turned to physical Framing
#### Description:

Create method crashes if Bar ahs concrete section set with null BarRebarIntent.

#### Steps to reproduce:

Call Create.FrameingElement accepting a bar, where the bar has a concrete section with null BarRebarIntent

#### Expected behaviour:

For the method to properly nullcheck the BarRebarIntet and not crash if it is null, but simple assume there is not reinforcement set on the section.

IsakNaslundBh:

Closes #2829

- Adding method for checking if a Bar has any reinforcement applied
- Calling the checking method when creating Bars from Framing

### Test files

https://burohappold.sharepoint.com/:f:/s/BHoM/EmAT8FrB1wRGkG6PmYkSWhIBfZWrqUF2nrGNlt6IHeZ8nA?e=RKU5Tp

### DIFF
--- a/Structure_Engine/Query/HasReinforcement.cs
+++ b/Structure_Engine/Query/HasReinforcement.cs
@@ -53,8 +53,8 @@ namespace BH.Engine.Structure
         /***************************************************/
 
         [Description("Returns true if the ConcreteSection has BarRebarIntent defined with at least one IBarReinforcement in it.  False if the ConcreteSection or BarRebarIntent is null or the IBarReinforcement count is zero.")]
-        [Input("bar", "The ConcreteSection to check for reinforcement.")]
-        [Output("hasReinforcement", "Returns true if the Bar has reinforcement assigned, false otherwise.")]
+        [Input("section", "The ConcreteSection to check for reinforcement.")]
+        [Output("hasReinforcement", "Returns true if the ConcreteSection has reinforcement assigned, false otherwise.")]
         public static bool HasReinforcement(this ConcreteSection section)
         {
             if (section == null)
--- a/Structure_Engine/Query/HasReinforcement.cs
+++ b/Structure_Engine/Query/HasReinforcement.cs
@@ -1,0 +1,74 @@
+﻿/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2022, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 
+ * GNU Lesser General Public License for more details.                          
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using BH.oM.Base;
+using BH.oM.Base.Attributes;
+using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Linq;
+using BH.oM.Structure.Elements;
+using BH.oM.Structure.Reinforcement;
+using BH.oM.Structure.SectionProperties;
+
+namespace BH.Engine.Structure
+{
+    public static partial class Query
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Returns true if the Bar has a concrete section with reinforcement intent defined with at least one IBarReinforcement in it. False if not, or if the Bar, ConcreteSection or BarRebarIntent is null.")]
+        [Input("bar", "The Bar to check for reinforcement.")]
+        [Output("hasReinforcement", "Returns true if the Bar has reinforcement assigned, false otherwise.")]
+        public static bool HasReinforcement(this Bar bar)
+        {
+            if (bar == null)
+                return false;
+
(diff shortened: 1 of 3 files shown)


Structure_Engine/Query/HasReinforcement.cs
```diff
@@ -0,0 +1,74 @@
...
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Returns true if the Bar has a concrete section with reinforcement intent defined with at least one IBarReinforcement in it. False if not, or if the Bar, ConcreteSection or BarRebarIntent is null.")]
+        [Input("bar", "The Bar to check for reinforcement.")]
+        [Output("hasReinforcement", "Returns true if the Bar has reinforcement assigned, false otherwise.")]
+        public static bool HasReinforcement(this Bar bar)
+        {
+            if (bar == null)
+                return false;
```
peterjamesnugent, line 47:
```suggestion
            if (bar.IsNull())
                return false;
```

Structure_Engine/Query/ReinforcementLayout.cs
```diff
@@ -333,7 +333,7 @@ private static bool CheckSectionAndExtractParameters(this ConcreteSection sectio
             longReif = new List<LongitudinalReinforcement>();
             tranReif = new List<TransverseReinforcement>();
 
-            if (section == null)
+            if (section == null || section.RebarIntent == null)
```
peterjamesnugent, line 336:
```suggestion
            if (section.IsNull() || section.RebarIntent.IsNull())
```

Structure_Engine/Query/HasReinforcement.cs
```diff
@@ -0,0 +1,74 @@
...
+    public static partial class Query
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Returns true if the Bar has a concrete section with reinforcement intent defined with at least one IBarReinforcement in it. False if not, or if the Bar, ConcreteSection or BarRebarIntent is null.")]
+        [Input("bar", "The Bar to check for reinforcement.")]
+        [Output("hasReinforcement", "Returns true if the Bar has reinforcement assigned, false otherwise.")]
+        public static bool HasReinforcement(this Bar bar)
```
peterjamesnugent, line 44:
```suggestion
        public static bool ExistsReinforcement(this Bar bar)
```

I wonder if this should be as suggested? Do we have a naming convention for this?

Structure_Engine/Query/HasReinforcement.cs
```diff
@@ -0,0 +1,74 @@
...
+
+namespace BH.Engine.Structure
+{
+    public static partial class Query
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Returns true if the Bar has a concrete section with reinforcement intent defined with at least one IBarReinforcement in it. False if not, or if the Bar, ConcreteSection or BarRebarIntent is null.")]
```
peterjamesnugent, line 41:
```suggestion
        [Description("Returns true if the Bar has a ConcreteSection with BarRebarIntent defined with at least one IBarReinforcement in it. False if the Bar, ConcreteSection or BarRebarIntent is null or the IBarReinforcement count is zero.")]
```

peterjamesnugent (changes requested):
Works fine for me, no errors even when chucking a `null` around in to some of the objects.

I have added the `IsNull()` checks around, may as well make a use of them and suggested changing the name of the method (but happy if it stays the same).

Structure_Engine/Query/HasReinforcement.cs
IsakNaslundBh: Actually, intentionally did not use those methods for this case. 

The reason for adding this method, was because the CreateFraming from a bar is crashing for concrete sections with default BarRebarIntent (null). What this method essentially want to check is if this is something you can extract rebars from (all non-null and some reinforcement actually defined) without throwing any errors that will just confuse things.

I could maybe agree using the IsNull for the bar itself, but everything in the ConcreteSection method is very intentionally not doing this (even for the section itself, as casting any other section to a ConcreteSection will result in a null, hence this can also be a valid null -> false case, which should not prompt any errors).

I think the IsNull methods are great for when things need to exit the method because of the null, and for those cases we ofc want to raise errors, but I do not see these methods as that type.

Does the above make sense?

Structure_Engine/Query/HasReinforcement.cs
IsakNaslundBh: We have a couple of other `HasXxxxxxx` methods in structure engine, and there are a few more sprinkled around. Also feels like a good couple with the `IsXxxxxx` methods that we have a significant amount of. But you are a native speaker, so happy to be voted down 🛩️ 😄

peterjamesnugent: If there's other examples dotted around happy to leave as is.

peterjamesnugent (dismissed):
LGTM.

IsakNaslundBh:
@BHoMBot check compliance

peterjamesnugent (approved):
LGTM.

IsakNaslundBh:
@BHoMBot check compliance
@BHoMBot check required
@BHoMBot check ready-to-merge

JosefTaylor (approved):
looks good!
Test script runs fine.
I notice that a default material for rebars is created, B500B, which is a european standard? Seems like if there's a default it should simply be the BHoM default steel material, rather than a region specific material. This appears to happen inside the BottomReinforcement method, so I'll raise it over there.

IsakNaslundBh:
@BHoMBot check compliance
@BHoMBot check required
@BHoMBot check ready-to-merge

IsakNaslundBh:
> I notice that a default material for rebars is created, B500B, which is a european standard? Seems like if there's a default it should simply be the BHoM default steel material, rather than a region specific material. This appears to happen inside the BottomReinforcement method, so I'll raise it over there.

I think it is intentionally picking a standard Reinforcing steel material, and, yeah B500B is one of the most common steels for reinforcement in the european region.

The standard steel that is picked is currently also a EU steel (S355), same with concrete (C30/37) when for example Steel/Concrete sections without providing any material.

Think the way it works, picking from a dataset, is worth keeping, but maybe now is the time to add some localisation setting somewhere, that allows users to set their localisation to for example "US", which then could be picked up by the DefaultMaterial method and simply pick from a different Dataset.

PR #2833, opened 2022-05-18
Data_Engine: Add Round method from TAS_Toolkit

robinflyman:
Closes [#308](https://github.com/BHoM/TAS_Toolkit/issues/308)

Moved Round method from TAS_Toolkit to Data_Engine.

### DIFF
--- a/Data_Engine/Modify/Round.cs
+++ b/Data_Engine/Modify/Round.cs
@@ -1,0 +1,51 @@
+﻿/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2022, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 
+ * GNU Lesser General Public License for more details.                          
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using System;
+using System.ComponentModel;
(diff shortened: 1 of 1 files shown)


Data_Engine/Modify/Round.cs
```diff
@@ -0,0 +1,51 @@
...
+        /**** Public Methods                            ****/
+        /***************************************************/
+        [PreviousVersion("5.2", "BH.Engine.Adapters.TAS.Modify.Round(System.Single, System.Int32)")]
+        [PreviousVersion("5.2", "BH.Engine.Adapters.TAS.Modify.Round(System.Double, System.Int32)")]
+
+        [Description("Returns a double number rounded to the specified number of decimal places")]
+        [Input("number", "A double to round")]
+        [Input("decimals", "The number of decimals to round to - default 3 decimal places (e.g. round 3.14159 to 3.141")]
+        [Output("number", "The double number rounded to the specified number of decimal places")]
+        public static double Round(this double number, int decimals = 3)
```
IsakNaslundBh, line 44: Just as a question - Do we need this method? From the code side, this should never be needed as it is just calling a common Math function, and all UIs we currently support has this functionality built in.

jamesramsden-bh: GH does not have a native round function (at least one that functions like this PR) so it would be useful to continue to support a Round function in BHoM. 

Agreed that default dp should be 0, not 3.

IsakNaslundBh: After offline conversation with @FraserGreenroyd , I understand this method is used quite extensively in scripts. For that reason, I think it is probably fine to keep it, but I would change the default of the decimal places (for the reasons stated above).

Would either:
1. Change the default to 0
2. Not provide a default at all

Think the 2nd option might be the safer route. Might force people to plug in some wires in existing scripts, but is probably the safer route

IsakNaslundBh: @jamesramsden-bh , realised I had forgotten to post, and had not seen your comment. Seems we are in agreement then :)

FraserGreenroyd, line 44:
```suggestion
        public static double Round(this double number, int decimals = 0)
```

FraserGreenroyd (changes requested):

FraserGreenroyd:
@BHoMBot check versioning
@BHoMBot check installer

jamesramsden-bh (approved):
Ok, moved over from TAS TK

FraserGreenroyd:
@BHoMBot check required

FraserGreenroyd:
@BHoMBot check ready-to-merge

FraserGreenroyd:
@BHoMBot this is a DevOps instruction. I am authorising dispensation to be granted on check ref. 6653123614

FraserGreenroyd (approved):

FraserGreenroyd:
@BHoMBot check ready-to-merge

PR #2835, opened 2022-05-19
Geometry_Engine: Adding methods required by IElement2D for PlanarSurface

### LINKED ISSUE #2812: Geometry_Engine: Add required methods for IElement2D for planar surface
#### Description:

Add support for methods required for the PlanarSurface to be an IElement2D, as highlighted in https://github.com/BHoM/BHoM/issues/426 .

Strategy to handle `InternalElements2D` is to create new temporary PlanarSurfaces that have ExternalBoundaries corresponding to the InternalBoundaries of the host PlanarSurface.

IsakNaslundBh:
### NOTE: Depends on 

https://github.com/BHoM/BHoM/pull/1390
   
Closes #2812

Adding methods required for IElement2D.

For planar surfaces, the InternalElement2Ds will be returned as a set of new PLanarSurfaces, with ExternalBoundaries correpsonding to the InternalBoundaries of the host PlanarSurface. This is to fullfill the requirement that the openings must also be IElement2D, which is not the case for the closed curves. This workaround alows for the methods to behave as expected.

Potential at one point to add some form of Region class, that is like the PlanarSurface, but without any openings. There is no practical use for that class though, that can not be achieved with a PlanarSurface with no openings, more than clarity for methods like this. For now, I think this is the best solution forward, even it it at first might look slightly unintuitive.

### Test files

https://burohappold.sharepoint.com/:f:/s/BHoM/EgJ3TMjsnNFBkM_sPjRKDewBcjco6-8OogC7UcSvS3lMKw?e=g8munn

### Additional comments

Reference to the wikipage outlining methods required by the IElement framework:

https://github.com/BHoM/documentation/wiki/IElement-required-extension-methods

### DIFF
--- a/Geometry_Engine/Create/IElement2D/NewInternalElement2D.cs
+++ b/Geometry_Engine/Create/IElement2D/NewInternalElement2D.cs
@@ -1,0 +1,53 @@
+/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2022, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 
+ * GNU Lesser General Public License for more details.                          
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using System;
+using BH.oM.Dimensional;
+using BH.oM.Geometry;
+
+using BH.oM.Base.Attributes;
+using System.ComponentModel;
+
+
+namespace BH.Engine.Geometry
+{
+    public static partial class Create
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Creates a new Element2D, appropriate to the input type. For this case the appropriate type for the PlanarSurface will be a new PlanarSurface with a null ExternalBoundary and empty list of InternalBoundaries. \n" +
+                     "Method required for any IElement2D that contians internal IElement2Ds.")]
+        [Input("surface", "PlanarSurface just used to determine the appropriate type of IElement2D to create.")]
+        [Output("opening", "The created Opening as a IElement2D.")]
+        public static IElement2D NewInternalElement2D(this PlanarSurface surface)
+        {
+            return new PlanarSurface(null, new System.Collections.Generic.List<ICurve>());
+        }
+
+        /***************************************************/
+    }
+}
+
+
+
--- a/Geometry_Engine/Create/IElement2D/NewInternalElement2D.cs
+++ b/Geometry_Engine/Create/IElement2D/NewInternalElement2D.cs
@@ -36,7 +36,7 @@ namespace BH.Engine.Geometry
         /**** Public Methods                            ****/
         /***************************************************/
 
-        [Description("Creates a new Element2D, appropriate to the input type. For this case the appropriate type for the PlanarSurface will be a new PlanarSurface with a null ExternalBoundary and empty list of InternalBoundaries. \n" +
+        [Description("Creates a a new PlanarSurface that can be used as internal element for another PlanarSurface. The resulting PlanarSurface will have a null ExternalBoundary and empty list of InternalBoundaries. The resulting PlanarSurface is casted to IElement2D for generalization purposes.  \n" +
                      "Method required for any IElement2D that contians internal IElement2Ds.")]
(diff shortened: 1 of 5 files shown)


IsakNaslundBh:
@BHoMBot check compliance

Geometry_Engine/Create/IElement2D/NewInternalElement2D.cs
```diff
@@ -0,0 +1,53 @@
...
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Creates a new Element2D, appropriate to the input type. For this case the appropriate type for the PlanarSurface will be a new PlanarSurface with a null ExternalBoundary and empty list of InternalBoundaries. \n" +
+                     "Method required for any IElement2D that contians internal IElement2Ds.")]
+        [Input("surface", "PlanarSurface just used to determine the appropriate type of IElement2D to create.")]
+        [Output("opening", "The created Opening as a IElement2D.")]
+        public static IElement2D NewInternalElement2D(this PlanarSurface surface)
```
alelom, line 43: The prefix `New` seems redundant in a method within the class `Create`.
```suggestion
        public static IElement2D InternalElement2D(this PlanarSurface surface)
```

alelom: More importantly, I really don't like that the input input `surface` is never used, or required by the method in any way, as the only requested data is the Type of the class. I would prefer that either: 
1. The method took a `Type` as input, or that 
2. The mapping between `typeof(PlanarSurface)` and a new instance of a PlanarSurface to be used as internal element was achieved in another way, e.g. via a static property somewhere
   ```cs
    public static class InternalElementMapping
    {
	    public static Dictionary<Type, IElement2D> Mapping { get; } = new Dictionary<Type, IElement2D>()
	    {
		    {typeof(PlanarSurface), (IElement2D)new PlanarSurface(null, new System.Collections.Generic.List<ICurve>())}
	    };
    }
   ```

Geometry_Engine/Create/IElement2D/NewInternalElement2D.cs
```diff
@@ -0,0 +1,53 @@
...
+
+namespace BH.Engine.Geometry
+{
+    public static partial class Create
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Creates a new Element2D, appropriate to the input type. For this case the appropriate type for the PlanarSurface will be a new PlanarSurface with a null ExternalBoundary and empty list of InternalBoundaries. \n" +
```
alelom, line 39: This is what I feel like the description should really be:
```suggestion
        [Description("Creates a a new PlanarSurface that can be used as internal element for another PlanarSurface. The resulting PlanarSurface will have a null ExternalBoundary and empty list of InternalBoundaries. The resulting PlanarSurface is casted to IElement2D for generalization purposes.  \n" +
```
Especially the last sentence is indicative of the problems of this approach I think - see comments below

Geometry_Engine/Create/IElement2D/NewInternalElement2D.cs
alelom: I now see that we have other cases of `NewInternalElement2D` methods - e.g. in the Analytical_Engine where we use Activator.CreateInstance to create the openings. If needed for current infrastructural reasons, my comments above can be dispensed - just leaving them there for reference as I don't particularly like the solution

alelom: For correctness, if those methods are required, I would at least add an overload for each of those methods that takes a `Type` as input, and have the other method call this new one.

alelom (changes requested):
Good in general, see comments

Geometry_Engine/Create/IElement2D/NewInternalElement2D.cs
IsakNaslundBh: Think you answered most of it yourself but will put some context here for reference.

It is basically a method that is required for some methods operating on IElement2Ds that can have cases where an opening (internal element 2d) needs to be cut in the host IElement2D. It is then required to generate a new IElement2D of a type acceptable for the host, and that is the job of this method.

The input type is to some extent used, for type switching via as dynamic/RunExtensionMethod. Calls are made as 
`IElement2d opening =  element2d.INewInternalElement2D();`

If this method was an instance method on the class (as in standard OO) with no arguments, it might have looked slightly better than what the current implementation does, but is basically doing the same thing, but in a way matching how the oM vs Engine is working.

Could also agree that the "New" is a bit redundant, alternatively, that this method should be a query method, but remember having that discussion years ago, before we landed in this solution. Happy for it to be changed, but not in this PR, as out of scope.

Also, not to sure I like the idea of changing to relying on type. That feels like it just would be a slowdown, for no real gain?

IsakNaslundBh: Have also added a link to the wiki about what methods are required by the IElement framework in the desciption to this PR. Sorry, should have done that when raising it :)

Geometry_Engine/Create/IElement2D/NewInternalElement2D.cs
IsakNaslundBh: Happy with this change, will push

alelom (dismissed):
Tested using test scripts, approving based on latest changes and discussion in comments

IsakNaslundBh:
@BHoMBot check compliance

IsakNaslundBh:
@BHoMBot check null-handling

IsakNaslundBh:
@BHoMBot check compliance

IsakNaslundBh:
@BHoMBot check required

alelom (approved):
Re-approving after null handling fixes

IsakNaslundBh:
@BHoMBot check ready-to-merge

PR #2838, opened 2022-05-20
Results_Engine: Fix FilterBy not filtering by all properties.

### LINKED ISSUE #2837: Results_ENgine: FilterResults can only filter by one proeprty type
#### Description:

Bug in the filter results class makes it so that it can only filter by one provided type, that is, either by resultcase or objectId, or timestep etc.

#### Steps to reproduce:

Try filtering by more than one property.

#### Expected behaviour:

For all filters to be applied, that is, if ResultsCases and ObejctIds is provided, both should be applied as filters.

IsakNaslundBh:

Closes #2837

Code change is hopefully quite self-evident. The method was always re-filtering the input collection, rather than stacking the filtering.

@alelom putting you as a reviewer as you been part of this part of the code before. Please let me know if you do not have capacity to review, and I will find someone else 😄 

### Test files

https://burohappold.sharepoint.com/:f:/s/BHoM/EnZ2F3XH5JFEnUylsv_fHE4BwlswbS4zG1J8a9qv-QQDjg?e=JgS9OB

### DIFF
--- a/Results_Engine/Query/FilterResults.cs
+++ b/Results_Engine/Query/FilterResults.cs
@@ -41,7 +41,8 @@ namespace BH.Engine.Results
                      "If TimeStepFilters contains items, only results of type ITimeStepResult will be returned that has a TimeStep matching any items in the TimeStepFilters.\n" +
                      "If ObjectIDFilters contains items, only results of type IObjectIdResult will be returned that has a ObjectId matching any items in the ObjectIDFilters.\n" +
                      "If NodeIDFilters contains items, only results of type IMeshElementResult will be returned that has a NodeId matching any items in the NodeIDFilters.\n" +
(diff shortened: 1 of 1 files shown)


FraserGreenroyd (changes requested):
Code reviewed, change looks sensible and makes sense.

Code changes support the issue described in #2837

However, I think the description now needs updating to match the method. The description reads that if such and such filter contains items, then only results of that type will be returned. But the way the code operates now, is that it will compound the queries as an `and` operation effectively, which is different to how the description describes it.

For example, if I have 100 results to filter, and I filter by both time step and node ID, then my returned results will be the 10 results that match both filters, of which in that scenario those results would have to be both `ITimeStepResult` and `IMeshElementResult` to get picked up.

Maybe I've misunderstood something because I've not run the test file, only read the code, but hopefully my query makes enough sense to discuss @IsakNaslundBh ?

IsakNaslundBh:
@BHoMBot check required
@BHoMBot check ready-to-merge

PR #2859, opened 2022-06-10
Environment_Engine: Created a method for SetRegionGeometry

### LINKED ISSUE #2851: Environment_Engine: New SetRegionGeometry
#### Description:

When it's necessary to manually edit the Region geometry (Perimeter), it's useful to have a quick way of matching the new, manually edited geometry back to the original Region then to set the new geometry as the Region Perimeter property value, i.e. to overwrite the previous geometry with the manually edited version.

My proposed method is to use the Location point from the original Room and use IsContaining to check whether that sits within the new, edited geometry. In GH long hand, that looks something like this: 
image

robinflyman:
Closes #2851

Created a new method for SetRegionGeometry.

### DIFF
--- a/Environment_Engine/Modify/SetRegionGeometry.cs
+++ b/Environment_Engine/Modify/SetRegionGeometry.cs
@@ -1,0 +1,57 @@
+﻿/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2022, the respective contributors. All rights reserved.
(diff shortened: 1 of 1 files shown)


robinflyman:
@BHoMBot check versioning

FraserGreenroyd:
@BHoMBot check required

FraserGreenroyd:
@BHoMBot check ready-to-merge

PR #2867, opened 2022-06-16
Environment_Engine: Hard coded acceptOnEdge value to true

### LINKED ISSUE #2866: Environment_Engine: Hard-code acceptOnEdges to true
The IsContaining method will always return false if tolerance > 0 on a bounding box with one dimension length equal to 0. Default tolerance is also > 0. Therefore, there is no need to have the Environment_Engine IsContaining check have the ability to set acceptOnEdges == false for checking between a point and a single panel.

jamesramsden-bh:
acceptOnEdge now hard-coded to true as per #2866 

   
Closes #2866

### DIFF
--- a/Environment_Engine/Query/IsContaining.cs
+++ b/Environment_Engine/Query/IsContaining.cs
@@ -71,15 +71,14 @@ namespace BH.Engine.Environment
         [Description("Defines whether an Environment Panel contains a provided point")]
         [Input("panel", "An Environment Panel to check with")]
         [Input("pt", "The point being checked to see if it is contained within the bounds of the panel")]
-        [Input("acceptOnEdges", "Decide whether to allow the point to sit on the edge of the panel, default false")]
         [Output("isContaining", "True if the point is contained within the panel, false if it is not")]
         [PreviousVersion("5.2", "BH.Engine.Environment.Query.IsContaining(BH.oM.Environment.Elements.Panel, BH.oM.Geometry.Point, System.Boolean)")]
-        public static bool IsContaining(this Panel panel, Point pt, bool acceptOnEdges = false, double tolerance = BH.oM.Geometry.Tolerance.Distance)
+        public static bool IsContaining(this Panel panel, Point pt, double tolerance = BH.oM.Geometry.Tolerance.Distance)
         {
             if (panel == null || pt == null)
                 return false;
 
-            return new List<Panel> { panel }.IsContaining(pt, acceptOnEdges, tolerance);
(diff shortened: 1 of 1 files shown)


jamesramsden-bh:
@BHoMBot check required

jamesramsden-bh:
@BHoMBot check ready-to-merge

FraserGreenroyd (changes requested):
I don't think this is how this was meant to be done - I was envisaging we would be changing line 111 to always be true instead, but allowing the optional input for users on the methods. That said, it doesn't look like that input is used in the rest of the code so maybe just removing it from both methods is acceptable and then setting it to `true` on the call to the geometry engine? Gimme a call if this makes little sense and we can go through the code without BHoMBot's annotations clouding things.

IsakNaslundBh:
> I don't think this is how this was meant to be done - I was envisaging we would be changing line 111 to always be true instead, but allowing the optional input for users on the methods. That said, it doesn't look like that input is used in the rest of the code so maybe just removing it from both methods is acceptable and then setting it to `true` on the call to the geometry engine? Gimme a call if this makes little sense and we can go through the code without BHoMBot's annotations clouding things.

Would agree on this. Would jsut hardcode the parameter being passed to IsCOntaining on line 112.

 Seems like the parameter is used on line 156 on main. There is also a call to IsContaining on line 149 for the polyline that potentially the parameter could be passed to as well, where I assume it should be doing the same check that is being done on line 156-168 (not sure on this though, would need to be checked)

IsakNaslundBh:
> This PR is being closed. After a brief investigation with @FraserGreenroyd , the IsContaining method was working with the 5.1 beta, and the only changes to the Environment_Engine in the meantime have been the addition of a tolerance, which don't seem to have affected the behaviour of the method within the Environment_Engine. A possibility is that there have been other changes elsewhere, e.g. in Geometry_Engine, which have had unintended consequences here. On that basis, we don't want to make changes to Environment_Engine as a workaround without first understanding what the changes elsewhere might be. @FraserGreenroyd is to investigate further.

Would diagree with this. The IsContaining method for Panels _was_ changed beyond the scope of of passing tolerance:

https://github.com/BHoM/BHoM_Engine/commit/d902bd8eee71cc195c626c5a94dc7552631b72c8

This commit added passing the `acceptOnEdge` boolean to the `IsContaining` method for boundingboxes, that was previously just using the default value (which is set to true). This to me looks like the clear cause of the change of the behaviour. The Geometry_Engine method in question (`IsContaining` for `BoundingBox`) has not had a change in behaviour in ages. Only change this milestone was added descriptions.

PR #2883, opened 2022-07-08
Structure_Engine: Add wall(panel) and floor(panel) and SurfaceElement(panel)

### LINKED ISSUE #2881: Physical_Engine: Floor(Panel) and Wall(Panel) to match IFramingElement(Bar, Usage)

### LINKED ISSUE #2882: Physical.Create.Wall(Construction, ExternalEdges, InternalEdges) ignores InternalEdges.

JosefTaylor:

Closes #2881 
Closes #2882   

### Test files

https://burohappold.sharepoint.com/:f:/r/sites/BHoM/02_Current/12_Scripts/02_Pull%20Request/BHoM/BHoM_Engine/Structure_Engine/%232883-AddSurfaceElementFromPanel?csf=1&web=1&e=cL8u2i

### Changelog

- Fixed bug in Physical.Create.Wall which ignored internal edges
- Added Wall(Panel), Floor(Panel) and ISurface(Panel, usage) to convert panels to physical surface elements
- Added IConstruction(ISurfaceProperty) to convert surface properties into Physical IConstructions. Convert to maintain total depth and material quantities, but may not accurately reflect geometry or structural properties of the SurfaceProperty. Ribbed properties will contain a void layer to reflect the missing concrete. Thus, it is not a 1:1 translation, until Physical IConstructions can represent the full range of objects represented by ISurfaceProperties.
- Added IVolumePerArea(Construction) matching the IVolumePerArea(SurfaceProperty) which gets the effective thickness of solid materials (important for comparison to MaterialComposition in material takeoff)
- Implemented IVolumePerArea() in SolidVolume() for ISurface and IOpening


JosefTaylor:
@BHoMBot check compliance

JosefTaylor:
@BHoMBot check compliance

JosefTaylor:
@BHoMBot check compliance

JosefTaylor:
@BHoMBot check required

JosefTaylor:
@BHoMBot check required

JosefTaylor:
@BHoMBot check null-handling

JosefTaylor:
@BHoMBot check null-handling

JosefTaylor:
@BHoMBot check null-handling

JosefTaylor:
@BHoMBot check compliance

JosefTaylor:
@BHoMBot fix documentation punctuation ref. 7567525610

JosefTaylor:
@BHoMBot check compliance

JosefTaylor:
@BHoMBot check required

Structure_Engine/Create/Physical/Floor.cs
```diff
@@ -0,0 +1,73 @@
...
+                construction = panel.Property.IConstruction();
+            //Get Location
+            PolyCurve externalEdges = Geometry.Create.PolyCurve(panel.ExternalEdges.Select(x => x.Curve));
+            List<PolyCurve> internalEdges = panel.Openings.Select(opening => Geometry.Create.PolyCurve(opening.Edges.Select(edge => edge.Curve))).ToList();
+            BHPE.Floor surfaceElement = Physical.Create.Floor(construction, externalEdges, internalEdges);
+            string name = panel.Name ?? "";
+            surfaceElement.Name = name;
+            if (panel.FindFragment<PanelRebarIntent>() != null || panel.Property.FindFragment<ReinforcementDensity>() != null)
+            {
+                Base.Compute.RecordWarning("The panel has reinforcement, but this is not yet implemented.");
```
pawelbaran, line 66:
```suggestion
                Base.Compute.RecordWarning("The panel has reinforcement, but embedding this information in the physical element is not yet implemented.");
```

Structure_Engine/Create/Physical/ISurface.cs
```diff
@@ -0,0 +1,75 @@
...
+            if (panel.IsNull())
+                return null;
+            if (structuralUsage == StructuralUsage2D.Undefined)
+            {
+                object result = panel.Property.PropertyValue("PanelType");
+                if (result != null)
+                {
+                    structuralUsage = (StructuralUsage2D)result;
+                }
+            }
```
pawelbaran, line 59: This looks like asking for a crash:
- if `panel.Property == null`
- if `PanelType` property has some unexpected value

I would consider adding more checks

Structure_Engine/Create/Physical/ISurface.cs
```diff
@@ -0,0 +1,75 @@
...
+            switch (structuralUsage)
+            {
+                case StructuralUsage2D.Wall:
+                    return Wall(panel);
+                case StructuralUsage2D.DropPanel:
+                case StructuralUsage2D.PileCap:
+                case StructuralUsage2D.Slab:
+                case StructuralUsage2D.Undefined:
+                default:
+                    return Floor(panel);
```
pawelbaran, line 70: I would rather record error and return null in those cases, to avoid confusion of someone setting structural usage of all elements to undefined. Alternatively, You could do the check if the panel is horizontal/vertical, then return floor/wall, and error if the panel is sloped.

Structure_Engine/Create/Physical/Wall.cs
```diff
@@ -0,0 +1,73 @@
...
+                construction = panel.Property.IConstruction();
+            //Get Location
+            PolyCurve externalEdges = Geometry.Create.PolyCurve(panel.ExternalEdges.Select(x => x.Curve));
+            List<PolyCurve> internalEdges = panel.Openings.Select(opening => Geometry.Create.PolyCurve(opening.Edges.Select(edge => edge.Curve))).ToList();
+            BHPE.Wall surfaceElement = Physical.Create.Wall(construction, externalEdges, internalEdges);
+            string name = panel.Name ?? "";
+            surfaceElement.Name = name;
+            if (panel.FindFragment<PanelRebarIntent>() != null || panel.Property.FindFragment<ReinforcementDensity>() != null)
+            {
+                Base.Compute.RecordWarning("The panel has reinforcement, but this is not yet implemented.");
```
pawelbaran, line 66: Same as above, would specify the warning a bit more

Structure_Engine/Create/Physical/Wall.cs
```diff
@@ -0,0 +1,73 @@
...
+{
+    public static partial class Create
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+        [Description("Creates a physical Wall from a Panel. The Wall will be assigned a Construction based on the SurfaceProperty of the Panel.")]
+        [Input("panel", "The Panel to use as the base for the framing element.")]
+        [Output("wall", "The created physical Wall based on the Panel element provided.")]
+        public static BHPE.Wall Wall(Panel panel)
```
pawelbaran, line 47: This could possibly become a generic method, taken that it is exactly the same as the Floor creation one? But not 200% sure, so feel free to ignore

Structure_Engine/Query/Construction.cs
```diff
@@ -0,0 +1,111 @@
...
+        [Output("construction", "The physical Construction to be used with ISurface such as Walls and Floors.")]
+        public static Construction Construction(this ISurfaceProperty surfaceProperty)
+        {
+            if (surfaceProperty.IsNull())
+                return null;
+            MaterialComposition comp = surfaceProperty.IMaterialComposition();
+            double volume = surfaceProperty.IVolumePerArea();
+            double thickness = surfaceProperty.ITotalThickness();
+            List<oM.Physical.Constructions.Layer> layers = new List<oM.Physical.Constructions.Layer>();
+            for (int i = 0; i < comp.Materials.Count(); i++)
```
pawelbaran, line 64: Do you know whether the panel layers are ordered by any means? Can imagine insulation landing at the core of the wall 😃

Structure_Engine/Query/Construction.cs
```diff
@@ -0,0 +1,111 @@
...
+                layers.Add(new oM.Physical.Constructions.Layer()
+                {Material = comp.Materials[i], Thickness = volume * comp.Ratios[i], Name = comp.Materials[i].Name});
+            if (volume < thickness)
```
pawelbaran, line 67: I would add a warning here - this is definitely a smell good to be aware of

Structure_Engine/Query/Construction.cs
```diff
@@ -0,0 +1,111 @@
...
+                layers.Add(new oM.Physical.Constructions.Layer()
+                {Material = null, Thickness = thickness - volume, Name = "void"});
+            return Physical.Create.Construction(surfaceProperty.Name, layers);
+        }
+
+        /***************************************************/
```
pawelbaran, line 73: Are the private methods used anywhere? Can't see a reference?

Structure_Engine/Create/Physical/ISurface.cs
pawelbaran: Structural panels do not have any info on whether they are walls or floors. So the trick would be that you can either:
- specify the surface type (wall/floor/roof etc.)
- let the algorithm decide what is what, with the caveat that anything that is not vertical or horizontal comes out as null, so you need to give it a bit more love before pushing

Not sure how many projects have only orthogonal panels, but would hope such cases could be sped up nicely.

Structure_Engine/Create/Physical/Wall.cs
pawelbaran: I meant you could try writing a method like `Create.Surface<T>(Panel panel) where T : ISurface`, and depending on type of T you create wall, floor etc.

Structure_Engine/Query/Construction.cs
pawelbaran: Right, something for empirical testing then

Physical_Engine/Create/Wall.cs
```diff
@@ -127,22 +106,21 @@ public static Wall Wall(IConstruction construction, ICurve edges, IEnumerable<IC
                 return null;
             }
 
-            PlanarSurface aPlanarSurface = Geometry.Create.PlanarSurface(edges);
+            List<ICurve> aInternalCurveList = null;
+            if (internalEdges != null && internalEdges.Count() > 0)
+                aInternalCurveList = internalEdges.ToList().ConvertAll(x => x as ICurve);
+            PlanarSurface aPlanarSurface = Geometry.Create.PlanarSurface(edges, aInternalCurveList);
```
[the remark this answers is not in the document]
JosefTaylor: this brings up an issue that neither of these methods are handling openings properly, I will try to fix.

Structure_Engine/Create/Physical/Wall.cs
JosefTaylor: Agreed

Structure_Engine/Query/Construction.cs
JosefTaylor: Yep, redundant!

Structure_Engine/Create/Physical/ISurface.cs
JosefTaylor: I've made it so that it first looks for a StructuralUsage, and if that's undefined (or null) it looks for the normal. If it's not horizontal or upright, it won't make it (until we have some other physical elements which can be sloped)

JosefTaylor (commented):
Made some changes, should resolve your comments @pawelbaran.

JosefTaylor:
@BHoMBot check required

Physical_Engine/Create/Wall.cs
```diff
@@ -87,62 +72,58 @@ public static Wall Wall(IConstruction construction, ICurve bottomEdge, double he
...
 
         /***************************************************/
-
         [Description("Creates physical wall object.")]
         [Input("construction", "Construction of the wall.")]
         [Input("edges", "External edges of the wall (Profile - planar closed curve).")]
         [Input("internalEdges", "Internal edges of wall (profile).")]
         [Output("wall", "A physical wall.")]
-        public static Wall Wall(IConstruction construction, ICurve edges, IEnumerable<ICurve> internalEdges)
+        public static Wall Wall(IConstruction construction, ICurve edges, IEnumerable<ICurve> internalEdges = null)
```
pawelbaran, line 97: why do you default `internalEdges` to `null` for walls but not for floors?

Physical_Engine/Query/SolidVolume.cs
```diff
@@ -85,15 +81,12 @@ public static double SolidVolume(this oM.Physical.Elements.ISurface surface)
 
             if (surface.Offset != Offset.Centre && !surface.Location.IIsPlanar())
                 Base.Compute.RecordWarning("The SolidVolume for non-Planar ISurfaces with offsets other than Centre is approxamite at best");
-
             double area = surface.Location.IArea();
             area -= surface.Openings.Sum(x => x.Location.IArea());
-            
-            return area * surface.Construction.IThickness();
+            return area * surface.Construction.IVolumePerArea();
```
pawelbaran, line 86: This is a good question... Depends on how we look at the panel: whether we want to get its outer volume or the actual volume of solids included, skipping voids. I do not know the answer to this question tbh

Structure_Engine/Create/Physical/ISurface.cs
```diff
@@ -0,0 +1,84 @@
...
+        [Input("structuralUsage", "The type of surface element to create. if Undefined, the type will be based on the panel's SurfaceProperty.")]
+        [Output("surfaceElement", "The created surface element based on the Panel element provided.")]
+        public static BHPE.ISurface ISurface(Panel panel, StructuralUsage2D structuralUsage = StructuralUsage2D.Undefined)
+        {
+            if (panel.IsNull())
+                return null;
+            if (structuralUsage == StructuralUsage2D.Undefined)
+            {
+                object result = panel.Property?.PropertyValue("PanelType");
+                if (result != null )
```
pawelbaran, line 57: Wouldn't that work and be safer?
```suggestion
                if (result is StructuralUsage2D)
```

Structure_Engine/Create/Physical/ISurface.cs
```diff
@@ -0,0 +1,84 @@
...
+                case StructuralUsage2D.Wall:
+                    return Wall(panel);
+                case StructuralUsage2D.DropPanel:
+                case StructuralUsage2D.PileCap:
+                case StructuralUsage2D.Slab:
+                    return Floor(panel);
+                case StructuralUsage2D.Undefined:
+                default:
+                    {
+                        if (Math.Abs(panel.Normal().DotProduct(Vector.ZAxis)) == 1)
```
pawelbaran, line 74: Please include some tolerance (refers to next line too):
```suggestion
                        if (1 - Math.Abs(panel.Normal().DotProduct(Vector.ZAxis)) <= BH.oM.Geometry.Tolerance.Angle)
```

Structure_Engine/Create/Physical/ISurface.cs
```diff
@@ -0,0 +1,84 @@
...
+                            return Floor(panel);
+                        else if (panel.Normal().DotProduct(Vector.ZAxis) == 0)
+                            return Wall(panel);
+                        else 
+                            return null;
```
pawelbaran, line 79: I would throw an error that some panels could not be identified because not horizontal/vertical, so the user is requested to enforce the type

Structure_Engine/Query/Construction.cs
pawelbaran: Actually now when I look at it, I would not involve `MaterialComposition` in this convert at all. Instead, I would rather work directly on `surfaceProperty.Layers` - it should be simpler plus voids could be maintained easily

pawelbaran (changes requested):
I had another round of review, a few new ideas and thoughts, hopefully driving the PR towards the good 👍 Yet again: the code is obfuscated by some sort of auto formatting, which 1. makes it hard to read 2. I would rather not merge because the code was formatted well before the changes. So it would be great if you could revert those changes (or just start the PR anew) before submitting the final version of the code.

JosefTaylor:
@FraserGreenroyd the bot got confused and tried to run the tests three times!

Physical_Engine/Create/Wall.cs
JosefTaylor: good catch, both should default to null.

Physical_Engine/Query/SolidVolume.cs
JosefTaylor: This aligns with the work we've already done in the Structural space, replacing AverageThickness with VolumePerArea and having a separate method TotalThickness which gets the outer extent of the panel property

This is chiefly useful for LCA work, material takeoff, and visualization, as Revit (for now) is going to ignore the material composition and just find a wall/floor type with the same name, so maintaining both volume (for material takeoff) and thickness (for display purposes) makes sense- I can't think of a better rule without implementing ribbed, waffle, etc. in the Physical space (which should eventually happen)

Structure_Engine/Create/Physical/ISurface.cs
JosefTaylor: Yes, thank you! I was looking for the right way to do this and couldn't find it, and it seemed like the null check and the default case would protect us enough, but this is clearer and probably a bit safer.

Structure_Engine/Create/Physical/ISurface.cs
JosefTaylor: Yeah, I'd considered this and left as-is, but that's the right move.

Structure_Engine/Create/Physical/ISurface.cs
JosefTaylor: agreed

Structure_Engine/Query/Construction.cs
JosefTaylor: That's a good point- using MaterialComposition simplifies a lot for the other surface types, but I had left room for specific methods for types which needed it, and Layered would benefit from maintaining the order of void layers.

JosefTaylor:
@pawelbaran There's no auto-formatting going on, perhaps I'm just too free with the carriage return. I'm not sure what you're seeing as formatting problems, but I don't see why they'd prevent merging.

JosefTaylor:
Ah, no robots involved, just me being messy. is the former (multi line) always preferred, or only when the line is long?

FraserGreenroyd:
> Ah, no robots involved, just me being messy. is the former (multi line) always preferred, or only when the line is long?

Multi-line is preferred when there's more than 1 property being assigned 😄

JosefTaylor:
@FraserGreenroyd @pawelbaran I'm noticing some more formatting that I think Pawel was looking at- it appears BHoMBot decided to remove a bunch of carriage returns and inline a bunch of property sets in [this commit](https://github.com/BHoM/BHoM_Engine/pull/2883/commits/0466db2ca84d589f9b833e0b6c4ce09aaa37b31e)

Structure_Engine/Create/Physical/Wall.cs
JosefTaylor: Tried this, and had some issues with how it came up in Grasshopper- I'd prefer to leave this to a later date, perhaps when more panel types are implemented in Physical oM

JosefTaylor:
@BHoMBot check compliance

Physical_Engine/Query/MaterialComposition.cs
```diff
@@ -195,10 +201,11 @@ private static MaterialComposition MaterialComposition(this Construction prop)
         {
             if (prop.Layers.Any(x => x.Material == null))
             {
-                Engine.Base.Compute.RecordError("The Construction MaterialComposition could not be queried as no Material has been assigned.");
-                return null;
+                Engine.Base.Compute.RecordWarning("At least one Material in a Layered surface property was null. MaterialConstruction excludes this layer, assuming it is void space.");
```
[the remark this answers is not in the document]
michaelhoehn: After testing, this method still throws an error "1. Cannot evaluate MaterialComposition because the List is empty. "  
With your proposal, you will have 2 messages instead of one to deal with.

michaelhoehn (changes requested):
I'm happy with the overall functionality. I have a couple general questions that I'd like addressed before approval. 

Additionally, please be sure to link your test script for posterity. I made a very basic one and included it in the same folder where yours exists. Please feel free to ignore it.

Structure_Engine/Create/Physical/ISurface.cs
pawelbaran: Such cases always bottom out sooner or later - been there too many times in the geometry engine to let it pass 🙈

Structure_Engine/Create/Physical/Wall.cs
pawelbaran: Sounds good to me - just in case, my idea was that we would have methods like `Create.Wall` and `Create.Floor` that would have only one line calling a private generic method `Create.ISurface<T>`. So from the UI perspective everything should stay the same. But happy for you to leave as is, actually I am not 100% sure whether my suggested change would be sustainable (thinking of minor differences in how floor or wall gets created)

Structure_Engine/Query/Construction.cs
pawelbaran: Right, I see your point now. I have only worked with constant thickness and layered surfaces, so my perspective was too narrow. However, in this case I would consider dynamic casting with the following implementations:
- dedicated, simple method for constant thickness
- dedicated method for layered
- the method you created for everything else, with a warning saying that ribs etc. are being simplified and voids can be lost

Would that work?

Physical_Engine/Query/SolidVolume.cs
pawelbaran: I would say we are dealing with a philosophical question here, which I personally cannot answer: both concepts of volume are correct, one included nested voids, the other not. So I cannot tell whether this change is right or wrong. Not sure if others can, so I would be fine with merging your proposal under the condition that it will be re-reviewed by @IsakNaslundBh as our main structural philosopher once he is back, before the Beta release. What do you think?

Physical_Engine/Query/MaterialComposition.cs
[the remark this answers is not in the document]
JosefTaylor: Hmm, the first error sounds better- this is if you feed an empty list of layers to the Layered SurfaceProperty? If you get the first error, the method should stop and not give you any more.
The second is a warning, as voids will often be included intentionally, but of course sometimes it could be unintentional.

Physical_Engine/Query/VolumePerArea.cs
```diff
@@ -0,0 +1,82 @@
...
+            return construction.Layers.Where(x => x.Material != null).Sum(x => x.Thickness);
+        }
+
+        /***************************************************/
+        /**** Private Methods                           ****/
+        /***************************************************/
+
+        private static double VolumePerArea(this object construction)
+        {
+            return 0; //Fallback method
```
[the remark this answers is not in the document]
JosefTaylor: yeah probably.

Physical_Engine/Query/SolidVolume.cs
```diff
@@ -150,8 +147,8 @@ public static double SolidVolume(this IOpening opening)
             }
 
             double area = opening.IArea();
-
             double thickness = 0;
+
             if (opening is Window)
                 thickness = (opening as Window).Construction.IThickness();
```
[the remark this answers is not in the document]
JosefTaylor: I hadn't done so as I'm not sure about the other voids, but if that's the desired behavior I'm happy to!

Structure_Engine/Create/Physical/Floor.cs
```diff
@@ -0,0 +1,83 @@
...
+ */
+
+using System.Collections.Generic;
+using System.Linq;
+using BH.oM.Structure.Elements;
+using BH.oM.Geometry;
+using BH.oM.Base.Attributes;
+using System.ComponentModel;
+using BHPE = BH.oM.Physical.Elements;
+using BHPC = BH.oM.Physical.Constructions;
```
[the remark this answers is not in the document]
JosefTaylor: TBH I'm copying previous work so go eat a bag of something nice.

Physical_Engine/Query/MaterialComposition.cs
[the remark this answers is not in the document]
JosefTaylor: I think I need to include a null check before this, so that if none of the layers have a material, it gives up immediately.

Physical_Engine/Query/VolumePerArea.cs
```diff
@@ -0,0 +1,82 @@
...
+        [Description("Gets the average thickness of the property for the purpose of calculating solid volume.")]
+        [Input("construction", "The property to evaluate the average thickness of.")]
+        [Output("volumePerArea", "The average thickness of the property for the purpose of calculating solid volume.", typeof(Length))]
+        public static double VolumePerArea(this Construction construction)
+        {
+            if (construction == null)
+                return 0;
+
+            if (construction.Layers.Any(x => x.Material == null))
+                Base.Compute.RecordWarning("At least one Material in a Construction was null. VolumePerArea excludes this layer, assuming it is void space.");
```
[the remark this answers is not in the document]
JosefTaylor: this has the null-check in front of it, which I'd like to use in the other spot as well. Just need to raise an error instead of dumping zero.

Structure_Engine/Query/Construction.cs
JosefTaylor: Yeah, I can add another for constant thickness. The warning comes up when volume < thickness, so I think there's nothing else needed there- I can re-word to specify that the void is a result of ribs/waffles/etc.

Structure_Engine/Create/Physical/Floor.cs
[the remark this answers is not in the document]
michaelhoehn: > 🙄 this is fine.

this is fine!

Physical_Engine/Query/SolidVolume.cs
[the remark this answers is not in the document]
michaelhoehn: I'm happy to leave it as is. This can be picked up at a later date as it's an additional feature.

JosefTaylor:
@BHoMBot check null-handling

michaelhoehn (approved):
I'm happy with the changes. Works as expected. Please make sure to resolve all comments still outstanding prior to check ready merge.

Structure_Engine/Query/Construction.cs
JosefTaylor: Added some additional null-checks around the shop, so the user should see when it happens, and get an idea of what's going on and if it's OK.

JosefTaylor:
@pawelbaran can you accept the changes?

JosefTaylor:
@BHoMBot check required

Structure_Engine/Create/Physical/Floor.cs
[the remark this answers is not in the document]
pawelbaran: Is this fine?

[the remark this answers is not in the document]
FraserGreenroyd: image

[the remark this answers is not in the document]
michaelhoehn: > > 🙄 this is fine.
> 
> this is fine!

this is fine?

pawelbaran (approved):
Happy with the code changes now - glad to have @michaelhoehn to test the functionality as well. The only potential open question is the solid volume, whether it should be area * thickness or area * volume per area. Marking @IsakNaslundBh to hopefully have a look at it and raise concerns if any. Thanks @JosefTaylor well done 👍

pawelbaran:
@BHoMBot check ready-to-merge

pawelbaran:
@BHoMBot check ready-to-merge

pawelbaran:
@BHoMBot check ready-to-merge

pawelbaran:
@BHoMBot check ready-to-merge

pawelbaran:
@BHoMBot check ready-to-merge

JosefTaylor:
@pawelbaran thanks for the final check, do you know what's up with BHoMBot?

pawelbaran:
@BHoMBot check ready-to-merge

pawelbaran:
@BHoMBot check ready-to-merge

pawelbaran:
@BHoMBot check ready-to-merge

PR #2894, opened 2022-08-19
BHoM_Engine: EventRecorded event added together with the correspondent EventArgs

### LINKED ISSUE #2893: BHoM_Engine: Extend the event logging mechanism to allow custom actions
#### Description:

As in the title, currently BHoM event logging mechanism has a really closed formula, in which one can only log an event to a static `Log` and get it out of there. As long as it is sufficient for the desktop solutions like Grasshopper, becomes an issue in case of async or distributed code. This is shown in the diagram below, where we have 4 different cases:
- single threaded desktop process
- parallelised desktop process
- web API
- server rendered web app

As on can see, the current concept of log works only in first two cases, while for the other two it does not give enough granularity, resulting in events from different sources being mixed in one log.

image

Important caveat here is the fact that making the log `ThreadStatic` would not work because async methods are often executed on more than one thread (the original thread is being released while awaiting for the tasks to finish, then a random thread from the pool is taken on callback). So in practice, it is virtually impossible to have a single `Log` that would serve all purposes (say, multithreaded desktop apps and async web controllers at the same time).

So instead of building a one solution to rule them all, I started thinking of extending the current mechanism by adding a `Action<Event>` field (or delegate), which would be triggered at the end of `BH.Engine.Base.Compute.RecordEvent` method and could be assigned by the external code to process the recorded event with full freedom. The assigned action could for example:
- write the event message to console
- write the event message to a file
- log the event in a task-specific log as explained in [this blog post](https://www.cazzulino.com/callcontext-netstandard-netcore.html) (I have tested this solution and it works like a charm, could consistently keep separate logs per each logical flow, even if the thread changed due to `await`)

Just in case, an event instead of `Action<Event>` will not work because it would need to be static, so each time an `Event` would be recorded and the event raised on _any_ thread, it would be visible to _all_ objects listening to it. And again, we get fully static behaviour where all events are seen by all threads.

**EDIT:** actually, both event and extra `Action<Event>` obviously exercise similar, static behaviour. In my initial experimentation I subscribed to the event too many times, all works great if done only once. `AsyncLocal` is the key to track the events per logical flow, BHoM_Engine will always be static with the current setup.

pawelbaran:

Closes #2893

### Test files

Pretty simple mechanism on the engine side - to be tested thoroughly by the consumers.

### Additional comments

See the details in https://github.com/BHoM/BHoM_Engine/issues/2893. Extending the diagrams from there with this PR and the concept of `CallContext` as in [this post](https://www.cazzulino.com/callcontext-netstandard-netcore.html), we land at the flow as below. This has been tested and works as expected.

image

On another note: not sure if `BH.Engine.Base.Compute` is the right place for an event, but it has not been dismissed by neither @IsakNaslundBh nor @alelom - hopefully the others will not mind too.


BHoM_Engine/Compute/EventRecorded.cs
```diff
@@ -0,0 +1,54 @@
...
+namespace BH.Engine.Base
+{
+    public static partial class Compute
+    {
+        /***************************************************/
+        /**** Public Events                             ****/
+        /***************************************************/
+
+        [Description("Gets raised every time an event gets recorded in the debug log (see BH.Engine.Compute.RecordEvent method).")]
+        public static event EventHandler<EventRecordedEventArgs> EventRecorded;
```
adecler, line 37: Minor detail: I would have this event in the `RecordEvent.cs` file and the content of the `OnEventRecorded` method directly at the only place it will ever be called.
It feels more in line with the BHoM file structure as this one doesn't actually contain a public Compute method.
Other than that, it looks good.

adecler (approved):
I am happy with the structure of the code.
There is very little else to review as the functionality is very straightforward.

pawelbaran:
@BHoMBot check required

pawelbaran:
@BHoMBot check unit-tests
@BHoMBot check ready-to-merge

PR #2924, opened 2022-10-06
Spatial_Engine: Add Orient method for IElements

### LINKED ISSUE #2923: Spatial_Engine: Add Orient method to IElements
#### Description:

Requested in an offline chat by @ssiimmoonnee - should be as easy as chaining `GetGeometry`, `Orient` and `SetGeometry`?

### LINKED ISSUE #2925: Facade_Engine: Transform of FrameEdge not cloning FrameEdgeProperty properly
#### Description:

https://github.com/BHoM/BHoM_Engine/blob/4217ecc209564943abf36da932493268cf2eb284/Facade_Engine/Modify/Transform.cs#L58-L78

Does not clone the property, and is hence modifying the incoming object. Proper cloning to be done

#### Steps to reproduce:

Apply a transform that includes a rotational aspect to a FrameEdge object (or any other facade object, such as Panel/Opening). Continously applying the modifies the incoming object and hence causes weird and uncontrollable behaviour.

#### Expected behaviour:

For incoming object not to be changed, and outcoming object to get the proper transformation applied.

IsakNaslundBh:

Closes #2923
Closes #2925 

- Adds method for Orienting Elements.
- Fix cloning issue with facade FrameEdge

### Test files

https://burohappold.sharepoint.com/:f:/s/BHoM/Eq_NayBBG7RNhDlbSgaQmD0BFSZdeP7w_QY68UITtETTnQ?e=QXSMRg

### Additional comments

Was debating with myself if providing null to from or to should default to world XY. That way it would be simpler to go from global to local, or vice versa. (not having to provide a global XY plane in one of the inputs). I opted out of it due to the Orient method in Geometry_Engine not behaving this way, but happy to change if we want to. @pawelbaran @al-fisher happy with comments on this.


IsakNaslundBh:
@BHoMBot check unit-tests

IsakNaslundBh:
@BHoMBot check unit-tests

IsakNaslundBh:
@BHoMBot check unit-tests

Spatial_Engine/Modify/Orient.cs
```diff
@@ -0,0 +1,74 @@
...
+            {
+                Base.Compute.RecordError($"The {nameof(from)} coordinate system is null. Unable to orient the element.");
+                return null;
+            }
+
+            if (to == null)
+            {
+                Base.Compute.RecordError($"The {nameof(to)} coordinate system is null. Unable to orient the element.");
+                return null;
+
```
pawelbaran, line 65: one blank too much 😉

Spatial_Engine/Modify/Orient.cs
```diff
@@ -0,0 +1,74 @@
...
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Orients the element from one coordinate system to another.")]
+        [Input("element", "The element to orient.")]
+        [Input("from", "The cartesian coordinate system to orient from.")]
+        [Input("to", "The cartesian coordinate system to orient to.")]
+        [Output("element", "The reoriented element.")]
+        public static IElement Orient(this IElement element, Cartesian from, Cartesian to)
```
pawelbaran, line 47: I am thinking if it would not be worth defaulting both `to` and `from` to the global CS? Personally I would find it handy in some circumstances

pawelbaran: Just read the additional comments in PR description - great minds think alike! 😃

IsakNaslundBh: Done

IsakNaslundBh:
@BHoMBot check unit-tests

Spatial_Engine/Modify/Orient.cs
```diff
@@ -0,0 +1,79 @@
...
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Orients the element from one coordinate system to another.")]
+        [Input("element", "The element to orient.")]
+        [Input("from", "The cartesian coordinate system to orient from. Deaults to global XY if nothing is provided.")]
+        [Input("to", "The cartesian coordinate system to orient to. Deaults to global XY if nothing is provided.")]
+        [Output("element", "The reoriented element.")]
+        public static IElement Orient(this IElement element, Cartesian from = null, Cartesian to = null)
```
pawelbaran, line 47: Personally I would consider adding a static `Cartesian.Global` or similar and defaulting `from` and `to` to that value. Now this method creates the risk of `null` being passed unnoticed, which is a no go for me I think.

IsakNaslundBh: there is not static one (otherwise I would have used it), can ofc switch to create one with standard axis though.

Also thought about the potential issue if the default values changed, but I have included all three null providing cases for coordinate system in the UT, so if it was ever changed, the UT would flag it.

If you still think that is to risky, can ofc update?

pawelbaran: I would create one, same as we have e.g. in `Basis.XY`. Not sure about the UT part, but I am thinking about a scenario where another method crashes and returns `null` and then it is picked by `Orient` and processed as if the input was valid - this is what I am afraid of.

IsakNaslundBh: Ah, I see. Not sure that is solvable from here though. Think the optional inputs need to be compile time constants, but will have a look.

pawelbaran: Ah, you are right, the coordinate system will not be constant compile time. In this case I am taking back my original suggestion - would rather be on the safe side than on the handy one.

IsakNaslundBh:
@BHoMBot check unit-tests

IsakNaslundBh:
@BHoMBot check compliance
@BHoMBot check null-handling

FraserGreenroyd:
@BHoMBot check versioning

FraserGreenroyd:
@BHoMBot check versioning

FraserGreenroyd:
@BHoMBot check versioning

pawelbaran (approved):
Happy to approve following the review made and discussion successfully resolved 👍

FraserGreenroyd:
@BHoMBot check versioning

FraserGreenroyd:
@BHoMBot this is a DevOps instruction. I am requesting a force check on: versioning

IsakNaslundBh:
@BHoMBot check core
@BHoMBot check serialisation
@BHoMBot check ready-to-merge

IsakNaslundBh:
@BHoMBot check core
@BHoMBot check serialisation
@BHoMBot check ready-to-merge

PR #2927, opened 2022-10-11
Geometry_Engine: Fix edgecases for boolean methods and SplitAtPoints

### LINKED ISSUE #2926: Geometry_Engine: Fix issues with edge cases for Boolean methods
#### Description:

For some edgecases boolean intersection and difference returns curves that are open, just above tolerance, and are hence culled.

For found edge cases this can be solved by closing the curves, before checking the areas.

#### Steps to reproduce:

Run test script (will be provided in PR)

#### Expected behaviour:

For edge cases to be handled

IsakNaslundBh:

Closes #2926

This turned a bit bigger than first expected, but the intention of this PR is to fix various edgecases for the BooleanIntersection and BooleanDifference methods.

- Add Close method for polycurve
- Close the curves before checking areas for BooleanUnion and BooleanDiffrence methods
- Complete rewrite of SplitAtPoints for Polycurve as it was malfunctioning for various cases (for example able to split a curve when flipped, but not when not flipped)
- Add initial distance check to curve for IsContaining as it for some edge cases did not properly catch a point (or exclude it) when the point was on the curve.

### Test files

Testscripts in these folders for BooleanIntersection, BooleanDifference and SplitAtPoints:

https://burohappold.sharepoint.com/:f:/s/BHoM/EsF7k48YhAZGnqtiqHvTLTQBHj5AFklfVD8R2eSmSg5f5w?e=eflnwh

BooleanDifference is showing 2 failures which are both tolerance based issues. First for PolyCurve the BHoM version gets an extra short Line segment (from closing the curve) that the Rhino equivalent does not produce. Before this PR that small part was not even included) In general, failures before was 6 on my machine, now down to this single one.
image

BooleanDifference for ICurve is showing one small difference as well, where the BHoMMethod is adding another small segment into the curve:
image
image

of 2mm. This, again, might not be 100% perfect, but this PR if for this method as well seeming to drop errors reported compared to main.

Boolean intersection script should be showing 0 errors (less than on main that had a few errors)
SplitAtPoints for PolyCurves should be showing 0 errors.

Link to testfiles for issues handled (edge cases) as well as new file for Close:

https://burohappold.sharepoint.com/:f:/s/BHoM/ElBY7oEIt5ROhzhGgA5n7kMB-jSM6L15qVIdVERkFo_BVw?e=Soso5B

Edge cases in testfiles have all been added to UTs.

Previous bug fix test file for SplitAtPoints:

https://burohappold.sharepoint.com/:f:/s/BHoM/Eudac92bFj1LumV-YZmgUZIBRFicNsjly4r5-yd0SL575w?e=bsQYpd

This case has also been added to the UT


IsakNaslundBh:
One item failing for BooleanIntersection in UTs. The failure is identified to be a simple shift of the inner list of SubParts for this case. As the methods are not stating anything about maintaining any order of the curves (nonsensical for most cases), I will regenerate the UT and push up a new one with the result past changes in this PR.

Initial case:
image

Branch:
image

IsakNaslundBh:
@BHoMBot check unit-tests

IsakNaslundBh:
@BHoMBot check compliance

IsakNaslundBh:
@BHoMBot check unit-tests

IsakNaslundBh:
@BHoMBot check compliance

IsakNaslundBh:
@BHoMBot check null-handling
@BHoMBot check versioning

Geometry_Engine/Modify/Close.cs
```diff
@@ -52,6 +52,37 @@ public static Polyline Close(this Polyline polyline, double tolerance = Toleranc
 
             return Create.Polyline(polylinePoints);
         }
+
+        /***************************************************/
+
+        [Description("Close open PolyCurve by adding Line segments between adjecent inner curves not within tolerance. Does not affect Closed curves. Inner Curves of polycurve are assumed to be Ordered such that the end point of inner Curve n should be the start point of Curve n+1 for the curve to be closed.")]
```
pawelbaran, line 58: I started rewriting the description and then realised your last sentence is clear enough. But keeping this comment for reference, please choose the one you like more (in yours there is typo in _adjacent_, just in case).
```suggestion
        [Description("Close open PolyCurve by adding Line segments between each pair of subsequent segment curves if the end point of the first does not lie within tolerance from start point of the second. Does not affect Closed curves.")]
```

Geometry_Engine/Modify/Close.cs
```diff
@@ -52,6 +52,37 @@ public static Polyline Close(this Polyline polyline, double tolerance = Toleranc
...
+        [Input("polyCurve", "PolyCurve to close.")]
+        [Input("tolerance", "Tolerance used in the method.")]
```
pawelbaran, line 60:
```suggestion
        [Input("tolerance", "Distance tolerance used in the method.")]
```

Geometry_Engine/Modify/SplitAtPoints.cs
```diff
@@ -212,111 +212,113 @@ public static List<PolyCurve> SplitAtPoints(this PolyCurve curve, List<Point> po
                 return new List<PolyCurve> { curve.DeepClone() };
```
pawelbaran, line 212: This `DeepClone` makes me upset, especially that there is a chance it was me who put it here long time ago 🙈

Geometry_Engine/Modify/SplitAtPoints.cs
```diff
@@ -212,111 +212,113 @@ public static List<PolyCurve> SplitAtPoints(this PolyCurve curve, List<Point> po
...
 
-            onCurvePoints = onCurvePoints.CullDuplicates(tolerance);
-
-            if (onCurvePoints.Count == 0)
-                return new List<PolyCurve> { curve.DeepClone() };
-
-            onCurvePoints = onCurvePoints.SortAlongCurve(curve);
-
-            foreach (ICurve crv in curve.SubParts())
+            List<Point> nonDuplicatePoints = points.CullDuplicates(tolerance);
```
pawelbaran, line 226: Shouldn't this happen at the beginning? Why is it call for multiple segments but not for one?

Geometry_Engine/Modify/SplitAtPoints.cs
```diff
@@ -212,111 +212,113 @@ public static List<PolyCurve> SplitAtPoints(this PolyCurve curve, List<Point> po
                 return new List<PolyCurve> { curve.DeepClone() };
 
             List<PolyCurve> result = new List<PolyCurve>();
-            List<ICurve> tmpResult = new List<ICurve>();
-            List<Point> subPoints = new List<Point>();
-            List<Point> onCurvePoints = new List<Point>();
 
-            foreach (Point p in points)
+            List<ICurve> subParts = curve.ISubParts().ToList(); //Subparts of PolyCurve
+            if (subParts.Count == 1)    //Single SubPart
```
pawelbaran, line 217: What's the benefit of having a dedicated case for a single segment? Couldn't it be processed by the general case?

Geometry_Engine/Modify/SplitAtPoints.cs
```diff
@@ -212,111 +212,113 @@ public static List<PolyCurve> SplitAtPoints(this PolyCurve curve, List<Point> po
...
-
-            foreach (ICurve crv in curve.SubParts())
+            List<Point> nonDuplicatePoints = points.CullDuplicates(tolerance);
+            PolyCurve prev = null;  //Polycurve to collect parts across segments
+            double sqTol = tolerance * tolerance;
+            for (int k = 0; k < subParts.Count; k++)
             {
-                if (crv is Arc)
+                ICurve crv = subParts[k];
+                List<Point> ptsOnSegment = nonDuplicatePoints.Where(x => x.IIsOnCurve(crv, tolerance)).ToList(); //Get points on the segment
```
pawelbaran, line 232: From a brief look at the other `SplitAtPoints` methods, `IsOnCurve` is called explicitly inside each of them. Which means that this call is likely to be redundant.

Geometry_Engine/Modify/SplitAtPoints.cs
```diff
@@ -212,111 +212,113 @@ public static List<PolyCurve> SplitAtPoints(this PolyCurve curve, List<Point> po
...
+
+                if (ptsOnSegment.Any()) //If any points on the current segment
                 {
-                    foreach (Point point in onCurvePoints)
+                    List<ICurve> split = new List<ICurve>();
+                    split.AddRange(SplitAtPoints(crv as dynamic, ptsOnSegment, tolerance)); //Split the segment. Assuming curves returned are ordered
```
pawelbaran, line 237: Couldn't you simply call `ISplitAtPoints` here?
```suggestion
                    List<ICurve> split = SplitAtPoints(crv as dynamic, ptsOnSegment, tolerance); //Split the segment. Assuming curves returned are ordered
```

Geometry_Engine/Modify/SplitAtPoints.cs
```diff
@@ -212,111 +212,113 @@ public static List<PolyCurve> SplitAtPoints(this PolyCurve curve, List<Point> po
...
+                            }
+                            prev = null;    //Split at end -> set prev to null as nothing to bring over to next segment
+                        }
+                    }
+                    else    //More than one split segment
+                    {
+                        if (prev != null)   //Prev contains parts from previous segment, unless split at endpoint
+                        {
+                            Point stPt = crv.IStartPoint();
+                            if (ptsOnSegment.Any(x => x.SquareDistance(stPt) < sqTol))  //If split at startpoint
```
pawelbaran, line 272: Not sure if faster, but I would consider switching to the suggestion below (also another step to get rid of `ptsOnSegment`, which I suggested not to extract a few comments ago)
```suggestion
                            if (split[0].IStartPoint().SquareDistance(stPt) < sqTol)  //If split at startpoint
```

Geometry_Engine/Modify/SplitAtPoints.cs
```diff
@@ -212,111 +212,113 @@ public static List<PolyCurve> SplitAtPoints(this PolyCurve curve, List<Point> po
...
+                                result.Add(prev);   //Add prev to return list
+                            }
+                        }
+                        else  //Prev null, means either this is first segment of curve or previous segment was split at endpoint -> nothing to bring over
+                            result.Add(new PolyCurve { Curves = new List<ICurve> { split[0] } });  //Simply add first split segment to return list
 
-            while (i <= onCurvePoints.Count)
-            {
-                List<ICurve> subResultList = new List<ICurve>();
+                        for (int s = 1; s < split.Count - 1; s++)   //Loop though and all segments but the first and last
```
pawelbaran, line 286:
```suggestion
                        for (int s = 1; s < split.Count - 1; s++)   //Loop through all segments but the first and last
```

Geometry_Engine/Modify/SplitAtPoints.cs
```diff
@@ -212,111 +212,113 @@ public static List<PolyCurve> SplitAtPoints(this PolyCurve curve, List<Point> po
...
+                        }
 
-                while (j < tmpResult.Count)
-                {
-                    subResultList.Add(tmpResult[j]);
-                    if (i < onCurvePoints.Count)
-                    {
-                        if (tmpResult[j].IEndPoint().IsEqual(onCurvePoints[i]) || (isClosedAndAntiClockwise && tmpResult[j].IStartPoint().IsEqual(onCurvePoints[i])))
+                        Point ePt = crv.IEndPoint();
+                        if (ptsOnSegment.Any(x => x.SquareDistance(ePt) < sqTol)) //If split at endpoint of current segment
```
pawelbaran, line 292: Same as above
```suggestion
                        if (split.Last().IEndPoint().SquareDistance(ePt) < sqTol) //If split at endpoint of current segment
```

Geometry_Engine/Modify/SplitAtPoints.cs
```diff
@@ -212,111 +212,113 @@ public static List<PolyCurve> SplitAtPoints(this PolyCurve curve, List<Point> po
...
-                    foreach (ICurve subCrv in result[result.Count - 1].ISubParts())
-                        subResultList.Add(subCrv);
-                    foreach (ICurve subCrv in result[0].ISubParts())
-                        subResultList.Add(subCrv);
-                    result.RemoveAt(0);
-                    result.RemoveAt(result.Count - 1);
-                    result.Add(new PolyCurve { Curves = subResultList.ToList() });
+                    if (prev != null)   //If prev is set
+                    {
+                        if (curve.IsClosed(tolerance) && result.Count != 0) //If curve is closed
```
pawelbaran, line 315: Am I correct that this line will become funky in case of disjoint `PolyCurve`? Also your algorithm of adding segments to `prev` may return weird stuff for disjoint curves. Maybe worth disallowing the method for anything that is not closed?

Geometry_Engine/Modify/SplitAtPoints.cs
```diff
@@ -212,111 +212,113 @@ public static List<PolyCurve> SplitAtPoints(this PolyCurve curve, List<Point> po
...
+                            result[0].Curves.InsertRange(0, prev.Curves);   //Insert the curves in prev to the start of the first curve, ensuring a segment looping around the start/end is added
+                        else
+                            result.Add(prev);   //If not, simply add the prev to the return list
+                    }
                 }
+            }
 
             return result;
```
pawelbaran, line 323: I would unwrap single-segment `PolyCurves` before returning (alternatively you could store all types in `result` while iterating through the segments, but this will require replacing single segments with `PolyCurves` when adding another segment to them)

Geometry_Engine/Query/IsContaining.cs
```diff
@@ -353,6 +353,15 @@ public static bool IsContaining(this PolyCurve curve, List<Point> points, bool a
                 if (pPt.SquareDistance(pt) > sqTol) // not on the same plane
                     return false;
 
+                double distance = pPt.Distance(curve);
```
pawelbaran, line 356: Wow, this is pretty brute force 😉 Hard to believe that this had not been captured in the past?

pawelbaran (changes requested):

Geometry_Engine/Modify/SplitAtPoints.cs
IsakNaslundBh: Mainly because those methods call `CullDuplicates` themselves anyway. Hence calling it here would be a wastefull call that would be of no real benefit.

pawelbaran: Right, makes sense. However, if it is a question of just one call then I would still opt for generalising the method for n >= 1 segments instead of the individual treatment of n == 1. But happy for you to make a final decision, you def have a better feeling as the originator 👍

Geometry_Engine/Modify/SplitAtPoints.cs
IsakNaslundBh: Can probably refactor it slightly to work, but think there are some cases that need to be handled slightly differently, for example, if it is split at start point, then the general case simply adds the segment to the list, while the special case for single segment instead adds the segment to a new "prev".

Slight difference in handling of end point split as well.

Basically, general cases handles 
1. Start split 
2. all middle splits
3. End split 

all separately

While the case for a single split it should only handle start _or_ end split case, if that makes sense?

Geometry_Engine/Modify/SplitAtPoints.cs
IsakNaslundBh: It kind of is redundant, yes, but there to check as a pre-check if a split will be made for the particular segment, if ptsOnSegment count is non-zero. 

Can refactor this slightly though to send in all points, and then, if a single item is returned, check if start and/or endpoints are within tolerance of any of the points instead. Might potentially be cleaner..

Geometry_Engine/Modify/SplitAtPoints.cs
pawelbaran: It does, but middle split is handled starting from `s = 1`, which will never be the case for a single split, so assuming that start/end should not be dependent on the number of resultant segments, it could be generalised?

Geometry_Engine/Modify/SplitAtPoints.cs
pawelbaran: Yeah, I think it may be cleaner, plus also potentially faster, assuming that point operations are lighter than duplicate `IsOnCurve` (run once to get `ptsOnSegment` and then called individually inside `SplitAtPoints`)

Geometry_Engine/Modify/SplitAtPoints.cs
IsakNaslundBh: same as above 😉

Geometry_Engine/Modify/SplitAtPoints.cs
IsakNaslundBh: You might very much be right in that this might do funky stuff for disjointed curves... Have not tested really actually.

With closed I guess you mean continuous curves? If so, maybe. Honestly not sure how well it worked for disjointed curves before.

As for this line, I can change it to something like:
```suggestion
                        if (curve.IStartPoint().SquareDistance(curve.IEndPoint()) < sqTol && result.Count != 0) //If curve is closed
```

Instead, which would then potentially work for disjointed curves (or at least as well as the other parts do...)

Geometry_Engine/Modify/SplitAtPoints.cs
IsakNaslundBh: Would require to change the return type of this method from PolyCurve to ICurve though. Not sure if we want to do that?

Geometry_Engine/Query/IsContaining.cs
IsakNaslundBh: Yeah, I know.. 

A bit unsure if what I did in https://github.com/BHoM/BHoM_Engine/pull/2802 caused some issue with this particular edge case. I am not sure how my changes there would though...

anyway, I think this might actually be a performance improvement for some cases (points on edges) as it skips the need for intersections etc. Ofc, will be a slowdown for other cases.

Basically, this _is_ attempted to be checked for further down, but it dos fail for some weird edgecases.

Geometry_Engine/Modify/SplitAtPoints.cs
IsakNaslundBh: Honestly, had it like that at first. (no special case). Main reason I added this was due to issues with closed single segmented subparts (read circle) where the strategy for prev stopped working as it lept around and attached to itself. 

Can be handled in other ways ofc, but this felt like the simplest way of handling it..

Geometry_Engine/Modify/SplitAtPoints.cs
pawelbaran: Right, good point. Me stupid 🤦‍♂️

Geometry_Engine/Modify/SplitAtPoints.cs
IsakNaslundBh: Had another look, and managed to refactor it all into something with less edge cases. THink this made the code cleaner (and significantly shorter) so thanks for pushing @pawelbaran .

Geometry_Engine/Modify/SplitAtPoints.cs
IsakNaslundBh: fixed now. No special case for single curve

Geometry_Engine/Modify/SplitAtPoints.cs
IsakNaslundBh: fixed now, with final refactoring

IsakNaslundBh:
@BHoMBot check unit-tests

Geometry_Engine/Modify/SplitAtPoints.cs
pawelbaran: Fair point - I would have nothing against, taken that `PolyCurve` implement `ICurve`. But curious about your opinion on changing the return type?

IsakNaslundBh: Would not really mind changing it in general, mostly cautios about changing it in case it cascades into affecting 55 other methods, as well as methods outside the BHoM_Engine. But in general, I would not be against it. Wrapping the split segments into single segment polycurves does not feel that pretty, fully agree on that.

Will check what effect a change of return type has on the engine to start with.

IsakNaslundBh:
> The code is much cleaner now, but UTs seem to fail. Also I would reiterate on my question regarding the return type: the more I think about it, the more I would recommend returning a collection of `ICurve`...

Have inspected the failing UT, and seem the latest iteration of the SplitAtPoint has changed back the behaviour of the UT mentoined above (https://github.com/BHoM/BHoM_Engine/pull/2927#issuecomment-1274474705) back to how it worked before. Will push an update, changing it back again.

IsakNaslundBh:
@BHoMBot check unit-tests

IsakNaslundBh:
@BHoMBot check installer

IsakNaslundBh:
@BHoMBot check unit-tests

Geometry_Engine/Modify/SplitAtPoints.cs
IsakNaslundBh: Yeah, prefer not to change it now though, as could impact elsewhere...

Geometry_Engine/Modify/SplitAtPoints.cs
IsakNaslundBh: Seems fine to change return type, so have actioned it and updated the UTs for SplitAtPoint to comply. Seemed as most methods were calling the interface method anyway, hence no need to change the code elsewhere, and fully agree that it is better to just return single arcs/lines rather than them wrapped in a polycurve for no real reason :)

IsakNaslundBh:
@BHoMBot check compliance
@BHoMBot check required

IsakNaslundBh:
@BHoMBot check unit-tests

IsakNaslundBh:
@BHoMBot check unit-tests

Geometry_Engine/Modify/SplitAtPoints.cs
```diff
@@ -50,41 +50,146 @@ public static List<Arc> SplitAtPoints(this Arc arc, List<Point> points, double t
...
+            }
+
+            //If any points on the full circle corresponding to the arc
+            if (cPts.Count > 0)
             {
-                Double startAng = arc.StartAngle;
-                Double endAng = arc.EndAngle;
-                Double tmpAng = 0;
-                Arc tmpArc;
+                double angleTolerance = tolerance / arc.Radius; //Allowable difference in angle for the points on the arc to be tolerance distance away from each other
```
pawelbaran, line 79: Are you 100% sure this value corresponds to the distance between two points on the arc (marked in blue)? I am not saying it is not, actually I have a blackout and can't find a definite answer 😆 

image

Geometry_Engine/Modify/SplitAtPoints.cs
```diff
@@ -50,41 +50,146 @@ public static List<Arc> SplitAtPoints(this Arc arc, List<Point> points, double t
...
+                Point projPt = point.Project(arcPlane);
+                double projSqDist = projPt.SquareDistance(point);
+                if (projSqDist > sqTol)
+                    continue;   //Not in plane
 
-            cPts.Add(arc.StartPoint());
-            cPts.Add(arc.EndPoint());
-            cPts = cPts.CullDuplicates(tolerance);
-            cPts = cPts.SortAlongCurve(arc, tolerance);
+                double inPlaneDist = Math.Abs(projPt.Distance(arcPlane.Origin) - arc.Radius);
```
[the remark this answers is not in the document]
IsakNaslundBh: Because the radius need to be subtracted.

`Math.Abs(projPt.SquareDistance(arcPlane.Origin) - arc.Radius * arc.Radius);`
Is not the same thing as 

`Math.Pow(Math.Abs(projPt.Distance(arcPlane.Origin) - arc.Radius), 2);`

unfortunately. 

Had a thought about it (and agree on the line 70 comment) but think this is the simplest and quickest it can be done.

Geometry_Engine/Modify/SplitAtPoints.cs
IsakNaslundBh: No, you are right, it is the length of the black, not the blue. Will try to update.

Geometry_Engine/Modify/SplitAtPoints.cs
```diff
@@ -50,41 +50,146 @@ public static List<Arc> SplitAtPoints(this Arc arc, List<Point> points, double t
...
+                    return new List<Arc> { arc.DeepClone() };
+
+                angles.Sort();  //Equal to sort along curve
+
+                //Remove duplicates within tolerance
+                //Both more efficient and accurate to handles this with the angles compared to removing duplicate input points
+                //Doing it this way ensures points on the curve are not within tolerance, rather than the input points, that could be
+                //Outside tolerance in relation to each other, but ending up on the same point on the angle
+                //The below does the equivalent of a DBScan (which is what the CullDuplicates for points relies on)
+                List<List<double>> angleGroups = new List<List<double>>() { new List<double> { angles[0] } };
```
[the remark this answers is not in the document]
IsakNaslundBh: It could. Not sure where to put it though.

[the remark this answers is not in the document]
IsakNaslundBh: As discussed offline, put as private method for now

IsakNaslundBh:
@BHoMBot check compliance
@BHoMBot check unit-tests

pawelbaran (approved):
Happy to approve after a thorough review and refactoring of refactoring 👍 well done @IsakNaslundBh

IsakNaslundBh:
@BHoMBot check required

IsakNaslundBh:
@BHoMBot check ready-to-merge

PR #2929, opened 2022-10-19
Geometry_Engine: implement GeometryHash

### LINKED ISSUE #2928: Geometry_Engine: Implement GeometryHash

alelom:

Closes #2928 

### Test files

https://burohappold.sharepoint.com/:f:/s/BHoM/EjL3UGZ1dPNJtyv8RSTZsLYBktwHrwDwRjxsNqh6aJPQrw?e=EDxHzY


al-fisher (changes requested):
Hey @alelom - looking good!
See comments below regarding `PolySurface`

Geometry_Engine/Query/GeometryHash.cs
```diff
@@ -0,0 +1,381 @@
...
+            PlanarSurface,
+            Extrusion,
+            Loft,
+            Pipe,
+            PolySurface,
+            NurbsSurface,
+            SurfaceTrim,
+            Mesh,
+            Mesh3D,
+            Plane
```
al-fisher, line 64: For some reason, when looking at this type list, my geometry brain sees this as the logical order - seemingly raising in complexity/degree of freeform.

In fact this is the order I would probably list the geometry if I was introducing their definitions in documentation. So makes sense to me at least! 😸 😕 

But offering this as a suggestion, happy to hear others alternate orders - it is all arbitrary in the end after all! @alelom @IsakNaslundBh 

```suggestion
            Point = 0,
            Plane,
            Line,
            Arc,
            Circle,
            Ellipse,
            NurbsCurve,
            PlanarSurface,
            Pipe,
            Extrusion,
            Loft,
            NurbsSurface,
            SurfaceTrim,
            Mesh,
            Mesh3D,
```

Geometry_Engine/Query/GeometryHash.cs
```diff
@@ -0,0 +1,381 @@
...
+            return result;
+        }
+
+        /***************************************************/
+
+        private static double[] GeometryHash(this PolySurface obj, double translationFactor)
+        {
+            translationFactor += (int)TypeTranslationFactor.PolySurface;
+
+            return obj.Surfaces.SelectMany(s => s.GeometryHash(translationFactor)).ToArray();
```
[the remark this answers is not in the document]
IsakNaslundBh: Agree with @al-fisher here.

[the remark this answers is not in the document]
alelom: Good point!

Geometry_Engine/Query/GeometryHash.cs
```diff
@@ -0,0 +1,381 @@
...
+            Arc,
+            Circle,
+            Ellipse,
+            Line,
+            NurbsCurve,
+            PlanarSurface,
+            Extrusion,
+            Loft,
+            Pipe,
+            PolySurface,
```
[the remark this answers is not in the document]
alelom: Good point.

Geometry_Engine/Query/GeometryHash.cs
alelom: Looks good to me! 😄 As I mentioned, the only reason I followed that specific order was to avoid confusion when writing down each type, following the order they appeared in in the Engine folder -- no reason not to make it more tidy now!

Geometry_Engine/Query/GeometryHash.cs
```diff
@@ -0,0 +1,409 @@
...
+
+namespace BH.Engine.Geometry
+{
+    public static partial class Query
+    {
+        /***************************************************/
+        /****               Public Methods              ****/
+        /***************************************************/
+
+        [Description("All geometry hash methods are implemented to be translational.")]
```
al-fisher, line 39:
```suggestion
        [Description("Returns the geometrical identity of any geometry primitive, useful for comparisons and diffing. " + 
        "Any transformations of underlying geometry performed in methods as part of the calculation of the Geometry Hash " +
        "are implemented to be translational only to ensure validity of any geometrical tolerance checking between two hashes downstream.")]
```

@alelom - I have had a go at a description for the public GeometryHash method - see what you think? 😄

Geometry_Engine/Query/GeometryHash.cs
```diff
@@ -0,0 +1,409 @@
...
+
+        [Description("The GeometryHash for a PolySurface is calculated as the GeometryHash of the individual surfaces.")]
+        private static double[] GeometryHash(this PolySurface obj, double translationFactor)
+        {
+            return obj.Surfaces.SelectMany(s => s.GeometryHash(translationFactor)).ToArray();
+        }
+
+        /***************************************************/
+
+        [Description("The GeometryHash for a PolySurface is calculated as the GeometryHash of its Curve3d and Curve2d, concatenated.")]
```
IsakNaslundBh, line 248:
```suggestion
        [Description("The GeometryHash for a SurfaceTrim is calculated as the GeometryHash of its Curve3d and Curve2d, concatenated.")]
```

Geometry_Engine/Query/GeometryHash.cs
```diff
@@ -0,0 +1,409 @@
...
+            List<double> concatenated = new List<double>();
+            for (int i = 0; i < obj.ControlPoints.Count(); i++)
+            {
+                double UKnotsSum = obj.UKnots.ToList().GetRange(i, obj.UDegree).Sum();
+                double VKnotsSum = obj.VKnots.ToList().GetRange(i, obj.VDegree).Sum();
+
+                double[] doubles = obj.ControlPoints[i].GeometryHash(UKnotsSum + VKnotsSum + obj.Weights[i] + translationFactor);
+                concatenated.AddRange(doubles);
+            }
+
```
IsakNaslundBh, line 218: Think we need to handle the SurfaceTrims here as well?

Geometry_Engine/Query/GeometryHash.cs
```diff
@@ -0,0 +1,409 @@
...
+            {
+                p.X + typeTranslationFactor + translationArray?.ElementAtOrDefault(0) ?? 0,
+                p.Y + typeTranslationFactor + translationArray?.ElementAtOrDefault(1) ?? 0,
+                p.Z + typeTranslationFactor + translationArray?.ElementAtOrDefault(2) ?? 0
+            };
+        }
+
+        /***************************************************/
+
+        private static double[] ToDoubleArray(this IEnumerable<Point> points, double typeTranslationFactor, double[] translationArray = null)
```
IsakNaslundBh, line 377: maybe no hard in keeping really, but do we use/need this method?

Geometry_Engine/Query/GeometryHash.cs
```diff
@@ -0,0 +1,409 @@
...
+
+
+        /***************************************************/
+        /****  Other methods                            ****/
+        /***************************************************/
+
+        // Fallback
+        private static double[] GeometryHash(this object obj, double translationFactor)
+        {
+            BH.Engine.Base.Compute.RecordError($"Could not find a {nameof(GeometryHash)} method for type {obj.GetType().FullName}.");
```
IsakNaslundBh, line 359: Potential future proof (in case a type of IGeometry is added outside the core, with its own implementation of GeometryHash in another Engine) is TryRunExtensionMethod here and return value if found, if not, return error.

Downside I guess is that that would be hashes outside our control, and unable to confirm that the implementation is valid, but also the case for other similar cases, such as IElement methods and similar.

Geometry_Engine/Query/GeometryHash.cs
```diff
@@ -0,0 +1,409 @@
...
+        /***************************************************/
+
+        private static double[] GeometryHash(this ISurface obj, double translationFactor)
+        {
+            return GeometryHash(obj as dynamic, translationFactor);
+        }
+
+        /***************************************************/
+
+        [Description("The GeometryHash for a Line is calculated as the GeometryHash of the External and Internal boundary curves, then concatenated.")]
```
IsakNaslundBh, line 168:
```suggestion
        [Description("The GeometryHash for a PlanarSurface is calculated as the GeometryHash of the External and Internal boundary curves, then concatenated.")]
```

Good to check through the types mentioned in the descs, or to simply not mention them at all, as inplicit from the target type for the method.

IsakNaslundBh (commented):
A few additional comments

Geometry_Engine/Query/GeometryHash.cs
alelom: Yes it's used on line 219 - thought I'd use it more but no harm in keeping it I think

Geometry_Engine/Query/GeometryHash.cs
alelom: Sorry I did copy paste and missed some types.

Geometry_Engine/Query/GeometryHash.cs
alelom: Good catch!

Geometry_Engine/Query/GeometryHash.cs
alelom: Why not! Sure I'll add that.

Geometry_Engine/Query/GeometryHash.cs
alelom: Much better!

Geometry_Engine/Query/GeometryHash.cs
```diff
@@ -0,0 +1,417 @@
...
+    public static partial class Query
+    {
+        /***************************************************/
+        /****               Public Methods              ****/
+        /***************************************************/
+
+        [Description("Returns the geometrical identity of any geometry primitive, useful for comparisons and diffing. " + 
+        "Any transformations of underlying geometry performed in methods as part of the calculation of the Geometry Hash " +
+        "are implemented to be translational only to ensure validity of any geometrical tolerance checking between two hashes downstream.")]
+        public static double[] IGeometryHash(this IBHoMObject bhomObj)
```
IsakNaslundBh, line 42: Shouldn't this be accepting `IGeometry` rather than `IBHoMObject`? Think we should handle the IBHoMObject outside the Geometry_Engine (imo, there should not be any methods accepting any type of BHoMObejct in the Geometry_Engine)

Geometry_Engine/Query/GeometryHash.cs
```diff
@@ -0,0 +1,417 @@
...
+        }
+
+        /***************************************************/
+
+        [Description("The GeometryHash for a Plane is the GeometryHash of its Origin point translated by its Normal vector.")]
+        private static double[] GeometryHash(this Plane obj, double translationFactor)
+        {
+            translationFactor += (int)TypeTranslationFactor.Plane;
+
+            return obj.Origin.Translate(obj.Normal).ToDoubleArray(translationFactor);
```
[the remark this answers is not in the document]
alelom: Yeah, I thought it might not be. Sure, happy to chat.

Geometry_Engine/Query/GeometryHash.cs
alelom: The reason I went for this solution:
1. I thought we'd like to expose `GeometryHash()` on top of every object, like we do for `Hash()` so my first choice was actually `IObject`
1. However, `IGeometry()` only accepts `IBHoMObject` and `CustomObject`. I agree with you, I don't like this either: this conflicts with 
   > imo, there should not be any methods accepting any type of BHoMObejct in the Geometry_Engine) 
1. I also am not sure we want to expose `GeometryHash()` for `IGeometry` directly: I think `GeometryHash()` makes more sense when paired with an object that is not geometry, because if you want to compare two geometries you can just do that. But don't feel strongly about this one, happy to revise.

Therefore, I went for `IBHoMObject`. I'd be more happy to change this to `IObject` honestly, so we replicate what `Hash()` targets.

Geometry_Engine/Query/GeometryHash.cs
```diff
@@ -0,0 +1,427 @@
...
+        [Description("The geometry hash of a Curve is obtained by first retrieving any Sub-part of the curve, if present." +
+            "The ISubParts() methods is able to return the 'primitive' curves that a curve is composed of. " +
+            "The GeometryHashes are then calculated for the individual parts and concatenated.")]
+        private static double[] GeometryHash(this ICurve curve, double translationFactor)
+        {
+            IEnumerable<ICurve> subParts = curve.ISubParts();
+
+            bool isMultipleCurves = subParts.Count() != 1;
+
+            return subParts.SelectMany(c => (double[])GeometryHash(c as dynamic, translationFactor, isMultipleCurves)).ToArray();
```
IsakNaslundBh, line 71: This we want the endpoint of the last curve included - at least for open curves, but probably fine to include for all. Would change to something like:
```suggestion
            List<ICurve> subParts = curve.ISubParts().ToList();

            List<double> hashes = new List<double>();

            //Add hash ignoring endpoint for all but last curve
            for (int i = 0; i < subParts.Count - 1; i++)
            {
                hashes.Add(GeometryHash(subParts[i] as dynamic, translationFactor, true));
            }
            //Include endpoint for hasing for last curve
            hashes.Add(GeometryHash(subParts.Last() as dynamic, translationFactor, false));

            return hashes.ToArray();
```

Could add a check for IsClosed for whether the endpoint of the last curve should be included or not, but then we start running in to various tolerance discussions that I think we want to avoid, so probably better to just always include it as above.

Geometry_Engine/Query/GeometryHash.cs
```diff
@@ -0,0 +1,427 @@
...
+        }
+
+        /***************************************************/
+
+        [Description("The GeometryHash for an Arc is calculated as the GeometryHash of the start, end and middle point of the Arc.")]
+        private static double[] GeometryHash(this Arc curve, double translationFactor, bool skipLastPoint = false)
```
IsakNaslundBh, line 77: Might change the name of this to skipEndPoint
```suggestion
        private static double[] GeometryHash(this Arc curve, double translationFactor, bool skipEndPoint= false)
```

Geometry_Engine/Query/GeometryHash.cs
```diff
@@ -0,0 +1,427 @@
...
+            if (!skipLastPoint)
+                hash = hash.Concat(curve.EndPoint().ToDoubleArray(translationFactor));
+
+            return hash.ToArray();
+        }
+
+        /***************************************************/
+
+        [Description("The GeometryHash for an Circle is calculated as the GeometryHash of the start, 1/3rd and 2/3rd points of the Circle.")]
+        private static double[] GeometryHash(this Circle curve, double translationFactor, bool skipLastPoint = false)
```
IsakNaslundBh, line 93:
```suggestion
        private static double[] GeometryHash(this Circle curve, double translationFactor, bool skipEndPoint = false)
```

Would probably, as written on the arc above, change this to skipEndPoint and have it do nothing for the case of circle and Ellipse.

Not sure it makes sense to skip a point for these kinds of curve, and by renaming it with the EndPoint instead of Last, it can be semantically more reasonable to just have the argument do nothing, as a circle and ellipse kind of by definition do not have endpoints.

Geometry_Engine/Query/GeometryHash.cs
```diff
@@ -0,0 +1,427 @@
...
+            if (!skipLastPoint)
+                hash = hash.Concat(curve.PointAtParameter(0.66).ToDoubleArray(translationFactor));
+
+            return hash.ToArray();
+        }
+
+        /***************************************************/
+
+        [Description("The GeometryHash for an Ellipse is calculated as the GeometryHash of the start, 1/3rd and 2/3rd points of the Ellipse.")]
+        private static double[] GeometryHash(this Ellipse curve, double translationFactor, bool skipLastPoint = false)
```
IsakNaslundBh, line 109: as above

Geometry_Engine/Query/GeometryHash.cs
```diff
@@ -0,0 +1,427 @@
...
+            if (!skipLastPoint)
+                hash = hash.Concat(curve.PointAtParameter(0.66).ToDoubleArray(translationFactor));
+
+            return hash.ToArray();
+        }
+
+        /***************************************************/
+
+        [Description("The GeometryHash for a Line is calculated as the GeometryHash of the start and end point of the Line.")]
+        private static double[] GeometryHash(this Line curve, double translationFactor, bool skipLastPoint = false)
```
IsakNaslundBh, line 125: as above

Geometry_Engine/Query/GeometryHash.cs
```diff
@@ -0,0 +1,427 @@
...
+               .Concat(curve.EndPoint().ToDoubleArray(translationFactor))
+               .ToArray();
+        }
+
+        /***************************************************/
+
+        [Description("The GeometryHash for a NurbsCurve is obtained by getting moving the control points " +
+            "by a translation factor composed by the weights and a subarray of the knot vector. " +
+            "The subarray is made by picking as many elements from the knot vector as the curve degree value.")]
+        private static double[] GeometryHash(this NurbsCurve curve, double translationFactor, bool skipLastPoint = false)
```
IsakNaslundBh, line 142: Just noting here that we are not doing anything with the argument for the general case here, which I think probably is fine, as it can be a bit more complicated to do. Just putting as a review comment for myself that I have looked at it :)

Geometry_Engine/Query/GeometryHash.cs
```diff
@@ -0,0 +1,427 @@
...
+        {
+            int curveDegree = curve.Degree();
+
+            if (curveDegree == 1)
+                return BH.Engine.Geometry.Create.Polyline(curve.ControlPoints).GeometryHash(translationFactor);
```
IsakNaslundBh, line 147: This there could be an edge case of an edgecase that could happen here when a nurbscurve of degree 1 is part of a polycurve, and then if that nurbscurve is changed to a polyline, then the subparts might start to act slightly differently with regards to the skip last point.

This can potentially be fixed by adding a SubParts method for the NurbsCurve, and return a list of Lines corresponding to the Polyline if the curve is Degree 1.

Do not think this is anything that needs to be handled in this PR, again mostly noting in case I read this again.

IsakNaslundBh (changes requested):
Some more I think final tweaks to the handling of endpoints of subparts for the curve case. See comments

IsakNaslundBh (changes requested):
Update required for NurbsSurface

Geometry_Engine/Query/GeometryHash.cs
```diff
@@ -0,0 +1,446 @@
...
+
+            return hash.ToArray();
+        }
+
+        /***************************************************/
+
+        [Description("The GeometryHash for an Circle is calculated as the GeometryHash of the start, 1/3rd and 2/3rd points of the Circle.")]
+        private static double[] GeometryHash(this Circle curve, double translationFactor, bool skipEndPoint = false)
+        {
+            // The input `skipEndPoint` is not used here because Ellipses cannot be part of Polycurves.
```
IsakNaslundBh, line 103: This is actually not correct. Circles and Ellipses can be part of PolyCurves.

The reason for not skipping end point (the way I look at it) is that ellipses do not have clear start and endpoints that could be seen as segments along a continuous curve. Would change comment to something like:
```suggestion
            // The input `skipEndPoint` is not used here because Circles do not have a clearly defined endpoint to be used in a chain of segment curves.
```
or something to that avail.

Geometry_Engine/Query/GeometryHash.cs
```diff
@@ -0,0 +1,446 @@
...
+
+            return hash.ToArray();
+        }
+
+        /***************************************************/
+
+        [Description("The GeometryHash for an Ellipse is calculated as the GeometryHash of the start, 1/3rd and 2/3rd points of the Ellipse.")]
+        private static double[] GeometryHash(this Ellipse curve, double translationFactor, bool skipEndPoint = false)
+        {
+            // The input `skipEndPoint` is not used here because Ellipses cannot be part of Polycurves.
```
IsakNaslundBh, line 119: As above
```suggestion
            // The input `skipEndPoint` is not used here because Ellipses do not have a clearly defined endpoint to be used in a chain of segment curves.
```

Geometry_Engine/Query/GeometryHash.cs
```diff
@@ -0,0 +1,446 @@
...
+        [Description("The GeometryHash for a Pipe is calculated as the GeometryHash of its centreline translated by its radius," +
+            "then concatenated with the GeometryHash of its centreline's StartPoint for extra reliability.")]
+        private static double[] GeometryHash(this Pipe obj, double translationFactor)
+        {
+            translationFactor += (int)TypeTranslationFactor.Pipe;
+
+            double[] result = obj.Centreline.GeometryHash(translationFactor + obj.Radius);
+
+            if (obj.Capped)
+                result.Concat(obj.Centreline.StartPoint().GeometryHash(translationFactor + obj.Radius));
```
IsakNaslundBh, line 268: Need to call interface method here. This just gets to the fallback method and reports an error.
```suggestion
                result.Concat(obj.Centreline.IStartPoint().GeometryHash(translationFactor + obj.Radius));
```

Geometry_Engine/Query/GeometryHash.cs
```diff
@@ -0,0 +1,446 @@
...
+
+        [Description("The GeometryHash for an Ellipse is calculated as the GeometryHash of the start, 1/3rd and 2/3rd points of the Ellipse.")]
+        private static double[] GeometryHash(this Ellipse curve, double translationFactor, bool skipEndPoint = false)
+        {
+            // The input `skipEndPoint` is not used here because Ellipses cannot be part of Polycurves.
+
+            translationFactor += (int)TypeTranslationFactor.Ellipse;
+
+            IEnumerable<double> hash = curve.StartPoint().ToDoubleArray(translationFactor)
+               .Concat(curve.PointAtParameter(0.33).ToDoubleArray(translationFactor))
```
IsakNaslundBh, line 124: Would update this branch from main to ensure this actually functions and a working UT can be produced, as this was added in https://github.com/BHoM/BHoM_Engine/pull/2937

IsakNaslundBh (changes requested):
A few more code comments.
Also, have a check of the UT test file. The pipe and Ellipse UTs are null which should be fixed with comments made below.

The Mesh type used for the UT creation is of type `BH.oM.Environment.Elements.Mesh`, but assume it really should be `BH.oM.Geometry.Mesh`?

Also, wondering if we should add a method for CompositeGeometry, which just could loop through the elements and add up?

Happy for this last point to be handled in a separate PR though.

Geometry_Engine/Query/GeometryHash.cs
```diff
@@ -0,0 +1,451 @@
...
+
+        [Description("The GeometryHash for a PolySurface is calculated as the GeometryHash of the individual surfaces.")]
+        private static double[] GeometryHash(this PolySurface obj, double translationFactor)
+        {
+            return obj.Surfaces.SelectMany(s => s.GeometryHash(translationFactor)).ToArray();
+        }
+
+        /***************************************************/
+
+        [Description("The GeometryHash for a SurfaceTrim is calculated as the GeometryHash of its Curve3d and Curve2d, concatenated.")]
```
IsakNaslundBh, line 283:
```suggestion
        [Description("The GeometryHash for a SurfaceTrim is calculated as the GeometryHash of its Curve3d.")]
```

Geometry_Engine/Query/GeometryHash.cs
```diff
@@ -0,0 +1,451 @@
...
+
+        [Description("The GeometryHash for a Point is simply an array of 3 numbers composed by the Point X, Y and Z coordinates.")]
+        private static double[] GeometryHash(this Point obj, double translationFactor)
+        {
+            return obj.ToDoubleArray(translationFactor);
+        }
+
+        /***************************************************/
+
+        [Description("The GeometryHash for a Point is simply an array of 3 numbers composed by the Point X, Y and Z coordinates.")]
```
IsakNaslundBh, line 380: this is just copy paste from point

IsakNaslundBh (changes requested):
minor desc tweaks

IsakNaslundBh (dismissed):
Happy with this now. Have made a few tests:
- Different line representations (line, polyline, nurbscurve with two ctrlPts) as well as Polycurves consisting of the above all coming out the same
- AS above for polylines working fine
- Moving Complex NurbsSurface and checking the hash, works fine
- Moving a Panel and checking the result
- UT covers most base cases.

Code looks sensible and have had quite deep discussion about the implementation for most types.

Approved

IsakNaslundBh:
@BHoMBot check compliance
@BHoMBot check unit-tests

alelom:
@BHoMBot check required

alelom:
@BHoMBot check compliance
@BHoMBot check unit-tests

IsakNaslundBh:
@BHoMBot check required

IsakNaslundBh:
@BHoMBot check unit-tests

IsakNaslundBh (approved):
Null handling correctly added and checked for.

Re-approving.

IsakNaslundBh:
Null-handling failure is due to this PR being behind main, where another method has been fixed and merged in. Will dispensate as this is the only failure.

Will also dispensate code-compliance as agreed with @FraserGreenroyd in thread above.

IsakNaslundBh:
@BHoMBot check ready-to-merge

IsakNaslundBh:
@BHoMBot this is a DevOps instruction. I am authorising dispensation to be granted on check ref. 9549259727

IsakNaslundBh:
@BHoMBot this is a DevOps instruction. I am authorising dispensation to be granted on check ref. 9549222836

PR #2930, opened 2022-10-20
Add DisplayText support on enums

adecler:
### NOTE: Depends on 
https://github.com/BHoM/BHoM/pull/1441

   
See https://github.com/BHoM/BHoM/issues/1440

### Test files
Not a file per se since there isn't obviously a case of `DisplayText` usage in the BHoM yet. But here's a working example:

image

image

Best is to add the attribute yourself to a random enum and test that it is well supported. Happy to jump to a call if it is easier for the review though.

### DIFF
--- a/BHoM_Engine/Compute/ParseEnum.cs
+++ b/BHoM_Engine/Compute/ParseEnum.cs
@@ -64,9 +64,11 @@ namespace BH.Engine.Base
                 return Enum.GetValues(enumType).OfType<Enum>()
                     .FirstOrDefault(x => {
                         FieldInfo fi = enumType.GetField(x.ToString());
-                        DescriptionAttribute[] attributes = fi.GetCustomAttributes(typeof(DescriptionAttribute), false) as DescriptionAttribute[];
+                        DisplayTextAttribute[] displayTexts = fi.GetCustomAttributes(typeof(DisplayTextAttribute), false) as DisplayTextAttribute[];
+                        DescriptionAttribute[] descriptions = fi.GetCustomAttributes(typeof(DescriptionAttribute), false) as DescriptionAttribute[];
 
-                        return attributes != null && attributes.Count() > 0 && attributes.First().Description == value;
+                        return (displayTexts != null && displayTexts.Count() > 0 && displayTexts.First().Text == value)
+                            || (descriptions != null && descriptions.Count() > 0 && descriptions.First().Description == value);
                     });
             }
         }
--- a/BHoM_Engine/Convert/ToText.cs
+++ b/BHoM_Engine/Convert/ToText.cs
@@ -161,7 +161,7 @@ namespace BH.Engine.Base
 
         /***************************************************/
 
-        public static string ToText(Enum item, bool includePath = false)
+        public static string ToText(this Enum item, bool includePath = false)
         {
             if (item == null)
                 return "null";
--- a/BHoM_Engine/Convert/ToText.cs
+++ b/BHoM_Engine/Convert/ToText.cs
@@ -22,6 +22,7 @@
 
 using System;
 using System.Collections.Generic;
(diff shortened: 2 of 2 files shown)


adecler:
@BHoMBot check compliance
@BHoMBot check versioning

FraserGreenroyd (commented):
image
Quick comment on an initial test @adecler - are we happy with the `DisplayText` value being displayed in the `One Locally Defined Value` area at the bottom?

Personally I am (because also I would expect people will put more appropriate display text items rather than what I've tested with) but just thought I'd raise it as a question while I test

FraserGreenroyd:
@BHoMBot check compliance

adecler:
> image Quick comment on an initial test @adecler - are we happy with the `DisplayText` value being displayed in the `One Locally Defined Value` area at the bottom?
> 
> Personally I am (because also I would expect people will put more appropriate display text items rather than what I've tested with) but just thought I'd raise it as a question while I test

Yes, this is exactly what the ToText methods are for. Great to see it already is being used on a wider scope. Ideally we would always like to see enum displayed in a human readable way and never show the code version if there is a `DisplayText` attribute present. looks like the dropdown itself need to be updated though to is uses the `DisplayText` too. But one thing at a time 😄

IsakNaslundBh:
> > image Quick comment on an initial test @adecler - are we happy with the `DisplayText` value being displayed in the `One Locally Defined Value` area at the bottom?
> > Personally I am (because also I would expect people will put more appropriate display text items rather than what I've tested with) but just thought I'd raise it as a question while I test
> 
> Yes, this is exactly what the ToText methods are for. Great to see it already is being used on a wider scope. Ideally we would always like to see enum displayed in a human readable way and never show the code version if there is a `DisplayText` attribute present. looks like the dropdown itself need to be updated though to is uses the `DisplayText` too. But one thing at a time 😄

A quick question on how this `DisplayText` attribute is to be used compared to the `Description` attribute (that already has support for in the menu in https://github.com/BHoM/BHoM_Engine/pull/1586 .

Assume the DisplayText is meant to be an alias, as a short alternative name (allowing white space etc)? If so I can see a use for both. Just wanted to make sure there is not to much overlap :)

adecler:
@IsakNaslundBh , check teh description of the issue: https://github.com/BHoM/BHoM/issues/1440
You'll see that people are already using teh Description attribute for actual (longer) descriptions. So this serves the different purpose of readability in the UIs / web tools.

FraserGreenroyd (approved):
I have tested both PRs by adding the new attribute to a couple of enum attributes and it is displaying well, in combination with displaying the default values where the attribute does not exist.

I am happy to see this merged, and am happy for now to not have unit tests for the methods on the basis that we currently do not have any enum values implementing the attribute in the oM - when that changes we will then want to add some unit tests for this method for the future.

adecler:
> I have tested both PRs by adding the new attribute to a couple of enum attributes and it is displaying well, in combination with displaying the default values where the attribute does not exist.
> 
> I am happy to see this merged, and am happy for now to not have unit tests for the methods on the basis that we currently do not have any enum values implementing the attribute in the oM - when that changes we will then want to add some unit tests for this method for the future.

Thanks @FraserGreenroyd . Yes, I agree on unit tests. They should be plenty of opportunities for the `DisplaytText` attribute within the BHoM already, so we should probably cover that within a wider exercise of improving enums throughout the BHoM. A frequent complain from reviewers I received when I started doing zero-code tools was that the enums where 'not displaying properly' in the interface (weird how people don't seem to like camel casing inside a UI 😄  ). So I am sure we need to do a similar exercise for our local BHoM UIs.

adecler:
@BHoMBot check unit-tests
@BHoMBot check ready-to-merge

PR #2957, opened 2022-11-29
Geometry_Engine: Add controls to external edges for mesh

### LINKED ISSUE #2956: Geometry_Engine: Add control over ExternalEdges for mesh if to be done by geometry or topology
#### Description:

Current version relies on geometriocal matching of vertices. Should also allow for instead using the mesh topology, which should be quicker, and for some mesh types give slightly different results.

SHould also change the equality check to a tolerance based square distance check.

IsakNaslundBh:

Closes #2956

Update external edge method for Mesh to support getting Edges out by topology or geometry.

### Test files

https://burohappold.sharepoint.com/:f:/s/BHoM/ElVIwxksipROv3GK94FTYQwBXbEHfFy1Z8gk-8bz7QtN1A?e=NwaeVy


IsakNaslundBh:
@BHoMBot check compliance

IsakNaslundBh:
@BHoMBot check compliance
@BHoMBot check unit-tests

IsakNaslundBh:
@BHoMBot check required

Geometry_Engine/Query/ExternalEdges.cs
```diff
@@ -102,38 +108,74 @@ public static List<ICurve> ExternalEdges(this PolySurface surface)
...
 
             if (mesh.Faces.Count < 1)
                 return null;
 
-            List<Line> edges = mesh.Faces.SelectMany(f => f.Edges(mesh)).ToList();
+            if (filterByTopology)
+            {
+                List<Tuple<int, int>> allEdgeTopologies = new List<Tuple<int, int>>();
+
+                //Creates tuples of indecies of edges of each face, ordering the index order so the lowest number is always first (to simplify grouping lower down)
```
pawelbaran, line 129:
```suggestion
                //Creates tuples of indices of edges of each face, ordering the index order so the lowest number is always first (to simplify grouping lower down)
```

Geometry_Engine/Query/ExternalEdges.cs
```diff
@@ -102,38 +108,74 @@ public static List<ICurve> ExternalEdges(this PolySurface surface)
...
                     {
-                        edges.RemoveAt(i); // shared edge so remove both
-                        edges.RemoveAt(n);
-                        i--;
-                        break;
+                        if ((edges[i].Start.SquareDistance(edges[n].Start) < sqTol && edges[i].End.SquareDistance(edges[n].End) < sqTol) || // edge[i] == edge[n]
+                            (edges[i].Start.SquareDistance(edges[n].End) < sqTol && edges[i].End.SquareDistance(edges[n].Start) < sqTol))   // edge[i] == edge[n].Reverse()
+                        {
+                            edges.RemoveAt(i); // shared edge so remove both
+                            edges.RemoveAt(n);
```
pawelbaran, line 163: Isn't it safer to wait with removal until the end of iteration? Otherwise you may get wrong results in case of nonmanifold topology?

Geometry_Engine/Query/ExternalEdges.cs
```diff
@@ -102,38 +108,74 @@ public static List<ICurve> ExternalEdges(this PolySurface surface)
...
+        [Input("filterByTopology", "If true, edges with unique topology are returned. If false, edges with unique geometry is returned. Toggle only renders different result for meshes with duplicate nodes.")]
+        [Input("tolerance", "Tolerance to be used for identifying duplicate edges. Only used if filterByTopology is false.")]
+        [Output("edges", "The external edge curves of the mesh.")]
+        public static List<Line> ExternalEdges(this Mesh mesh, bool filterByTopology = false, double tolerance = Tolerance.Distance)
         {
             if (mesh == null)
                 return null;
 
             if (mesh.Faces.Count < 1)
                 return null;
```
pawelbaran, line 123: I would consider returning empty list here - empty list of faces should return an empty list of edges?

pawelbaran (changes requested):
I have left a few minor comments to the code. Additionally, I would rather spend 1 minute to add missing input descriptions (should be really simple in this case) and be done with it 🙈

Geometry_Engine/Query/ExternalEdges.cs
IsakNaslundBh: Sure, can get behind that

Geometry_Engine/Query/ExternalEdges.cs
IsakNaslundBh: Yeah, you are right. Had not changed this really from what it was before, But might as well get it fixed I guess.

Scope for a more deep refactoring of this though, making use of some more cleaver datastructures such as the domain trees. Might have a go at that as well.

pawelbaran:
@BHoMBot check compliance

IsakNaslundBh:
@BHoMBot check compliance

IsakNaslundBh:
@BHoMBot check compliance

pawelbaran (approved):
Happy with the code and functionality, with one minor comment: during testing I have found out that `ExternalEdges` for meshes returns only the edges that adjoin to only one face, which makes perfect sense, but is counterintuitive in case of three-dimensional meshes, i.e. for a cuboidal mesh the method would return an empty collection, while one could possibly expect the edges of the cuboid. This could be explained a bit better in the description to avoid confusion on the user's side.

FraserGreenroyd:
@BHoMBot check required
@BHoMBot check unit-tests

FraserGreenroyd:
@BHoMBot check ready-to-merge

PR #2965, opened 2022-12-20
Reflection_Engine: Subtypes query added

### LINKED ISSUE #2964: Reflection_Engine: Add a method to query subtypes of a type
#### Description:

In some cases (mainly in adapter actions - see https://github.com/BHoM/Mongo_Toolkit/issues/166) it is needed to get all subtypes of a type, so would be good to add a method that does it.

pawelbaran:

Closes #2964

### Test files

On [SharePoint](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?RootFolder=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F02%5FPull%20Request%2FBHoM%2FBHoM%5FEngine%2FReflection%5FEngine%2F%232965%2DSubtypesQuery&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4)

### Changelog

`ImplementingTypes` and `Subtypes` are essentially the same method, but I did not dare to merge them taken slightly different use cases (at least in my mind). However, happy to do it if requested in the review.


FraserGreenroyd (commented):
I can see the primary differences between the methods is that one is specifically for interfaces and the other (the new one, `SubTypes`) is not.

Originally, I would say I wouldn't see the point of the second given the BHoM inheritance structure prohibits object inheritance, however we have broken that rule in strategic places, so I can see a reason for wanting to know the subtypes of non-interface types.

But I'm not sure we need two methods to achieve this. I think a `bool` flag on `ImplementingTypes` for `interfaceOnly` would suffice? While I like the attempt to keep it tidy by having one call the other, I'm just not convinced the use case for individual methods is that strong as opposed to a modification of the existing one?

Though maybe the default bool could cause confusion so I can also see an argument for having them as separate methods... Happy to see what others think for a final decision I think.

pawelbaran:
Hmm, I got confused now @FraserGreenroyd. The facts as I see them:
- both methods return only non-abstract, non-interface BHoM types
- `ImplementingTypes` method accepts only interface types, while `Subtypes` will be happy with any input of type `Type`
- technically `ImplementingTypes` is the same method as `Subtypes`, just more strict on the input properties

...`bool` input parameter seems a bit redundant to me in this case, I was more considering whether we need `ImplementingTypes` altogether. Or am I missing something?

Reflection_Engine/Query/Subtypes.cs
```diff
@@ -0,0 +1,62 @@
+﻿/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2022, the respective contributors. All rights reserved.
```
FraserGreenroyd, line 3:
```suggestion
 * Copyright (c) 2015 - 2023, the respective contributors. All rights reserved.
```

FraserGreenroyd (changes requested):
Happy with the discussion from above - could we generate a unit test for this before merge though @pawelbaran ?

pawelbaran:
I will generate UTs on Monday @FraserGreenroyd - thanks!

Reflection_Engine/Query/Subtypes.cs
```diff
@@ -0,0 +1,62 @@
...
+
+            return implemented;
+        }
+
+        /***************************************************/
+    }
+}
+
+
+
```
FraserGreenroyd, line 62:
```suggestion
```

FraserGreenroyd (changes requested):

FraserGreenroyd:
@BHoMBot check unit-tests

FraserGreenroyd:
@BHoMBot check unit-tests
@BHoMBot check ready-to-merge

FraserGreenroyd (approved):
Code works and looks good, happy to deploy to support the Mongo work as this can be merged independently.

FraserGreenroyd:
@BHoMBot check ready-to-merge

FraserGreenroyd:
@BHoMBot this is a DevOps instruction. I am requesting neutral checks on: ready-to-merge

PR #2998, opened 2023-01-24
BHoM_Engine: Add methods to record events from exceptions

### LINKED ISSUE #2011: Reflection_Engine: Add record event methods that accept Exceptions
#### Description:

Suggestion to adding additional RecordEvent methods, RecordError, RecordWarning and RecordNote methods that accepts an exception as an input.

This way we could centralise extraction of the exception message, together with some additional error text provided together with the exception.

Can the also use the call stack from the exception rather than extracting the call stack ourselves, which might help with debugging.

@adecler @al-fisher @alelom @FraserGreenroyd

FraserGreenroyd:
Fixes #2011 

### Additional comments
Code review should be suitable for these methods.

### DIFF
--- a/BHoM_Engine/Compute/RecordError.cs
+++ b/BHoM_Engine/Compute/RecordError.cs
@@ -23,6 +23,8 @@
 using BH.oM.Base.Debugging;
 using BH.oM.Base.Attributes;
 using System.Linq;
+using System.ComponentModel;
+using System;
 
 namespace BH.Engine.Base
 {
@@ -32,13 +34,26 @@ namespace BH.Engine.Base
         /**** Public Methods                            ****/
         /***************************************************/
 
+        [Description("Record an error within the BHoM logging system.")]
+        [Input("message", "The message to be recorded for this error.")]
+        [Output("success", "True if the error has been successfully recorded as a BHoM Event.")]
         public static bool RecordError(string message)
         {
             return RecordEvent(new Event { Message = message, Type = EventType.Error });
         }
+
+        [Description("Record an error with details of a C# exception within the BHoM logging system.")]
+        [Input("exception", "The C# exception being caught to provide the error and stack information for.")]
+        [Input("message", "An optional additional message which will be displayed first in the event log.")]
+        [Output("success", "True if the error has been successfully recorded as a BHoM Event.")]
+        public static bool RecordError(Exception exception, string message = "")
+        {
+            if (!string.IsNullOrEmpty(message))
+                message = $"{message}\n\n{exception.ToString()}";
+            else
+                message = exception.ToString();
+
+            return RecordEvent(new Event { Message = message, StackTrace = exception.StackTrace, Type = EventType.Error });
+        }
     }
 }
-
-
-
-
--- a/BHoM_Engine/Compute/RecordError.cs
+++ b/BHoM_Engine/Compute/RecordError.cs
@@ -50,12 +50,7 @@ namespace BH.Engine.Base
         [Output("success", "True if the error has been successfully recorded as a BHoM Event.")]
         public static bool RecordError(Exception exception, string message = "")
         {
-            if (!string.IsNullOrEmpty(message))
-                message = $"{message}\n\n{exception.ToString()}";
-            else
-                message = exception.ToString();
-
-            return RecordEvent(new Event { Message = message, StackTrace = exception.StackTrace, Type = EventType.Error });
+            return RecordEvent(exception, message, EventType.Error);
         }
     }
 }
--- a/BHoM_Engine/Compute/RecordError.cs
+++ b/BHoM_Engine/Compute/RecordError.cs
@@ -42,6 +42,8 @@ namespace BH.Engine.Base
             return RecordEvent(new Event { Message = message, Type = EventType.Error });
         }
 
+        /***************************************************/
+
         [Description("Record an error with details of a C# exception within the BHoM logging system.")]
         [Input("exception", "The C# exception being caught to provide the error and stack information for.")]
         [Input("message", "An optional additional message which will be displayed first in the event log.")]
--- a/BHoM_Engine/Compute/RecordEvent.cs
+++ b/BHoM_Engine/Compute/RecordEvent.cs
@@ -57,7 +57,11 @@ namespace BH.Engine.Base
             Exception e = exception;
             while(e != null)
             {
-                exceptionMessage += $"{e.Message}{Environment.NewLine}{Environment.NewLine}";
+                if (!string.IsNullOrEmpty(exceptionMessage))
+                    exceptionMessage += $"{Environment.NewLine}{Environment.NewLine}";
+
+                exceptionMessage += $"{e.Message}";
+
                 e = e.InnerException;
             }
 
--- a/BHoM_Engine/Compute/RecordEvent.cs
+++ b/BHoM_Engine/Compute/RecordEvent.cs
@@ -46,6 +46,22 @@ namespace BH.Engine.Base
 
         /***************************************************/
 
+        [Description("Record an event with details of a C# exception within the BHoM logging system.")]
+        [Input("exception", "The C# exception being caught to provide the event and stack information for.")]
+        [Input("message", "An optional additional message which will be displayed first in the event log.")]
+        [Output("success", "True if the event has been successfully recorded as a BHoM Event.")]
+        public static bool RecordEvent(Exception exception, string message = "")
+        {
+            if (!string.IsNullOrEmpty(message))
+                message = $"{message}\n\n{exception.ToString()}";
+            else
+                message = exception.ToString();
+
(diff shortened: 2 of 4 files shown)


FraserGreenroyd:
@BHoMBot check compliance

BHoM_Engine/Compute/RecordEvent.cs
```diff
@@ -46,6 +46,22 @@ public static bool RecordEvent(string message, EventType type = EventType.Unknow
...
+        [Input("message", "An optional additional message which will be displayed first in the event log.")]
+        [Output("success", "True if the event has been successfully recorded as a BHoM Event.")]
+        public static bool RecordEvent(Exception exception, string message = "")
+        {
+            if (!string.IsNullOrEmpty(message))
+                message = $"{message}\n\n{exception.ToString()}";
+            else
+                message = exception.ToString();
+
+            return RecordEvent(new Event { Message = message, StackTrace = exception.StackTrace, Type = EventType.Unknown });
```
IsakNaslundBh, line 60: Would get out the message from the exception here, including inner exception, as well as providing settable event type as a parameter, so this can be called by the other methods.

```suggestion
                public static bool RecordEvent(Exception exception, string message = "", EventType type = EventType.Unknown)
        {
            string exceptionMessage = "";

            Exception e = exception;
            while (e != null)
            { 
                exceptionMessage += e.Message + Environment.NewLine;
                e = e.InnerException;
            }

            if (!string.IsNullOrEmpty(message))
                message = $"{message}\n\n{exceptionMessage}";
            else
                message = exceptionMessage;

            return RecordEvent(new Event { Message = message, StackTrace = exception.StackTrace, Type = type });

```

Also need to update the method bellow, that logs the events, as that currently is overwriting the stack trace

BHoM_Engine/Compute/RecordError.cs
```diff
@@ -32,13 +34,28 @@ public static partial class Compute
...
+        [Input("message", "An optional additional message which will be displayed first in the event log.")]
+        [Output("success", "True if the error has been successfully recorded as a BHoM Event.")]
+        public static bool RecordError(Exception exception, string message = "")
+        {
+            if (!string.IsNullOrEmpty(message))
+                message = $"{message}\n\n{exception.ToString()}";
+            else
+                message = exception.ToString();
 
+            return RecordEvent(new Event { Message = message, StackTrace = exception.StackTrace, Type = EventType.Error });
```
IsakNaslundBh, line 58: With changes to the RecordEvent, can simply change to this:

```suggestion
            return RecordEvent(exception, message, EventType.Error);
```

Holds true for Warning and Not as well

IsakNaslundBh (changes requested):
Some potential centralisation fo logic to single method RecordEvent that all other new methods can call.

Also, if we want to make use of the stacktrace from the event, we need to change the main RecordEvent method that currently overwrites it.

FraserGreenroyd:
@BHoMBot check compliance

FraserGreenroyd:
@BHoMBot check compliance

BHoM_Engine/Compute/RecordEvent.cs
```diff
@@ -46,6 +46,31 @@ public static bool RecordEvent(string message, EventType type = EventType.Unknow
...
+                exceptionMessage += $"{e.Message}{Environment.NewLine}{Environment.NewLine}";
+                e = e.InnerException;
+            }
+
+            if (!string.IsNullOrEmpty(message))
+                message = $"{message}\n\n{exceptionMessage}";
+            else
+                message = exceptionMessage;
+
+            return RecordEvent(new Event { Message = message, StackTrace = exception.StackTrace, Type = EventType.Unknown });
```
alelom, line 69: Just leaving here my personal opinion, nothing wrong with the PR. The mechanism that we currently have and that this PR is pushing is well working and well suited to UIs.
My personal preference would be to store the exception in the Event, and have the `RecordEvent` mechanism simply recording the exception itself, possibly with a message. When data extracted from the exception is needed (e.g. the StackTrace, or the concatenated InnerExceptions) we could have another method that does the necessary processing to the Event.Exception data as required by the target context (e.g. the UI). This would allow to deal with the recorded events in different ways in different contexts (i.e. console applications) and would also expose more exception data that may be useful to filter certain events if required by the target context.

alelom, line 69: Do we need to pass in the `type` from the inputs here?
```suggestion

            return RecordEvent(new Event { Message = message, StackTrace = exception.StackTrace, Type = type  });
```

BHoM_Engine/Compute/RecordEvent.cs
```diff
@@ -46,6 +46,31 @@ public static bool RecordEvent(string message, EventType type = EventType.Unknow
...
+        [Input("message", "An optional additional message which will be displayed first in the event log.")]
+        [Output("success", "True if the event has been successfully recorded as a BHoM Event.")]
+        public static bool RecordEvent(Exception exception, string message = "", EventType type = EventType.Unknown)
+        {
+            string exceptionMessage = "";
+
+            Exception e = exception;
+            while(e != null)
+            {
+                exceptionMessage += $"{e.Message}{Environment.NewLine}{Environment.NewLine}";
```
pawelbaran, line 60: @IsakNaslundBh already did a proper review of the PR, so only a minor code comment from me. From what I understand, `exceptionMessage` will always end with 2 line breaks. This may not look ideal neither in logs nor in the UI, so I would consider trimming the string at some point or slightly rewriting the code to something around below:

```suggestion
                if (!string.IsNullOrWhiteSpace(exceptionMessage ))
                    exceptionMessage += $"{Environment.NewLine}{Environment.NewLine}";
                
                exceptionMessage += e.Message;
```

This comment has been written without running the code, so please ignore if I missed a point 🙈

BHoM_Engine/Compute/RecordEvent.cs
FraserGreenroyd: If we were to do this I would prefer it be a new issue as that would involve changing the event object itself and broadens the scope of this work considerably. I think that falls better into the logging discussions that I've had with @pawelbaran recently too and is probably one to discuss in more detail during our upcoming hack sessions, so if it's ok with you @alelom I'll mark this as resolved on the basis of being out of scope for this PR but one for us to discuss at the upcoming hack sessions (feel free to unresolve if you disagree!) 😄

BHoM_Engine/Compute/RecordEvent.cs
FraserGreenroyd: Makes sense, implemented.

FraserGreenroyd:
@BHoMBot check compliance

FraserGreenroyd:
@BHoMBot check compliance

kprusicka (approved):
Looks good to me

BHoM_Engine/Compute/RecordEvent.cs
```diff
@@ -46,6 +46,35 @@ public static bool RecordEvent(string message, EventType type = EventType.Unknow
 
         /***************************************************/
 
+        [Description("Record an event with details of a C# exception within the BHoM logging system.")]
+        [Input("exception", "The C# exception being caught to provide the event and stack information for.")]
+        [Input("message", "An optional additional message which will be displayed first in the event log.")]
+        [Output("success", "True if the event has been successfully recorded as a BHoM Event.")]
+        public static bool RecordEvent(Exception exception, string message = "", EventType type = EventType.Unknown)
```
IsakNaslundBh, line 53: Might as well get this input arg added I think

IsakNaslundBh (changes requested):
nitty gritty desc, apart from that, from just code inspection, looks good. Will test and if all clear, approve

BHoM_Engine/Compute/RecordEvent.cs
alelom: Absolutely! The comment was just to record my opinion and to have further discussions down the line. Thanks!

BHoM_Engine/Compute/RecordEvent.cs
FraserGreenroyd: Agreed, added.

FraserGreenroyd:
@BHoMBot check compliance

FraserGreenroyd:
@BHoMBot check versioning
@BHoMBot check required

FraserGreenroyd:
@BHoMBot check null-handling

FraserGreenroyd:
@BHoMBot check compliance

FraserGreenroyd:
@BHoMBot check versioning

FraserGreenroyd:
@BHoMBot check core
@BHoMBot check serialisation
@BHoMBot check unit-tests

IsakNaslundBh (approved):
Tested, and functioning as intended

FraserGreenroyd:
@BHoMBot check ready-to-merge

FraserGreenroyd:
@BHoMBot this is a DevOps instruction. I am requesting neutral checks on: null-handling

Issue #943, opened 2019-05-07
Geometry_Engine: Remove unnecessary clones


LMarkowski:
e.g. in
```
        public static BoundingBox Bounds(this BoundingBox boundingBox)
        {
            return boundingBox.Clone();
        }
```

FraserGreenroyd:
Well it no longer clones but now it deep clones instead 🤦‍♂️ so should probably still fix it...

PR #2999, opened 2023-01-25
Geometry_Engine: Add missing methods for ellipse

### LINKED ISSUE #314: Geometry_Engine: Ellipse methods not existing
Collection of methods not existing for ellipse class needed to at least get their method declaration done as the corresponding method exist for ICurve:

Query:

 - [x] Area
 - [x] Clone
 - [x] ClosestPoint
 - [x] ControlPoints
 - [x] DiscontinuityPoints
 - [x] EndDir
 - [x] EndPoint
 - [x] IsClosed
 - [x] IsContaining
 - [x] IsInPlane
 - [x] IsLinear
 - [x] IsPlanar
 - [x] IsPolyLinear
 - [x] Length
 - [x] LineIntersections
 - [ ] MeshIntersections
 - [x] ParameterAtPoint
 - [x] PlaneIntersections
 - [x] PointAtParameter
 - [x] StartDIr

Compute:

 - [x] FitPlane
 - [x] SortAlongCurve

Modify:

 - [x] Flip
 - [x] Mirror
 - [x] Project
 - [x] Rotate
 - [x] Scale
 - [ ] SplitAtPoints
 - [x] Transform
 - [x] Translate

IsakNaslundBh:

Closes #314

Adding all of the (relatively) low hanging methods mentioned in the issue.

Intentionally skipping:
- MeshIntersection - which we do not generally support for most curve types
- SplitAtPoint - requires either Nurbs implementation for split or EllipticalArc class to be added.

For both of those cases I think the issue is fine to be closed and for those two cases to be handled separately.

### Test files

Test file testing through general cases, and also attempting to find as many edge cases that I could think of.
Testing through ellipses of a wide range of aspect ratios and size. 

~~Only real failure noted in the script is for some plane intersections for ellipse with ridiculous aspect ratio of 1 : 10 000 000, which is still close to correct. This failure seem to simply be a numerical tolerance issue, that I attempted to fix as much as possible, but could not figure out to get it better. Basically, the maths are correct, but for some of these extreme cases, the method might give slightly erroneous results for the intersection methods. Overall, though, they are working much better than the built in rhino versions.~~

**EDIT** fixed the plane intersection - should be fine now

https://burohappold.sharepoint.com/:f:/s/BHoM/EkzmvWVps5FOnYoqdle_gBMBCwtmchnT5Qp6iWCbWFTivg?e=CCgMBX

### Additional comments

Have unit tests ready to be pushed based on all the tests preformed in the testfile. Will need to update existing test datasets, so holding of pushing until review is done, to make sure I only have to push them up once.


IsakNaslundBh:
@BHoMBot check compliance

IsakNaslundBh:
@BHoMBot check compliance

IsakNaslundBh:
@BHoMBot check compliance

IsakNaslundBh:

IsakNaslundBh:

IsakNaslundBh:

IsakNaslundBh:

IsakNaslundBh:

IsakNaslundBh:
@BHoMBot check compliance
@BHoMBot check null-handling

IsakNaslundBh:
@BHoMBot check compliance
@BHoMBot check null-handling

Geometry_Engine/Query/ClosestPoint.cs
```diff
@@ -33,9 +37,13 @@ public static partial class Query
         /**** Public Methods - Vectors                  ****/
         /***************************************************/
 
+        [Description("Finds the closest point in the provided set of points in relation to the provided Point.")]
+        [Input("cloud", "The set of points from which to find the closest to the provided point.")]
+        [Input("point", "The reference point. The point in the cloud with the smallest distance to this point will be returned.")]
+        [Output("closePt", "The point in the cloud with the smallest distance to the provided point.")]
```
pawelbaran, line 43: Refers to all methods in this file

```suggestion
        [Output("closestPt", "The point in the cloud with the smallest distance to the provided point.")]
```

Geometry_Engine/Query/ClosestPoint.cs
```diff
@@ -106,6 +136,37 @@ public static Point ClosestPoint(this Circle circle, Point point, double toleran
...
+        [Output("closePt", "The point on the Ellipse with the smallest distance to the provided point.")]
+        public static Point ClosestPoint(this Ellipse ellipse, Point point, double tolerance = Tolerance.Distance)
+        {
+            if (ellipse.IsNull() || point.IsNull())
+                return null;
+
+            //Tranform the point to the local coordinates of the ellipse
+            //After tranformation can see it as the ellipse centred in the origin with first axis long global x and second axis along global y
+            Cartesian coordinateSystem = Create.CartesianCoordinateSystem(ellipse.Centre, ellipse.Axis1, ellipse.Axis2);
+            TransformMatrix transform = Create.OrientationMatrixLocalToGlobal(coordinateSystem);
```
pawelbaran, line 152: Sanity check: shouldn't it be global to local, taken we are transforming from global to local?

Geometry_Engine/Query/ClosestPoint.cs
```diff
@@ -106,6 +136,37 @@ public static Point ClosestPoint(this Circle circle, Point point, double toleran
...
+            Point ptLoc = point.Transform(transform);
+
+            //Get the closest point in local coordinates of the ellipse
+            Point closePt = ClosestPointEllipseLocal(ellipse.Radius1, ellipse.Radius2, ptLoc, tolerance);
+
+            //Tranform back to global coordinates
+            transform = Create.OrientationMatrixGlobalToLocal(coordinateSystem);
```
pawelbaran, line 159: Same as above, just opposite 😉

Geometry_Engine/Query/ClosestPoint.cs
```diff
@@ -202,6 +277,144 @@ private static Point ClosestPoint(this IGeometry geometry, Point point)
...
+
+            //When h is equal to 1, the ellipse is a line
+            //The algorithm below will not be able to handle to elongated ellipses, hence 
+            //pointless to evaluate.
+            if (min == 0 || aspectRatio > 1e20)
+            {
+                //Raise a warning when b is not exactly equal to 0
+                if (b != 0)
+                {
+                    Base.Compute.RecordWarning("The aspect ratio of the provided Ellipse is to large to be able to accurately evaluate the Closest point. Point on line between vertex and co-vertex returned.");
```
pawelbaran, line 316:
```suggestion
                    Base.Compute.RecordWarning("The aspect ratio of the provided Ellipse is too large to be able to accurately evaluate the Closest point. Point on line between vertex and co-vertex returned.");
```

Geometry_Engine/Query/ClosestPoint.cs
```diff
@@ -181,11 +253,14 @@ public static Point ClosestPoint(this PlanarSurface surface, Point point)
...
 
-
         /***************************************************/
         /**** Public Methods - Interfaces               ****/
         /***************************************************/
 
+        [Description("Finds the closest point on the geometry in relation to the provided Point.")]
+        [Input("geometry", "The geometry on which to find the closest point in relation to the provided point.")]
+        [Input("point", "The reference point. The point on the geometry with the smallest distance to this point will be returned.")]
+        [Output("closePt", "The point on the geometry with the smallest distance to the provided point.")]
```
pawelbaran, line 263: No fun that the tolerance is not passed here

Geometry_Engine/Query/Distance.cs
```diff
@@ -85,41 +108,81 @@ public static double Distance(this Point point, Plane plane)
...
+        [Input("infiniteSegment", "If true, distance will be computed to the closest point on the infinite line. If false, the distance will be compated to the closest point on the finite line segment.")]
+        [Output("dist", "The Euclidean distance between the Point and the Line.", typeof(Length))]
         public static double Distance(this Point point, Line line, bool infiniteSegment = false)
         {
             return point.Distance(line.ClosestPoint(point, infiniteSegment));
         }
 
         /***************************************************/
 
+        [Description("Computes and returns the Square distance between the Point and the closest point on the Line.")]
```
pawelbaran, line 123:
```suggestion
        [Description("Computes and returns the square distance between the Point and the closest point on the Line.")]
```

Geometry_Engine/Query/EndDir.cs
```diff
@@ -50,20 +57,40 @@ public static Vector EndDir(this Circle circle)
 
         /***************************************************/
 
+        [Description("Gets the tangent vector at the end of the Ellipse.")]
+        [Input("ellipse", "The Ellipse from which to get the tangent at the end point.")]
+        [Output("endTan", "The tangent vector at the end of the Ellipse.")]
+        public static Vector EndDir(this Ellipse ellipse)
+        {
+            //Ellipse starts and ends at the first axis point, and move towards the second axis, hence second axis is the tangent at start and end
```
pawelbaran, line 65: This could possibly be said in the description instead of a code comment

Geometry_Engine/Query/IsContaining.cs
```diff
@@ -167,6 +169,154 @@ public static bool IsContaining(this Circle curve, List<Point> points, bool acce
...
+            if (minRad < tolerance && !acceptOnEdge)
+            {
+                //If the minimum radius is lower than tolerance, the points will always be either on the edge or outside
+                return false;  
+            }
+
+            //Tranform the point to the local coordinates of the ellipse
+            //After tranformation can see it as the ellipse centred in the origin with first axis long global x and second axis along global y
+            Cartesian coordinateSystem = Create.CartesianCoordinateSystem(curve.Centre, curve.Axis1, curve.Axis2);
+            TransformMatrix transform = Create.OrientationMatrixLocalToGlobal(coordinateSystem);
```
pawelbaran, line 195: Again, is it only me to whom the name sounds quite opposite to what it actually does? 🙈

Geometry_Engine/Query/IsContaining.cs
```diff
@@ -167,6 +169,154 @@ public static bool IsContaining(this Circle curve, List<Point> points, bool acce
...
+            }
+
+
+            //To handle tolerance in relation to edge, a first pass is made that finds points clearly outside/inside the ellipse
+            //Then a second pass is made to check if the points near the edge actually are on the edge or not
+            //Implemented in this fashion for efficency reasons, as the first loop is significantly less expensive than the iterative method of finding
+            //The closest point on an ellipse
+            List<Point> ptsNearEdge = new List<Point>();
+
+            //Scale factors for creation of ellipses that are guarantiued to contain points on the edge.
```
pawelbaran, line 221:
```suggestion
            //Scale factors for creation of ellipses that are guaranteed to contain points on the edge.
```

Geometry_Engine/Query/IsContaining.cs
```diff
@@ -167,6 +169,154 @@ public static bool IsContaining(this Circle curve, List<Point> points, bool acce
...
+            //A point that is on the edge is guarantiued to be inside the ellipse scaled up by the scaleOutside and guarantiued to be outside an ellipse scaled by the scaleInside
```
pawelbaran, line 222:
```suggestion
            //A point that is on the edge is guaranteed to be inside the ellipse scaled up by the scaleOutside and guaranteed to be outside an ellipse scaled by the scaleInside
```

Geometry_Engine/Query/IsPolylinear.cs
```diff
@@ -46,27 +51,51 @@ public static bool IsPolylinear(this PolyCurve curve)
...
         }
 
         /***************************************************/
 
+        [Description("Returns if the curve is polylinear or not. For a Ellipse this always returns false.")]
+        [Input("curve", "The Ellipse to check if it is polylinear.")]
+        [Output("isLinear", "Return true if the curve is polylinear. For a Ellipse this always return false.")]
+        public static bool IsPolylinear(this Ellipse curve)
+        {
+            //Could argue that the ellipse is polylinear if one of the radii is 0, but given the use of this method,
```
pawelbaran, line 79: This is a Pandora's box - same could be said about linearity of an infinitely large arc (or even a circle, why not). So supportive towards this approach

Geometry_Engine/Query/LineIntersection.cs
```diff
@@ -258,6 +259,160 @@ public static List<Point> LineIntersections(this Circle circle, Line line, bool
...
+        [Description("Calculates and returns the intersection points of a Line and a Ellipse.")]
+        [Input("ellipse", "Ellipse to intersect with the Line.")]
+        [Input("line", "Line to intersect with the Ellipse.")]
+        [Input("useInfiniteLine", "If true or if the Infinite property of the Line is true, a intersection point found that is outside the domain of the Lines start and end point is accepted. If false, and the Lines Infinite property is false, the found intersection point needs to be on the finite line segment, between or on the start and end point.")]
+        [Input("tolerance", "Distance tolerance to be used in the method. Used for checking if the intersection point is within acceptable distance from the Lines.", typeof(Length))]
+        [Output("intersections", "The intersection points of the Ellipse and the Line.")]
+        public static List<Point> LineIntersections(this Ellipse ellipse, Line line, bool useInfiniteLine = false, double tolerance = Tolerance.Distance)
+        {
+            if (ellipse.IsNull() || line.IsNull())
+                return new List<Point>();
```
pawelbaran, line 271: I would return `null` output for `null` inputs

Geometry_Engine/Query/PlaneIntersection.cs
```diff
@@ -177,6 +204,94 @@ public static List<Point> PlaneIntersections(this Circle curve, Plane plane, dou
...
 
+        [Description("Computes and returns the intersection point(s) between the Ellipse and the Plane. Returns an empty list if the plane is parallel with the curve plane, including the case where the curve is fully within the plane.")]
+        [Input("curve", "The Ellipse to intersect with the Plane.")]
+        [Input("plane", "The Plane to intersect with the Ellipse.")]
+        [Input("tolerance", "Distance tolerance to be used in the method.", typeof(Length))]
+        [Output("interPts", "The resulting intersection points between the Ellipse and the Plane.")]
+        public static List<Point> PlaneIntersections(this Ellipse curve, Plane plane, double tolerance = Tolerance.Distance)
+        {
+            if (curve.IsNull())
+                return new List<Point>();
```
pawelbaran, line 215: As above, I would return `null` in this case

Geometry_Engine/Query/PlaneIntersection.cs
```diff
@@ -177,6 +204,94 @@ public static List<Point> PlaneIntersections(this Circle curve, Plane plane, dou
...
+                        return new List<Point>();   //Not overlapping and not tangential
+                }
+
+                sqrtTerm = Math.Sqrt(sqrtTerm);
+                ts.Add(2 * Math.Atan((b - sqrtTerm) / (a - c)));
+                ts.Add(2 * Math.Atan((b + sqrtTerm) / (a - c)));
+            }
+
+            //Function for getting the point on the ellipse based ont the angle parameter
+            Func<double, Point> elipsePt = t => curve.Centre + Math.Cos(t) * curve.Radius1 * curve.Axis1 + Math.Sin(t) * curve.Radius2 * curve.Axis2;
```
pawelbaran, line 269:
```suggestion
            Func<double, Point> ellipsePt = t => curve.Centre + Math.Cos(t) * curve.Radius1 * curve.Axis1 + Math.Sin(t) * curve.Radius2 * curve.Axis2;
```

Geometry_Engine/Query/PlaneIntersection.cs
```diff
@@ -177,6 +204,94 @@ public static List<Point> PlaneIntersections(this Circle curve, Plane plane, dou
...
+
```
pawelbaran, line 270: on blank could go

Geometry_Engine/Query/StartDir.cs
```diff
@@ -50,20 +57,40 @@ public static Vector StartDir(this Circle circle)
 
         /***************************************************/
 
+        [Description("Gets the tangent vector at the start of the Ellipse.")]
+        [Input("ellipse", "The Ellipse from which to get the tangent at the start point.")]
+        [Output("startTan", "The tangent vector at the start of the Ellipse.")]
+        public static Vector StartDir(this Ellipse ellipse)
+        {
+            //Ellipse starts and ends at the first axis point, and move towards the second axis, hence second axis is the tangent at start and end
```
pawelbaran, line 65: As in the previous comment

Geometry_Engine/Query/StartDir.cs
```diff
@@ -34,13 +35,19 @@ public static partial class Query
...
         public static Vector StartDir(this Arc arc)
         {
             return arc.CoordinateSystem.Y.Rotate(arc.StartAngle, arc.CoordinateSystem.Z);
         }
 
         /***************************************************/
 
+        [Description("Gets the tangent vector at the start of the Circle.")]
+        [Input("circle", "The Circle from which to get the tangent at the start point.")]
+        [Output("startTan", "The tangent vector at the start of the Circle.")]
```
pawelbaran, line 50: Shouldn't it be `startDir` to fit into method naming convention? Refers to all methods here, also end dir

Geometry_Engine/Query/TangentAtParameter.cs
```diff
@@ -35,57 +37,130 @@ public static partial class Query
...
+        [Output("tan", "The tangent vector at the provided parameter.")]
+        [PreviousInputNames("t", "parameter")]
+        public static Vector TangentAtParameter(this Ellipse curve, double t, double tolerance = Tolerance.Distance)
+        {
+            if (curve.IsNull())
+                return null;
+
+            if (t < 0)
+                t = 0;
+            if (t > 1)
```
pawelbaran, line 99:
```suggestion
            else if (t > 1)
```

pawelbaran (changes requested):
I have gone through the code - lots of good math in here 👍 Overall happy with the code itself, but left a few minor suggestions. Functionality to be stress tested tomorrow!

Geometry_Engine/Compute/SortAlongCurve.cs
```diff
@@ -69,6 +83,42 @@ public static List<Point> SortAlongCurve(this List<Point> points, Circle circle,
...
+        public static List<Point> SortAlongCurve(this List<Point> points, Ellipse ellipse, double distanceTolerance = Tolerance.Distance, double angleTolerance = Tolerance.Angle)
+        {
+            if(points == null || points.Count == 0)
+                return new List<Point>();
+
+            if (ellipse.IsNull())
+                return new List<Point>();
+
+            if (ellipse.Radius1 <= distanceTolerance && ellipse.Radius2 <= distanceTolerance)
+                return points.Select(p => p.DeepClone()).ToList();
```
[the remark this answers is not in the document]
IsakNaslundBh: Thought about that as well, and generally agree. For this first run I basically just did what the other methods did, even if I kind of thought it was unnecessary. 

Happy to remove for this method, but slightly hesitant to do it on the others, just in case there are some weird edge case where another method somewhere assumes that the returned points will be new points, and free to modify. Alternatively we can keep the deep clone on this method, for now, and then remove in separate run, while resolving https://github.com/BHoM/BHoM_Engine/issues/943 .

What do you think?

Geometry_Engine/Compute/SortAlongCurve.cs
```diff
@@ -69,6 +83,42 @@ public static List<Point> SortAlongCurve(this List<Point> points, Circle circle,
...
 
+        [Description("Sorts a set of points along the provided curve. Sorting is done based on the curve parameter at the position of the closest point on the curve in relation to each provided Point.")]
+        [Input("points", "The list of points to sort.")]
+        [Input("ellipse", "The Ellipse to sort along.")]
+        [Input("distanceTolerance", "Tolerance to be used for distance computations in the method.", typeof(Length))]
+        [Input("angleTolerance", "Tolerance to be used for angle computations in the methods.", typeof(Angle))]
+        [Output("sortedPts", "The provided points sorted along the curve.")]
+        public static List<Point> SortAlongCurve(this List<Point> points, Ellipse ellipse, double distanceTolerance = Tolerance.Distance, double angleTolerance = Tolerance.Angle)
+        {
+            if(points == null || points.Count == 0)
```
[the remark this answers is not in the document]
IsakNaslundBh: Yeah, can change to that for null collection, but return empty collection for empty input.

Geometry_Engine/Compute/SortAlongCurve.cs
```diff
@@ -69,6 +83,42 @@ public static List<Point> SortAlongCurve(this List<Point> points, Circle circle,
...
+                return new List<Point>();
+
+            if (ellipse.IsNull())
+                return new List<Point>();
```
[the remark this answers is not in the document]
IsakNaslundBh: Slightly less sure here, but could argue it both ways really. Happy to change to null return if you feel strongly about it though :)

Geometry_Engine/Query/ClosestPoint.cs
IsakNaslundBh: Ha, yeah, these always confuse me as well, but, no, should be like this. The transformation of LocalToGlobal would put the ellipse in the world origin, with Axis1 aligned with global X and Axis2 aligned with global Y. We do not need to make that transformation for the ellipse itself, as we can just assume it is there, but applying the same transformation to the point will essentially put it in the same reference R3, where it can be related to the world system as if it was the ellipse system.

Also, if it should be the other way around, the test scripts would fail miserably! :D (And in all honestly, might have had it the wrong way around first, with it failing, and then switching 😄 )

Geometry_Engine/Query/ClosestPoint.cs
IsakNaslundBh: as above :)

Geometry_Engine/Query/ClosestPoint.cs
IsakNaslundBh: Not disagreeing. Just did not want to interfere to much with other methods in this PR, with versioning etc.

Can ofc add it in this PR, but would almost like to handle it separately. We really do need to run through all methods though and ensure that the tolerance is added as a input parameter when required, as well as properly passed.

If you are fine with it, I would prefer to do it as a separate exercise though.

Geometry_Engine/Query/EndDir.cs
IsakNaslundBh: Yeah, happy to move it up :)

Geometry_Engine/Query/IsContaining.cs
IsakNaslundBh: Again, as above. I think the name actually is correct, when looking at it from the perspective of the ellipse, but does sound weirder when applying that same transformation to some other object!

Geometry_Engine/Query/LineIntersection.cs
IsakNaslundBh: As previous, could argue both ways. Lets grab a quick chat when you are in :)

Geometry_Engine/Query/StartDir.cs
IsakNaslundBh: My thinking here was to use another name, to add information. Basically use a synonym as it could be called either (and we could have called the methods StartTan and EndTan instead really).

Can change if you disagree though.

Geometry_Engine/Compute/SortAlongCurve.cs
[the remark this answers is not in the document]
pawelbaran: Yeah that would be most logical IMO

Geometry_Engine/Compute/SortAlongCurve.cs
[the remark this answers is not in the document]
pawelbaran: I would return `null` - from what I understand, in case of sorting the input objects should never change, only their order. So if it is impossible to deduct the valid order, it does not mean the points disappear

Geometry_Engine/Compute/SortAlongCurve.cs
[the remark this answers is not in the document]
pawelbaran: Makes sense, this is exactly what I thought when reading the code. I think it is fine to leave it as is for now, no harm other than performance in the end

Geometry_Engine/Query/ClosestPoint.cs
pawelbaran: This is... confusing 😃 but it only proves intuition is not always the best advisor - I took it wrong based on the name itself.

Geometry_Engine/Query/ClosestPoint.cs
pawelbaran: Sure thing, I only made this comment to highlight the fact and add it to the master to do list - absolutely not asking to fix it here 👍

Geometry_Engine/Query/StartDir.cs
pawelbaran: I think I would try being consistent, also not sure how many users would be confused by tangent, while direction is very obvious even if less precise. But no strong opinion here - I am just a guest in this toolkit so I do not have the right to push conventions 😉

pawelbaran:
I have gone through the test file - it could sometimes be cleaner (😉), but finally I think I managed to comprehend. I found one failure that concerns me, added a dedicated script to the PR script folder you shared. Besides that, could you please either delete or leave a not what is the part below doing? Thanks!

image

IsakNaslundBh:
> I have gone through the test file - it could sometimes be cleaner (😉), but finally I think I managed to comprehend. I found one failure that concerns me, added a dedicated script to the PR script folder you shared. Besides that, could you please either delete or leave a not what is the part below doing? Thanks!
> 

Thanks for testing @pawelbaran . Tried to be relatively clean, and explain what parts where doing, but sorry, should have put a not on that section that I used to make some initial tests and reasoning regarding the plane intersection algorithm. Was thinking of deleting it, but left it in case I needed to come back to it.

Have left a note on it now.

Also, thanks for spotting the failure cases for ClosestPoint. (or point at parameter really, but was due to ClosestPoint). Have slightly tweaked that method now, which should give better results for more cases, both for the extreme as well as more normal cases.

Please let me know if you find any more issues! Thought I had tested really quite substantially, but as shown by the above, not always easy to catch all cases yourself, especially when they are down to numerical idiosyncrasies!

pawelbaran:
I have tried to break the code once again, without success this time 😢 Happy with how my comments were addressed, all clear after the offline chat we had with @IsakNaslundBh. Again, this PR is a really good piece of math, chapeau bas @IsakNaslundBh!

pawelbaran (dismissed):
...and approval with text to make @BHoMBot happy 😉

IsakNaslundBh:
@BHoMBot check compliance

IsakNaslundBh:
@BHoMBot check compliance
@BHoMBot check unit-tests
@BHoMBot check required

IsakNaslundBh:
@BHoMBot check unit-tests

IsakNaslundBh:
@BHoMBot check compliance
@BHoMBot check required

IsakNaslundBh:
Requesting dispensation for the null handling as the only method failing is a method returning a boolean, where both true and false can be valid returns, and expected valued, even for a crash. Reference this old issue: https://github.com/BHoM/admin/issues/7

The only safe way to handle currently, until we support nullable primitives formally, is to just let it crash.

FraserGreenroyd:
@BHoMBot this is a DevOps instruction. I am authorising dispensation to be granted on check ref. 11130365886

FraserGreenroyd (approved):
Only changes since @pawelbaran approval have been to unit tests - approving on the basis that @pawelbaran approval is still valid.

FraserGreenroyd:
@BHoMBot check ready-to-merge

IsakNaslundBh:
Thanks @FraserGreenroyd  :)

PR #3026, opened 2023-03-15
BHoM_Engine: `IsNumeric()` to be explicit about whether enum types should be considered numeric

### LINKED ISSUE #3025: BHoM_Engine: IsNumeric() fails to recognise when an input object is an `enum` type, causing Diffing to fail in edge cases
#### Description:

As title. This is because IsNumeric() rightly recognise the enum as an integer, but fails to check if it is an enum.
Because this method is used by the diffing in `NumericalDifferenceInclusion()` to establish whether a type is a number or not, it causes the diffing to fail in edge cases where the input is an enum.

#### Expected behaviour:

IsNumeric should clarify whether it considers `enums` as a numeric type or not. 
The method was created having in mind to check whether an input type is of any of the C# numeric types (Int, double, etc) without considering that an enum could be cast to an integer.

We could add an optional parameter `considerEnumsAsNumeric` which I think should be defaulting to `false`.

#### Test file(s):

Tested with the Diffing Test procedure on BHoM_6.1.230315.beta.0. The test that identified this issue is in DiffingTest_Prototypes where a Node was randomly created and a CustomData entry ended up being a Constraint6DOF; the test then checked against numerical differences in properties, which highlighted this issue.

alelom:

Closes #3025 

This PR adds a new `bool EnumAsNumeric = true` optional parameter to `IsNumeric()` and `IsNumericIntegralType()` methods. This is needed to clarify whether `enums` should be considered a numeric number or not. In certain contexts like Diffing (see #3025), `IsNumeric()` is used to parse a wide variety of types, and `IsNumeric` needs this option to clarify what it does, because we had Diffing failures in certain edge cases.

The default value of this new parameter `EnumAsNumeric` is set to `true`, so it keeps the same behaviour by default, avoiding problems with any existing reference to the method. The diffing engine method that needs this parameter was changed to pass `false` in.

### Test files

Test with the newly added unit tests.

### Changelog

- Provided a new `EnumAsNumeric` optional input to `IsNumeric()` and `IsNumericIntegralType()` methods.

### DIFF
--- a/BHoM_Engine/Query/IsNumeric.cs
+++ b/BHoM_Engine/Query/IsNumeric.cs
@@ -29,6 +29,7 @@ namespace BH.Engine.Base
 {
     public static partial class Query
     {
+        [PreviousVersion("6.1", "BH.Engine.Base.Query.IsNumeric(System.Type)")]
         [Description("Determine whether a type is a numeric type.")]
         [Input("type", "Type that we want to check if it is numeric type or not.")]
         [Input("enumsAsNumbers", "(Optional, defaults to true) Whether `enum` types should be considered to be numeric.")]
--- a/BHoM_Engine/Query/IsNumeric.cs
+++ b/BHoM_Engine/Query/IsNumeric.cs
@@ -31,8 +31,9 @@ namespace BH.Engine.Base
     {
         [Description("Determine whether a type is a numeric type.")]
         [Input("type", "Type that we want to check if it is numeric type or not.")]
+        [Input("enumsAreNumbers", "(Optional, defaults to false) Whether `enum` types should be considered to be numeric.")]
         [Output("isNumeric", "True if the object is a numeric Type, false if not.")]
-        public static bool IsNumeric(this Type type)
+        public static bool IsNumeric(this Type type, bool enumsAreNumbers = false)
         {
             switch (Type.GetTypeCode(type))
             {
@@ -95,6 +96,7 @@ namespace BH.Engine.Base
                 case TypeCode.Int32:
                 case TypeCode.Int64:
                     return true;
+                    return !type.IsEnum || enumsAreNumbers;
                 default:
                     return false;
             }
--- a/BHoM_Engine/Query/IsNumeric.cs
+++ b/BHoM_Engine/Query/IsNumeric.cs
@@ -31,9 +31,9 @@ namespace BH.Engine.Base
     {
         [Description("Determine whether a type is a numeric type.")]
         [Input("type", "Type that we want to check if it is numeric type or not.")]
-        [Input("enumsAreNumbers", "(Optional, defaults to false) Whether `enum` types should be considered to be numeric.")]
+        [Input("enumsAreNumbers", "(Optional, defaults to true) Whether `enum` types should be considered to be numeric.")]
         [Output("isNumeric", "True if the object is a numeric Type, false if not.")]
-        public static bool IsNumeric(this Type type, bool enumsAreNumbers = false)
+        public static bool IsNumeric(this Type type, bool enumsAreNumbers = true)
         {
             switch (Type.GetTypeCode(type))
             {
--- a/BHoM_Engine/Query/IsNumeric.cs
+++ b/BHoM_Engine/Query/IsNumeric.cs
@@ -31,9 +31,9 @@ namespace BH.Engine.Base
     {
         [Description("Determine whether a type is a numeric type.")]
         [Input("type", "Type that we want to check if it is numeric type or not.")]
-        [Input("enumsAreNumbers", "(Optional, defaults to true) Whether `enum` types should be considered to be numeric.")]
+        [Input("enumsAsNumbers", "(Optional, defaults to true) Whether `enum` types should be considered to be numeric.")]
         [Output("isNumeric", "True if the object is a numeric Type, false if not.")]
-        public static bool IsNumeric(this Type type, bool enumsAreNumbers = true)
+        public static bool IsNumeric(this Type type, bool enumsAsNumbers = true)
         {
             switch (Type.GetTypeCode(type))
             {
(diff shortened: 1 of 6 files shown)


BHoM_Engine/Query/IsNumeric.cs
```diff
@@ -47,59 +48,10 @@ public static bool IsNumeric(this Type type)
...
-                case TypeCode.Byte:
-                case TypeCode.SByte:
-                case TypeCode.UInt16:
-                case TypeCode.UInt32:
-                case TypeCode.UInt64:
-                case TypeCode.Int16:
-                case TypeCode.Int32:
-                case TypeCode.Int64:
-                    return true;
+                    return !type.IsEnum || enumsAreNumbers;
```
IsakNaslundBh, line 51: Could do this instead?
Guess functionality wise, will do the same, but could maybe be clearer (and doing exactly the same as the other two methods)
```
            switch (Type.GetTypeCode(type))
            {
                case TypeCode.Byte:
                case TypeCode.SByte:
                case TypeCode.UInt16:
                case TypeCode.UInt32:
                case TypeCode.UInt64:
                case TypeCode.Int16:
                case TypeCode.Int32:
                case TypeCode.Int64:
                    return !type.IsEnum || enumsAreNumbers;
                case TypeCode.Decimal:
                case TypeCode.Double:
                case TypeCode.Single:
                    return true;
                default:
                    return false;
            }

```

BHoM_Engine/Query/IsNumeric.cs
```diff
@@ -31,8 +31,9 @@ public static partial class Query
     {
         [Description("Determine whether a type is a numeric type.")]
         [Input("type", "Type that we want to check if it is numeric type or not.")]
+        [Input("enumsAreNumbers", "(Optional, defaults to true) Whether `enum` types should be considered to be numeric.")]
```
IsakNaslundBh, line 34: Happy for you to keep as is, but wondering if `enumsAsNumbers` is a clearer name?

If you disagree, please feel free to ignore this comment.

IsakNaslundBh (commented):
Think change looks sensible, and reason for it also makes sense. Code changes generally looks good as well, have a minor comment, see below.

Also, versioning is missing on the changed methods needs to be added.

adecler:
Just a comment and not a request for change but I would prefer for `AreEnumsNumeric` to be false by default. 
For example, they do not support any mathematical operation. For charting as well, I would treat enums as strings and not number (e.g. stack charts don't make sense for enums)

alelom:
> Just a comment and not a request for change but I would prefer for `AreEnumsNumeric` to be false by default. For example, they do not support any mathematical operation. For charting as well, I would treat enums as strings and not number (e.g. stack charts don't make sense for enums)

Me too! Which is what I wrote on the original issue. I was about to make it default to `true` but then I thought that it would be risky to change the default behaviour so close to the beta closure. Therefore, I'd rather re-change it once the beta is delivered.

adecler:
> I thought that it would be risky to change the previous behaviour so close to the beta closure (the code on `main` of `IsNumeric()` recognises enums as numeric -- which I agree is not desirable). Therefore, I'd rather switch the default value to `false` once the beta is delivered.

Fully agree on the strategy. Just thought I would add my vote for the false as default in the long run.

pawelbaran (commented):
I have 2 comments on this PR:
- it seems that you missed a [few calls](https://github.com/search?q=org%3ABHoM+isnumeric&type=code) to `IsNumeric` - particularly the one in Revit_Toolkit looks like it should take `false`, not sure about base `Query.Hash` (it looks fine?)
- wouldn't it be sensible for `IsNumeric` to call `return IsNumericIntegralType || IsNumericFloatingPointType`?

pawelbaran:
@BHoMBot check core
@BHoMBot check compliance
@BHoMBot check unit-tests

alelom:
Thanks @pawelbaran,

> it seems that you missed a [few calls](https://github.com/search?q=org%3ABHoM+isnumeric&type=code) to IsNumeric - particularly the one in Revit_Toolkit looks like it should take false, not sure about base Query.Hash (it looks fine?)

Checking them one by one:
- https://github.com/BHoM/BHoM_Engine/blob/66d4a7219c97d961d32d06b1c42c94d40df3e95e/Diffing_Engine/Query/NumericalDifferenceInclusion.cs: covered by this PR
- https://github.com/BHoM/BHoM_Engine/blob/66d4a7219c97d961d32d06b1c42c94d40df3e95e/BHoM_Engine/Query/Hash.cs no adjustment needed as enum check already covered in same line
- [RDF_Engine/Compute/ToWebVOWLJson.cs](https://github.com/BHoM/RDF_Prototypes/blob/8b380a30fb092bb9bf685dc540c5710fdbaf13a3/RDF_Engine/Compute/ToWebVOWLJson.cs) not an actual reference but a comment

This leaves us with only:
[Revit_Engine/Query/HashString.cs](https://github.com/BHoM/Revit_Toolkit/blob/07f2733e4af7c2c3ed80f4b9beef9e13dbbdfe1e/Revit_Engine/Query/HashString.cs)
With my testing, I never found an issue in the Revit_Engine diffing related to `IsNumeric()`, although I agree it can potentially be a very rare issue in future, but to avoid burdening BHoM_Engine and this PR, I would do it separately to this beta -- let me know what you think, otherwise I can quickly raise it, no problem.

pawelbaran (approved):
Happy with @alelom's response - I have looked into [Revit_Engine/Query/HashString.cs](https://github.com/BHoM/Revit_Toolkit/blob/07f2733e4af7c2c3ed80f4b9beef9e13dbbdfe1e/Revit_Engine/Query/HashString.cs) and the value should never be an `enum` under normal circumstances, so no major need for a fix now.

One final request @alelom: could you please make sure the default value of `enumsAsNumbers` gets set to `false` in the new milestone? Thanks!

alelom:
> One final request @alelom: could you please make sure the default value of `enumsAsNumbers` gets set to `false` in the new milestone? Thanks!

Sure, raised issue here: https://github.com/BHoM/BHoM_Engine/issues/3028

FraserGreenroyd:
@BHoMBot this is a DevOps instruction. I am authorising dispensation to be granted on check ref. 12165517821

PR #3042, opened 2023-04-14
Serialiser_Engine: Move away from Mongo.Bson

### LINKED ISSUE #3041: Serialiser_Engine: Move away from Mongo.Bson
#### Description:
Task to capture the required tasks for upgrading Serialiser Engine to move away from Mongo BSON if possible, or fixing the current issues highlighted in the Serialiser Engine. This includes some work on investigating options and risk (to existing infrastructure, efficiency, etc.) vs reward structure, and then possible tests and updates if it goes ahead.

Agreement for coding to be done by one person with daily touchdowns during the development prior to the PR, with others also on call for immediate discussion as required.

Coding: @adecler 
Review and support: @IsakNaslundBh and @FraserGreenroyd

adecler:

Closes #3041 

### Test files
So far, only a first version of serialisation done.

[NewSerialiser_Test.zip](https://github.com/BHoM/BHoM_Engine/files/11234036/NewSerialiser_Test.zip)

image
The Json results that have different length are because of people not following conventions or because the type is sometimes recorded with AssemblyQualifiedName for no reason in the old system.
Only one object returns a different result when deserialised with the old deserialiser but Pawel is aware of it and is going to fix it.

[NewDeserialiser_Test.zip](https://github.com/BHoM/BHoM_Engine/files/11275451/NewDeserialiser_Test.zip)

image


Serialiser_Engine/Compute/Serialise/IObject.cs
```diff
@@ -0,0 +1,61 @@
...
+
+            writer.WriteName("_t");
+            writer.WriteString(value.GetType().FullName);
+
+            foreach (PropertyInfo prop in value.GetType().GetProperties())
+            {
+                writer.WriteName(prop.Name);
+                ISerialise(prop.GetValue(value), writer);
+            }
+            writer.WriteEndDocument();
```
IsakNaslundBh, line 56: Would be a change to behaviour, but might be good to add a `writer.AddVersion()` here?

As I think we have discussed before, good to have access to that on all levels when it comes to versioning.

That would ofc mean that you might have issues de-serialising files using an older version of the BHoM, that have been serialised with this (not, forward compatible) so I guess could be something we could but a pin in and add in 6.3

Serialiser_Engine/Compute/Serialise/IObject.cs
```diff
@@ -0,0 +1,61 @@
...
+                writer.WriteNull();
+                return;
+            }
+
+            writer.WriteStartDocument();
+
+            writer.WriteName("_t");
+            writer.WriteString(value.GetType().FullName);
+
+            foreach (PropertyInfo prop in value.GetType().GetProperties())
```
IsakNaslundBh, line 51: Know you know, @adecler and good that it is not done in these first commits, but putting comment for my own memory if nothing else.

Could quite easily optimise this loop by caching `Func<object,object>` for each type being serialised, making use of the `ToFunc()` method in the base engine, and store in a `(Concurrent)Dictionary` with the type or `value.GetType().FullName` as key.

Assume that would speed up serialisation by quite a bit when you have multiple objects of the same type.

IsakNaslundBh (commented):
Generally really like this so far. 

Serialisation code looks significantly cleaner and easier to read compared to what was there before.

Did put some comments on the IObject serialiser, for discussion/reminder to myself, and things I am sure you are already aware of.

Now just hoping de-serialisation can be close to this clean... :)

Serialiser_Engine/Compute/Serialise/IObject.cs
adecler: For the moment, I am focusing on producing the exact same content than the current version. Makes it easier to debug and catch problems. I would recommend doing anything that alters the output json in a later PR to keep things simple.

Serialiser_Engine/Compute/Serialise/IObject.cs
adecler: I know 😄 

In the spirit of keeping things as simple as possible to debug, I have stayed away from that kind of optimisation for now.

adecler:
@FraserGreenroyd , @IsakNaslundBh , let have a chat about the deserialiser. There is still a few additional types that need to be covered but, right now, the performances are becoming a bit of a concern. The plan was to focus on clarity and simplicity and I think we achieved that but the difference in computational speed is too large to be confident that a few optimisations and pre-compilations will be enough to bridge the gap.

BHoM_Engine/Create/Type/Type.cs
```diff
@@ -51,6 +51,18 @@ public static Type Type(string name, bool silent = false)
```
IsakNaslundBh, line 51: Think this is the clear pinch point in the deserialisation process.

Did some profiling, and some testing on a set of random Bars, and from that, the majority of the time for the de-serialisation was highlighted to be the linq query of creating a new dictionary in the BHoMTypeDictionary method:

https://github.com/BHoM/BHoM_Engine/blob/27983e8f22b39b98778cfdb4b6ee5e46af56f67c/BHoM_Engine/Query/TypeDictionary.cs#L39-L42

if we simply skip that method here, and directly just access the underlying concurrent dictionary from the Global variable, by changing the line here to:

```
            ConcurrentDictionary<string, List<Type>> typeDictionary = Global.BHoMTypeDictionary;
```

we get a speedup (for my test case) that brings the de-serialisation to the same level as the old system. I ran a few tests, and they both compare, with some cases favouring the old, and some the new. 

One option if we dont want to directly access the concurrent dictionary here is to change the return type of the BHoMTypeDictionary to `IReadOnlyDictionary` and simply return the concurrent dictionary. Not sure if that could have other implications elsewhere though, as on BHoM_UI and similar.

IsakNaslundBh (commented):
Se comment below.

Might be other pinch points for some other cases that I have not tested, but seem to literally be the linq `ToDictionary` call in the `BHoMTypeDictionary` taking up roughly 90% of the execution time of the new de-serialisation.

With that changed, old and new are very comparable, and that is before we have even attempted to optimise the new version through things like pre-compiling property getters and setters as well as constructors for immutable types.

Tried running through your testscript as well, and that is now giving _as_ good results as my simpler testcase with a set of Bars, but it is on a factor 3 level, rather than a factor 20-30 level:

image


Again, with the only change made being to access the concurrent dictionary directly from Global, rather than via the method.

So, still slower, but ofc, that is before _any_ level of optimisation ahs been done.

Also, still getting quite a few failures (95 types I think) but guess you are aware of that.

Anyway, happy to have a chat when you are in.

IsakNaslundBh:
@BHoMBot check versioning

Serialiser_Engine/Compute/Serialise/Short.cs
```diff
@@ -0,0 +1,46 @@
...
+
+namespace BH.Engine.Serialiser
+{
+    public static partial class Compute
+    {
+
+        /*******************************************/
+        /**** Public Methods                    ****/
+        /*******************************************/
+        private static void Serialise(this short value, BsonDocumentWriter writer, Type targetType)
```
FraserGreenroyd, line 38:
```suggestion
        /**** Private Methods                   ****/
        /*******************************************/
        
        private static void Serialise(this short value, BsonDocumentWriter writer, Type targetType)
```

Serialiser_Engine/Compute/Serialise/String.cs
```diff
@@ -0,0 +1,46 @@
...
+
+namespace BH.Engine.Serialiser
+{
+    public static partial class Compute
+    {
+
+        /*******************************************/
+        /**** Public Methods                    ****/
+        /*******************************************/
+        private static void Serialise(this string value, BsonDocumentWriter writer, Type targetType)
```
FraserGreenroyd, line 39:
```suggestion
        /**** Private Methods                   ****/
        /*******************************************/
        
        private static void Serialise(this string value, BsonDocumentWriter writer, Type targetType)
```

Serialiser_Engine/Compute/Serialise/Tuple.cs
```diff
@@ -0,0 +1,126 @@
...
+using System.Reflection;
+using System.Text;
+
+namespace BH.Engine.Serialiser
+{
+    public static partial class Compute
+    {
+
+        /*******************************************/
+        /**** Public Methods                    ****/
```
FraserGreenroyd, line 39:
```suggestion

        /*******************************************/
        /**** Private Methods                   ****/
```

Serialiser_Engine/Compute/Serialise/Tuple.cs
```diff
@@ -0,0 +1,126 @@
...
+                value.Item3.ISerialise(writer, typeof(T3));
+                value.Item4.ISerialise(writer, typeof(T4));
+                value.Item5.ISerialise(writer, typeof(T5));
+                writer.WriteEndArray();
+            });
+        }
+
+        /*******************************************/
+
+
```
FraserGreenroyd, line 124:
```suggestion
```

Serialiser_Engine/Compute/Serialise/Type.cs
```diff
@@ -0,0 +1,117 @@
...
+                    }
+                }
+                else if (value.Namespace.StartsWith("BH.oM"))
+                    writer.WriteString(value.FullName);
+                else if (value.AssemblyQualifiedName != null)
+                    writer.WriteString(value.AssemblyQualifiedName);
+                else
+                    writer.WriteString(""); //TODO: is that even possible?
+
+
```
FraserGreenroyd, line 97:
```suggestion
```

Serialiser_Engine/Convert/OldBson.cs
```diff
@@ -0,0 +1,235 @@
...
+                    dic.Remove("BHoM_Guid");
+                }
+                co.CustomData = dic;
+                return co;
+            }
+            else
+                return obj;
+        }
+
+
```
FraserGreenroyd, line 112:
```suggestion
```

Serialiser_Engine/Convert/OldBson.cs
```diff
@@ -0,0 +1,235 @@
...
+        private static bool m_TypesRegistered = false;
+        private static readonly object m_RegisterTypesLock = new object();
+
+        /*******************************************/
+    }
+}
+
+
+
+
```
FraserGreenroyd, line 235:
```suggestion
```

Serialiser_Engine/Convert/OldJson.cs
```diff
@@ -0,0 +1,143 @@
+/*
```
FraserGreenroyd, line 1: How necessary is this file for the ongoing strategy/plan?

FraserGreenroyd: If we intent to keep it, we should put some deprecation warnings on it to avoid people using it in the long run and sleep walking into problems?

FraserGreenroyd (changes requested):

.ci/unit-tests/Serialiser_Engine_Tests/Serialiser_Engine_Tests.csproj
```diff
@@ -0,0 +1,87 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0-windows</TargetFramework>
```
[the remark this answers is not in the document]
IsakNaslundBh: Was to be able to load up some oMs and Engines relying on WPF to be able to check that they still could serialise/deserialise.

.ci/unit-tests/Serialiser_Engine_Tests/Serialiser_Engine_Tests.csproj
```diff
@@ -0,0 +1,87 @@
...
+    <FileVersion>6.2.0.0</FileVersion>
+    <AssemblyVersion>6.0.0.0</AssemblyVersion>
+    <Description>https://github.com/BHoM/BHoM_Engine</Description>
+    <ImplicitUsings>enable</ImplicitUsings>
+	<UseWPF>true</UseWPF>
```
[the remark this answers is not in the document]
IsakNaslundBh: As above

BHoM_Engine/Create/Type/Type.cs
```diff
@@ -84,6 +89,17 @@ public static Type Type(string name, bool silent = false)
         }
 
         /***************************************************/
+
+        private static Dictionary<string, Type> m_ExplicitSystemTypes = new Dictionary<string, Type>
```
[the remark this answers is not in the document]
IsakNaslundBh: Not currently, no. List below is essentially the types that did not automatically from `type = System.Type.GetType(name);` unless the string provided was assembly qualified, and this was the simplest workaround we found to that issue.

First aimed to keep this in the Serialiser_Engine, but that lead to problems with things like generics and collections, hence opted to move here.

Can add a comment above if helpful

Playground/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System.Reflection;
```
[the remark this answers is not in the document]
IsakNaslundBh: @adecler think this is your initial test project, can I leave it with you to delete?

Serialiser_Engine/Compute/Serialise/DateTime.cs
```diff
@@ -0,0 +1,48 @@
...
+{
+    public static partial class Compute
+    {
+
+        /*******************************************/
+        /**** Public Methods                    ****/
+        /*******************************************/
+        private static void Serialise(this DateTime value, BsonDocumentWriter writer, Type targetType)
+        {
+            // TODO: is this right? will lose precision after round trip
```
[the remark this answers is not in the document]
IsakNaslundBh: @adecler you put this comment here, you have any idea?

Serialiser_Engine/Convert/OldJson.cs
IsakNaslundBh: long terms strategy for this file is for it to be removed. Think that can be done in a separate PR though.

The risk of people using this I would say is quite low, given that most calls to the serialiser go via some mechnism that will point at the new system, but guess we could raise warnings if it is called or similar.

But from my point of view, I think it is fairly safe to keep it in for now, and then remove in a separate PR.

.ci/unit-tests/Serialiser_Engine_Tests/Serialiser_Engine_Tests.csproj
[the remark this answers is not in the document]
FraserGreenroyd: Ok, but can it now be reset to those away from WPF, given this particular repository has no dependency on such things?

FraserGreenroyd:
@BHoMBot check compliance

.ci/unit-tests/Serialiser_Engine_Tests/Serialiser_Engine_Tests.csproj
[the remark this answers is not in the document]
FraserGreenroyd: Tagging @IsakNaslundBh to notify of the new question above.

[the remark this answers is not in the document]
FraserGreenroyd: Resolved via call with @IsakNaslundBh to only use WPF if in Test config mode.

IsakNaslundBh:
@BHoMBot check compliance

IsakNaslundBh:
@BHoMBot  check code-compliance

FraserGreenroyd:
@BHoMBot check versioning

FraserGreenroyd:
@BHoMBot check compliance

Serialiser_Engine/Compute/Serialise/DateTime.cs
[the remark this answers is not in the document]
IsakNaslundBh: Regarding this, this is behaving the same way as the previous serialiser was, see:

https://github.com/BHoM/BHoM_Engine/blob/a498583b58fa4513cdb541a1a6f0e27bc704d256/Serialiser_Engine/Objects/BsonSerializers/ObjectSerializer.cs#L86-L88

Hence think this can be marked as resolved, and if any issues is found with this, then a separate issue can be opened and resolved in a separate PR.

IsakNaslundBh:

FraserGreenroyd:

FraserGreenroyd:
@BHoMBot check compliance

FraserGreenroyd:
@BHoMBot check serialisation
@BHoMBot check versioning

IsakNaslundBh:
@BHoMBot check versioning

FraserGreenroyd:
@BHoMBot check core
@BHoMBot check null-handling

FraserGreenroyd:
@BHoMBot this is a DevOps instruction. I am authorising dispensation to be granted on check ref. 13549104346

FraserGreenroyd:
@BHoMBot this is a DevOps instruction. I am authorising dispensation to be granted on check ref. 13549104003

alelom:
@pawelbaran identified an issue on the serialization of the `Diff` object. This issue can be worked around for the time being and should not be regarded as a blocker in our opinion. I am logging it here for future discussion.

The `Diff` object owns a `DiffingConfig` property, which itself owns a property of type `Func<>`. This is required to obtain a dependency decoupling when passing diffing configuration from downstream repo to the base Diffing_Engine. The serializer is unable to serialize/deserialize or ignore the Func, but the rest of the Diff object is serialized/deserialized correctly. This behaviour is consistent to what we used to have in the earlier version, but the exception was "swallowed" and not displayed (quoting @pawelbaran), hence this error was not discovered before, also because the Func object is needed for the back-end and not manually created/queried in UIs. However, the error hiding ("swallowing") is not wanted in the current version of the serializer, and the exception is now surfacing in the UIs when serializing/deserializing `Diff`s. In appearance, the Diff serializes just fine (with the exception of the `Func` property), but an error gets thrown to UIs.

Points to be discussed:
- The current RevitDiffing wraps the function passed into the `Func<>` property into an anonymous lambda. This is unnecessary and may be the source of the serialization issue (@pawelbaran).
- Modifying the `Func<>` property to `MethodBase` may help, as tested by @pawelbaran, but it comes with lesser type-strong checking on the arguments.
- Do we need to support for `Func<>` in the serializer? If it is needed, in case we do _not_ want to add `Func` support, this means that a new compliance rule should be set to detect the usage of `Func<>`s in BHoM objects; additionally, this would put into question what other base C# types may want to not support in BHoM because they may misbehave during serialization.

pawelbaran:
To follow up on @alelom's comment, I would opt for supporting `Func<>` in the same scope as we do for `MethodBase` - the question is how much effort would that require, hopefully @IsakNaslundBh could estimate?

FraserGreenroyd (approved):
This PR has been tested by various members of the dev team over the last week, with minor issues found and ironed out with final checks today showing this is suitable for alpha deployment to continue testing and find any further issues.

This is a great enhancement to our Serialiser Engine process and makes debugging much easier :smile: great work team! @adecler @IsakNaslundBh @pawelbaran @alelom

FraserGreenroyd:
@BHoMBot check ready-to-merge

PR #3044, opened 2023-04-19
Base_Engine: RunExtensionMethods to be able to call methods with default params

### LINKED ISSUE #3021: BHoM_Engine: enhance capability of ExtensionMethod mechanism to support finding methods with optional parameters
#### Description:

As discussed with @IsakNaslundBh and @al-fisher , the current ExtensionMethod can be improved in two interrelated ways:
- support optional parameters
- the find of existing methodinfos by key can be improved in speed

This came out of the discovery of some `IGeometry3D()` implementations not working anymore, i.e., not called by the base `BH.Engine.Base.IGeometry3D()` method. For example, [IGeometry3D for Bar](https://github.com/BHoM/BHoM_Engine/blob/develop/Structure_Engine/Query/Geometry3D.cs#L45) is not invoked by the base engine method, although it used to be correctly found and invoked. My theory is that the work around the `TryRunExtensionMethods` and subsequent optimisations may have introduced a change that made them unfindable. The code causing this is:

https://github.com/BHoM/BHoM_Engine/blob/66d4a7219c97d961d32d06b1c42c94d40df3e95e/BHoM_Engine/Query/ExtensionMethodToCall.cs#L83-L85

here the extension method are found matching not only by name, but also by the number of parameters passed to the parent invoker, which is compared with the number of parameters owned by the target method. This prevents to find extension methods that have default parameters. We verified with @IsakNaslundBh that the dynamic dispatch works in situations where methods with default parameters are specified, for example:

```cs

public void ISomeExtensionMethod(BHoMObject obj)
{ 
   SomeExtensionMethod(obj as dynamic);
}

public void SomeExtensionMethod(BHoMObject obj)
{ some logic }

public void SomeExtensionMethod(Bar bar, bool someOptionalParameter = false)
{ some logic }

```

in this case, if one invokes the `ISomeExtensionMethod` and passes in a `Bar`, the third and final method is then invoked by the dynamic dispatch. Therefore, with @IsakNaslundBh we've agreed that the base ExtensionMethod should replicate this exact behaviour, so some correction is needed.

An initial workaround attempted by @IsakNaslundBh was to replace the ExtensionMethodToCall loop with:

```cs

// Make sure the type of parameters is matching, skipping first as already used to extract parameters
bool matchingTypes = true;
for (int i = 1; i < paramInfo.Length; i++)
{
    if (i >= parameters.Length)
    {
        if (!paramInfo[i].Attributes.HasFlag(ParameterAttributes.HasDefault))
        {
            //Parameter does not match, abort for this method
            matchingTypes = false;
            break;
        }
    }
    else if (!paramInfo[i].ParameterType.IsAssignableFromIncludeGenerics(parameters[i].GetType()))
    {
        //Parameter does not match, abort for this method
        matchingTypes = false;
        break;
    }
}

```

however more work will be needed to ensure this is the best approach:
- there are some areas around this that can be improved by more efficiently storing the MethodInfo key
- to improve speed, we should store a Func delegate for all possible combinations of a method with variable number of optional parameters passed to it

### LINKED ISSUE #3022: Structure_Engine: Geometry3D extension methods are not being invoked by the base engine IGeometry3D()
#### Description:

As title. This is due to some change introduced by the work around the TryRunExtensionMethods and subsequent optimisations, as explained in #3021. This issue will be automatically solved once #3021 is solved.

#### Steps to reproduce:

Try using the `BH.Engine.Base.IGeometry3D` method on a `Bar`. It will return nothing, despite there exist a `BH.Engine.Structure.Query.Geometry3D` method.

#### Expected behaviour:

`BH.Engine.Base.IGeometry3D` method on a `Bar` should work.

#### Test file(s):

Can be tested as above.

IsakNaslundBh:

Closes #3021
Closes #3022

Fix RunExtensionMethod and TryRunExtensionMethod able to run methods with additional parameters that have default values, for example (as shown in tests) the Geometry3d method.

This aligns with behaviour of the `as dynamic` behaviour.

Whilst doing this required to change the flow and method call order, as with new system, it is required to be able to store multiple functions refering to the same method, in the case of it being called via run extention method with all or just some of the parameters provided.

Whilst doing this, a change to make TryRunExtensionMethod to main method of execution made sense, hence the core call logic has been moved there.

This also removes some methods no longer used by the internal logic as they where identified to not be used anywhere in the code base.

### Test files

Have run through all UTs available for IElement workflows which all pass.

There is also some added targeted UTs added to the test project.

Test script to verify that `IGeometry3D()` works for the method with default argument `Geometry3D(Bar bar, bool onlyOuterExtrusion = true)`: [geometry3Dtest.zip](https://github.com/BHoM/BHoM_Engine/files/11323978/geometry3Dtest.zip)


IsakNaslundBh:
@BHoMBot check compliance

IsakNaslundBh:
@BHoMBot check installer

IsakNaslundBh:
@BHoMBot check compliance
@BHoMBot check versioning

alelom:
@BHoMBot check required

alelom (approved):
Tested with the mentioned test script on Geometry3D, the provided unit tests, and Diffing functions that require extension methods to work (i.e. RevitDiffing). Works well: successfully enables the support for default inputs (e.g. for the Geometry3D method for `Bar`), and retains functionality of existing extension methods.

pawelbaran (approved):
Our Revit code does not explicitly call `RunExtensionMethod`, but I made a sanity check and run tests, without issues. So happy to approve the change from Revit perspective 👍

FraserGreenroyd:
@BHoMBot check unit-tests
@BHoMBot check ready-to-merge

FraserGreenroyd:
@BHoMBot this is a DevOps instruction. I am authorising dispensation to be granted on check ref. 13011227380

PR #3086, opened 2023-06-14
Fix serialiser crash on failed

### LINKED ISSUE #3066: Serialiser_Engine: Deserialising lists throws massive errors if things fail
#### Description:
When trying to deserialise some unit tests in this repository to resolve #3065 , Rhino/Grasshopper is crashing completely when deserialising the JSON when trying to deserialise a List of Objects which have been upgraded/don't exist (highlighted from the LCA refactor).

#### Steps to reproduce:
Read in a JSON file for `.ci\Datasets\Facade_Engine\Compute\UValueOpeningAW.json` and try to deserialise with `FromJson` in Grasshopper - Rhino/Grasshopper will crash with an unrecoverable error.

#### Expected behaviour:
Rhino/Grasshopper doesn't crash - but errors are reported back to the UI appropriately.

adecler:

Closes #3066 

As detailed in the issue, the BHoM UI was crashing on the deserialisation of a specific dataset. It appears that the 'failed' boolean that is passes up from properties to parents was used to force a different deserialisation behaviour inside the `SetProperties` method. 
 - The issue is that that boolean is never reset between properties. This means one incorrect property would trigger a failing behaviour on all the following ones. 
 - Deserialisation is also capable of restoring partial objects by saving the remaining properties inside CustomData. Those objects were still marked as 'failed' though causnig the whole chain of parents to come back as Custom objects. 

As the only place where the 'failed' boolean is used for a conditional check is inside the `SetProperty` method (where it is incorrectly used), I would recommend to get rid of that boolean altogether. We agreed with @FraserGreenroyd  to do this in two phases:
- Remove the use of the 'failed' boolean inside the `SetProperty` method (this PR)
-  Remove the 'failed' boolean as an argument of all methods (all are private). Issue raised here: https://github.com/BHoM/BHoM_Engine/issues/3087

### Test files
 - [x] Make sure the versioning check is passing
 - [x] Test on older versioning datasets (at least a year coverage)
 - [x] Test on datasets found in the BHoM folder

### DIFF
--- a/Serialiser_Engine/Compute/Deserialise/IObject.cs
+++ b/Serialiser_Engine/Compute/Deserialise/IObject.cs
@@ -119,17 +119,6 @@ namespace BH.Engine.Serialiser
                         {
                             object propertyValue = item.Value.IDeserialise(prop.PropertyType, ref failed, prop.GetValue(value), version, isUpgraded);
 
-                            if (failed)
-                            {
-                                if (!isUpgraded && TryUpgrade(doc, version, out IObject upgraded))
-                                    return upgraded;
-                                else
-                                {
-                                    Base.Compute.RecordError($"Failed to deserialise property {item.Name} for object of type {value?.GetType().Name ?? "unknown type"}. Custom object returned in its place.");
-                                    return DeserialiseDeprecatedCustomObject(doc, ref failed, version, false);
-                                }
-                            }
-
                             if (CanSetValueToProperty(prop.PropertyType, propertyValue))
                             {
                                 prop.SetValue(value, propertyValue);


adecler:
@BHoMBot check versioning

FraserGreenroyd:
@BHoMBot check compliance
@BHoMBot check serialisation

FraserGreenroyd:
@BHoMBot check core
@BHoMBot check unit-tests
@BHoMBot check installer
@BHoMBot check null-handling

FraserGreenroyd (approved):
I have reviewed the code and it's only deleted lines which is always nice to tidy up 😄 

I have then performed the following actions on this branch:

 - Deserialised Test Sets for Versioning from version 3.3 to 6.1 for objects and methods
   - Some failures are noted during this deserialisation, however, they are either known issues as a result of the new serialisation framework, or not related to this particular work
   - Despite failures, every single test set JSON file deserialised without crashing Rhino/Grasshopper
   - Objects were returned where possible
 -  Deserialised every dataset housed within BHoM_Datasets
   - All datasets deserialised without crashing Rhino/Grasshopper
   - Some errors in deserialisation as a result of failing to convert types for objects I do not currently have compiled on my machine, with all other datasets deserialising correctly

In addition, the dataset issue described in the #3066 has also stopped crashing Rhino entirely and deserialises to a dataset but with errors reported as expected.

Thus, I am happy to merge this for wider testing during our 6.2 beta close out sprint this sprint, but this improves the behaviour discussed 😄

FraserGreenroyd:
@BHoMBot check ready-to-merge

PR #3097, opened 2023-06-26
Serialiser_Engine: IsValidJson checks if string is in valid format

### LINKED ISSUE #1427: Serialiser_Engine: Add methods to check if a string is a valid json
#### Description:

Currently the `FromJson` takes a `string in and tryes to output a valid object. If the string is not a valid json, it returns an error..

When calling this methods from the outside, such as in the `HTTPAdapter.Pull(GetRequest request PullType pulType, ActionConfig, actionsConfig)`, there is no standard way of checking if the string is a valid json, and the component turns red, if an e.g. an xml is pulled.

desaiwangBH:

Closes #1427  

- checks if json is valid
- only checks if formatting is correct, does not validate schema

### Additional comments

As per Fraser's suggestions, creating a new method inside Query that is similar to Convert/ToJson to check if json is valid. The method returns true if the input string is a valid json, false otherwise.


Serialiser_Engine/Query/Json.cs
```diff
@@ -0,0 +1,63 @@
...
+            return Regex.IsMatch(str, @"^[\],:{}\s]*$");
+
+        }
+
+        //JSON checking trailing coomas regex based on: https://stackoverflow.com/questions/34344328/json-remove-trailing-comma-from-last-object
+        private static bool HasTrailingCommas(this string str)
+        {
+            return Regex.IsMatch(str, @"\,(?!\s*?[\{\[\""\w])");
+        }
+
```
FraserGreenroyd, line 60:
```suggestion
```

Serialiser_Engine/Query/Json.cs
```diff
@@ -0,0 +1,63 @@
...
+    public static partial class Query
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        //JSON validation based on code from https://github.com/prototypejs/prototype/blob/560bb59414fc9343ce85429b91b1e1b82fdc6812/src/prototype/lang/string.js#L699
+        
+        [Description("Checks if a string is a valid Json")]
+        [Input("string", "String to check.")]
```
FraserGreenroyd, line 38:
```suggestion
        [Input("str", "String to check validity of.")]
```

Serialiser_Engine/Query/Json.cs
```diff
@@ -0,0 +1,63 @@
...
+
+namespace BH.Engine.Serialiser
+{
+    public static partial class Query
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        //JSON validation based on code from https://github.com/prototypejs/prototype/blob/560bb59414fc9343ce85429b91b1e1b82fdc6812/src/prototype/lang/string.js#L699
```
FraserGreenroyd, line 35:
```suggestion
```

Serialiser_Engine/Query/Json.cs
```diff
@@ -0,0 +1,63 @@
...
+        
+        [Description("Checks if a string is a valid Json")]
+        [Input("string", "String to check.")]
+        [Output("validity", "True if string is a json, false otherwise.")]
+        public static bool IsValidJson(this string str)
+        {
```
FraserGreenroyd, line 41:
```suggestion
        {
            //JSON validation based on code from https://github.com/prototypejs/prototype/blob/560bb59414fc9343ce85429b91b1e1b82fdc6812/src/prototype/lang/string.js#L699
```

FraserGreenroyd (changes requested):

Serialiser_Engine/Query/Json.cs
```diff
@@ -0,0 +1,63 @@
...
+            str = Regex.Replace(str, @"""[^""\\\n\r]*""|true|false|null|-?\d+(?:\.\d*)?(?:[eE][+\-]?\d+)?", "]");
+            str = Regex.Replace(str, @"(?:^|:|,)(?:\s*\[)+", "");
+
+            return Regex.IsMatch(str, @"^[\],:{}\s]*$");
+
+        }
+
+        //JSON checking trailing coomas regex based on: https://stackoverflow.com/questions/34344328/json-remove-trailing-comma-from-last-object
+        private static bool HasTrailingCommas(this string str)
+        {
```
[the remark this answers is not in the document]
desaiwangBH: commas spelled wrong, I updated as suggested on my local branch and pushed to origin

adecler (commented):
A few things from the top of my head that are not valid Json and yet are returning true:
 - `"A"`
 - `[{`
 - `[{:}{,}]`

desaiwangBH:
> A few things from the top of my head that are not valid Json and yet are returning true:
> 
> * `"A"`
> * `[{`
> * `[{:}{,}]`

@adecler 
According to [JsonLint](https://jsonlint.com/), `"A"` is a valid json. However, `A` would is not a valid json, but that is already returning false.

As for the rest of the cases, I realized that there was some issues with the regex. So instead I reverted the IsValidJson method logic back to using BSON, while adding a check that does return single values as valid ("A", true, null, 3, for example). However, I also noticed that BSON has some intelligence when parsing, so for [{\"name\": \"John\", city:"NYC"}] IsValidJson will return true even though city is not inside quotes/is not a string. 

I argue this is ok, since the `FromJson` also uses BSON methods, so all jsons that are determined to be valid can be parsed.

Also added unit tests. Let me know if there's anything else to do.

FraserGreenroyd:
Following offline discussion with @adecler and @desaiwangBH it has been agreed that this PR isn't needed anymore with the following reasons provided:

 - The Serialiser_Engine has come along a fair bit since January 2020 when #1427 was originally raised
 - The Adapters have likewise undergone improvement following work done by @alelom since January 2020
 - The JSON format used by BHoM, while needing to be valid JSON, also needs to be valid objects to deserialise to, as such `"A"`, while valid JSON, is not going to be able to be parsed by BHoM, so saying it is valid or invalid JSON doesn't help matters
 
Thanks @desaiwangBH for the research into the issue, we'll keep the code on these commits should we change our mind in the future, but as of today, this is the decision reached as a collective 😄

PR #3107, opened 2023-07-14
Geometry_Engine: Fix bug where SortAlongCurve for a Polyline returns duplicate points

### LINKED ISSUE #3104: Geometry_Engine: SortAlongCurve method returns duplicate points
#### Description:

The `SortAlongCurve` method (specifically for `Polyline` returns more points than input. We can use LINQ to cull them but that should not be needed given the description of the output: 
https://github.com/BHoM/BHoM_Engine/blob/fa8a87f1ad6905adfe4c82ca040ac27fa5002f71/Geometry_Engine/Compute/SortAlongCurve.cs#L174

#### Steps to reproduce:

See script.

#### Expected behaviour:

Number of points returned to match the input points.

#### Test file(s):

[#3104-SortAlongCurve](https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/01_Issue/BHoM/BHoM_Engine/Geometry_Engine/%233104-SortAlongCurve.gh?csf=1&web=1&e=zkOUFi)

peterjamesnugent:

Closes #3104 

### Test files

https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/01_Issue/BHoM/BHoM_Engine/Geometry_Engine/%233104-SortAlongCurve.gh?csf=1&web=1&e=zkOUFi

### Changelog

Fixed a bug where a closed `Poyline` was returning more `Point`s than were being input.

### Additional comments

UnitTests only include `Ellipse` so I have added the `Polyline` tests.


peterjamesnugent:
@BHoMBot check required

Geometry_Engine/Compute/SortAlongCurve.cs
```diff
@@ -209,6 +209,9 @@ public static List<Point> SortAlongCurve(this List<Point> points, Polyline curve
             else if (tmpResult.Count == 1)
                 return tmpResult;
 
+            // For a close polyline there will be a duplicate point which needs to be removed
+            if (tmpResult.Count > points.Count - 1)
+                tmpResult = tmpResult.Distinct().ToList();
```
FraserGreenroyd, line 214: Is it safer to check only if the last point is the same as the first point for BHoM closed polylines? I wouldn't expect there to be more points along the curve than input but maybe there is a case for that?

peterjamesnugent: Have updated to check `Curve.IsClosed()` and tested for open and closed `Polyline`. 

The `points` are not always the same as the `ControlPoints` from the `Polyline`.

peterjamesnugent:
@BHoMBot check unit-tests

pawelbaran (approved):
Happy to approve the changes because they fix the issue stated. However, I have looked into the code and ended up with a few question marks such as:
- points that do not lie on the curve get culled without even a warning (I would argue they should be sorted based on the closest point on curve, while a warning about being skipped sounds like a bare minimum)
- `IsOnCurve` is called multiple times, while it could be done only once (pair each point with the closest curve and work with each of such sets independently)

Of course this is just a side comment, out of scope of this PR. But curious about your opinion on the above points @peterjamesnugent

peterjamesnugent:
@pawelbaran agree, if we are culling we need to make that clear in the description. Given that you can use `IsOnCurve` prior to using `SortAlongCurve`, I think we should project points and then sort - then we just need to use `IsOnCurve` when we go through `Polyline.SubParts`.

pawelbaran:
Project the points, but still sort the original input, right? So that the original input points get sorted based on their correspondent points on curve, would that make sense or am I overcomplicating?

peterjamesnugent:
Yes exactly.

pawelbaran:
https://github.com/BHoM/BHoM_Engine/issues/3108 raised 👍 

Meanwhile, happy to get this PR merged as soon as the checks pass.

FraserGreenroyd (approved):
My comment was addressed, happy to approve for merge.

FraserGreenroyd:
@BHoMBot check compliance
@BHoMBot check required

FraserGreenroyd:
@BHoMBot check ready-to-merge

FraserGreenroyd:

PR #3134, opened 2023-08-07
Serialiser_Engine: Force invariant representation of number in serialisation

### LINKED ISSUE #3130: Serialiser_Engine: Serialisation of decimals, doubles, and floats should be using only English representation
#### Description:
Currently, `Parse()` and TryParse()` in the serialiser are culture dependent. This means that exchanging json data between a UK computer and a Dutch computer for example is likely causing the extraction of incorrect numbers. 

For example, `double.Parse("0.0045");` would give `45` in a Dutch computer.

GH seems to be set to English mode anyways but Chrome isn't so this impacts the zero-code tools.

adecler:

Closes #3130

 - Overall, the serialisation of numbers was already relying on Mongo to do things correctly but there was a few cases where we were doing a 'manual' conversion to/from strings.

 - I used `CultureInfo.InvariantCulture` instead of `en-GB` because it is [more robust to changes](https://learn.microsoft.com/en-us/dotnet/api/system.globalization.cultureinfo.invariantculture?view=net-7.0) 

 -  Integers don't need formatting constraints as they don't change across cultures. I also checked other types like `Timespan` and `DateTime` and they don't need any modification as the first is invariant across cultures for default formatting and the second is represented as a long anyways.

### DIFF
--- a/Serialiser_Engine/Compute/Deserialise/Decimal.cs
+++ b/Serialiser_Engine/Compute/Deserialise/Decimal.cs
@@ -25,6 +25,7 @@ using MongoDB.Bson;
 using MongoDB.Bson.IO;
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Text;
 
 namespace BH.Engine.Serialiser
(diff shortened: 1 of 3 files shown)


FraserGreenroyd:
@BHoMBot check serialisation

FraserGreenroyd (approved):
Have reviewed the code and the changes make sense, and I cannot think of any other places currently where we would need to do this within the core engine.

I think it's worth capturing a [wiki page](https://bhom.xyz/documentation) on this convention for future reference, but that can be done independently of deploying this fix for ZCT or other platforms requiring it.

FraserGreenroyd:
@BHoMBot check compliance
@BHoMBot check required

FraserGreenroyd:
@BHoMBot this is a DevOps instruction. I am requesting neutral checks on: versioning, unit-tests, ready-to-merge

PR #3138, opened 2023-08-14
Structure_Engine: Add engine methods for foundation classes

### LINKED ISSUE #3137: Structure_oM: Add associated Engine methods for Foundation classes
#### Description:

See https://github.com/BHoM/BHoM/issues/1543 for definition of classes

peterjamesnugent:
### NOTE: Depends on 

https://github.com/BHoM/BHoM/pull/1545
   
Closes #3137 

### Test files

https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/01_Issue/BHoM/BHoM/Structure_oM/%231543-FoundationClasses/Foundation%20classes%20and%20associated%20methods.gh?csf=1&web=1&e=Q17tVv

### Changelog

- Added necessary methods for `PadFoundation` and `PileFoundation` to work with `IElement2D` and `IElementM` respectively;
- Added `Create` methods for `PadFoundation`, `PileFoundation` and `PileGroup`
- Updated `IsNull` methods to include new foundation objects;
- Added `GrahamScan` method to determine the perimeter of a `List<Point>` which is used in the `Create` methods for a `PileFoundation`.


peterjamesnugent:
@BHoMBot check required

Structure_Engine/Modify/Transform.cs
```diff
@@ -249,6 +249,33 @@ public static FEMesh Transform(this FEMesh mesh, TransformMatrix transform, doub
...
         /***************************************************/
+
+        /***************************************************/
+
+        [Description("Transforms the PadFoundation's edges, openings and basis by the transform matrix. Only rigid body transformations are supported.")]
+        [Input("padFoundation", "PadFoundation to transform.")]
+        [Input("transform", "Transform matrix.")]
+        [Input("tolerance", "Tolerance used in the check whether the input matrix is equivalent to the rigid body transformation.")]
+        [Output("transformed", "Modified PadFoundation with unchanged properties, but transformed edges and basis.")]
+        public static PadFoundation Transform(this PadFoundation padFoundation, TransformMatrix transform, double tolerance = Tolerance.Distance)
```
pawelbaran, line 260: Shouldn't we transform the edges here in the first place?

Structure_Engine/Query/OutlineElements1D.cs
```diff
@@ -0,0 +1,58 @@
...
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Returns the outline 1D elements of a PadFoundation.")]
+        [Input("padFoundation", "The PadFoundation to find the outline elements for.")]
+        [Output("outlineElements", "A collection of outline 1D elements.")]
+        public static List<IElement1D> OutlineElements1D(this PadFoundation padFoundation)
+        {
+            if (padFoundation.IsNull())
+                return new List<IElement1D>();
```
pawelbaran, line 50: I would rather return `null` here - how can a `null` object return a collection? As long as it is not a well established pattern (however arguable it feels to me now)

pawelbaran (commented):
Good stuff 👍 but same as in the oM, I left quite a few comments - hopefully they will be more help than distraction.

Geometry_Engine/Compute/GrahamScan.cs
```diff
@@ -0,0 +1,141 @@
...
+            if (pts.IsNullOrEmpty())
+            {
+                Base.Compute.RecordError("The point list is null or empty.");
+                return null;
+            }
+
+            // Find the point with the lowest Y coordination
+            IOrderedEnumerable<Point> orderedPts = pts.OrderBy(pt => pt.Y);
+
+            Point p = pts.First();
```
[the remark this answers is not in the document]
peterjamesnugent: It needs to be sorted to find the lowest Y coordinate, so I kept it as an `IOrderedEnumerable<Point> so we can check whether there is more than one point with the lowest Y coordinate. If there are more than one, we can `ThenBy` to sort by X. Otherwise, just to cast `ToList()`.

Geometry_Engine/Compute/GrahamScan.cs
```diff
@@ -0,0 +1,141 @@
...
+
+            // Check if there is more than one point with the lowest Y
+            if (Math.Abs(p.Y - pts[1].Y) < Tolerance.MicroDistance)
```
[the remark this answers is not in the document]
peterjamesnugent: Agree on all three, will ammend,  include the tolerance input and check for > 3 points.

Geometry_Engine/Compute/GrahamScan.cs
```diff
@@ -0,0 +1,141 @@
...
+                p = pts.First();
+            }
+            else
+                pts = orderedPts.ToList();
+
+            // Remove the P from the list of points
+            pts.Remove(p);
+
+            // Sort by the increasing order of the angle the point and P make with the x-axis
+            pts = pts.OrderBy(pt => Math.Atan2(pt.Y - p.Y, pt.X - p.X)).ToList(); // TODO: If several points have the same angle - break ties by increasing distance
```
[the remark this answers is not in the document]
peterjamesnugent: I have already actioned the TODO - so removed.

I have updated to include the dot product instead of the `atan` - good spot!

Geometry_Engine/Compute/GrahamScan.cs
```diff
@@ -0,0 +1,141 @@
...
+            };
+
+            // Due to the grouping, it is necessary to add the first point to close the polyline
+            if (duplicateAngle)
+                selPts.Add(selPts.First());
+
+            return selPts;
+        }
+
+        private static List<Point> GrahamScanInt(ref List<Point> pts, ref List<Point> selPts)
```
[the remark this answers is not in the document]
peterjamesnugent: Renamed to `GrahamScanSolver` as this method is the one doing the removing, and sorting.

Structure_Engine/Create/Elements/PadFoundation.cs
```diff
@@ -0,0 +1,53 @@
...
+        }
+
+        /***************************************************/
+
+        [Description("Creates a PadFoundation from its fundamental parts and a basis.")]
+        [Input("width", "The width of the PadFoundation aligned with Global X.")]
+        [Input("length", "The length of the PadFoundation aligned with Global Y.")]
+        [InputFromProperty("property")]
+        [Input("coordinates", "The Cartesian coordinate system to control the position and orientation of the PadFoundation.")]
+        public static PadFoundation PadFoundation(double width, double length, ConstantThickness thickness, Basis orientation)
```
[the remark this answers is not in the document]
peterjamesnugent: The idea being that you can create a single `PadFoundation` that can then be replicated `n` times over a layout with different orientations.

Structure_Engine/Create/Elements/PileFoundaton.cs
```diff
@@ -0,0 +1,56 @@
...
+        public static PileFoundation PileFoundation(List<PileGroup> pileGroups, ConstantThickness property, double offset, Basis orientation)
+        {
+            List<Point> pts = new List<Point>();
+
+            foreach (PileGroup pileGroup in pileGroups)
+            {
+                pts.AddRange(pileGroup.PileLayout.Points);
+            }
+
+            pts.Add(pts.First());
```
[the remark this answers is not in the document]
peterjamesnugent: To close the `Polyline` - don't mind either way as long as it's clear.

peterjamesnugent:
@BHoMBot check required

peterjamesnugent:

Structure_Engine/Create/Elements/PadFoundation.cs
```diff
@@ -0,0 +1,60 @@
...
+using BH.oM.Structure.Elements;
+using BH.oM.Structure.SurfaceProperties;
+using BH.Engine.Base;
+using BH.Engine.Geometry;
+using BH.Engine.Spatial;
+
+
+namespace BH.Engine.Structure
+{
+    public static partial class Create
```
pawelbaran, line 20: Curious if it would not be worth having `Create` methods taking following inputs:
- outlines as a collection of `ICurve`
- thickness property as `double` and `IMaterialFragment`

I will never use these so please fill free to ignore, just trying to put myself in the shoes of a GH user

Structure_Engine/Modify/SetGeometry.cs
```diff
@@ -0,0 +1,63 @@
...
+    {
+        /******************************************/
+        /****            IElement1D            ****/
+        /******************************************/
+
+        [Description("Modifies the geometry of a Pile to be the provided curve. The Piles other properties are unaffected.")]
+        [Input("pile", "The Pile to modify the geomerty of.")]
+        [Input("curve", "The new geometry curve for the Pile. Note a Pile is defined by the TopNode and BottomNode, the start and the end of the curve will be used.")]
+        [Output("element1D", "A IElement1D with the properties of 'element1D' and the location of 'curve'.")]
+        public static Pile SetGeometry(this Pile pile, ICurve curve)
```
pawelbaran, line 45: I would check if the curve is `Line` to avoid false positives

Structure_Engine/Modify/Transform.cs
```diff
@@ -249,6 +249,44 @@ public static FEMesh Transform(this FEMesh mesh, TransformMatrix transform, doub
...
+
+            List<Edge> edges = new List<Edge>();
+            foreach (Edge edge in result.TopOutline)
+            {
+                Edge clone = edge.ShallowClone();
+                clone.Curve = clone.Curve.ITransform(transform);
+                edges.Add(clone);
+            }
+
+            result.TopOutline = edges;
```
pawelbaran, line 283: What is happening to the orientation angle here? Should we orient it too? Probs easiest to look up how it is done for `Panel`

Structure_Engine/Objects/EqualityComparers/SectionPropertyComparer.cs
```diff
@@ -0,0 +1,65 @@
...
+ */
+
+
+using BH.oM.Structure.SectionProperties;
+using System.Collections.Generic;
+using System;
+
+namespace BH.Engine.Structure
+{
+    public class SectionPropertyComparer : IEqualityComparer<ISectionProperty>
```
pawelbaran, line 30: Is there a way of testing the comparers (for section and surface properties)? I think most knowledge about this concept is currently gone on leave together with @IsakNaslundBh, so I would not be overly confident with reviewing/approving this bit tbh.

Structure_Engine/Query/Elements0D.cs
```diff
@@ -0,0 +1,51 @@
...
+        /******************************************/
+        /****            IElement1D            ****/
+        /******************************************/
+
+        [Description("Gets the Element0Ds of a Pile, which for the case of a Pile means getting the StartNode and EndNode. Method necessary for IElement pattern.")]
+        [Input("pile", "The Pile to extract IElement0ds from.")]
+        [Output("element0Ds", "The list of Elements0D of the Pile, i.e. the StartNode and EndNode.")]
+        public static List<IElement0D> Elements0D(this Pile pile)
+        {
+            return new List<IElement0D> { pile.TopNode, pile.BottomNode };
```
pawelbaran, line 42: I would think of adding as many null checks/handlers as possible in the basic interface methods - such minor negligence likes to bite back at most unexpected moments

Structure_Engine/Query/VolumetricMaterialTakeoff.cs
```diff
@@ -0,0 +1,115 @@
...
+    public static partial class Query
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Gets the volumetric material takeoff from the PileFoundation object. The takeoff will contain materials and volumes from the PileCap and Piles.")]
+        [Input("pileFoundation", "The PileFoundation object to extract the volumetric material takeoff from.")]
+        [Output("volTakeoff", "The volumetric material takeoff based on buildup of the PileFoundation object.")]
+        public static VolumetricMaterialTakeoff VolumetricMaterialTakeoff(this PileFoundation pileFoundation)
```
pawelbaran, line 46: Are you sure we want/need the material takeoffs for analytical elements? My question originates from the fact that for some reason we do not have them for bars and panels, curious if it was not intended to for example push the users to use `Physical` objects for material purposes. I am 100% ignorant in the topic, so please feel free to ignore 👍

Geometry_Engine/Compute/GrahamScan.cs
```diff
@@ -0,0 +1,151 @@
...
+            {
+                Base.Compute.RecordError("The point list must be greater than three to determine the Convex Hull.");
+                return pts;
+            }
+
+            // Check that the points are all wihtin the same plane 
+            Plane plane = pts.FitPlane(tolerance);
+            if (!Query.IsInPlane(pts, Plane.XY))
+            {
+                Base.Compute.RecordError("The point list needs to be within the XY Plane.");
```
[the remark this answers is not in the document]
peterjamesnugent: That was a hangover when I tried (unsuccessfully) to make it orientation agnostic. Have opted for Option 3 because the `Plane` is needed to get the local axis to `Orient()` them.

Geometry_Engine/Compute/GrahamScan.cs
```diff
@@ -0,0 +1,151 @@
...
+                // Get points with lowest Y coordinate and select the point with the lowest X
+                pts = orderedPts.ThenBy(x => x.X).ToList();
+                p = pts.First();
+            }
+
+            // Remove the P from the list of points
+            pts.Remove(p);
+
+            // Calculate the angles between p and each pt
+            List<double> angles = pts.Select(pt => pt - p).Select(v => v.DotProduct(Vector.XAxis) / v.Length()).ToList();
```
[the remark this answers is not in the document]
peterjamesnugent: Only if the duplicate point is `p` - I have added it to the start of the method so that the point list can't be > 3 but full of duplicates.

Geometry_Engine/Compute/GrahamScan.cs
```diff
@@ -0,0 +1,151 @@
...
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Implements the GrahamScan algorithm to determine the convex hull of a list of points contained within the XY Plane.")]
+        [Input("pts", "The points to determine the convex hull contained within the XY Plane.")]
+        [Input("tolerance", "Geometrical tolerance to be used in the method.", typeof(Length))]
+        [Output("c", "The convex hull of the point list, no repeat points are returned.")]
+        public static List<Point> GrahamScan(List<Point> pts, double tolerance = Tolerance.MicroDistance)
```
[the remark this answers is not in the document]
peterjamesnugent: I had thought `MicroDistance` was 1E-03, my mistake! Have updated to `Tolerance.MacroDistance` - I generally use closest mm.

Structure_Engine/Create/Elements/PadFoundation.cs
peterjamesnugent: I think it's quite similar to the first `Create` method, can always be added in the future.

Structure_Engine/Modify/SetGeometry.cs
peterjamesnugent: I have made it clear in the description that you can provide any `ICurve` but only the start and end points will be used. You could have a scenario where the new `curve` is defined by a straight `Polyline` made up of many straight lines - and that would get rejected because it's not a `Line`.

Structure_Engine/Modify/SetOutlineElements1D.cs
```diff
@@ -0,0 +1,65 @@
...
+                     "The Method will return a new PadFoundation with the provided edges as the TopSurface.")]
+        [Input("padFoundation", "The PadFoundation to update the ExternalEdge of.")]
+        [Input("edges", "A list of IElement1Ds which all should be of a type of ICurve.")]
+        [Output("padFoundation", "A new PadFoundation with TopSurface matching the provided edges.")]
+        public static PadFoundation SetOutlineElements1D(this PadFoundation padFoundation, IEnumerable<IElement1D> edges)
+        {
+            if (padFoundation.IsNull())
+                return padFoundation;
+
+            List<Edge> externalEdges = edges.Cast<Edge>().ToList();
```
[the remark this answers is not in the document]
peterjamesnugent: Have changed the definition of TopOutline to `PolyCurve` and this casts to a `Polycurve` and joins them. The description makes clear that the curve should be closed and no planarity checks are made.

Structure_Engine/Objects/EqualityComparers/SectionPropertyComparer.cs
peterjamesnugent: This is the one bit I was really no confident on - I think the reason we don't have `SectionPropertyComparer` and `SurfacePropertyComparer` is because how many levels deep you have to compare (and then you'd have to have a `MaterialFragmentComparer` too). 

I only stuck these in because they're needed for `IsMergeableProperties` but I wonder if we can just return `false` if that method is absolutely required.

Structure_Engine/Query/VolumetricMaterialTakeoff.cs
peterjamesnugent: The `VolumetricMaterialTakeoff` does work for `Bar` and `Panel` though  - I'd been keen to have them included for `Structure.Elements` in general so material take offs can be done when optioneering different structural typologies.

Geometry_Engine/Compute/GrahamScan.cs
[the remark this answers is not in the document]
pawelbaran: I would be careful here tbh - for structural elements I think we always use `Tolerance.Distance` because 1e-3 may be not sensitive enough (would be good to have @IsakNaslundBh around to confirm), so using 1e-3 here would be a bit of a risk factor + an oddity too IMO. But the decision is yours, feel free to go either way as long as it works for you 👍

Geometry_Engine/Compute/GrahamScan.cs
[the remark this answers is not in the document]
pawelbaran: I think it could be done if you take normal of the plane fit and create a `Cartesian` coordinate system  with arbitrary X and Y, to then use it to orient to and back from the global. Wouldn't that work?

Geometry_Engine/Compute/GrahamScan.cs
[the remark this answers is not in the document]
peterjamesnugent: It's ultimately a method in the Geometry_Engine, so happy to adopt whatever default is used on other methods - looks like be `Tolerance.Distance`.

Structure_Engine/Modify/SetGeometry.cs
pawelbaran: Right - then maybe you could call `IIsLinear` query and throw an error if it returns `false`?

Structure_Engine/Objects/EqualityComparers/SectionPropertyComparer.cs
pawelbaran: No clue 🙈 Let's try to chat it through offline, maybe we can come up with something that would make us more confident

Structure_Engine/Query/VolumetricMaterialTakeoff.cs
pawelbaran: Right, I got confused by the fact that there had not been any `VolumetricMaterialTakeoff` method in Structure_Engine. But following that fact, if `Bar` works with a dedicated overload, why would you need it for `Pile`?

Structure_Engine/Objects/EqualityComparers/SectionPropertyComparer.cs
peterjamesnugent: As per discussion, I will remove the two methods for `Pile` and `PadFoundation` and we will rely on the `IElement1D` and `IElement2D` methods in the `Spatial_Engine`. The onus will be on the user to group similar `PadFoundation` and `Pile` (or use the `Filter` methods in `Data_Engine`.

Structure_Engine/Query/VolumetricMaterialTakeoff.cs
peterjamesnugent: Remove `VolumetricMaterialTakeoff` and add `SolidVolume` method for `Pile`/`PileFoundation` and `MaterialComposition` for `Pile`/`PileFoundation`. 

The `PadFoundation` are already covered for `IAreaElement` which `PadFoundation` implements.

peterjamesnugent:
@BHoMBot check required

peterjamesnugent:

peterjamesnugent:
@BHoMBot check required

Chrisshort92 (approved):
Functionality, terminology and application reviewed in Grasshopper with test file. Full script tested and passed to merge.

peterjamesnugent:
@BHoMBot check required

peterjamesnugent:
@BHoMBot check required

peterjamesnugent:
@BHoMBot check unit-tests

peterjamesnugent:
@BHoMBot check required

peterjamesnugent:
@BHoMBot check unit-tests

FraserGreenroyd:

FraserGreenroyd (approved):
Approving based on @Chrisshort92 review - following commits by @peterjamesnugent have been compliance updates.

FraserGreenroyd:
@BHoMBot check ready-to-merge

PR #3144, opened 2023-08-14
Serialiser_Engine: Improve size of JSON by excluding empty IBHoMObject properties

### LINKED ISSUE #3017: Serialiser_Engine: Improve the size of the Json string by ignoring empty values for IBHoMObject properties
#### Description:
This means:
 - Name
 - Fragments
 - Tags
 - CustomData

Something to be discussed...

FraserGreenroyd:
Fixes #3017 

Test file available [here](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?newTargetListUrl=%2Fsites%2FBHoM%2F02%5FCurrent&viewpath=%2Fsites%2FBHoM%2F02%5FCurrent%2FForms%2FAllItems%2Easpx&id=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F02%5FPull%20Request%2FBHoM%2FBHoM%5FEngine%2FSerialiser%5FEngine%2F%233144%2DJSONSize&viewid=a5a93cba%2Dfcca%2D46cc%2Da31a%2D80b8d5cbfd7b).

Test file is very simple - serialise any object implementing `IBHoMObject` (should be most - I went with an Environment oM Panel) without providing a Name or other property implemented from `IBHoMObject` - see that the JSON doesn't contain the key value pair for those properties. Then provide values to those properties and see them get included.

I opted to go with storing the property names on load so that if you're serialising a collection of `IBHoMObjects`, we don't lose time querying the same 5 property names over and over again. Also opted to query the `IBHoMObject` properties so that if we ever change them, we don't have to remember to update the list here.

I did consider the option of whether we would want to include empty properties if they were overridden (as @IsakNaslundBh has occasionally done to override the `Name` property so he can add his own description to it for certain Structural objects), but decided against it for now on the basis that this will only exclude them if they're empty - so even if they're overridden, if they have no value assigned, do we really want them taking up space in the JSON string? I'm arguing for not - but we could go the other way if there's strong argument to.

### DIFF
--- a/Serialiser_Engine/Compute/Serialise/IBHoMObject.cs
+++ b/Serialiser_Engine/Compute/Serialise/IBHoMObject.cs
@@ -53,11 +53,33 @@ namespace BH.Engine.Serialiser
             foreach (PropertyInfo prop in value.GetType().GetProperties())
             {
                 var propertyValue = prop.GetValue(value);
-                if ((propertyValue == null || string.IsNullOrEmpty(propertyValue.ToString())) && m_PropertyNamesToIgnore.Contains(prop.Name))
-                    continue; //Don't write out null or empty properties if they're part of the base IBHoMObject properties
+                bool include = true;
 
-                writer.WriteName(prop.Name);
-                ISerialise(propertyValue, writer, prop.PropertyType);
+                switch (prop.Name)
+                {
+                    case "Name":
+                        include = !string.IsNullOrEmpty(propertyValue.ToString()); //If string is not null or empty - include the property
+                        break;
+                    case "Fragments":
+                        include = ((FragmentSet)propertyValue).Count > 0; //If fragment count is greater than 0, include it
+                        break;
+                    case "Tags":
+                        include = ((HashSet<string>)propertyValue).Count > 0; //As with fragments
+                        break;
+                    case "CustomData":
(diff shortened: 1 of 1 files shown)


adecler (changes requested):
The code doesn't cover the case for the `IEnumerable` properties (i.e. `Tags`, `Fragments`, And `CustomData`) as they will show in json even if empty:

image

In this case, I wonder if it wouldn't be simpler to use a switch-case on `prop.Name` to explicitly list  the properties to treat differently (simpler and more transparent) ?

FraserGreenroyd:
image

🤦‍♂️ much cleaner! Have implemented the `switch` - though this does mean we'll need to keep it updated if we ever change the `IBHoMObject` properties - but given we haven't done that in years, I think that's a minimal overhead compared to the greater transparency this gives.

FraserGreenroyd:
@BHoMBot check compliance
@BHoMBot check required

FraserGreenroyd:
@BHoMBot check installer

FraserGreenroyd:
@BHoMBot this is a DevOps instruction. I am requesting neutral checks on: versioning, unit-tests, installer, ready-to-merge

PR #3191, opened 2023-10-06
BHoM_Engine: RunExtensionMethodAsync and TryRunExtensionMethodAsync added

### LINKED ISSUE #3140: BHoM_Engine: add async version of RunExtensionMethod
#### Description:

As in the title, async versions of `RunExtensionMethod` and `TryRunExtensionMethod` methods would be useful.

pawelbaran:

Closes #3140

### Test files

Write a simple `public static async` Engine method taking `BHoMObject` and try running it via script uploaded [here](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?RootFolder=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F02%5FPull%20Request%2FBHoM%2FBHoM%5FEngine%2FBHoM%5FEngine%2F%233191%2DRunExtensionMethodAsync&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4).

### Additional comments

The methods added are to large extent a rewrite of `TryRunExtensionMethod`. However, `async` methods are not allowed to have `out` parameters, therefore I changed the method signature compared to synchronous flavour - happy to discuss if anyone has different/better ideas :+1:

### DIFF
--- a/BHoM_Engine/Compute/RunExtensionMethod.cs
+++ b/BHoM_Engine/Compute/RunExtensionMethod.cs
@@ -20,12 +20,10 @@
  * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
  */
 
+using BH.oM.Base;
 using BH.oM.Base.Attributes;
-using System;
-using System.Collections.Concurrent;
 using System.ComponentModel;
-using System.Linq;
-using System.Reflection;
+using System.Threading.Tasks;
 
 namespace BH.Engine.Base
 {
@@ -60,6 +58,39 @@ namespace BH.Engine.Base
         {
             if(TryRunExtensionMethod(target, methodName, parameters, out object result))
                 return result;
+            else
+                return null;
+        }
+
+        /***************************************************/
+
+        [Description("Asynchronously runs an extension method accepting a single argument based on a provided object and method name.\n" +
+                     "Finds the method via reflection the first time it is run, then compiles it to a function and stores it for subsequent calls.")]
+        [Input("target", "The object to find and run the extension method for.")]
+        [Input("methodName", "The name of the method to be run.")]
+        [Output("result", "The result of the method execution. If no method was found, null is returned.")]
+        public static async Task<object> RunExtensionMethodAsync(object target, string methodName)
+        {
+            Output<bool, object> result = await TryRunExtensionMethodAsync(target, methodName);
+            if (result.Item1)
+                return result.Item2;
+            else
+                return null;
+        }
+
+        /***************************************************/
+
+        [Description("Asynchronously runs an extension method accepting multiple arguments based on a provided main object and method name and additional arguments.\n" +
+                     "Finds the method via reflection the first time it is run, then compiles it to a function and stores it for subsequent calls.")]
+        [Input("target", "The first of the argument of the method to find and run the extention method for.")]
+        [Input("methodName", "The name of the method to be run.")]
+        [Input("parameters", "The additional arguments of the call to the method, skipping the first argument provided by 'target'.")]
+        [Output("result", "The result of the method execution. If no method was found, null is returned.")]
+        public static async Task<object> RunExtensionMethodAsync(object target, string methodName, object[] parameters)
+        {
+            Output<bool, object> result = await TryRunExtensionMethodAsync(target, methodName, parameters);
+            if (result.Item1)
+                return result.Item2;
             else
                 return null;
         }
--- a/BHoM_Engine/Compute/TryRunExtensionMethod.cs
+++ b/BHoM_Engine/Compute/TryRunExtensionMethod.cs
@@ -20,19 +20,15 @@
  * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
  */
 
-using BH.oM.Base.Debugging;
-using System;
-using System.Collections.Generic;
-using System.Diagnostics;
-using System.IO;
-using System.Linq;
-using System.Runtime.CompilerServices;
-using System.Reflection;
-using System.Text;
-using System.Threading.Tasks;
-using System.ComponentModel;
+using BH.oM.Base;
 using BH.oM.Base.Attributes;
+using System;
 using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Linq;
+using System.Reflection;
+using System.Threading.Tasks;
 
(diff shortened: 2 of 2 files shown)


alelom (commented):
This looks good to me. Because this is a "backend" base method, could you please add an Unit test in the BHoM_Engine_Tests solution too? 

It can be a simple test that runs any existing extension method available in the base BHoM_Engine asynchronously, like `Geometry()` for example.

This is where the test project is: https://github.com/BHoM/BHoM_Engine/tree/develop/.ci/unit-tests/Base_Engine_Tests
and this new test should sit in a new `Compute` folder, in a file called like the method you're testing. Just follow the example of the [IsNumeric unit test](https://github.com/BHoM/BHoM_Engine/blob/develop/.ci/unit-tests/Base_Engine_Tests/Query/IsNumeric.cs), for instance.

pawelbaran:
> This looks good to me. Because this is a "backend" base method, could you please add an Unit test in the BHoM_Engine_Tests solution too?
> 
> It can be a simple test that runs any existing extension method available in the base BHoM_Engine asynchronously, like `Geometry()` for example.
> 
> This is where the test project is: https://github.com/BHoM/BHoM_Engine/tree/develop/.ci/unit-tests/Base_Engine_Tests and this new test should sit in a new `Compute` folder, in a file called like the method you're testing. Just follow the example of the [IsNumeric unit test](https://github.com/BHoM/BHoM_Engine/blob/develop/.ci/unit-tests/Base_Engine_Tests/Query/IsNumeric.cs), for instance.

Thanks @alelom for having a look. I was thinking about adding a unit test before I made one up in GH, but the problem is that all Engine methods are synchronous, and it is not possible to call one via `RunExtensionMethodAsync` - you naturally get `MethodName does not contain a definition for 'GetAwaiter'` exception. So I would need to add a dummy async extension method to the engine or call `RunExtensionMethodAsync` inside `RunExtensionMethodAsync`, none of which I was not a big fan of. What do you think?

FraserGreenroyd:
> So I would need to add a dummy async extension method to the engine or call `RunExtensionMethodAsync` inside `RunExtensionMethodAsync`, none of which I was not a big fan of. What do you think?

Haven't read everything through yet, but would agree being against adding a dummy method to the engine for the sake of testing 👍

alelom:
> > This looks good to me. Because this is a "backend" base method, could you please add an Unit test in the BHoM_Engine_Tests solution too?
> > It can be a simple test that runs any existing extension method available in the base BHoM_Engine asynchronously, like `Geometry()` for example.
> > This is where the test project is: https://github.com/BHoM/BHoM_Engine/tree/develop/.ci/unit-tests/Base_Engine_Tests and this new test should sit in a new `Compute` folder, in a file called like the method you're testing. Just follow the example of the [IsNumeric unit test](https://github.com/BHoM/BHoM_Engine/blob/develop/.ci/unit-tests/Base_Engine_Tests/Query/IsNumeric.cs), for instance.
> 
> Thanks @alelom for having a look. I was thinking about adding a unit test before I made one up in GH, but the problem is that all Engine methods are synchronous, and it is not possible to call one via `RunExtensionMethodAsync` - you naturally get `MethodName does not contain a definition for 'GetAwaiter'` exception. So I would need to add a dummy async extension method to the engine or call `RunExtensionMethodAsync` inside `RunExtensionMethodAsync`, none of which I was not a big fan of. What do you think?

It's actually well possible in DotNet -- just wrap the synchronous method in a `Task.Run()` invocation. E.g.
```cs
var result = await Task.Run(() => MySyncMethod());
```

alelom:
> > It's actually well possible in DotNet -- just wrap the synchronous method in a `Task.Run()` invocation. E.g.
> > ```cs
> > var result = await Task.Run(() => MySyncMethod());
> > ```
> 
> Of course, if you have a chance to call `Task.Run` - here we find the method by name and then await the method call:
> 
> https://github.com/BHoM/BHoM_Engine/blob/181d3c2c3a53afb629473a9c71b14e2213d4bf61/BHoM_Engine/Compute/TryRunExtensionMethod.cs#L141

Yeah sure I meant maybe you could go ahead with a dummy method like:

```cs

public async Task<IGeometry> GeometryAsync(IObject obj)
{
    return await Task.Run(() => obj.Geometry());
}

```

pawelbaran:
> Yeah sure I meant maybe you could go ahead with a dummy method like:
> 
> ```cs
> public static async Task<IGeometry> GeometryAsync(this IObject obj)
> {
>     return await Task.Run(() => obj.Geometry());
> }
> ```
> 
> and use that for the invocation in the test. Or something similar

Yeah, but then it needs to be added to a legit engine project to be reflected in the pool parsed by `RunExtenstionMethod`. So we end up with the situation commented by @FraserGreenroyd [here](https://github.com/BHoM/BHoM_Engine/pull/3191#issuecomment-1752660415), or I would need to create a dedicated engine for this one method...

alelom:
> > Yeah sure I meant maybe you could go ahead with a dummy method like:
> > ```cs
> > public static async Task<IGeometry> GeometryAsync(this IObject obj)
> > {
> >     return await Task.Run(() => obj.Geometry());
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and use that for the invocation in the test. Or something similar
> 
> Yeah, but then it needs to be added to a legit engine project to be reflected in the pool parsed by `RunExtenstionMethod`. So we end up with the situation commented by @FraserGreenroyd [here](https://github.com/BHoM/BHoM_Engine/pull/3191#issuecomment-1752660415), or I would need to create a dedicated engine for this one method...

Apologies, it all came back to my mind – this is what I was thinking of having you do. However, I just remembered that I had prepared a "Test Engine" exactly for this purpose – mainly, for testing backend methods like the RunExtensionMethod. I had even added some unit tests, and they were all running fine locally. However, we had immense trouble making them run automatically via CI/CD when BHoMBot attempted to run them via NUnit. I ended up removing them after weeks of trying. 

tl;dr: let's not add this unit test method. Sorry for the long back and forth, I think it was useful to record this exchange anyways.

FraserGreenroyd:
@BHoMBot check compliance
@BHoMBot check required

FraserGreenroyd:

alelom (approved):
Approving based on stated procedure and our conversation above.

FraserGreenroyd:
@BHoMBot check ready-to-merge

PR #3193, opened 2023-10-12
Analytical_Engine: Resolve Dictionary Key Duplication Error

### LINKED ISSUE #3192: Analytics_Engine: Graph Dictionary Key Duplication Error
#### Description:

The graph method at: 
https://github.com/BHoM/BHoM_Engine/blob/eb756492dbf5274f069bdcbbcac9641c6e19107b/Analytical_Engine/Create/Graph/Graph.cs#L126-L154
fails on 
https://github.com/BHoM/BHoM_Engine/blob/eb756492dbf5274f069bdcbbcac9641c6e19107b/Analytical_Engine/Create/Graph/Graph.cs#L150
given there is no check to see if the key already exists in the dictionary.

Discovery shows that:

- [ ] Safeguard Dictionary Add method by checking first with `.ContainsKey()`

travispotterBH:
Checks against Dictionary Duplicated Key added

Closes #3192

### DIFF
--- a/Analytical_Engine/Create/Graph/Graph.cs
+++ b/Analytical_Engine/Create/Graph/Graph.cs
@@ -159,7 +159,6 @@ namespace BH.Engine.Analytical
                 {
                     graph.Entities.Add(((IBHoMObject)n).BHoM_Guid, ((IBHoMObject)n));
                 }
-
             });
 
             graph.Relations = relations;
--- a/Analytical_Engine/Create/Graph/Graph.cs
+++ b/Analytical_Engine/Create/Graph/Graph.cs
@@ -66,7 +66,7 @@ namespace BH.Engine.Analytical
         }
 
         /***************************************************/
-        
+
         [Description("Create a graph from a collection of IBHoMObjects, a collection of IRelations, property names and decimal places to determine unique graph entities.")]
         [Input("entities", "Optional collection of IBHoMOBjects to use as Graph entities. Entities can include DependencyFragments to determine the Graph Relations.")]
         [Input("relations", "Optional collection of IRelations to use as Graph Relations. Relations can include sub Graphs containing the entities to be used in the Graph.")]
@@ -91,24 +91,24 @@ namespace BH.Engine.Analytical
             m_MatchedObjects = Query.UniqueEntitiesReplacementMap(clonedEntities, comparisonConfig);
 
             //convert dependency fragments attached to entities and add to relations
-            clonedEntities.ForEach(ent => clonedRelations.AddRange(ent.ToRelation())); 
+            clonedEntities.ForEach(ent => clonedRelations.AddRange(ent.ToRelation()));
 
             //add to graph
             graph.Relations.AddRange(clonedRelations);
 
             //add unique objects
-            foreach (KeyValuePair< Guid, IBHoMObject> kvp in m_MatchedObjects)
+            foreach (KeyValuePair<Guid, IBHoMObject> kvp in m_MatchedObjects)
             {
- 
+
                 if (!graph.Entities.ContainsKey(kvp.Value.BHoM_Guid))
                     graph.Entities.Add(kvp.Value.BHoM_Guid, kvp.Value);
-                
+
             }
             graph = graph.UniqueEntities(m_MatchedObjects);
 
             graph.UniqueEntityNames();
 
-            if(graph.Relations.Count == 0)
+            if (graph.Relations.Count == 0)
                 Base.Compute.RecordWarning("No Relations have been defined for this graph.");
 
             return graph;
@@ -129,6 +129,12 @@ namespace BH.Engine.Analytical
             if (entities == null)
                 entities = new List<IElement0D>();
 
+            if (!entities.All(e => e is IBHoMObject && e is IGeometry))
+            {
+                Base.Compute.RecordError("All entities must be of type IBHoMObject and IGeometry.");
+                return null;
+            }
+
             List<IElement0D> entitiesCloned = entities.DeepClone();
 
             List<IRelation> relations = new List<IRelation>();
@@ -145,9 +151,18 @@ namespace BH.Engine.Analytical
                 };
                 relations.AddRange(RelationsToAdd(relation, relationDirection));
             }
+
             Graph graph = new Graph();
-            
-            entitiesCloned.ForEach(n => graph.Entities.Add(((IBHoMObject)n).BHoM_Guid, ((IBHoMObject)n)));
+
+            entitiesCloned.ForEach(n =>
+            {
+                if (!graph.Entities.ContainsKey(((IBHoMObject)n).BHoM_Guid))
+                {
+                    graph.Entities.Add(((IBHoMObject)n).BHoM_Guid, ((IBHoMObject)n));
+                }
+
+            });
+
             graph.Relations = relations;
             graph.UniqueEntityNames();
 
@@ -173,7 +188,7 @@ namespace BH.Engine.Analytical
             {
                 Point p = Geometry.Create.RandomPoint(m_Rnd, boundingBox);
                 IElement0D entity = prototypeEntity.ClonePositionGuid(p);
-                
+
                 if (!ToCloseToAny(entities, entity, tolerance))
                     entities.Add(entity);
 
@@ -253,7 +268,7 @@ namespace BH.Engine.Analytical
                                 Target = c.BHoM_Guid
                             };
                             graph.Relations.AddRange(RelationsToAdd(relation, relationDirection));
-                                
+
                         }
                     }
                 }
@@ -275,7 +290,7 @@ namespace BH.Engine.Analytical
                 entity = prototypeEntity.ClonePositionGuid(point);
                 entities.Add(entity);
             }
-                
+
             return entity;
         }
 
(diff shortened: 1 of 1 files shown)


Analytical_Engine/Create/Graph/Graph.cs
```diff
@@ -91,24 +91,24 @@ public static Graph Graph(List<IBHoMObject> entities = null, List<IRelation> rel
...
 
             //add to graph
             graph.Relations.AddRange(clonedRelations);
 
             //add unique objects
-            foreach (KeyValuePair< Guid, IBHoMObject> kvp in m_MatchedObjects)
+            foreach (KeyValuePair<Guid, IBHoMObject> kvp in m_MatchedObjects)
             {
- 
+
```
FraserGreenroyd, line 102:
```suggestion
```

Analytical_Engine/Create/Graph/Graph.cs
```diff
@@ -129,6 +129,12 @@ public static Graph Graph<T>(List<ICurve> connectingCurves, T prototypeEntity, L
             if (entities == null)
                 entities = new List<IElement0D>();
 
+            if (!entities.All(e => e is IBHoMObject && e is IGeometry))
```
FraserGreenroyd, line 132: This would now exclude objects which were of type [INode](https://github.com/BHoM/BHoM/blob/develop/Analytical_oM/Elements/INode.cs) from working with the graphing capabilities of Analytical_Engine, which is fairly key for this engine to remain capable of working on objects from its own linked oM in my opinion.

Analytical_Engine/Create/Graph/Graph.cs
```diff
@@ -145,9 +151,18 @@ public static Graph Graph<T>(List<ICurve> connectingCurves, T prototypeEntity, L
...
-            
-            entitiesCloned.ForEach(n => graph.Entities.Add(((IBHoMObject)n).BHoM_Guid, ((IBHoMObject)n)));
+
+            entitiesCloned.ForEach(n =>
+            {
+                if (!graph.Entities.ContainsKey(((IBHoMObject)n).BHoM_Guid))
+                {
+                    graph.Entities.Add(((IBHoMObject)n).BHoM_Guid, ((IBHoMObject)n));
+                }
+
```
FraserGreenroyd, line 163:
```suggestion
```

FraserGreenroyd (changes requested):

FraserGreenroyd:
@travispotterBH also please ensure the PR title matches the engine being modified - I've updated for this PR but just a reminder for future 😄 

Also, for work like this, I would refrain from running the heavy bot checks until after a review is provided, as it risks blocking the bot for other work to check unfinished work, so unless you're trying to resolve a versioning issue for example, or you need an installer out to support testing, I would recommend running compliance checks at the early stages of a PR and then moving towards the heavy checks once the PR starts to have consensus just to keep bot resource available for all 😄

Analytical_Engine/Create/Graph/Graph.cs
```diff
@@ -129,6 +128,12 @@ public static Graph Graph<T>(List<ICurve> connectingCurves, T prototypeEntity, L
             if (entities == null)
                 entities = new List<IElement0D>();
 
+            if (!entities.All(e => e is IBHoMObject && e is IGeometry))
```
travispotterBH, line 131:
```suggestion
            if (!entities.All(e => e is IBHoMObject))
```

Analytical_Engine/Create/Graph/Graph.cs
```diff
@@ -129,6 +128,12 @@ public static Graph Graph<T>(List<ICurve> connectingCurves, T prototypeEntity, L
             if (entities == null)
                 entities = new List<IElement0D>();
 
+            if (!entities.All(e => e is IBHoMObject))
```
FraserGreenroyd, line 131: Given the incoming `entities` is already a `List<IBHoMObject>` I'm not sure what benefit this if statement will have.

This feels a little bit like trying to put a plaster on a flesh wound - from reading through the issue there is an issue with certain object types and invalid casts, but it feels like this PR is solving a specific test case for the `Fitting` objects, but without considering the remainder of the algorithm(s) or uses.

Might be best to have a call on this and look to leverage some resourcing mechanisms cause I think this requires a more detailed review/overhaul than one PR can sufficiently patch?

FraserGreenroyd (changes requested):

Analytical_Engine/Create/Graph/Graph.cs
travispotterBH: IElement0D never interfaces with nor inherits from BHoMObject or IBHoMObject....

FraserGreenroyd: Which would suggest no `IElement0D` could ever end up in this list unless the object itself also implements `IBHoMObject` - but this check is still fairly redundant given the input object is of type `List<IBHoMObject>` already?

travispotterBH: https://github.com/BHoM/BHoM/blob/3264411efd671aeb808f79ab6d29d7cc4f7ba3cb/Geometry_oM/Vector/Point.cs#L30

Not a BHoMObject. 

Is IElement0D

travispotterBH: Our Anemic Domain Model, with use of extension methods only, allows for the problem where since interfaces cannot enforce properties or methods as a contract, that it is entirely possible that an object may satisfies the polymorphic relationship in terms of "is a" relationship, but does make it entirely possible that the desired contract of props and methods is not satisfied.

FraserGreenroyd: Aye, but [this line](https://github.com/BHoM/BHoM_Engine/blob/develop/Analytical_Engine/Create/Graph/Graph.cs#L75) shows the input `entities` is of type `List<IBHoMObject>`, which means you wouldn't be able to have a `Point` in the list in the first place. Therefore, this `if` statement checking whether all the objects in list `entities` are `IBHoMObject`s is redundant because you couldn't get a Point object in the list to start with? E.G:

image

Bit small screenshot, apologies, but because `Point` is not an `IBHoMObject`, we can't add it to collections of `IBHoMObject` which the input of this method is. If the input was `object` or `IObject` or `IElement0D` then I think the if statement would be needed 😄

travispotterBH: But taking a step back, it would be good to know why the Guid is so important, which is why the object needs to be a BHoMObject as in lines here: https://github.com/BHoM/BHoM_Engine/blob/eb756492dbf5274f069bdcbbcac9641c6e19107b/Analytical_Engine/Create/Graph/Graph.cs#L140-L145

This is the root issue maybe that I have given too much weight to

FraserGreenroyd: Agreed - it looks like the `Guid` was being used to uniquely identify objects when they might be more complex than just 3 vertices (as a point is), but given `BHoM_Guid` cannot be guaranteed to be the same every time, matching based on that is usually a recipe for disaster.

I would encourage us to consider the `Relation` object and maybe refactor it if needed. I think we could refactor it do be `IElement0D` for the Source and Target and use the whole object rather than just the `Guid` - though maybe the `Guid` was used for speed because some `IElement0D` objects can be data heavy when passing around/using within the system.

FraserGreenroyd:
During mid-sprint touchdown 23.10.23 - @travispotterBH to reduce scope of PR following discussion with @FraserGreenroyd to just handle the dictionary adding error.

travispotterBH:
Updated PR and Issue to only add Dictionary Contains Key Check

Analytical_Engine/Create/Graph/Graph.cs
```diff
@@ -145,9 +144,17 @@ public static Graph Graph<T>(List<ICurve> connectingCurves, T prototypeEntity, L
...
             Graph graph = new Graph();
-            
-            entitiesCloned.ForEach(n => graph.Entities.Add(((IBHoMObject)n).BHoM_Guid, ((IBHoMObject)n)));
+
+            entitiesCloned.ForEach(n =>
+            {
+                if (!graph.Entities.ContainsKey(((IBHoMObject)n).BHoM_Guid))
+                {
+                    graph.Entities.Add(((IBHoMObject)n).BHoM_Guid, ((IBHoMObject)n));
+                }
```
FraserGreenroyd, line 155:
```suggestion
                var bo = (IBHoMObject)n;
                if (bo != null && !graph.Entities.ContainsKey(bo.BHoM_Guid))
                {
                    graph.Entities.Add(bo.BHoM_Guid, bo);
                }
```

FraserGreenroyd (changes requested):
Happy with the scope now, one minor change requested to protect as belt and braces.

FraserGreenroyd (approved):
Looks good now following discussions, approving on code review. Cheers @travispotterBH 😄

FraserGreenroyd:
@BHoMBot check compliance

FraserGreenroyd:
@BHoMBot check required

FraserGreenroyd:
@BHoMBot check unit-tests
@BHoMBot check ready-to-merge

FraserGreenroyd:

PR #3202, opened 2023-10-31
Physical_Engine: Add relevant engine methods associated with the new foundation objects

### LINKED ISSUE #3201: Physical_Engine: Create associated engine methods with Physical foundation objects
#### Description:

`Create` and `IsNull` methods for `Pile`, `PadFoundation` and `PileFoundation`.

See https://github.com/BHoM/BHoM/issues/566 and https://github.com/BHoM/BHoM/issues/1155 for further information on foundation objects.

peterjamesnugent:
### NOTE: Depends on 

https://github.com/BHoM/BHoM/pull/1559
   
Closes #3201 

### Test files

https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/01_Issue/BHoM/BHoM_Engine/Physical_Engine/%233201%20-%20Foundation%20engine%20methods/%233201%20Foundation%20methods.gh?csf=1&web=1&e=ArT00o

### Changelog

- Added `Create` methods for `PadFoundation` and `PileFoundation` objects;
- Updated descriptions for `Pile` create method and ammended `Create` method;
- Added `IsNull` methods for `Pile`, `PadFoundation`, `PileFoundation`, `IFramingElementProperty` and `IConstruction`
- Added relevant `IElement1D`methods  for `Pile` including `Geometry()` (other methods covered by `IFramingElement` inheritance;
- Added relevant `IElement2D` methods for `PadFoundation` including `OutlineElements1D`, `SetOutlineElements1D` and `HasMegeableProprties()`
- Added relevant `IElementM` methods for `PadFoundation`, `Pile` and `PileFoundation` including `SolidVolume`, `MaterialComposition` and `HasMergeableProperties`;
- Added `IsWithinPileCap` method to check whether the `Pile.Location` top exists within the 3D geometry of the `PadFoundation`;
- Added `EmbedVolume` that calculates the volume of the `Pile` that is embed within the `PadFoundation` - when a `PileFoundation` is used this method is utilised within `SolidVolume()`.


Physical_Engine/Create/PadFoundation.cs
```diff
@@ -0,0 +1,56 @@
...
+    public static partial class Create
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Creates a physical PadFoundation element. To generate elements compatible with structural packages refer to BH.oM.Structure.Elements.PadFoundation.")]
+        [Input("location", "PlanarSurface defining the top face of the PadFoundation.")]
+        [Input("construction", "Construction of the PadFoundation, containing its thickness and Material.")]
+        [Input("name", "The name of the Pile, default empty string.")]
```
pawelbaran, line 43:
```suggestion
        [Input("name", "The name of the PadFoundation, default empty string.")]
```

Physical_Engine/Create/PadFoundation.cs
```diff
@@ -0,0 +1,56 @@
...
+        [Output("padFoundation", "The created physical padFoundation.")]
```
pawelbaran, line 44:
```suggestion
        [Output("padFoundation", "The created physical PadFoundation.")]
```

Physical_Engine/Create/PileFoundation.cs
```diff
@@ -0,0 +1,58 @@
...
+{
+    public static partial class Create
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Creates a physical PileFoundation element. To generate elements compatible with structural packages refer to BH.oM.Structure.Elements.PileFoundation.")]
+        [Input("pileCap", "PadFoundation defining the pile cap and the extents of Piles.")]
+        [Input("piles", "Construction of the PadFoundation, containing its thickness and Material.")]
```
pawelbaran, line 44:
```suggestion
        [Input("piles", "Collection of Pile elements defining the piles belonging to the created PadFoundation.")]
```

Physical_Engine/Create/PileFoundation.cs
```diff
@@ -0,0 +1,58 @@
...
+namespace BH.Engine.Physical
+{
+    public static partial class Create
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Creates a physical PileFoundation element. To generate elements compatible with structural packages refer to BH.oM.Structure.Elements.PileFoundation.")]
+        [Input("pileCap", "PadFoundation defining the pile cap and the extents of Piles.")]
```
pawelbaran, line 43: What do you mean by extents of piles?

Physical_Engine/Create/PileFoundation.cs
```diff
@@ -0,0 +1,58 @@
...
+        [Input("piles", "Construction of the PadFoundation, containing its thickness and Material.")]
+        [Input("name", "The name of the Pile, default empty string.")]
```
pawelbaran, line 45:
```suggestion
        [Input("name", "The name of the PileFoundation, default empty string.")]
```

pawelbaran (changes requested):
A few minor comments on descriptions. Three other things worth raising:
- Could you please provide a test file including null inputs? I think it is worth double checking if the error messages are correct and consistent
- I started thinking if the pile foundation should not be `IImmutable`, with Create method to check if the piles lie within the outline of the pile cap etc.
- Another follow-up thought is whether we should not define conventions for pile length, i.e. should the pile curve extend to the top surface of a pile cap? How should pile length  be calculated then? Should it be shortened by depth of the pile? Differences between `Physical` and `Structural`?
    image

Physical_Engine/Create/PileFoundation.cs
peterjamesnugent: The `PileFoundation` sets the region in which `Pile`s can be located i.e. a `Pile` cannot be located outside the `PileCap`.

pawelbaran: But am I correct that it is just textual information, we do not block it programmatically on the level of creation? So that in practice we stick to garbage in == garbage out? It actually corresponds to one of my questions left in the review (the `IImmutable` one).

peterjamesnugent:
@pawelbaran:
- Test script added;
- I have added a `WithinPileCap()` method to check whether the `Pile`s exist within the 3D geometry of the `PadFoundation`;
- This is used on the `Create` method, which I think is a useful check, I cannot think of any reason why you would want a `Pile` located outside the `PileCap`;
- Pile length should be measured from the top of the `Pile` to its base, typically there is a cut off where the `Pile` extends in to the `PileCap`. I think the Structural and Physical calculations would be the same.

Physical_Engine/Create/PileFoundation.cs
peterjamesnugent: So with the `WithinPileCap` we do block is programmatically - but as below I cannot think of any good reason why you would. Therefore, it's useful to flag to the user as it is likely in error.

peterjamesnugent:
- `Convert` methods for Structure to Physical and Physical to Structure
- IElement1D and IElement2D methods for Physical

Physical_Engine/Query/EmbedVolume.cs
```diff
@@ -0,0 +1,93 @@
...
+        [Description("Determines whether the Piles are within the volume of the Piles that are embed in to the pile cap.")]
+        [Input("padFoundation", "The PadFoundation that defines the extents for the Piles to be located.")]
+        [Input("piles", "The Piles to calculate the embeded volume.")]
+        [Input("tolerance", "Distance tolerance to be used in the method.", typeof(Length))]
+        [Output("embedVolume", "True if all Piles are contained within the boundary and depth of the PileCap. False if one or more Piles are not.")]
+        public static double EmbedVolume(this PadFoundation padFoundation, List<Pile> piles, double tolerance = Tolerance.MicroDistance)
+        {
+            if (padFoundation.IsNull())
+                return 0;
+            else if (piles == null)
```
pawelbaran, line 54: I can see this check with an error being repeated a few times - wouldn't it be worth writing an `IsNull` overload for `IEnumerable<Pile>`?

Physical_Engine/Query/EmbedVolume.cs
```diff
@@ -0,0 +1,93 @@
...
+                return 0;
+
+            if (!IsWithinPileCap(padFoundation, piles, tolerance))
+            {
+                Base.Compute.RecordError("The list of Piles must be within the extents and depth of the pile cap,.");
+                return 0;
+            }
+
+            // Get the top of the piles
+            List<Point> tops = piles.Select(p => p.Location.IControlPoints().OrderBy(pt => pt.Z).First()).ToList();
```
pawelbaran, line 69:
```suggestion
            List<Point> tops = piles.Select(p => p.Location.IEndPoints().OrderBy(pt => pt.Z).Last()).ToList();
```

Physical_Engine/Query/EmbedVolume.cs
```diff
@@ -0,0 +1,93 @@
...
+
+            // Offset the surface for the padfoundation
+            PlanarSurface bottom = padFoundation.Location.Translate(new Vector() { Z = -padFoundation.Construction.IThickness() });
```
pawelbaran, line 72: Since we already know the piles are within the outline of a pile cap, it should be enough to work off `Plane` instead of `PlanarSurface`, which is much cheaper computationally.
```suggestion
            Plane bottom = padFoundation.Location.FitPlane();
            if (bottom.Normal.Z < 0)
                bottom.Normal *= -1;
            
            bottom.Origin -= bottom.Normal * padFoundation.Construction.IThickness();
```

Physical_Engine/Query/EmbedVolume.cs
```diff
@@ -0,0 +1,93 @@
...
+            double embedVolume = 0;
+
+            //For each pile, get the intersection with the bottom surface, get the line extending in to the pile cap and calculate the solid volume
+            for (int i = 0; i < tops.Count; i++)
+            {
+                Point pileIntersect = bottom.ClosestPoint(tops[i]);
+                ICurve split = piles[i].Location.ISplitAtPoints(new List<Point>() { pileIntersect })[0];
+                Pile embedPile = piles[i].ShallowClone();
+                embedPile = (Pile)embedPile.SetGeometry(split);
+                embedVolume += embedPile.SolidVolume();
```
pawelbaran, line 83:
```suggestion
                Point top = tops[i];
                Point projected = top.Project(bottom);
                if (top.Z > projected.Z)
                {
                    Pile embedPile = (Pile)embedPile.SetGeometry(new Line { Start = projected, End = top });
                    embedVolume += embedPile.SolidVolume();
                }
```

Physical_Engine/Query/Geometry.cs
```diff
@@ -59,6 +59,39 @@ public static oM.Geometry.ISurface Geometry(this oM.Physical.Elements.ISurface s
 
         /***************************************************/
 
+        [Description("Gets the geometry of a Pile as a single line. Method required for automatic display in UI packages.")]
+        [Input("pile", "Pile to get the line geometry from.")]
+        [Output("curve", "The curve defining the Pile.")]
+        public static IGeometry Geometry(this Pile pile)
+        {
+            return pile.IsNull() ? null : pile.Location;
+        }
```
pawelbaran, line 68: This method should not be needed, taken there is a method for `IFramingElement`

Physical_Engine/Query/Geometry.cs
```diff
@@ -59,6 +59,39 @@ public static oM.Geometry.ISurface Geometry(this oM.Physical.Elements.ISurface s
...
+
+        /***************************************************/
+
+        [Description("Gets the geometry of a PadFoundation as a single curve. Method required for automatic display in UI packages.")]
```
pawelbaran, line 72: It is not a curve, rather a surface. I would copy the description from `ISurface` 😉

Physical_Engine/Query/Geometry.cs
```diff
@@ -59,6 +59,39 @@ public static oM.Geometry.ISurface Geometry(this oM.Physical.Elements.ISurface s
...
+        [Description("Gets the geometry of a PileFoundation as a surface representing the pile cap and curves representing the piles. Method required for automatic display in UI packages.")]
+        [Input("pileFoundation", "PileFoundation to get the geometry from.")]
+        [Output("geometry", "The geometry defining the PadFoundation.")]
+        public static IGeometry Geometry(this PileFoundation pileFoundation)
+        {
+            List<IGeometry> geometry = new List<IGeometry>();
+            geometry.Add(pileFoundation.PileCap.Geometry());
+            geometry.AddRange(pileFoundation.Piles.Select(x => x.Geometry()));
+
+            return Engine.Geometry.Create.CompositeGeometry(geometry);
```
pawelbaran, line 89: Not sure if really needed, but I would consider culling nulls from the collection, just in case

Physical_Engine/Query/HasMergeablePropertiesWith.cs
```diff
@@ -128,6 +128,29 @@ public static bool HasMergeablePropertiesWith(this IOpening element, IOpening ot
...
+        [Output("equal", "True if the PadFoundation non-geometrical property is equal to the point that they could be merged into one object.")]
+        public static bool HasMergeablePropertiesWith(this PadFoundation element, PadFoundation other)
+        {
+            if (element.IsNull() || other.IsNull())
+                return false; //If either is null, then it can probably can't have its properties merged
+
+            if (element.Construction == other.Construction)
+                return true;
+
+            if (element.Construction.IsNull() || other.Construction.IsNull())
```
pawelbaran, line 143: What if both constructions are null? Isn't it valid to merge properties then?

Physical_Engine/Query/IsWithinPileCap.cs
```diff
@@ -0,0 +1,112 @@
...
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Determines whether the Piles are within the extents defined by the PadFoundation (both depth and extents). Piles located on the boundary are accepted within tolerance" +
+            "if their profile does not intersect with the pad foundation boundary.")]
+        [Input("padFoundation", "The PadFoundation that defines the extents for the Piles to be located.")]
+        [Input("piles", "The Piles to check if they are located within the extents of the PadFoundation.")]
+        [Input("tolerance", "Distance tolerance to be used in the method. Piles are deemed to be on the edge of the PadFoundation if they are within this distance from the curve.", typeof(Length))]
+        [Output("withinPileCap", "True if all Piles are contained within the boundary and depth of the PileCap. False if one or more Piles are not.")]
+        public static bool IsWithinPileCap(this PadFoundation padFoundation, List<Pile> piles, double tolerance = Tolerance.MicroDistance)
```
pawelbaran, line 50: Why `Tolerance.MicroDistance` and not `Distance`?

Physical_Engine/Query/IsWithinPileCap.cs
```diff
@@ -0,0 +1,112 @@
...
+        {
+            if (padFoundation.IsNull())
+                return false;
+            else if (piles == null)
```
pawelbaran, line 54: As above, we could have an overload of `IsNull` method taking a collection of piles, then the next check (if any of the piles is null) could be called from within - wouldn't it be neater?

Physical_Engine/Query/IsWithinPileCap.cs
```diff
@@ -0,0 +1,112 @@
...
+            {
+                Base.Compute.RecordError("The list of Piles are null, the method will return false.");
+                return false;
+            }
+            else if (piles.Any(x => x.IsNull()))
+                return false;
+
+            // Get the top of the piles
+            List<Point> tops = piles.Select(p => p.Location.IControlPoints().OrderBy(pt => pt.Z).Reverse().First()).ToList();
```
pawelbaran, line 63:
```suggestion
            List<Point> tops = piles.Select(p => p.Location.IEndPoints().OrderBy(pt => pt.Z).Last()).ToList();
```

Physical_Engine/Query/IsWithinPileCap.cs
```diff
@@ -0,0 +1,112 @@
...
+
+            // Project all points to top surface of the pad
+            Plane padTop = new Plane() { Normal = padFoundation.Location.Normal(), Origin = padFoundation.Location.Centroid() };
```
pawelbaran, line 66: Should be computationally cheaper
```suggestion
            Plane padTop = padFoundation.Location.FitPlane() };
```

Physical_Engine/Query/IsWithinPileCap.cs
```diff
@@ -0,0 +1,112 @@
...
+
+            // Check if all piles are within the curve of the pad
+            if(topOutline.IIsContaining(pTops,true,tolerance))
+            {
+                // Check if the piles are within the depth of the pad
+                if(tops.Any(pt => pt.Z > maxZ || pt.Z < minZ))
+                {
+                    Base.Compute.RecordError("One or more the Piles tops is not located within the depth of the PileCap.");
+                    return false;
+                }
```
pawelbaran, line 83: I made an attempt to enable tilted pads plus avoid computationally heavy containment check if possible - see below. Also worth checking if construction/thickness is set to the cap before running the check, and raise a warning about skipping the check if not?

```suggestion
            double padThickness = padFoundation.Construction.IThickness();
            for (int i = 0; i < tops.Count; i++)
            {
                if (tops[i].Distance(pTops[i]) + tolerance > padThickness)
                {
                    Base.Compute.RecordError("One or more the Piles tops is not located within the depth of the PileCap.");
                    return false;
                }
            }
            
            // check containment after making sure the piles are within the thickness
            if(topOutline.IIsContaining(pTops,true,tolerance))
```

Physical_Engine/Query/IsWithinPileCap.cs
```diff
@@ -0,0 +1,112 @@
...
+                else
+                {
+                    // Check if the profile pushes the pile outside the curve of the pad
+                    for(int i = 0; i < pTops.Count; i++)
+                    {
+                        ConstantFramingProperty property = (ConstantFramingProperty)piles[i].Property;
+                        ICurve profile = property.Profile.Edges[0].ITranslate(Engine.Geometry.Create.Vector(pTops[i])).IRotate(pTops[i],Vector.ZAxis,property.OrientationAngle);
```
pawelbaran, line 90: Why do you only take one edge? Would also be good to check if the edges are actually assigned, and raise a warning about skipped check if not?

Physical_Engine/Query/IsWithinPileCap.cs
```diff
@@ -0,0 +1,112 @@
...
+                        if(profile.ICurveIntersections(topOutline).Count > 0)
```
pawelbaran, line 91: What if the pile intersects with the edge of a pile cap? Shouldn't that be allowed?

Physical_Engine/Query/OutlineElements1D.cs
```diff
@@ -84,6 +84,21 @@ public static List<IElement1D> OutlineElements1D(this IOpening opening)
...
+        [Input("padFoundation", "The PadFoundation to get the external outline elements from.")]
+        [Output("curves", "The curves defining the external boundery of the PadFoundation.")]
+        public static List<IElement1D> OutlineElements1D(this PadFoundation padFoundation)
+        {
+            if (padFoundation.IsNull())
+                return null;
+
+            PlanarSurface surface = padFoundation.Location;
+
+            return surface.ExternalBoundary.ISubParts().ToList<IElement1D>();
```
pawelbaran, line 97:
```suggestion
            return padFoundation.Location?.ExternalBoundary?.ISubParts()?.ToList<IElement1D>();
```

Physical_Engine/Query/SolidVolume.cs
```diff
@@ -136,7 +138,46 @@ public static double SolidVolume(this IOpening opening)
...
+
+        /***************************************************/
+
+        [Description("Returns the solid volume of a PadFoundation based on the thickness and area.")]
+        [Input("padFoundation", "The PadFoundation to get the volume from.")]
+        [Output("volume", "The PadFoundation solid material volume.", typeof(Volume))]
+        public static double SolidVolume(this PadFoundation padFoundation)
+        {
+            if (padFoundation.IsNull())
+                return double.NaN;
```
pawelbaran, line 150: Seeing double.NaN being returned here I am wondering if not revisiting `EmbedVolume` method whether it shouldn't return `NaN` in any scenario?

Physical_Engine/Query/VolumetricMaterialTakeoff.cs
```diff
@@ -149,6 +149,45 @@ public static VolumetricMaterialTakeoff VolumetricMaterialTakeoff(this Door open
...
+
+        [Description("Gets the volumetric material takeoff from a PadFoundation.")]
+        [Input("padFoundation", "The physical PadFoundation object to extract the volumetric material takeoff from.")]
+        [Output("volTakeoff", "The volumetric material takeoff for the PadFoundation, made of up the volume and materiality of the construction and surface area.")]
+        public static VolumetricMaterialTakeoff VolumetricMaterialTakeoff(this PadFoundation padFoundation)
+        {
+            if (padFoundation.IsNull())
+                return null;
+
+            return TakeOff(padFoundation.Location.IArea(), padFoundation.Construction);
```
pawelbaran, line 162:
```suggestion
            return TakeOff(padFoundation.Location.IArea(), padFoundation.Construction);
```

pawelbaran (changes requested):

Physical_Engine/Modify/SetOutlineElements1D.cs
```diff
@@ -81,6 +81,22 @@ public static IOpening SetOutlineElements1D(this IOpening opening, List<IElement
...
+        [Input("outlineElements1D", "One dimensional elements defining the external boundery geometry, PadFoundation has no edge properties and all IElements1D will be treated as their geometry. They must be closed and planar.")]
+        [Output("padFoundation", "The PadFoundation with new location surface.")]
+        public static PadFoundation SetOutlineElements1D(this PadFoundation padFoundation, List<IElement1D> outlineElements1D)
+        {
+            if (padFoundation.IsNull())
+                return null;
+
+            PadFoundation clone = padFoundation.ShallowClone();
+            ICurve outline = Geometry.Compute.IJoin(outlineElements1D.Select(x => x.IGeometry()).ToList()).Single();
+            clone.Location = Geometry.Create.PlanarSurface(outline);
```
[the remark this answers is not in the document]
peterjamesnugent: This is for a `PadFoundation`, so there are no `Pile`s. 

The `PileFoundation` is neither an `IElement1D` or `IElement2D` - issue raised here for composite IElement interfaces: https://github.com/BHoM/BHoM/issues/1554

Physical_Engine/Query/EmbedVolume.cs
peterjamesnugent: I have replaced with the `Engine.Base.IsNullOrEmpty()` method which takes a `List<T>`.

Physical_Engine/Query/EmbedVolume.cs
peterjamesnugent: I can't find an `IEndPoints()` method?

Physical_Engine/Query/Geometry.cs
peterjamesnugent: I even used the `IFramingElement` method on the test script 🤦‍♂️

Physical_Engine/Query/Geometry.cs
peterjamesnugent: Is it necessary? The `Create` for `PileFoundation` uses the `WithinPileCap()` which calls `IsNull()`.

Physical_Engine/Query/HasMergeablePropertiesWith.cs
peterjamesnugent: Is that not covered by L140?

Physical_Engine/Query/IsWithinPileCap.cs
peterjamesnugent: I can't find an `IEndPoints()` method?

Physical_Engine/Query/HasMergeablePropertiesWith.cs
pawelbaran: of course it is, me blind

Physical_Engine/Query/IsWithinPileCap.cs
pawelbaran: Ha, because we do not have it 😃 I was 100% convinced we do. Shouldn't we add it? Could be useful for the future + you could save a few ticks. But up to you, happy with what we have if you do not want to inflate this PR any further. Just please consider the 2nd part of the suggestion (1st vs last - alternatively `OrderByDescending`)

Physical_Engine/Modify/SetOutlineElements1D.cs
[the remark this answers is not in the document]
pawelbaran: Ahh right - apologies, I did my best not to confuse `PileFoundation` and `PadFoundation` at any point, did not succeed...

Physical_Engine/Query/IsWithinPileCap.cs
FraserGreenroyd: I would argue/agree that adding `IEndPoints` should be a separate PR/issue so as not to scope creep this one too badly unless it's desperately needed by this PR 😄

Physical_Engine/Query/IsWithinPileCap.cs
peterjamesnugent: We only need the outer edge (which is the first one) after some limited testing.

Physical_Engine/Query/IsWithinPileCap.cs
peterjamesnugent: Unless the profile of the pile is infinitesimally small, it is not valid. There is a check just below this that takes in to account the profile to see if it pushes it outside the extents of the pile cap.

Physical_Engine/Query/SolidVolume.cs
peterjamesnugent: My preference would be to return zero and raise a warning - I was just replicating what had been done with the other methods.

Physical_Engine/Query/VolumetricMaterialTakeoff.cs
peterjamesnugent: Was this comment complete? Looks the same to me.

Physical_Engine/Query/SolidVolume.cs
FraserGreenroyd: The problem with returning 0 @peterjamesnugent is that a user might use the method, get 0 because of an error, but that 0 can be legitimately used in downstream calculations. Even with the raising of a warning/error, users might not check it, or even be able to begin working out which pile from a hundred is causing the error, and the downstream calculation is calculating so no harm no foul right? That's how users think at least, so it's best to return something that cannot be an acceptable output from the method if something isn't working imo.

Physical_Engine/Query/IsWithinPileCap.cs
pawelbaran: Sorry let me confirm visually - is the pile (red) on cap (blue) valid?
image

Physical_Engine/Query/IsWithinPileCap.cs
pawelbaran: I am afraid that would only be valid for round piles:
image

Physical_Engine/Query/VolumetricMaterialTakeoff.cs
pawelbaran: I do not recall attempting to make any comment on these lines - looks like a missclick or some other glitch on my end to me 🤦‍♂️ Apologies

Physical_Engine/Query/Geometry.cs
peterjamesnugent: Fair point, I have added null checks in.

Physical_Engine/Query/IsWithinPileCap.cs
peterjamesnugent: Cool, raised here: https://github.com/BHoM/BHoM_Engine/issues/3207

Physical_Engine/Query/IsWithinPileCap.cs
peterjamesnugent: Good point, I have added a `IJoin()` and then sorted by length to get the outer curve.

Physical_Engine/Query/IsWithinPileCap.cs
peterjamesnugent: As it stands no, I can't think of a good way to handle that as it varies for the different section types. 

We could justify not allowing it as you generally would offset the pile from the edge.

Physical_Engine/Query/SolidVolume.cs
peterjamesnugent: Cool, happy with @FraserGreenroyd's response and will leave the code as is.

Physical_Engine/Query/IsWithinPileCap.cs
pawelbaran: Sounds good to me, no strong opinion here, just checking if the current solution is intentional

Physical_Engine/Query/IsWithinPileCap.cs
peterjamesnugent: Have rearranged to include your suggestion - I don't think we can get away from including the `IsContaining()`.

peterjamesnugent:
@BHoMBot check project-compliance

peterjamesnugent:

Physical_Engine/Query/Geometry.cs
```diff
@@ -59,6 +62,41 @@ public static oM.Geometry.ISurface Geometry(this oM.Physical.Elements.ISurface s
...
+        }
+
+        [Description("Gets the geometry of a PileFoundation as a surface representing the pile cap and curves representing the piles. Method required for automatic display in UI packages.")]
+        [Input("pileFoundation", "PileFoundation to get the geometry from.")]
+        [Output("geometry", "The geometry defining the PadFoundation.")]
+        public static IGeometry Geometry(this PileFoundation pileFoundation)
+        {
+            List<IGeometry> geometry = new List<IGeometry>();
+            IGeometry pileCapGeometry = pileFoundation.PileCap.Geometry();
+            if(!pileCapGeometry.IsNull())
```
pawelbaran, line 80: I would rather simple check `== null` here - otherwise you will get an error plus a warning in case of a null geometry

Physical_Engine/Query/Geometry.cs
```diff
@@ -59,6 +62,41 @@ public static oM.Geometry.ISurface Geometry(this oM.Physical.Elements.ISurface s
...
+                geometry.Add(pileCapGeometry);
+            else
+                Base.Compute.RecordWarning("Null geometry for the pile cap was excluded.");
+
+            List<ICurve> pileGeometry = pileFoundation.Piles.Select(x => x.Geometry()).ToList();
+            foreach(ICurve curve in pileGeometry)
+            {
+                if (!curve.IsNull())
```
pawelbaran, line 88: Same as above

Physical_Engine/Query/IsWithinPileCap.cs
```diff
@@ -0,0 +1,117 @@
...
+        {
+            if (padFoundation.IsNull())
+                return false;
+            else if (piles.IsNullOrEmpty())
+                return false;
+            else if (piles.Any(x => x.IsNull()))
+                return false;
+            else if (padFoundation.Construction.IThickness() == 0)
+            {
+                Base.Compute.RecordError("The PadFoundation does not contain a thickness, the method will return false.");
```
pawelbaran, line 61: This error sounds weird in the UI:
image

Maybe we could remove the part about returning false?

Physical_Engine/Query/IsWithinPileCap.cs
```diff
@@ -0,0 +1,117 @@
...
+            ICurve topOutline = padFoundation.Location.ExternalBoundary;
+
+            // Check if all piles are within the curve of the pad
+            if (topOutline.IIsContaining(pTops, true, tolerance))
+            {
+                    // Check if the profile pushes the pile outside the curve of the pad
+                    for (int i = 0; i < pTops.Count; i++)
+                    {
+                        ConstantFramingProperty property = (ConstantFramingProperty)piles[i].Property;
+                        ICurve profile = Engine.Geometry.Compute.IJoin(property.Profile.Edges.ToList()).OrderBy(c => c.Length()).Last()
```
pawelbaran, line 93: Do we need to join and take the largest outline? Maybe we could check each segment separately (this is what is being done under the hood for `PolyCurve` anyways)

Also worth checking if the profile is actually assigned - I have just started creating a test script and got a null reference exception that I needed to debug to find out why did it crash

Physical_Engine/Query/IsWithinPileCap.cs
```diff
@@ -0,0 +1,117 @@
...
+                            .ITranslate(Engine.Geometry.Create.Vector(pTops[i])).IRotate(pTops[i], Vector.ZAxis, property.OrientationAngle);
```
pawelbaran, line 94: As discussed offline with @peterjamesnugent, we should not be doing translations and rotations in here, should be a centralised returning `TransformMatrix`, same as in structures:
https://github.com/BHoM/BHoM_Engine/blob/474c6194f5107f73a0dc0273c06ce8bb47f67744/Structure_Engine/Query/BarSectionTranformation.cs#L43

As discussed, we will rather keep the code as is and sort it out in a separate issue/PR thread, as long as @peterjamesnugent does not change his mind 😉

Physical_Engine/Query/Geometry.cs
```diff
@@ -59,6 +62,41 @@ public static oM.Geometry.ISurface Geometry(this oM.Physical.Elements.ISurface s
 
         /***************************************************/
 
+        [Description("Gets the geometry of a PadFoundation as a surface. Method required for automatic display in UI packages.")]
+        [Input("padFoundation", "PadFoundation to get the surface geometry from.")]
+        [Output("surface", "The surface defining the PadFoundation.")]
+        public static IGeometry Geometry(this PadFoundation padFoundation)
+        {
+            return padFoundation.IsNull() ? null : padFoundation.Location;
```
pawelbaran, line 70: Same as below, it is in general better to simply check if the value is `null` in `Geometry` queries rather than calling `IsNull` to avoid irrelevant errors.
```suggestion
            return padFoundation?.Location;
```

pawelbaran (changes requested):
Final comments following the discussion with @peterjamesnugent

Physical_Engine/Query/IsWithinPileCap.cs
peterjamesnugent: https://github.com/BHoM/BHoM_Engine/issues/3209

FraserGreenroyd:
@BHoMBot check core

Physical_Engine/Query/IsWithinPileCap.cs
```diff
@@ -0,0 +1,129 @@
...
+                    foreach (ICurve curve in profile)
+                    {
+                        ICurve transformedCurve;
+
+                        // because the profile will be in the XY plane
+                        transformedCurve = curve.ITranslate(Engine.Geometry.Create.Vector(pTops[i])).IRotate(pTops[i], Vector.ZAxis, property.OrientationAngle + Math.PI / 2);
+
+                        if (transformedCurve.ICurveIntersections(topOutline).Count > 0)
+                        {
+                            Base.Compute.RecordError("One or more the Pile profiles is located outside the edge of the pile cap.");
```
pawelbaran, line 110:
```suggestion
                            Base.Compute.RecordError("One or more of the Pile profiles is located outside the edge of the pile cap.");
```

Physical_Engine/Query/IsWithinPileCap.cs
```diff
@@ -0,0 +1,129 @@
...
+            List<Point> pTops = tops.Select(x => x.Project(padTop)).ToList(); //project points only used for IsContaining
+
+
+            // Get the thickness of the PadFoundation to compare against the distance coordinates of the pile tops
+            double padThickness = padFoundation.Construction.IThickness();
+            for (int i = 0; i < tops.Count; i++)
+            {
+                if (tops[i].Distance(pTops[i]) + tolerance > padThickness)
+                {
+                    Base.Compute.RecordError("One or more the Piles tops is not located within the depth of the PileCap.");
```
pawelbaran, line 80:
```suggestion
                    Base.Compute.RecordError("One or more of the Piles tops is not located within the depth of the PileCap.");
```

Physical_Engine/Query/IsWithinPileCap.cs
```diff
@@ -0,0 +1,129 @@
...
+                        {
+                            Base.Compute.RecordError("One or more the Pile profiles is located outside the edge of the pile cap.");
+                            return false;
+                        }
+                    }
+                }
+            }
+            else
+            {
+                Base.Compute.RecordError("One or more the Piles centre points is located outside the extents of the PileCap.");
```
pawelbaran, line 118:
```suggestion
                Base.Compute.RecordError("One or more of the Piles centre points is located outside the extents of the PileCap.");
```

Physical_Engine/Query/IsWithinPileCap.cs
```diff
@@ -0,0 +1,129 @@
...
+        {
+            if (padFoundation.IsNull())
+                return false;
+            else if (piles.IsNullOrEmpty())
+                return false;
+            else if (piles.Any(x => x.IsNull()))
+                return false;
+            else if (padFoundation.Construction.IThickness() == 0)
+            {
+                Base.Compute.RecordError("The PadFoundation is not assigned a thickness (within the Construction), the method will return false as it cannot determine the thickness" +
```
pawelbaran, line 61: As discussed the other day, I would try to avoid 'false' word
```suggestion
                Base.Compute.RecordError("The PadFoundation does not have thickness assigned (within the Construction), therefore piles are considered as outside of the pile cap." +
```

pawelbaran (changes requested):
Minor tweaks suggested to the error messages, besides all good 👍 Exciting! And happy/relieved we made it 😉 @peterjamesnugent could you please approve https://github.com/BHoM/BHoM/pull/1559, so that I could approve here and we get both PRs merged?

peterjamesnugent:
@BHoMBot check required

peterjamesnugent:
@BHoMBot check required

peterjamesnugent:
@BHoMBot check versioning check installer

peterjamesnugent:
@BHoMBot check unit-tests

FraserGreenroyd:

pawelbaran (approved):
Happy to approve 👍 a bit of a journey with this PR, but I am really happy where we landed together @peterjamesnugent. Good stuff!

pawelbaran:
@BHoMBot check ready-to-merge

Issue #3199, opened 2023-10-25
Environment_Engine: Refactor ElementInSpace and Space IsContaining methods from Space to IRegion and move to Spatial_Engine


enarhi:
#### Description:
As per the title, the methods for IsContaining and element in space checks should be refactored and moved to the Spatial_Engine namespace for wider usage beyond Environment_Engine.

PR #3203, opened 2023-10-31
Environment_Engine: Refactor of IsContaining and ElementsInSpace methods to Geometry and Analytical namespaces as appropriate

### LINKED ISSUE #3198: Environment_Engine: Refactor IsContaining method to take geometry and move from Environment_Engine to Geometry_Engine
#### Description:
The current core IsContaining check relies on Environment panels, but should be refactored to take in List<ICurve> rather than List<Panel> and then moved to the Spatial_Engine.

enarhi:

Closes #3198 and #3199 

Methods for IsContaining were added to Geometry_Engine, and the ElementsInSpace method was moved to analytical. Both now work off polylines and regions respectively to allow for wider usage outside of the Environment_Engine.

### Test files
[Test Script](https://burohappold.sharepoint.com/:f:/r/sites/BHoM/02_Current/12_Scripts/02_Pull%20Request/BHoM/BHoM_Engine/Environment_Engine/%233199-IsContaining_Refactor?csf=1&web=1&e=o1EJ7L)

### DIFF
--- a/Analytical_Engine/Compute/ExtrudeToVolume.cs
+++ b/Analytical_Engine/Compute/ExtrudeToVolume.cs
@@ -1,0 +1,104 @@
+/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2023, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 
+ * GNU Lesser General Public License for more details.                          
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+using BH.Engine.Geometry;
+using BH.oM.Geometry;
+
+using BH.oM.Base.Attributes;
+using System.ComponentModel;
+
+using BH.oM.Analytical.Elements;
+
+namespace BH.Engine.Analytical
+{
+    public static partial class Compute
+    {
+        [Description("Takes a region with a Floor Perimeter and creates a collection of polylines which represent the closed volume of the region.")]
+        [Input("region", "A region with a floor perimeter to extrude into a collection of panels.")]
+        [Input("height", "The height of the region, as a double, to calculate the ceiling level of the region. This will be used as the Z value of the perimeter + the given height.")]
+        [Input("tolerance", "The degree of tolerance on the angle calculation for collapsing the regions perimeter to a polyline. Default is equal to BH.oM.Geometry.Tolerance.Angle.")]
+        [Output("polylines", "A collection of polylines which represent the closed volume of the region.")]
+        [PreviousInputNames("region", "room, space")]
+        public static List<Polyline> ExtrudeToVolume(this IRegion region, double height, double tolerance = BH.oM.Geometry.Tolerance.Angle)
+        {
+            if (region == null)
+                return new List<Polyline>();
+
+            Polyline floor = region.Perimeter.ICollapseToPolyline(tolerance);
+            return floor.ExtrudeToVolume(region.Name, height);
+        }
+
+        [Description("Takes a polyline perimeter and creates a collection of polylines which represent the closed volume of a space.")]
+        [Input("pLine", "A polyline perimeter to extrude into a collection of panels.")]
+        [Input("connectingSpaceName", "The name of the space the panels will enclose.")]
+        [Input("height", "The height of the space, as a double, to calculate the ceiling level of the room. This will be used as the Z value of the perimeter + the given height.")]
+        [Output("polylines", "A collection of polylines which represent the closed volume of the region.")]
+        public static List<Polyline> ExtrudeToVolume(this Polyline pLine, string connectingSpaceName, double height)
+        {
+            if (pLine == null)
+                return new List<Polyline>();
+
+            List<Point> floorPoints = pLine.ControlPoints;
+            List<Point> checkPoints = floorPoints.CullDuplicates();
+
+            if (floorPoints.Count != (checkPoints.Count + 1))
+            {
+                BH.Engine.Base.Compute.RecordError("The polyline has duplicate control points and cannot be extruded to a volume.");
+                return new List<Polyline>();
+            }                     
+
+            List<Polyline> panels = new List<Polyline>();
+
+            Polyline floorPanel = pLine;
+            panels.Add(floorPanel);
+            
+            List<Point> roofPoints = new List<Point>();
+            foreach (Point p in floorPoints)
+                roofPoints.Add(new Point { X = p.X, Y = p.Y, Z = p.Z + height });
+
+            Polyline roofPanel = new Polyline { ControlPoints = roofPoints };
+            panels.Add(roofPanel);
+
+            for (int a = 0; a < floorPoints.Count - 1; a++)
+            {
+                List<Point> panelPoints = new List<Point>();
+                panelPoints.Add(floorPoints[a]);
+                panelPoints.Add(new Point { X = floorPoints[a].X, Y = floorPoints[a].Y, Z = floorPoints[a].Z + height });
+                panelPoints.Add(new Point { X = floorPoints[a + 1].X, Y = floorPoints[a + 1].Y, Z = floorPoints[a + 1].Z + height });
+                panelPoints.Add(floorPoints[a + 1]);
+                panelPoints.Add(floorPoints[a]);
+
+                panels.Add(new Polyline {ControlPoints = panelPoints });
+            }
+
+            return panels;
+        }
+    }
+}
+
+
--- a/Analytical_Engine/Query/IsContaining.cs
+++ b/Analytical_Engine/Query/IsContaining.cs
@@ -31,6 +31,8 @@ using BH.oM.Geometry;
 using BH.Engine.Geometry;
 using BH.oM.Base;
 using BH.oM.Base.Attributes;
+using BH.oM.Dimensional;
+using BH.Engine.Spatial;
 using System.ComponentModel;
 
 namespace BH.Engine.Analytical
@@ -80,6 +82,40 @@ namespace BH.Engine.Analytical
                 Item1 = results,
                 Item2 = notContained,
             };
+        }
+
+        [Description("Defines whether an Environment Space contains each of a provided list of points.")]
+        [Input("region", "An Environment Space object defining a perimeter to build a 3D volume from and check if the volume contains the provided point.")]
+        [Input("regionHeight", "The height of the region.", typeof(BH.oM.Quantities.Attributes.Length))]
+        [Input("points", "The points being checked to see if it is contained within the bounds of the 3D volume.")]
+        [Input("acceptOnEdges", "Decide whether to allow the point to sit on the edge of the region, default false.")]
+        [Output("isContaining", "True if the point is contained within the region, false if it is not.")]
+        public static List<bool> IsContaining(this IRegion region, double regionHeight, List<Point> points, bool acceptOnEdges = false)
+        {
+            List<Polyline> panelsFromSpace = region.ExtrudeToVolume(regionHeight);
+            return panelsFromSpace.IsContaining(points, acceptOnEdges);
+        }
+
+        [Description("Defines whether an Environment Space contains a provided Element.")]
+        [Input("region", "An Environment Space object defining a perimeter to build a 3D volume from and check if the volume contains the provided element.")]
+        [Input("regionHeight", "The height of the region.", typeof(BH.oM.Quantities.Attributes.Length))]
+        [Input("elements", "The elements being checked to see if they are contained within the bounds of the 3D volume.")]
+        [Input("acceptOnEdges", "Decide whether to allow the element's point to sit on the edge of the region, default false.")]
+        [Input("acceptPartialContainment", "Decide whether to count elements overlapping the region as contained.")]
+        [Output("isContaining", "True if the point is contained within the region, false if it is not.")]
+        public static List<bool> IsContaining(this IRegion region, double regionHeight, List<IElement> elements, bool acceptOnEdges = false, bool acceptPartialContainment = false)
+        {
+            List<Polyline> polylines = region.ExtrudeToVolume(regionHeight);
+
+            List<List<Point>> pointLists = new List<List<Point>>();
+
+            foreach (IElement elem in elements)
+            {
+                List<Point> points = elem.IControlPoints();
+                pointLists.Add(points);
+            }
+
+            return BH.Engine.Geometry.Query.IsContaining(polylines, pointLists, acceptOnEdges, acceptPartialContainment);
         }
     }
 }
--- a/Analytical_Engine/Query/IsContaining.cs
+++ b/Analytical_Engine/Query/IsContaining.cs
@@ -93,7 +93,7 @@ namespace BH.Engine.Analytical
         public static List<bool> IsContaining(this IRegion region, double regionHeight, List<Point> points, bool acceptOnEdges = false)
         {
             List<Polyline> panelsFromSpace = region.ExtrudeToVolume(regionHeight);
-            return panelsFromSpace.IsContaining(points, acceptOnEdges);
+            return panelsFromSpace.IsContaining(points, acceptOnEdges, Tolerance.Distance);
         }
 
         [Description("Defines whether an Environment Space contains a provided Element.")]
--- a/Geometry_Engine/Query/IsContaining.cs
+++ b/Geometry_Engine/Query/IsContaining.cs
@@ -829,12 +829,6 @@ namespace BH.Engine.Geometry
                 return false;
             }
 
-            if (planes.Any(x => x == null))
-            {
-                BH.Engine.Base.Compute.RecordError("Cannot query if a provided plane is null.");
-                return false;
-            }
-
             if (!BH.Engine.Geometry.Query.IsContaining(boundingBox, point, true, tolerance))
                 return false;
 
--- a/Geometry_Engine/Query/IsContaining.cs
+++ b/Geometry_Engine/Query/IsContaining.cs
@@ -809,6 +809,34 @@ namespace BH.Engine.Geometry
         }
 
         /***************************************************/
+        /**** Public Methods - Interfaces               ****/
+        /***************************************************/
+
+        [Description("Checks if the collection of Points are all contained within the curve. If a single Point is outside the curve, the method will return false. Points not in the plane of the curve are deemed to be outside.")]
+        [Input("curve", "The PolyCurve to check if it is containing all of the provided points. If the PolyCurve is not closed or planar, the method will return false.")]
+        [Input("points", "The points to check if they are all contained within the curve. If a single point is outside the curve or not in the plane of the curve the method will return false.")]
+        [Input("acceptOnEdge", "If true and the curve is closed, points that are within the tolerance distance away from the curve are demmed to be inside it. If false, only points that are inside and not within tolerance distance away from the curve are deemed to be inside.")]
+        [Input("tolerance", "Distance tolerance to be used in the method. Points are deemed to be on the edge of the curve if they are within this distance from the curve.", typeof(Length))]
+        [Output("isContaining", "Returns true if all of the provided points are inside the curve.")]
+        public static bool IIsContaining(this ICurve curve, List<Point> points, bool acceptOnEdge = true, double tolerance = Tolerance.Distance)
+        {
+            return IsContaining(curve as dynamic, points, acceptOnEdge, tolerance);
+        }
+
+        /***************************************************/
+
+        [Description("Checks if a curve is contained within another the curve. The curves need to be co-planar for the method to be able to return true.")]
+        [Input("curve1", "The curve to check if it is containing the second curve. Needs to be closed and coplanar with the second curve.")]
+        [Input("curve2", "The curve to check if it is contained within the the first curve. Needs to be coplanar with the first curve.")]
+        [Input("acceptOnEdge", "If true, a the inner curve is allowed to touch the outer curve within tolerance. If false, all points of the second curve needs to be fully inside the first curve and are not allowed to be within tolerance distance from the first curve.")]
+        [Input("tolerance", "Distance tolerance to be used in the method. Points on the second are deemed to be on the edge of the curve if they are within this distance from the curve.", typeof(Length))]
+        [Output("isContaining", "Returns true if the second curve is inside the first curve.")]
+        public static bool IIsContaining(this ICurve curve1, ICurve curve2, bool acceptOnEdge = true, double tolerance = Tolerance.Distance)
+        {
+            return IsContaining(curve1 as dynamic, curve2 as dynamic, acceptOnEdge, tolerance);
+        }
+
+        /***************************************************/
         /**** Private Methods - Closed Volume Polylines ****/
         /***************************************************/
 
@@ -836,7 +864,7 @@ namespace BH.Engine.Geometry
 
             if (uniqueEdges.Any(p => point.IsOnCurve(p, tolerance)))
                 return acceptOnEdges;
-            
+
             double rayLength = (boundingBox.Max - boundingBox.Min).Length() + 1;
             Vector rVec;
             Line line = null;
@@ -868,11 +896,11 @@ namespace BH.Engine.Geometry
 
                 Polyline pLine = closedVolume[x];
 
-                if  (BH.Engine.Geometry.Query.IsContaining(boundingBox, intPt, true, tolerance) && BH.Engine.Geometry.Query.IsContaining(pLine, new List<Point> { intPt }, false, tolerance))
+                if (BH.Engine.Geometry.Query.IsContaining(boundingBox, intPt, true, tolerance) && BH.Engine.Geometry.Query.IsContaining(pLine, new List<Point> { intPt }, false, tolerance))
                     intersectPoints.Add(intPt);
             }
 
-            bool isContained = intersectPoints.CullDuplicates().Count % 2 != 0;
+            bool isContained = intersectPoints.CullDuplicates(tolerance).Count % 2 != 0;
 
             return isContained; //If the number of intersections is odd the point is outsde the space
         }
@@ -883,7 +911,7 @@ namespace BH.Engine.Geometry
             List<Line> lines = new List<Line>();
             foreach (Line line in outlines.SelectMany(x => x.SubParts()))
             {
-                if (lines.All(x => !x.IsSimilar(line, tolerance)))
+                if (lines.All(x => !x.IsSimilar(line, sqTol)))
                     lines.Add(line);
             }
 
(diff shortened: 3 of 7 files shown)


Geometry_Engine/Query/IsContaining.cs
```diff
@@ -709,6 +709,182 @@ public static bool IsContaining(this Cuboid cuboid, IGeometry geometry, bool acc
...
+        /**** Public Methods - Closed Volume Polylines  ****/
+        /***************************************************/
+
+        [Description("Checks if a point lies within a collection of polylines describing a closed volume using their primitive planes and bounds.")]
+        [Input("panels", "A collection of Polylines outlining each of the faces of the closed volume to check with.")]
+        [Input("point", "The point to check to see if it is contained within the bounds of the panels.")]
+        [Input("acceptOnEdges", "Decide whether to allow the point to sit on the edge of the panel, default false.")]
+        [Input("tolerance", "Distance tolerance to use to determine intersections.")]
+        [Output("isContaining", "True if the point is contained within the bounds of the panels, false if it is not for each point provided.")]
+        public static bool IsContaining(this List<Polyline> panels, Point point, bool acceptOnEdges = false, double tolerance = BH.oM.Geometry.Tolerance.Distance)
```
FraserGreenroyd, line 722: Calling this input `panels` might be confusing to standard geometry users. I'm wondering if we call it something like `closedVolume` instead to indicate it's a collection of polylines representing a closed volume (so helping with the description)?

Structure_Engine/Modify/SetAnalysisParameters.cs
```diff
@@ -82,7 +82,7 @@ public static void SetAnalysisParameters(this SawnTimber timber, TimberAnalysisS
                 timber.YoungsModulus = new Vector { X = timber.E_0_k, Y = timber.E_90_Mean, Z = timber.E_90_Mean };
                 timber.ShearModulus = new Vector { X = timber.G_Mean, Y = timber.G_Mean, Z = timber.G_Mean };
 
-                message = $"{nameof(SawnTimber)} materials does not contain characteristic values for youngs modulus perpendicular. Mean value of perpendicular stiffness used." + Environment.NewLine;
+                message = $"{nameof(SawnTimber)} materials does not contain characteristic values for youngs modulus perpendicular. Mean value of perpendicular stiffness used." + System.Environment.NewLine;
```
FraserGreenroyd, line 85: This suggests the Environment namespace has somehow started to bleed into Structures if the compiler was complaining about this? If so we should investigate and prevent

FraserGreenroyd (changes requested):
Minor comments. Somehow I think this will escape needing versioning though which would be cool if so - but let's maybe run the check to be sure.

FraserGreenroyd:
@BHoMBot check compliance

FraserGreenroyd (changes requested):
Let's fix compliance before checking versioning though 😉 

Otherwise from code review this all looks good and in line with what we discussed @enarhi so should be quick enough to settle out.

Geometry_Engine/Query/IsContaining.cs
pawelbaran: Agreed with @FraserGreenroyd - I would actually consider going a step further and suggest using `BoundaryRepresentation` as input, indicating we are speaking of a closed volume here. However, this would only make sense if converting the outlines to a Brep does not introduce any friction into existing code ofc.

Geometry_Engine/Query/IsContaining.cs
```diff
@@ -709,6 +709,182 @@ public static bool IsContaining(this Cuboid cuboid, IGeometry geometry, bool acc
...
+            if (!BH.Engine.Geometry.Query.IsContaining(boundingBox, point, true, tolerance))
+                return false;
+
+            //We need to check one line that starts in the point and end outside the bounding box
+            Vector vector = new Vector() { X = 1, Y = 0, Z = 0 };
+            //Use a length longer than the longest side in the bounding box. Change to infinite line?
+            Line line = new Line() { Start = point, End = point.Translate(vector * (((boundingBox.Max - boundingBox.Min).Length()) * 10)) };
+
+            bool isInPlane = false;
+            do
```
pawelbaran, line 836: What if there point lies in planes along `(1,0,0)` and `(0.5,0.5,0.5)` at the same time? Won't this loop go forever?

Actually, can't you simply take an arbitrary vector that has absolute value of dot product with each of plane normal larger than tolerance?

Geometry_Engine/Query/IsContaining.cs
```diff
@@ -709,6 +709,182 @@ public static bool IsContaining(this Cuboid cuboid, IGeometry geometry, bool acc
...
+                }
+            }
+            while (isInPlane);
+
+            //Check intersections
+            List<Point> intersectPoints = new List<Point>();
+
+            for (int x = 0; x < planes.Count; x++)
+            {
+                if (planes[x] != null)
```
pawelbaran, line 855: Should we allow for `null` planes at all? To me getting that far with a null in our inputs would be hm, suspicious 😉

Geometry_Engine/Query/IsContaining.cs
```diff
@@ -709,6 +709,182 @@ public static bool IsContaining(this Cuboid cuboid, IGeometry geometry, bool acc
...
+                {
+                    if ((BH.Engine.Geometry.Query.PlaneIntersection(line, planes[x], false, tolerance)) == null)
+                        continue;
+
+                    List<Point> intersectingPoints = new List<oM.Geometry.Point>();
+                    intersectingPoints.Add(BH.Engine.Geometry.Query.PlaneIntersection(line, planes[x], true, tolerance));
```
pawelbaran, line 861: Why do you first intersect with a finite line and then with infinite? Shouldn't the result be the same as we already ruled out `null` intersections 3 lines above?

Geometry_Engine/Query/IsContaining.cs
```diff
@@ -709,6 +709,182 @@ public static bool IsContaining(this Cuboid cuboid, IGeometry geometry, bool acc
...
+                    Polyline pLine = panels[x];
+
+                    if (intersectingPoints != null && BH.Engine.Geometry.Query.IsContaining(pLine, intersectingPoints, true, tolerance))
```
pawelbaran, line 864: What if the ray will hit a corner between two surfaces? Won't it return a false negative? Wouldn't `acceptOnEdges = false` in this call to `IsContaining` help against it?

Also `intersectingPoints` cannot be `null` here

One more thought: in order to optimise I would first check if the point fits into the bbox first before calling computationally heavy `IsContaining`

Geometry_Engine/Query/IsContaining.cs
```diff
@@ -709,6 +709,182 @@ public static bool IsContaining(this Cuboid cuboid, IGeometry geometry, bool acc
...
+                        intersectPoints.AddRange(intersectingPoints);
+                }
+            }
+
+            bool isContained = !((intersectPoints.CullDuplicates().Count % 2) == 0);
```
pawelbaran, line 869: I would rather type it as `x != 0` instead of `!(x == 0)` - more readable

Geometry_Engine/Query/IsContaining.cs
```diff
@@ -709,6 +709,182 @@ public static bool IsContaining(this Cuboid cuboid, IGeometry geometry, bool acc
...
+
+            if (!isContained && acceptOnEdges)
+            {
+                //Check the edges in case the point is on the edge of the BE
+                foreach (Polyline p in panels)
+                {
```
pawelbaran, line 875: I would rather see a call `IsOnCurve(Polyline, Point)` here, and if it does not exist then I would add such method 😉

pawelbaran (changes requested):
I had a read through the code and raised a few questions as in the comments - hopefully they make sense, please shout if you prefer to grab a call instead of typing @enarhi 👍

Structure_Engine/Modify/SetAnalysisParameters.cs
enarhi: It does, but I don't understand how?

Geometry_Engine/Query/IsContaining.cs
```diff
@@ -709,6 +709,182 @@ public static bool IsContaining(this Cuboid cuboid, IGeometry geometry, bool acc
...
+            {
+                BH.Engine.Base.Compute.RecordError("Cannot query if a collection of panels contains a point if the panels are null.");
+                return false;
+            }
+
+            List<Plane> planes = new List<Plane>();
+            foreach (Polyline be in panels)
+                planes.Add(be.IControlPoints().FitPlane(tolerance));
+
+            List<Point> ctrPoints = panels.SelectMany(x => x.IControlPoints()).ToList();
```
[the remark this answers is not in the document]
enarhi: Resolved, redundant call removed.

Geometry_Engine/Query/IsContaining.cs
```diff
@@ -709,6 +709,182 @@ public static bool IsContaining(this Cuboid cuboid, IGeometry geometry, bool acc
...
+            BoundingBox boundingBox = BH.Engine.Geometry.Query.Bounds(ctrPoints);
+
+            return isContained = IsContaining(panels, planes, boundingBox, point, acceptOnEdges);
```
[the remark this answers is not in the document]
enarhi: No points for guessing which method isn't covered by the test script...fixed.

Geometry_Engine/Query/IsContaining.cs
enarhi: Renamed, leaving as closedVolume for now, a followup issue/pr can be created if we want to move from current handling to a proper brep based method for closed volumes.

Geometry_Engine/Query/IsContaining.cs
```diff
@@ -709,6 +709,182 @@ public static bool IsContaining(this Cuboid cuboid, IGeometry geometry, bool acc
...
+            {
+                BH.Engine.Base.Compute.RecordError("Cannot query if a collection of panels contains a point if the panels are null.");
+                return new List<bool>() { false };
+            }
+
+            List<Plane> planes = new List<Plane>();
+            foreach (Polyline be in panels)
+                planes.Add(be.IControlPoints().FitPlane(tolerance));
+
+            List<Point> ctrPoints = panels.SelectMany(x => x.IControlPoints()).ToList();
```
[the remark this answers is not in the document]
enarhi: Fixed.

Geometry_Engine/Query/IsContaining.cs
```diff
@@ -709,6 +709,182 @@ public static bool IsContaining(this Cuboid cuboid, IGeometry geometry, bool acc
...
+        /***************************************************/
+
+        [Description("Checks if a point lies within a collection of polylines describing a closed volume using their primitive planes and bounds.")]
+        [Input("panels", "A collection of Polylines outlining each of the faces of the closed volume to check with.")]
+        [Input("point", "The point to check to see if it is contained within the bounds of the panels.")]
+        [Input("acceptOnEdges", "Decide whether to allow the point to sit on the edge of the panel, default false.")]
+        [Input("tolerance", "Distance tolerance to use to determine intersections.")]
+        [Output("isContaining", "True if the point is contained within the bounds of the panels, false if it is not for each point provided.")]
+        public static bool IsContaining(this List<Polyline> panels, Point point, bool acceptOnEdges = false, double tolerance = BH.oM.Geometry.Tolerance.Distance)
+        {
```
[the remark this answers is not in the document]
enarhi: Point method removed, using List<Point> now.

Geometry_Engine/Query/IsContaining.cs
```diff
@@ -709,6 +709,182 @@ public static bool IsContaining(this Cuboid cuboid, IGeometry geometry, bool acc
...
+
+        /***************************************************/
+
+        [Description("Checks if a point lies within a collection of polylines describing a closed volume contains each of a provided list of list of points.")]
+        [Input("panels", "A collection of Polylines outlining each of the faces of the closed volume to check with.")]
+        [Input("pointLists", "The List of Lists of points to check to see if each List of points are contained within the bounds of the panels.")]
+        [Input("acceptOnEdges", "Decide whether to allow the points to sit on the edge of the panel, default false.")]
+        [Input("acceptPartialContainment", "Decide whether to allow some of the points to sit outside the panels as long as at least one is within them.")]
+        [Output("isContaining", "True if the points of each sublist are contained within the bounds of the panels, false if it is not for each sublist of points provided.")]
+        public static List<bool> IsContaining(this List<Polyline> panels, List<List<Point>> pointLists, bool acceptOnEdges = false, bool acceptPartialContainment = false, double tolerance = BH.oM.Geometry.Tolerance.Distance)
```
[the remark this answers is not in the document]
enarhi: Fixed, all methods ultimately call List<List<Point>>

Geometry_Engine/Query/IsContaining.cs
enarhi: Replaced with call as noted.

Geometry_Engine/Query/IsContaining.cs
enarhi: fixed

Geometry_Engine/Query/IsContaining.cs
enarhi: fixed

Geometry_Engine/Query/IsContaining.cs
enarhi: fixed

Geometry_Engine/Query/IsContaining.cs
enarhi: checking at start of method now

Geometry_Engine/Query/IsContaining.cs
enarhi: fixed

enarhi:
@BHoMBot check compliance

Geometry_Engine/Query/IsContaining.cs
```diff
@@ -709,6 +709,158 @@ public static bool IsContaining(this Cuboid cuboid, IGeometry geometry, bool acc
...
+                 rVec = BH.Engine.Geometry.Create.RandomVector(seed); 
+                
+                isInPlane = false;
+                foreach (Plane p in planes)
+                    isInPlane = isInPlane || rVec.IsInPlane(p, tolerance);
+
+                if (isInPlane)
+                {
+                    seed += 1;
+                }
```
pawelbaran, line 828: You need to normalise the random vector.

Besides that I could not stop myself from rewriting the code according to my own likings 😉 But it is a matter of personal preference, so please feel free to ignore 👍 
```suggestion
                int seed = 1;
                rVec = BH.Engine.Geometry.Create.RandomVector(seed).Normalise();                 
                isInPlane = planes.Any(p => rVec.IsInPlane(p, tolerance));

                if (isInPlane)
                    seed ++;
```

Geometry_Engine/Query/IsContaining.cs
```diff
@@ -709,6 +709,158 @@ public static bool IsContaining(this Cuboid cuboid, IGeometry geometry, bool acc
...
+            List<Point> intersectPoints = new List<Point>();
+
+            for (int x = 0; x < planes.Count; x++)
+            {
+                Point intPt = PlaneIntersection(line, planes[x], true, tolerance);
+                if (intPt == null)
+                    continue;
+                Polyline pLine = closedVolume[x];
+
+                if  (BH.Engine.Geometry.Query.IsContaining(boundingBox, intPt, true, tolerance) && BH.Engine.Geometry.Query.IsContaining(pLine, new List<Point> { intPt }, false, tolerance))
```
pawelbaran, line 845: I am thinking abstract about the problem in the context of call to `IsContaining(Polyline)` query, we have 3 interesting intersection scenarios:
- point inside solid, ray not hitting any edge
- point inside solid, ray hitting edges
- point outside solid, ray hitting edges

The 1st will always yield correct results, 2nd will yield wrong results if `acceptOnEdges` is set to `false` and the ray leaves the solid through the edges, while 3rd will return false if `acceptOnEdges` is set to `true` and the ray hits an edge.

The above makes perfect sense if you look at how the method for `Polyline` was written:
https://github.com/BHoM/BHoM_Engine/blob/851d1acd212c1aa4123d24b4f3749154af958146/Geometry_Engine/Query/IsContaining.cs#L337

...writing similar for 2d panels instead of lines sounds like lots of fun, so I started thinking how to avoid it - what came to my mind is that you could add one more check against the random vector to make sure that the ray does not hit the outlines of any panel, and if it does, then pick another random. What do you think?

Follow up thought: would be good to start from checking if the point itself does not lie on the edge - can help you skip a few heave computations + will let you avoid infinite loop in case of implementation of the suggestion above.

Geometry_Engine/Query/IsContaining.cs
```diff
@@ -709,6 +709,158 @@ public static bool IsContaining(this Cuboid cuboid, IGeometry geometry, bool acc
...
+
+            //Use a length longer than the longest side in the bounding box. Change to infinite line?
+            Line line = new Line() { Start = point, End = point.Translate(rVec * (((boundingBox.Max - boundingBox.Min).Length()) * 10)) };
+
+            //Check intersections
+            List<Point> intersectPoints = new List<Point>();
+
+            for (int x = 0; x < planes.Count; x++)
+            {
+                Point intPt = PlaneIntersection(line, planes[x], true, tolerance);
```
pawelbaran, line 840: Are you aware your line will not be treated as infinite here? To achieve that, you would first need to set `Infinite = true`. I think we do not need it here, but just FYI

Geometry_Engine/Query/IsContaining.cs
```diff
@@ -709,6 +709,158 @@ public static bool IsContaining(this Cuboid cuboid, IGeometry geometry, bool acc
...
+
+            bool isContained = intersectPoints.CullDuplicates().Count % 2 != 0;
+
+            if (!isContained && acceptOnEdges)
+            {
+                //Check the edges in case the point is on the edge of the BE
+                foreach (Polyline p in closedVolume)
+                {
+                    if (point.IsOnCurve(p, tolerance))
+                        isContained = true;
```
pawelbaran, line 857: You missed exit from the loop. You can make the code even shorter btw:
```C#
if (!isContained && acceptOnEdges)
    isContained  = closedVolume.Any(p => point.IsOnCurve(p, tolerance));

```

```suggestion
                    if (point.IsOnCurve(p, tolerance))
                    {
                        isContained = true;
                        break;
                    }
```

pawelbaran (changes requested):
Please don't hate me @enarhi, getting me to review geometry never ends up without comments 😃

Geometry_Engine/Query/IsContaining.cs
```diff
@@ -709,6 +709,158 @@ public static bool IsContaining(this Cuboid cuboid, IGeometry geometry, bool acc
...
+
+        /***************************************************/
+
+        [Description("Checks if a list of points lies within a collection of polylines describing a closed volume using their primitive planes and bounds.")]
+        [Input("closedVolume", "A collection of Polylines outlining each of the faces of the closed volume to check with.")]
+        [Input("points", "The points to check to see if it is contained within the bounds of the panels.")]
+        [Input("acceptOnEdges", "Decide whether to allow the point to sit on the edge of the panel, default false.")]
+        [Input("tolerance", "Distance tolerance to use to determine intersections.")]
+        [Output("isContaining", "True if the point is contained within the bounds of the panels, false if it is not for each point provided.")]
+        public static List<bool> IsContaining(this List<Polyline> closedVolume, List<Point> points, bool acceptOnEdges = false, double tolerance = BH.oM.Geometry.Tolerance.Distance)
```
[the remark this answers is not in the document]
enarhi: Added additional method for what you were expecting this one to do, as this one returns a list of bools for each point in the list of points, the new one returns one bool for the list of points with the partialContainment option.

Geometry_Engine/Query/IsContaining.cs
```diff
@@ -709,6 +709,158 @@ public static bool IsContaining(this Cuboid cuboid, IGeometry geometry, bool acc
...
+                BH.Engine.Base.Compute.RecordError("Cannot query if a collection of panels contains a point if the panels are null.");
+                return new List<bool>() { false };
+            }
+
+            List<Point> ctrPoints = new List<Point>();
+            List<Plane> planes = new List<Plane>();
+            foreach (Polyline be in closedVolume)
+            {
+                List<Point> srfPts = be.IControlPoints();
+                planes.Add(srfPts.FitPlane(tolerance));
```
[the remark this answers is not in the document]
enarhi: Added

Geometry_Engine/Query/IsContaining.cs
enarhi: Suggest this is created as a separate follow up issue,

Geometry_Engine/Query/IsContaining.cs
enarhi: Updated.

Structure_Engine/Modify/SetAnalysisParameters.cs
enarhi: Fixed itself, change removed now.

Geometry_Engine/Query/IsContaining.cs
```diff
@@ -709,6 +709,158 @@ public static bool IsContaining(this Cuboid cuboid, IGeometry geometry, bool acc
...
+                ctrPoints.AddRange(srfPts);
+            }
+
+            BoundingBox boundingBox = BH.Engine.Geometry.Query.Bounds(ctrPoints);
```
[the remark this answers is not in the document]
enarhi: added.

Geometry_Engine/Query/IsContaining.cs
```diff
@@ -709,6 +709,158 @@ public static bool IsContaining(this Cuboid cuboid, IGeometry geometry, bool acc
...
+        private static bool IsContaining(this List<Polyline> closedVolume, List<Plane> planes, BoundingBox boundingBox, Point point, bool acceptOnEdges = false, double tolerance = BH.oM.Geometry.Tolerance.Distance)
+        {
+            //Return if point is null even without checking boundingBox.IsContaining(point)
+            if (point == null)
+            {
+                BH.Engine.Base.Compute.RecordError("Cannot query if a collection of panels contains a point if the point is null.");
+                return false;
+            }
+
+            if (planes.Any(x => x == null))
```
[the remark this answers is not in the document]
enarhi: Removed.

Geometry_Engine/Query/IsContaining.cs
enarhi: Modified per suggestion.

enarhi:
@BHoMBot check compliance

Analytical_Engine/Compute/ExtrudeToVolume.cs
```diff
@@ -0,0 +1,104 @@
...
+        [Description("Takes a region with a Floor Perimeter and creates a collection of polylines which represent the closed volume of the region.")]
+        [Input("region", "A region with a floor perimeter to extrude into a collection of panels.")]
+        [Input("height", "The height of the region, as a double, to calculate the ceiling level of the region. This will be used as the Z value of the perimeter + the given height.")]
+        [Input("tolerance", "The degree of tolerance on the angle calculation for collapsing the regions perimeter to a polyline. Default is equal to BH.oM.Geometry.Tolerance.Angle.")]
+        [Output("polylines", "A collection of polylines which represent the closed volume of the region.")]
+        [PreviousInputNames("region", "room, space")]
+        public static List<Polyline> ExtrudeToVolume(this IRegion region, double height, double tolerance = BH.oM.Geometry.Tolerance.Angle)
+        {
+            if (region == null)
+                return new List<Polyline>();
```
pawelbaran, line 50: @FraserGreenroyd do you know if we landed with a consistent convention around such cases? It tempts me to ask for a change to `return null`, so that `null` provided yields `null` in return

Analytical_Engine/Compute/ExtrudeToVolume.cs
```diff
@@ -0,0 +1,104 @@
...
+
+        [Description("Takes a polyline perimeter and creates a collection of polylines which represent the closed volume of a space.")]
+        [Input("pLine", "A polyline perimeter to extrude into a collection of panels.")]
+        [Input("connectingSpaceName", "The name of the space the panels will enclose.")]
+        [Input("height", "The height of the space, as a double, to calculate the ceiling level of the room. This will be used as the Z value of the perimeter + the given height.")]
+        [Output("polylines", "A collection of polylines which represent the closed volume of the region.")]
+        public static List<Polyline> ExtrudeToVolume(this Polyline pLine, string connectingSpaceName, double height)
+        {
+            if (pLine == null)
+                return new List<Polyline>();
```
pawelbaran, line 64: Same as above

Analytical_Engine/Query/IsContaining.cs
```diff
@@ -81,6 +83,40 @@ public static partial class Query
...
+
+        [Description("Defines whether an Environment Space contains each of a provided list of points.")]
+        [Input("region", "An Environment Space object defining a perimeter to build a 3D volume from and check if the volume contains the provided point.")]
+        [Input("regionHeight", "The height of the region.", typeof(BH.oM.Quantities.Attributes.Length))]
+        [Input("points", "The points being checked to see if it is contained within the bounds of the 3D volume.")]
+        [Input("acceptOnEdges", "Decide whether to allow the point to sit on the edge of the region, default false.")]
+        [Output("isContaining", "True if the point is contained within the region, false if it is not.")]
+        public static List<bool> IsContaining(this IRegion region, double regionHeight, List<Point> points, bool acceptOnEdges = false)
+        {
+            List<Polyline> panelsFromSpace = region.ExtrudeToVolume(regionHeight);
```
pawelbaran, line 95: This method could be much faster if you simply check whether the points fall within height domain and outline?

Geometry_Engine/Query/IsContaining.cs
```diff
@@ -709,6 +709,168 @@ public static bool IsContaining(this Cuboid cuboid, IGeometry geometry, bool acc
...
+        /***************************************************/
+
+        [Description("Checks if a list of points lies within a collection of polylines describing a closed volume using their primitive planes and bounds.")]
+        [Input("closedVolume", "A collection of Polylines outlining each of the faces of the closed volume to check with.")]
+        [Input("points", "The points to check to see if it is contained within the bounds of the panels.")]
+        [Input("acceptOnEdges", "Decide whether to allow the point to sit on the edge of the panel, default false.")]
+        [Input("acceptPartialContainment", "Decide whether to allow some of the points to sit outside the panels as long as at least one is within them.")]
+        [Input("tolerance", "Distance tolerance to use to determine intersections.")]
+        [Output("isContaining", "True if the points are contained within the bounds of the panels, false if they are not.")]
+        public static bool IsContaining(this List<Polyline> closedVolume, List<Point> points, bool acceptOnEdges = false, bool acceptPartialContainment = false, double tolerance = BH.oM.Geometry.Tolerance.Distance)
```
pawelbaran, line 751: Do we need both overloads taking a list of points, with `acceptPartialContainment` and without?

Geometry_Engine/Query/IsContaining.cs
```diff
@@ -709,6 +709,168 @@ public static bool IsContaining(this Cuboid cuboid, IGeometry geometry, bool acc
...
+            }
+
+            List<Point> ctrPoints = new List<Point>();
+            List<Plane> planes = new List<Plane>();
+            foreach (Polyline be in closedVolume)
+            {
+                List<Point> srfPts = be.IControlPoints();
+                Plane fitPlane = srfPts.FitPlane(tolerance);
+                if (fitPlane != null)
+                    planes.Add(fitPlane);
```
pawelbaran, line 784: Does it make sense to run the containment check if a plane is missing? Wouldn't we end up with different collection counts for `closedVolume` and `planes`?

Geometry_Engine/Query/IsContaining.cs
pawelbaran: Nope, at least from how I read the code of `PlaneIntersection` query: you need to have both `true` parameter and `line.Infinite` set to `true`

pawelbaran (commented):
The code looks good, although I have left a few more comments and questions. Regarding the potential bug related to rays cutting the corners of a solid, agreed it could become a separate issue. However, before we do that I will try to crack it myself before the whole thought process evaporates from my head and commit if successful.

Analytical_Engine/Query/IsContaining.cs
enarhi: Leaving out of this PR, can be picked up as a separate issue if necessary.

Geometry_Engine/Query/IsContaining.cs
enarhi: Resolved per latest commit from @pawelbaran

Geometry_Engine/Query/IsContaining.cs
enarhi: Resolved per latest commit from @pawelbaran

Geometry_Engine/Query/IsContaining.cs
enarhi: Resolved per latest commit from @pawelbaran

Geometry_Engine/Query/IsContaining.cs
enarhi: Yes, one checks all points for a single containment value, the other checks a list of points for individual containment and returns a list of result values.

enarhi:
@BHoMBot check compliance

Analytical_Engine/Compute/ExtrudeToVolume.cs
FraserGreenroyd: Based on our [cheat sheet](https://bhom.xyz/documentation/Development/Concepts/Null-Handling/#cheat-sheet) either an empty list or a `null` is acceptable for this scenario. Given the propensity of issues by returning `null` from the `ExtrudeToVolume` I think previously it was decided to return an empty list when the method existed within the Environment Engine for further calculations which is probably why it's doing the same here now.

Analytical_Engine/Compute/ExtrudeToVolume.cs
enarhi: Okay as per comment above.

Geometry_Engine/Query/IsContaining.cs
enarhi: Resolved with latest changes.

enarhi:
@BHoMBot check required

enarhi:
@BHoMBot check versioning

Geometry_Engine/Query/IsContaining.cs
```diff
@@ -709,6 +709,193 @@ public static bool IsContaining(this Cuboid cuboid, IGeometry geometry, bool acc
...
+                if (intPt == null)
+                    continue;
+
+                Polyline pLine = closedVolume[x];
+
+                if  (BH.Engine.Geometry.Query.IsContaining(boundingBox, intPt, true, tolerance) && BH.Engine.Geometry.Query.IsContaining(pLine, new List<Point> { intPt }, false, tolerance))
+                    intersectPoints.Add(intPt);
+            }
+
+            bool isContained = intersectPoints.CullDuplicates().Count % 2 != 0;
```
pawelbaran, line 875:
```suggestion
            bool isContained = intersectPoints.CullDuplicates(tolerance).Count % 2 != 0;
```

Geometry_Engine/Query/IsContaining.cs
```diff
@@ -709,6 +709,193 @@ public static bool IsContaining(this Cuboid cuboid, IGeometry geometry, bool acc
...
+            return isContained; //If the number of intersections is odd the point is outsde the space
+        }
+
+        private static List<Line> UniqueSegments(this List<Polyline> outlines, double tolerance)
+        {
+            double sqTol = tolerance * tolerance;
+            List<Line> lines = new List<Line>();
+            foreach (Line line in outlines.SelectMany(x => x.SubParts()))
+            {
+                if (lines.All(x => !x.IsSimilar(line, tolerance)))
```
pawelbaran, line 886: Nothing more fun than correcting own self 😃 
```suggestion
                if (lines.All(x => !x.IsSimilar(line, sqTol)))
```

Geometry_Engine/Query/IsContaining.cs
```diff
@@ -709,6 +709,193 @@ public static bool IsContaining(this Cuboid cuboid, IGeometry geometry, bool acc
...
+                if  (BH.Engine.Geometry.Query.IsContaining(boundingBox, intPt, true, tolerance) && BH.Engine.Geometry.Query.IsContaining(pLine, new List<Point> { intPt }, false, tolerance))
+                    intersectPoints.Add(intPt);
+            }
+
+            bool isContained = intersectPoints.CullDuplicates().Count % 2 != 0;
+
+            return isContained; //If the number of intersections is odd the point is outsde the space
+        }
+
+        private static List<Line> UniqueSegments(this List<Polyline> outlines, double tolerance)
```
pawelbaran, line 880: Apologies @enarhi for leaving mess behind, but could we move this method to the private section in the file?

enarhi: Moved.

Geometry_Engine/Query/IsContaining.cs
enarhi: Fixed.

Geometry_Engine/Query/IsContaining.cs
enarhi: FIxed.

enarhi:
@BHoMBot check versioning

enarhi:
@BHoMBot check versioning

enarhi:
@BHoMBot check compliance

pawelbaran (approved):
Nice addition to the engine, glad we ironed out the potential bugs ahead of merging - thanks @enarhi!

pawelbaran:
@FraserGreenroyd is your change request still valid or can we proceed with merging?

enarhi:
@BHoMBot check required

Tom-Kingstone (approved):
@FraserGreenroyd asked me to go through the test procedures for the IES and XML toolkits, including the new automated test procedures. None of the changes in this PR break any of the existing functionality and all of these tests pass, so I am happy to approve for this case.

FraserGreenroyd (approved):
I've reviewed the code and run a couple of test scripts. @Tom-Kingstone has run larger test procedures against dependent toolkits which required the `IsContaining` method originally to ensure that functionality isn't broken which he's reporting is the case, so happy to approve for merge.

FraserGreenroyd:

FraserGreenroyd:
@BHoMBot check ready-to-merge

Issue #3264, opened 2024-01-25
Geometry_Engine: Add HashArray method for core geometry types


pawelbaran:
#### Description:

As in the title, `HashArray` seems to be missing for a few core geometry types:
- `TransformMatrix`
- `Cartesian`

alelom:
The reason why we didn't add GeometryHash for many geometric types is because we thought it would never get used with them. In particular, we added support for "non-conceptual" types: everything except Vector, TransformMatrix, Cartesian, and probably a few more. Anyways, it can make sense, for completion's sake, to add support for all.

pawelbaran:
Yes, I see immediate use for transforms, so definitely worth adding 👍

Issue #3255, opened 2024-01-17
Base_Engine: Hash to leverage the new aggregating GeometryHash method for geometry types


alelom:
#### Description:

This depends on #3251 . 

Currently, the Hash method traverses all properties of geometric types to create their signature. This can be extremely slow, and can instead leverage the GeometryHash method(s). In particular, when we will have implemented the aggregating GeometryHash method that returns a single number (#3251), we can use it instead of this property traversal for geometric types in the main `Hash` method. We can expose this as the default behaviour, but also have an option in the `ComparisonConfig` to toggle the old property traversal behaviour for Geometric types, if required for some reason. Further, we can expose the GeometryHash optional arguments such as the translation tolerance as properties of the `ComparisonConfig` class. 

This will greatly increase the speed (and accuracy) of the Hash method for geometric types. Further, we will be able to always use the main Hash method for all use cases – including geometric types. This would make things more consistent and simpler to use.

Thoughts? @pawelbaran @al-fisher

pawelbaran:
Looks like the great minds think alike! 😃 I had the same thought when I first learnt about `GeometryHash` - why aren't we calling it from the `Hash` method?! So yes, 100% support, especially that I do not think there is any precedent where we would persist the hashes at the moment - most users compare on the fly, i.e. if what they think is the same still comes back as the same, it is not a breaking change 👍

al-fisher:
Yes - 100% 😸 Love this

PR #3257, opened 2024-01-18
Geometry_Engine: implement GeometryHash as a single value

alelom:
### NOTE: Depends on 

Not a dependency, but a dependant: https://github.com/BHoM/DiffingTests_Prototypes/pull/17
   
### Issues addressed by this PR

Closes #3251

PR #3268, opened 2024-01-29
Diffing_Engine: Hash and Diff to leverage new GeometryHash

### LINKED ISSUE #3269: Geometry_Engine: `GeometryHash()` to support the options from `BaseComparisonConfig`
#### Description:

Because `BaseComparisonConfig` has the role of determining the configuration used when identifying an object (for comparisons), it should be provided as an input to all comparison and hashing methods. We need to introduce it in the `GeometryHash()` method and, consequently, in the `HashArray()` (formerly called "`ToDoubleArray()`") method.

This is particularly needed when we implement the support for using the GeometryHash method from the main Hash and Diffing methods. In fact, in order to add this support, we need to provide an new "UseGeometryHash" option to the BaseComparisonConfig (see BHoM/BHoM#1587). This last option needs to be orthogonal to the rest of the options in the BaseComparisonConfig, meaning, we need to make the rest of the options relevant even when setting "UseGeometryHash" to true.

An alternative would be to _not_ implement the BaseComparisonConfig options in the GeometryHash method, but this means that, if we want to implement BHoM/BHoM#1587, the UseGeometryHash option would not be orthogonal to the rest of the options, requiring to make the BaseComparisonConfig IImmutable and introduce checks whenever it is used, in order to ensure that the combination of its property is consistent. For example, we'd not be able to allow a workflow where TypeExceptions includes a Geometric Type and UseGeometryHash is set to true.  
This alternative however adds a lot of burden for existing downstream implementations, and risks regarding usage of Hash and Diffing, because ensuring consistency is not easy. After careful evaluation, I determined that it is more convenient to implement support for all of the BaseComparisonConfig options in the GeometryHash rather than going for this alternative. In fact, it is possible to add this support without impacting performance (only a minor impact is expected when the Config includes geometrical settings, e.g. if TypeExceptions include a Geometric type).

alelom:
### NOTE: Depends on 

https://github.com/BHoM/BHoM/pull/1594
   
Closes https://github.com/BHoM/BHoM_Engine/issues/3255
Closes #3269

_The following was also implemented in this PR in order to be able to test properly the implementation of #3255._
Closes https://github.com/BHoM/BHoM_Engine/issues/3264  

This PR makes the base `Hash()` use the `GeometryHash()` method by default (and consequently, the Diffing can also leverage it via the `DiffWithHash()` method). This can give a significant speed boost for an average of 33% less computation time, and a potential 99% for heavy geometry inputs. The average 33% speed improvement was measured via the [DiffingTests_Prototypes](https://github.com/BHoM/DiffingTests_Prototypes/pull/18) `DiffProfiling` project and the results are as below:
<details>

___________________

### Profiling results, running the `DiffingProfiling` project
The profiling considered an _average_ use case with a mixture of geometrical and non-geometrical objects.
- Before (all on `develop`):  
2024-01-29 20_45_06-C__Users_alombardi_Documents_GitHub_DiffingTests_Prototypes_DiffingProfiling_bin
- When on this PR [(BHoM_Engine #3268)](https://github.com/BHoM/BHoM_Engine/pull/3268):  
2024-01-29 20_24_42-Select C__Users_alombardi_Documents_GitHub_DiffingTests_Prototypes_DiffingProfil

This profiling was repeated 20 times and gives an average value of about 33% less time.

___________________

</details>

In order to make this possible, the `BH.Engine.Geometry.Query.HashArray()` method now makes use of the `BaseComparisonConfig` options, bringing it on par to the base `BH.Engine.Base.Query.Hash()` method in terms of identification potential for Geometry objects. This is a requirement for any method that return an object signature (Hash). See #3269 for more information.

### Test files

Same tests used in https://github.com/BHoM/BHoM_Engine/pull/3257, but calling `BH.Engine.Base.Hash()` instead of `BH.Engine.Geometry.GeometryHash()`.

Please also run all tests in https://github.com/BHoM/DiffingTests_Prototypes/pull/18. 
They should all pass. Please note that the default value of the new GeometryHash option is set to `true` in the ComparisonConfig (see https://github.com/BHoM/BHoM/pull/1594). The fact that all tests pass in the DiffingTests_Prototypes means that the new GeometryHash integration is fully compatible with the `Hash()` and `Diffing()` workflows and gives the same results, while also giving a performance boost.

### Changelog

- The base `BH.Engine.Base.Query.Hash()` method now leverages the `BH.Engine.Geometry.Query.GeometryHash()` method by default. This can give a significant speed boost, for an average of 33% less computation time, and a potential 99% for heavy geometry inputs.
- The `BH.Engine.Geometry.Query.HashArray()` method now makes use of the `BaseComparisonConfig` options, bringing it on par to the base `BH.Engine.Base.Query.Hash()` method in terms of identification potential for Geometry objects.


pawelbaran:
@BHoMBot check compliance

Geometry_Engine/Query/HashArray.cs
```diff
@@ -66,100 +86,173 @@ public static double[] IHashArray(this IGeometry igeometry)
         [Description("The geometry hash of a Curve is obtained by first retrieving any Sub-part of the curve, if present." +
             "The ISubParts() methods is able to return the 'primitive' curves that a curve is composed of. " +
             "The GeometryHashes are then calculated for the individual parts and concatenated.")]
-        private static double[] HashArray(this ICurve curve, double translationFactor)
+        private static double[] HashArray(this ICurve curve, double translationFactor, BaseComparisonConfig comparisonConfig = null, string fullName = null)
         {
+            if (comparisonConfig?.TypeExceptions?.Any(t => typeof(ICurve).IsAssignableFrom(t)) ?? false)
+                return default;
+
+            if (curve == null)
```
pawelbaran, line 94: I would consider putting it as the 1st check to be made in the method to avoid redundant checks against type exceptions. Minor thing, but may have some impact in case of massive workflows with complex configs?

Geometry_Engine/Query/HashArray.cs
```diff
@@ -66,100 +86,173 @@ public static double[] IHashArray(this IGeometry igeometry)
...
+                return default;
+
             List<ICurve> subParts = curve.ISubParts().ToList();
 
+            if (!subParts.Any())
```
pawelbaran, line 99: That looks pretty complex as for invalid curves 😉 The only chances for this `if` to be hit is empty `PolyCurve` or `Polyline` with less than 2 control points...

Geometry_Engine/Query/HashArray.cs
```diff
@@ -66,100 +86,173 @@ public static double[] IHashArray(this IGeometry igeometry)
...
+                    skipEndPoint: true,
+                    comparisonConfig: comparisonConfig,
+                    fullName: fullName.AppendPropertyName($"[{i}]"))
+                );
             }
-            //Include endpoint for hasing for last curve
-            hashes.AddRange(HashArray(subParts.Last() as dynamic, translationFactor, false));
+
+            //Include endpoint for hashing for last curve
+            hashes.AddRange(HashArray(subParts.Last() as dynamic,
```
pawelbaran, line 126: What about the cases of disjoint `PolyCurves`? They theoretically are valid.

Geometry_Engine/Query/HashArray.cs
```diff
@@ -66,100 +86,173 @@ public static double[] IHashArray(this IGeometry igeometry)
...
+                return default;
+
             translationFactor += (int)TypeTranslationFactor.Arc;
 
-            IEnumerable<double> hash = curve.StartPoint().HashArray(translationFactor)
-               .Concat(curve.PointAtParameter(0.5).HashArray(translationFactor));
+            IEnumerable<double> hash = curve.StartPoint().HashArray(translationFactor,
+                    comparisonConfig,
+                    fullName.AppendPropertyName($"{fullName}.{nameof(StartPoint)}"))
+               .Concat(curve.PointAtParameter(0.5).HashArray(translationFactor,
```
pawelbaran, line 150: Is that used to enable avoiding checking curves for particular control points? If so, I would argue it is impractical, would be happy to skip this level of complexity in favour of code readability, assuming that no one will ever compare arcs ignoring their mid point (especially that the consumer would somehow need to know that arcs get sampled at parameter equal to 0.5)

Geometry_Engine/Query/HashArray.cs
```diff
@@ -66,100 +86,173 @@ public static double[] IHashArray(this IGeometry igeometry)
...
-                   .Concat(curve.PointAtParameter(0.66).HashArray(translationFactor));
+            IEnumerable<double> hash = curve.StartPoint().HashArray(translationFactor,
+                        comparisonConfig,
+                        fullName.AppendPropertyName($"{nameof(StartPoint)}"))
+                   .Concat(curve.PointAtParameter(0.33).HashArray(translationFactor,
+                        comparisonConfig,
+                        fullName.AppendPropertyName($"{nameof(PointAtParameter)}(33e-2)"))
+                   ).Concat(curve.PointAtParameter(0.66).HashArray(translationFactor,
+                        comparisonConfig,
+                        fullName.AppendPropertyName($"{nameof(PointAtParameter)}(66e-2)"))
```
pawelbaran, line 181: Following up ty my previous comment, here it is even more valid: the consumer would need to know the circles are sampled at 1/3, plus would need to know that a value needed to ignore it is 0.33 rather than 0.3333. Of course assuming that I have not misunderstood the code 😉

pawelbaran (commented):
Great stuff @alelom, although not trivial to keep track of 🤯 I did my best to review it properly and took the following steps:
- read the code changes - see comments, pardon my ignorance if any of them does not make sense, a few points are based more on intuition rather than in-depth debugging
- ran GH test scripts I have built for this concept on previous occasions - all passed

pawelbaran:
Test failures are gone after switching to the right branch in the diffing test repo, apologies and thanks for help @alelom!

Geometry_Engine/Query/HashArray.cs
alelom: Agreed, applied in e360db11d702c8219f104b7448a521e131f849fe

Geometry_Engine/Query/HashArray.cs
alelom: I had experienced this in my tests -- if we remove this some test can fail, especially the ones that generate geometry at random. Doesn't hurt to keep.

Geometry_Engine/Query/HashArray.cs
```diff
@@ -45,13 +45,33 @@ public static partial class Query
...
+        [Input("comparisonConfig", "Configurations on how the hash array is computed, with options for numerical approximation, type exceptions and many others.")]
+        [Input("fullName", "Name of the property that holds the target object to calculate the hash array for. This name will be used to seek any matching custom configuration to apply against the `comparisonConfig` input.")]
+        [Output("hashArray", "Array of numbers representing a unique signature of the input geometry.")]
+        public static double[] IHashArray(this IGeometry igeometry, BaseComparisonConfig comparisonConfig, string fullName = null)
         {
             if (igeometry == null)
                 return new double[] { };
 
-            return HashArray(igeometry as dynamic, 0);
+            fullName = fullName ?? igeometry.GetType().FullName;
```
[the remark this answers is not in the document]
alelom: Because if `fullName` is null, then the input object is not considered a property of some other object, but a stand alone object, whose name is its FullName.

Geometry_Engine/Query/HashArray.cs
alelom: After conversation with @pawelbaran we've agreed this is a thing to tackle separately from this PR, as this requires improving the mechanism that was originally designed for GeometryHash/ToHashArray. Raised issue here: https://github.com/BHoM/BHoM_Engine/issues/3272

Geometry_Engine/Query/HashArray.cs
alelom: The middle point is required to compare two arcs. The GeometryHash for an Arc is calculated as the GeometryHash of the start, end and middle point of the Arc.

Geometry_Engine/Query/HashArray.cs
alelom: Similar response as above, although here we took a more arbitrary decision. As per its description, the GeometryHash for a Circle is calculated as the GeometryHash of the start, 1/3rd and 2/3rd points of the Circle. This is because you need at least 3 points to make sure two Circles are equal. Also this is not the subject of this PR, which is again only adapting the existing mechanism to support BaseComparisonConfig. Happy for you to raise a separate issue if you think that needs improving.

pawelbaran: I do not question picking the points as unique definitions of curves. My question was pointed at the value of appending the name of such arbitrary property as point at 0.66 parameter - the consumer would never know that a point can be excluded from hash computation by setting such value in properties to ignore. Therefore, we would be fine with simply passing the original value of `fullName` I believe (assuming that no one cares about that name so deep down the algorithm). However, it probably makes sense to keep consistent across the entire call stack, so happy with the code as is! 👍

pawelbaran (approved):
Happy to approve following code review, testing and offline discussions with @alelom. Good stuff, thanks!

Geometry_Engine/Query/HashArray.cs
alelom: Ah sorry I didn't understand! Yes you're right then -- it's may be a bit overzealous, I just went straight for the most flexible mechanism.

Geometry_Engine/Query/HashArray.cs
alelom: Ah sorry I didn't understand! Yes you're right then -- it's may be a bit overzealous, I just went straight for the most flexible mechanism.

alelom:
@BHoMBot check required

FraserGreenroyd:

FraserGreenroyd:

PR #3281, opened 2024-02-14
BHoM_Engine: Use NetStandard version of DeepCloner in Nugets rather than NetFramework

### LINKED ISSUE #3223: Fix missing dependencies on DeepCloner or replace DeepCloner with something that supports net standard 2.0
#### Description:
At the moment, DeepCloner is only referenced for .NET Framework on the nuget package. This means that we end up with crashed on the Zero Code Tool side. Here's an example: 

image

This means we end up having to manually add its NuGet in the ZCT, hoping that the tree shaker will not remove it on publish.
We do get the following message when adding the package though:

image

#### Steps to reproduce:
A good ZCT to look at to reproduce the problem is the Structure Embodied Carbon Tool

#### Expected behaviour:
We shouldn't have to manually fix dependencies for NuGet packages.

FraserGreenroyd:
Fixes #3223 

This is based on [this Nuget report](https://www.nuget.org/packages/DeepCloner/0.10.2#supportedframeworks-body-tab) which highlights the `netstandard1.3` is a supported version for DeepCloner, and as such has a computed support package up to `net5.0` and `net6.0` crucially for the tools being prototyped across the BHoM community.

Ultimate proof will be in the nuget produced at the end of this sprint, hopefully will resolve the warning and error.

### DIFF
--- a/.ci/BHoMBot/Nuget/BHoM.Engine.nuspec
+++ b/.ci/BHoMBot/Nuget/BHoM.Engine.nuspec
@@ -23,7 +23,7 @@
         <dependency id="Mono.Reflection" version="2.0.0" />
         <dependency id="CompareNETObjects" version="4.67.0" />
       </group>
-      <group targetFramework="net472">
+      <group targetFramework="netstandard1.3">
           <dependency id="DeepCloner" version="0.10.2" />
       </group>
     </dependencies>


FraserGreenroyd:
@BHoMBot check compliance
@BHoMBot check core
@BHoMBot check versioning
@BHoMBot check null-handling
@BHoMBot check serialisation

albinber (approved):
Change make sense, approving

PR #3286, opened 2024-02-15
BHoM_Engine: Enhance the logging functionality by including ability to suppress logging if desired

### LINKED ISSUE #1840: Base_Engine: Centralisation and Standardisation of errors, warnings and notes
#### Description:

Sorry if this has already been raised elsewhere(!!) Having just made up another warning text, I've been thinking for a while it could be useful to standardise some common warnings into a format - e.g. IsNurbsCurve or IsPlanar, concatenating (or not depending what the format(s) we decide) things like the guid.

FraserGreenroyd:
Fixes #1840 

### Test files
Available [here](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?newTargetListUrl=%2Fsites%2FBHoM%2F02%5FCurrent&viewpath=%2Fsites%2FBHoM%2F02%5FCurrent%2FForms%2FAllItems%2Easpx&id=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F02%5FPull%20Request%2FBHoM%2FBHoM%5FEngine%2FBHoM%5FEngine%2F%233286%2DLogging&viewid=a5a93cba%2Dfcca%2D46cc%2Da31a%2D80b8d5cbfd7b)

### Changelog
 - Add ability to suppress recording events
 - Add ability to unsuppress recording events (events are recorded by default, this is to allow switching back on after being switched off)
 - Add ability to toggle throwing error events as exceptions or not

### Additional comments

I have noted some `ToDo` around documentation and potential further discussion. One item of further discussion is on defaulting throwing exceptions. Currently this is being defaulted to off, however, @alelom raised a point during discussion that we could default to on with BHoM_UI handling turning this off when it initialises to keep the UX on the UI side clean. Happy to debate, however, scope of this PR is limited to adding the functionality so I would do a separate joint PR with here and BHoM_UI if we decide we want to do that after this PR.


BHoM_Engine/Compute/SwitchOffRecording.cs
```diff
@@ -0,0 +1,85 @@
...
+{
+    public static partial class Compute
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Switch off the entire logging system used within BHoM. Any part of the code base which tries to log any note, warning, error, or other event into the log system will be housed in the silent recorder and not displayed to users. By default, all recording systems are turned on when BHoM is initialised.")]
+        [Input("areYouSure", "Set this to true if you are sure you want to turn off all log recording. This boolean exists so that if this component is placed on a BHoM UI accidently, it does not turn off the system unless then expressly requested.")]
+        public static void SwitchOffRecording(bool areYouSure)
```
travispotterBH, line 39: I have a though on this PR as a whole. We are relying on the user to remember to turn back on the logging system the use of code where this is implemented could have effects where during an entire session the logging has been turned off. 

I'm considering an approach where we use the "using" statement.
We can create a class that implements IDisposable, called EventsSupressor, and an enum.
So by using "using" we only allow suppression between a scope and the suppression is removed once the scope has been exceeded. This makes it so that you can only have your suppression last for the lifetime of your method, rather than affecting the whole system after your method has been called, you you've forgotten to renable

``` csharp
using System;
using System.Collections;
using System.Collections.Generic;
using System.Linq;

main();

void main (params string[] args){
    Supression suppressionConfig = Supression.Warnings & Supression.Errors & Supression.Notes;

    using(var eventSupressor = new EventSupressor(suppressionConfig)){
        // do everything you want to do
    }

   // outside of the scope the supression has been returned to normal 
}

enum Supression{
    Warnings,
    Errors,
    Notes
}


class EventSupressor : IDisposable
{
    List<string> m_SupressedLog = new List<string>();

    public EventSupressor(Supression config)
    {
      //logic that turns off supresses 
        if(config.HasFlag(Supression.Warnings)){
           //supress warnings 
        }

        if(config.HasFlag(Supression.Errors)){
           //supress errors
        }

        if(config.HasFlag(Supression.Notes)){
           //supress Notes
        }
    }

    public void Dispose()
    {
      //logic that turns returns logging to proper state
      System.Console.WriteLine("Logging Supression has been ended.");
    }

```

travispotterBH (changes requested):
Generally naming could be better, but also, I question the approach, and believe that we should protect against accidentally affecting the logging of event/warnings/errors outside of the place of implementation of this code.

FraserGreenroyd:
> Generally naming could be better, but also, I question the approach, and believe that we should protect against accidentally affecting the logging of event/warnings/errors outside of the place of implementation of this code.

Thanks for the speedy feedback @travispotterBH , much appreciated 🥇 

I've gone through and updated the language from `switch off` to `suppress` as I think this is easier to handle within the code base. For example, it changed the variable `areWeSwitchedOff` to `suppressEvents` - turning it from a question variable to a statement variable and more in keeping with easy code readability. Hopefully this will also please @alelom (as he expressed the `areWeSwitchedOff` variable name could be clearer - which I think it now is) 😄 

For your suggestion in the larger comment, I would agree it could be nice, however, it is key to note that [BHoM does not attempt to standardise exact processes, because these must be flexible](https://bhom.xyz/documentation/#bhom-tries-to-avoid-common-pitfalls-of-other-aec-frameworks) and that philosophy should align to our code when used by developers as much as it aligns for our users. Our logging system has been working well for a while and this PR scope is not to refactor the entire aspect, but to enhance it to provide flexibility that we don't currently have, which hopefully this does (particularly as we grow out some of our tools in the ZeroCode Prototypes, and AI/ML prototypes @alelom is working with) 😄 

I do agree that there is a risk of the logging system being turned off for longer than it ought to be, and some form of automatic turn on would be nice. However, as a developer, I get frustrated when people try to protect me from myself, as I believe @adecler also feels from previous conversations. I would rather have the power to turn it off for as long as I want and if I forget to turn it back on then that's on me. Having an additional `using` block also risks the codebase drifting from being accessible to [any professional in the AEC industry, regardless of their level of computational proficiency](https://bhom.xyz/documentation/Basics/Technical-philosophy-of-the-BHoM/), as it'll add a layer of complexity to potentially all methods as opposed to very defined systems.

The risk of the user turning off the logging system is inherent with how we handle our code, by virtue of reflecting all methods into all our UIs. We deliberately decided that [BHoM functionality looks the same whether you use it from a programming script, a Grasshopper script, an Excel spreadsheet, or any other interface that can expose C#](https://bhom.xyz/documentation/Basics/Technical-philosophy-of-the-BHoM/#the-basic-architecture), and this isn't a philosophy point that I think we should move away from at this stage. However, the risk is there and to mitigate against it, I have opted to not just exclude recording events entirely - they just get suppressed from being seen by users but they still get logged in the `suppressedLog`. I have also added a method to `RetrieveSuppresedLog()`, so that if anyone does accidently leave the suppression state on for too long, they can retrieve the events back to the main log 😄 

I also expect we will be able to improve on this with iterations as we do with all things so I would limit the scope of this PR to introducing this functionality, and while it ought to be good functionality, we shouldn't necessarily aim to resolve all the issues we can foresee with it in this work - thus while I've marked the comment as resolved I'm very happy to continue debating it in an issue if you'd like 😄

pawelbaran:
I am a big fan of the concept in general, as some of you may know 😉 I am with @travispotterBH when it comes to switching vs suppressing, definitely the latter sounds better to me. To further improve what we have now, I would consider start/stop suppressing instead of suppress/unsuppress (GitHub highlights 'unsuppress' in red as nonexistent in English 🙈). But this is just my opinion as a non-native speaker, so no prob with getting voted down

FraserGreenroyd:
@BHoMBot check compliance

alelom (approved):
Code changes make sense and work well (tried script). 

Happy to approve also following our conversation @FraserGreenroyd . 

I'm adding a few more related thoughts below, not directly connected to this PR -- can be subject for separate conversations. The tl;dr version is that I think that this change is useful, and that we can bring this further by saving the original `Exception` in the Event object (rather than just its `StackTrace`). 

<details> 

Keeping the original exception would allow us to re-raise it when needed, in certain contexts (namely, in any non-UI context).

Since at the moment exceptions are swallowed by the event then, when you call BHoM methods from a non-UI context, this is limiting, because you don't directly see if an error occurred during a method call. Think about when writing NUnit tests, or when calling from terminal or another non-UI runtime like RhinoCompute. 

We can already _optionally_ pass the `Exception` to RecordError, however if and when the Exception is re-raised, we lose the original type of Exception, while only the StackTrace is retained. Having the original Exception re-thrown would be particularly useful to write exception-targeting tests, which is particularly useful for TDD. 

However, this would imply that need to we make sure that we always pass the original Exception to the RecordError method, in particular when calling RecordError from a `catch` clause.

</details>

BHoM_Engine/Compute/SuppressRecording.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        [Description("Suppress the logging system used within BHoM. Any part of the code base which tries to log notes, warnings, or errors into the log system will be housed in the suppressed log and not displayed to users depending on which systems you've chosen to suppress. By default, all recording systems are turned on when BHoM is initialised.")]
+        [Input("suppressErrors", "Determine whether to suppress BHoM Events of type ERROR from the log. Set to true to suppress these events.")]
+        [Input("suppressWarnings", "Determine whether to suppress BHoM Events of type WARNING from the log. Set to true to suppress these events.")]
+        [Input("suppressNotes", "Determine whether to suppress BHoM Events of type NOTE from the log. Set to true to suppress these events.")]
+        public static void StartSuppressRecordingEvents(bool suppressErrors = false, bool suppressWarnings = false, bool suppressNotes = false)
+        {
+            m_SuppressError = suppressErrors;
+            m_SuppressWarning = suppressWarnings;
+            m_SuppressNote = suppressNotes;
```
Tom-Kingstone, line 45: Reading the description of this method makes it seem like setting to true would lock `m_SuppressError` to true until `StopSuppressingRecordingEvents` is called, but actually it can be set to false and that will stop suppressing errors. As discussed with @FraserGreenroyd, there is a potential for a situation where a method suppresses all events, and then another method calls to suppress warnings but not errors, in which case errors would not be suppressed any more.

Tom-Kingstone (changes requested):

FraserGreenroyd:
When I rerun compliance checks, I will be dispensating code compliance because of [this issue](https://github.com/BHoM/Test_Toolkit/issues/466)

FraserGreenroyd:
@BHoMBot check versioning
@BHoMBot check installer

Tom-Kingstone (approved):
After testing, my changes were addressed correctly, and works as expected now. Happy to approve

FraserGreenroyd:

FraserGreenroyd:
@BHoMBot this is a DevOps instruction. I am authorising dispensation to be granted on check ref. 21651490134

albinber (approved):
Code look good and after testing everything work as intended. Happy to approve

FraserGreenroyd:
@BHoMBot check ready-to-merge

FraserGreenroyd:
@BHoMBot check ready-to-merge

PR #3293, opened 2024-02-23
Geometry_Engine: Tolerance introduced to the check of 2-dimensionality of the input point set

### LINKED ISSUE #3288: Geometry_Engine: Apply tolerance to preliminary check in FitLine
#### Description
~~https://github.com/BHoM/BHoM_Engine/pull/3280 changes the `FitLine` method in a way that makes `tolerance` input redundant. It would be good to remove it, rather in a separate PR to avoid the valuable feature being blocked due to versioning and cross dependencies.~~

Actually, the tolerance is needed to verify whether the input is at least 2-dimensional (in simple words, whether all input points do not overlap). Therefore removal of tolerance is not a solution, we rather need to make use of it instead.

pawelbaran:

Closes #3288

### Test files

On [SharePoint](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?RootFolder=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F02%5FPull%20Request%2FBHoM%2FBHoM%5FEngine%2FGeometry%5FEngine%2F%233293%2DMakeUseOfToleranceInFitLine&FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4)

### DIFF
--- a/Geometry_Engine/Compute/FitLine.cs
+++ b/Geometry_Engine/Compute/FitLine.cs
@@ -37,7 +37,7 @@ namespace BH.Engine.Geometry
 
         [Description("Fits a line into a set of points using Orthogonal Least Squares algorithm.")]
         [Input("points", "Points into which the line is meant to be fit.")]
-        [Input("tolerance", "Distance tolerance to be used in the process of fitting.")]
+        [Input("tolerance", "Distance tolerance to be used to check whether a line can be fit into the input points (i.e. whether the points are not coincident).")]
         [Output("fitLine", "Line fit into the input set of points based on the Orthogonal Least Squares algorithm.")]
         public static Line FitLine(this IEnumerable<Point> points, double tolerance = Tolerance.Distance)
         {
--- a/Geometry_Engine/Compute/FitLine.cs
+++ b/Geometry_Engine/Compute/FitLine.cs
@@ -35,7 +35,7 @@ namespace BH.Engine.Geometry
         /****              Public Methods               ****/
         /***************************************************/
 
-        [Description("Fits a line into a set of points using Orthogonal Least Squares algorithm.")]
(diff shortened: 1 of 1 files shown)


pawelbaran:
@BHoMBot check compliance

pawelbaran:
@BHoMBot check required

peterjamesnugent (commented):
Happy with the test script, and that it works as expected. Very clear and easy to follow.

Would it be worth adding an error message if we are going to return null?

pawelbaran:
Thanks @peterjamesnugent!

> Would it be worth adding an error message if we are going to return null?

Concerning the above question, I would rather not do that because e.g. `IsCollinear` is meant to return `true` in case of `FitLine() == null`, assuming that coincident points are collinear. Would that make sense to you?

Geometry_Engine/Compute/FitLine.cs
```diff
@@ -37,13 +37,13 @@ public static partial class Compute
 
         [Description("Fits a line into a set of points using Orthogonal Least Squares algorithm.")]
```
peterjamesnugent, line 38:
```suggestion
        [Description("Fits a line into a set of points using Orthogonal Least Squares algorithm. Will return null if the number of non-duplicate points is less than two.")]
```

Maybe something like this if we don't want to raise an error?

peterjamesnugent (approved):
Happy changes have been implementd and code is unchanged.

FraserGreenroyd:
@BHoMBot check compliance

FraserGreenroyd:
@BHoMBot check required

FraserGreenroyd:
@BHoMBot check ready-to-merge

FraserGreenroyd:

PR #3296, opened 2024-02-26
Settings_Engine: Ensure SaveSettings also updates settings in memory

### LINKED ISSUE #3295: Settings_Engine: Ensure that SaveSettings also updates settings loading in memory
Currently `SaveSettings` will update settings on disc, but not the settings in memory, while `UpdateSettings` will update the settings but not save them.

This means, when working with settings programmatically, you need to call both to ensure later calls to `GetSettings` return the ones saved - which is extra work for developer users.

So instead, if we `SaveSettings`, let's ensure we also `UpdateSettings` at the same time. However, keep `UpdateSettings` as is, to allow for people to update their in-memory settings without overwriting the saved settings for on-the-fly setting operations.

FraserGreenroyd:
Fixes #3295

### DIFF
--- a/Settings_Engine/Compute/SaveSettings.cs
+++ b/Settings_Engine/Compute/SaveSettings.cs
@@ -63,6 +63,8 @@ namespace BH.Engine.Settings
                 return false;
             }
 
+            Modify.UpdateSettings(settings); //Update the settings in memory as well so the next call to GetSettings has these settings which were saved
+
             return true;
         }
 


FraserGreenroyd:
@BHoMBot check compliance
@BHoMBot check required

Tom-Kingstone (approved):
Addition makes sense as a user otherwise would have to reset whenever changing settings.

FraserGreenroyd:

FraserGreenroyd:
@BHoMBot check ready-to-merge

PR #3363, opened 2024-06-26
Structure_Engine: Fixed bug where PanelToFEMesh returned an error 

peterjamesnugent:
### Issues addressed by this PR

Closes #3360 
Closes #3365

### Test files

[PanelToFEMesh](https://burohappold.sharepoint.com/:u:/s/BHoM/EW7jAkBEjXJLkQms-JGmrF0BvlpHxA0_H-mVl_42zf0HtA?e=keCf04)
[DiscontinuityPoints](https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/01_Issue/BHoM/BHoM_Engine/Geometry_Engine/%233365-DiscontinuityPointsForPolyCurve.gh?csf=1&web=1&e=g5pYK0)

### Changelog

- Fixed by bug whereby one `Edge` existed in the opposite direction to the `Panel` would caush an error for `PanelToFEMesh`;
- Fixed bug whereby `DiscontinuityPoints()` did not work for `PolyCurve`;
- Updated unit tests for `DiscontinuityPoints()` to include open `PolyCurve`s;
- Replaced unit tests for `FEMeshToPanel` and `PanelToFEMesh` as the data was empty.

Got the branch name wrong 🤐

PR #3364, opened 2024-06-27
Serialiser_Engine: Update NullOrEmptyCheck to just check for Null

### LINKED ISSUE #3357: Serialiser_Engine: Deserialisation of a serialised object with an empty name resulting in a null
#### Description:

The only one still failing locally is the `Cellular` section, and that is a serialisation issue rather than a UT issue.

It is coming from the change to how the name is handled in https://github.com/BHoM/BHoM_Engine/pull/3144 . Where the obejct returned has an empty name, but when serialised and de-serialised it comes back with a null name.

Could potentially set names on deserialization to empty strings rather than null, but that could cause similar issues the other way around. Other otption would be to roll back the change for the name made in the linked PR, or make it onyl skip name if null, rather than NullOrEmpty.

#### Steps to reproduce:

See test file.

#### Expected behaviour:

Unit Test to pass.

#### Test file(s):

https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/01_Issue/BHoM/BHoM_Engine/Serialiser_Engine/Cellular%20Section%20UT%20Fail.gh?csf=1&web=1&e=CYt4pK

@al-fisher, @michaelhoehn, @pawelbaran this is what we discussed in the call.

peterjamesnugent:

Closes #3357

### Test files

[Unit tests](https://burohappold.sharepoint.com/:u:/r/sites/BHoM/02_Current/12_Scripts/01_Issue/BHoM/BHoM_Engine/Serialiser_Engine/Cellular%20Section%20UT%20Fail.gh?csf=1&web=1&e=CYt4pK)
[Overriden objects and non-overriden objects](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4&id=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F02%5FPull%20Request%2FBHoM%2FBHoM%5FEngine%2FSerialiser%5FEngine%2F%233364%2DDeserialisationOfObjectsWithAnEmptyName)

### Changelog

- Updated `NullOrEmptyCheck` in `Serialise` method for `IBHoMObject`s to just check `null` for `Name` properties;
- Updated unit tests for `CellularSectionFromBase` following changes to the `Serialse` method;


peterjamesnugent:
@BHoMBot check compliance

peterjamesnugent:
@BHoMBot check unit-tests

Serialiser_Engine/Compute/Serialise/IBHoMObject.cs
```diff
@@ -58,7 +58,7 @@ private static void Serialise(this IBHoMObject value, BsonDocumentWriter writer,
                 switch (prop.Name)
                 {
                     case "Name":
-                        include = !string.IsNullOrEmpty(value.Name);
+                        include = !(value.Name == null);
```
pawelbaran, line 61:
```suggestion
                        include = value.Name != "";
```

pawelbaran (commented):
Actually, I second to the comment of @IsakNaslundBh - dealing with name may lead to issues in case the default value is overridden (see script [here](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4&id=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F02%5FPull%20Request%2FBHoM%2FBHoM%5FEngine%2FSerialiser%5FEngine%2F%233364%2DDeserialisationOfObjectsWithAnEmptyName), all overrides can be quickly looked up as [here](https://github.com/search?q=org%3ABHoM+%22override+string+Name%22&type=code)). Also in the proposed shape, values equal to `null` get deserialised to empty, so `FromJson != ToJson` (see code comment with a fix and test in script above - however, it still does not address the general problem of overridden defaults).

So to summarise, personally I would rather always serialise the name as an optimal tradeoff between risk/effort and volume/performance (storing an empty name is literally a few characters in total).

peterjamesnugent:
Added the check suggested by @pawelbaran as that solves the majority of cases serialising/deserialising. This does leave a mismatch for overriden properties though as commented above. 

Perhaps we should add an additional check for unit-tests that looks for null names or any string properties that is null so they can be updated rather than have to patch over the `Serialise` and `Deserialise` methods?

pawelbaran:
Are you against removing `case "Name"` from the code altogether @peterjamesnugent? Maybe it got lost in our lengthy comments, but I think both @IsakNaslundBh as well as me opted for such solution - would be good to understand why do you still keep it 👍

pawelbaran (approved):
Happy to approve the code changes, I think what we have now is an optimal trade-off. I see 2 alternatives here:
- allowing for bugs/inconsistencies we are aware of (in case of overridden default) - we ruled it out, not great
- checking whether `value.Name` is a default value of the property - I did quick research and it looks like this would require instantiating a dummy object, which is a massive overhead plus potential risk (think of `IImmutable`)

...so I think what we have now is best value that we can get for now, better alternatives to be rather searched in a separate thread 👍

adecler (approved):
I would follow Ockham on this one. If removing code solves the problem, this is probably a good direction to consider.
More seriously, the exclusion of some properties from json was purely a minor but straightforward optimisation. If it stops being straightforward, then the best approach is indeed to remove it.

Some quick test confirms that this change
 - doesn't have a significant impact on the size of the json (`BHoM_Guid`, `_t`, and `_bhomVersion` alone are already much longer)
 - does maintain the name value even when it is null

image

Although more testing is always good, I don't see any reason to have anymore than this simple one as long as the bot checks are passing so happy to approve.

pawelbaran:
@BHoMBot check required

pawelbaran:
@BHoMBot check unit-tests

pawelbaran:
@peterjamesnugent do you have any clue if the unit tests failure is caused by this PR?

pawelbaran:
Dispensated unit tests check on the basis of the fact that the errors ([here](@BHoMBot this is a DevOps instruction. I am authorising dispensation to be granted on check ref. 26892311972)) are exactly the same as in other BHoM_Engine PRs (e.g. [here](https://github.com/BHoM/BHoM_Engine/pull/3363/checks?check_run_id=26990406140)).

pawelbaran:
@BHoMBot check ready-to-merge

PR #3369, opened 2024-07-04
Geometry_Engine: BooleanDifference secured against single-dimensional regions

pawelbaran:
### Issues addressed by this PR

Closes #3368

### Test files

On [SharePoint](https://burohappold.sharepoint.com/sites/BHoM/02_Current/Forms/AllItems.aspx?FolderCTID=0x0120008122C8891F89054B8ACED0196C70DFC4&id=%2Fsites%2FBHoM%2F02%5FCurrent%2F12%5FScripts%2F02%5FPull%20Request%2FBHoM%2FBHoM%5FEngine%2FGeometry%5FEngine%2F%233369%2DNullReferenceInBooleanOperations).

PR #3371, opened 2024-07-10
Serialiser_Engine: fix for MessageForDeleted to be picked up for properties of types that have multiple matching types

### LINKED ISSUE #3370: Serialiser_Engine: MessageForDeleted is not picked up for properties of types that have multiple matching types
#### Description:

Edge case of an edge case reported under https://github.com/BHoM/BHoM_Engine/issues/3343 - if a property is removed with `MessageForDeleted`, its parent type will version correctly only if there is exactly one matching type for it.

However, it may happen that there is more than one assembly containing that particular type (which currently happens only with Revit stuff, where we have version-specific assemblies), and then the type is not found, leading to the object being deserialised to `CustomObject`.

pawelbaran:

Closes #3370

### DIFF
--- a/BHoM_Engine/Create/Type/Type.cs
+++ b/BHoM_Engine/Create/Type/Type.cs
@@ -78,15 +78,14 @@ namespace BH.Engine.Base
             else if (types.Count == 1)
                 return types[0];
             else if (types.Count > 1 && takeFirstIfMultiple)
-                return types.OrderBy(x => x.Assembly.FullName).First();
-            else if (!silent)
             {
-                string message = "Ambiguous match: Multiple types correspond the the name provided: \n";
-                foreach (Type type in types)
-                    message += "- " + type.FullName + "\n";
+                if (!silent)
+                    Compute.RecordWarning($"Ambiguous match: Multiple types correspond the the name provided:\n{string.Join("\n", types.Select(x => x.FullName))}");
 
-                Compute.RecordError(message);
+                return types.OrderBy(x => x.Assembly.FullName).First();
             }
+            else if (!silent)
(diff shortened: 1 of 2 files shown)


pawelbaran:
@BHoMBot check serialisation
@BHoMBot check versioning

pawelbaran:
@BHoMBot check required

adecler (commented):
Thanks @pawelbaran ,

The change makes sense. As discussed, I would just add a warning message for the case where multiple matching types where found to still provide good traceability: 
image

pawelbaran:
Thanks @adecler for the review, I have now addressed the comment 👍

pawelbaran:
@BHoMBot check required

adecler (approved):
The corrections are matching the request change so happy to approve.

As a side note, the if-else sequence is a bit weird here. After ruling out `count == 1`, there is still a check for `count > 1` but only when `takeFirstIfMultiple` is true. The final case, will just assume there is more than one type. This is correct since the access to the `BHoMTypeDictionary` would have returned false if not types were found. So no bug or error and it was like that before this PR, just a little confusing. So still happy to approve.

pawelbaran:
@BHoMBot check unit-tests
@BHoMBot check ready-to-merge

pawelbaran:
@BHoMBot check unit-tests

pawelbaran:
I will dispensate the failing unit test check on the basis of the fact that it has been failing consistently for a longer period of time, as can be seen [here](https://github.com/BHoM/BHoM_Engine/pull/3364/checks?check_run_id=26892311972) or [here](https://github.com/BHoM/BHoM_Engine/pull/3369/checks?check_run_id=27077776498) - each time the failure message is exactly the same.

PR #3344, opened 2024-05-06
BHoM_Engine: creation of type in case of multiple matching types fixed

pawelbaran:
### Issues addressed by this PR

Closes #3343

PR #3389, opened 2024-08-05
Type matching based on unqualified names enabled in Create.EngineType

### LINKED ISSUE #3385: BHoM_Engine: ExtractAssembly does not capture engine types
#### Description:

...due to an obvious typo `IsOmNamespace` vs `IsEngineNamespace` introduced in https://github.com/BHoM/BHoM_Engine/pull/3341:
https://github.com/BHoM/BHoM_Engine/blob/f1e91381d8918eed3dd173d815f360c0bc9723ae/BHoM_Engine/Compute/ExtractAssembly.cs#L111

FYI @adecler @IsakNaslundBh

### LINKED ISSUE #3386: BHoM_Engine: Create.EngineType only works for matching fully qualified types
#### Description:

Currently `Create.Type` method has the ability to match types based on their unqualified names, i.e. `Type.FullName` rather than `Type.AssemblyQualifiedName`:
https://github.com/BHoM/BHoM_Engine/blob/f1e91381d8918eed3dd173d815f360c0bc9723ae/BHoM_Engine/Create/Type/Type.cs#L57

This is a powerful feature extremely helpful in case of multi-version builds producing many assemblies with same/similar types, e.g. Revit code. The same is needed in `Create.EngineType` to enable versioning of Revit engine assemblies.

FYI @adecler @IsakNaslundBh

pawelbaran:

Closes #3385
Closes #3386

### Test files

Passing `serialisation` and `versioning` checks should be enough of verification.

### Additional comments

Fix for #3385 is almost self-explanatory, while #3386 requires a bit of explanation I believe. The origin of the bug is renaming of Revit assemblies that led me to finding out that `Create.EngineType` matches types based on fully qualified name - after renaming the assemblies the old and new names are different, resulting in a failure. Two steps were required to address that:
1. Enabling matching by unqualified name
2. Enabling picking first type in case of finding multiple matching types, similar to what we did with @IsakNaslundBh in https://github.com/BHoM/BHoM_Engine/pull/3344

### DIFF
--- a/BHoM_Engine/Create/Type/EngineType.cs
+++ b/BHoM_Engine/Create/Type/EngineType.cs
@@ -64,7 +64,7 @@ namespace BH.Engine.Base
             else
             {
                 //Unique method not found in list, check if it can be extracted using the system Type
-                Type type = System.Type.GetType(name, silent);
+                Type type = System.Type.GetType(name, false);
                 if (type == null && !silent)
                 {
                     if (types.Count == 0)


pawelbaran:
@BHoMBot check ready-to-merge

PR #3410, opened 2024-09-04
Diffing_Engine: Do not include static properties and fields in differences

### LINKED ISSUE #3406: Missing skipping static properties for ObejctDifferences for when checking results of unit tests
#### Definition of the test :

In the refactoring to remove duplicate methods done in https://github.com/BHoM/Test_Toolkit/pull/474 to make use of the [ObjectDifferences](https://github.com/BHoM/BHoM_Engine/blob/develop/Diffing_Engine/Query/ObjectDifferences.cs) in the Diffing_Engine, a small detail was missed, which is that the method that was removed ruled out static properties and static fields.

The method in the Diffing_Engine does include these, which has a quite detrimental effect on comparing some system classes, so far, namely the `System.Drawing.Color` class, that has a significant [list of other `System.Drawing.Color` properties](https://github.com/dotnet/runtime/blob/5535e31a712343a63f5d7d796cd874e563e5ac14/src/libraries/System.Drawing.Primitives/src/System/Drawing/Color.cs#L21-L310), which then means that the comparison goes recursive until the max limit of properties to compare is hit.

Like color.Red.Red.Red.Red.Red...........Red.Red.Red.Red.A

To fix this, we need to either add a bool on the BaseComparisonConfig to control if static properties and fields should be included or not, or just always rule them out in the methods in the Diffing_Enigine. First option is probably better, as giving better flexibility.

Found this by realising that the [Trasform UT](https://github.com/BHoM/BHoM_Engine/blob/develop/.ci/Datasets/Graphics_Engine/Modify/Transform.json) in Graphics_Engine had started to take 10 minutes to evaluate, and realised it was the Colors that took 70ms per colour, which ofc adds up when you compare 7k vertecies.

So, to fix I see two options:

1

- Add boolean options for static fields and properties to the base config
- Make use of the booleans in methods in the diffing engine
- Set the value to include to false for the config sent by the Test_Toolkit for CheckTest

2

- Replicate hte previous behaviour of the method in Test_Toolkit that always rules out static properties

IsakNaslundBh:

Closes #3406

Rule out static properties from inclusion in comparison. (see issue for a bit more detail).
Minor cleanup of unused using's.

This reinstates the behaviour for testing prior to this:
https://github.com/BHoM/Test_Toolkit/pull/474

### Test files

Check unit-tests. This should give a not insignificant improvement in speed for running through them, especially for tests containing System.Drawing.Color.

Also, @alelom , great if you can test this through your diffing tests.

### Additional comments

Example run before this PR:

https://github.com/BHoM/BHoM_Engine/runs/29579805351

### DIFF
--- a/Diffing_Engine/Query/DifferentProperties.cs
+++ b/Diffing_Engine/Query/DifferentProperties.cs
@@ -29,7 +29,6 @@ using System.ComponentModel;
 using BH.oM.Base.Attributes;
 using BH.oM.Diffing;
 using BH.oM.Base;
-using kellerman = KellermanSoftware.CompareNetObjects;
 using System.Reflection;
 using BH.Engine.Base;
 
--- a/Diffing_Engine/Query/NumericalDifferenceInclusion.cs
+++ b/Diffing_Engine/Query/NumericalDifferenceInclusion.cs
@@ -29,7 +29,6 @@ using System.ComponentModel;
 using BH.oM.Base.Attributes;
 using BH.oM.Diffing;
 using BH.oM.Base;
-using kellerman = KellermanSoftware.CompareNetObjects;
 using System.Reflection;
 using BH.Engine.Base;
 using BH.Engine.Reflection;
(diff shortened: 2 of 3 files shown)


IsakNaslundBh:
@BHoMBot check unit-tests

IsakNaslundBh:
@BHoMBot check serialisation

alelom (approved):
This works well. When profiling with >1000 elements having static properties (e.g. with 1000 Planes), this gives an improvement of about 50% less time.

Before and after comparison (from the DiffingTests_Prototypes):
image

IsakNaslundBh:
@BHoMBot check required
@BHoMBot check compliance

IsakNaslundBh:
@BHoMBot check ready-to-merge

PR #3447, opened 2024-12-15
Structure_Engine: flipping of bars fixed to take release into account

### LINKED ISSUE #3446: Structure_Engine: Bar.Flip() does not flip the release
#### Description:

As in the title, only end nodes are flipped.

pawelbaran:

Closes #3446

### Test files

On [SharePoint]()

### DIFF
--- a/Structure_Engine/Modify/Flip.cs
+++ b/Structure_Engine/Modify/Flip.cs
@@ -27,6 +27,7 @@ using BH.oM.Base.Attributes;
 using BH.oM.Quantities.Attributes;
 using System.ComponentModel;
 using BH.Engine.Base;
+using BH.oM.Structure.Constraints;
 
 
 namespace BH.Engine.Structure
@@ -50,6 +51,15 @@ namespace BH.Engine.Structure
             Node tempNode = flipped.Start;
             flipped.Start = flipped.End;
             flipped.End = tempNode;
+
+            BarRelease flippedRelease = bar.Release?.ShallowClone();
+            if (flippedRelease != null)
+            {
+                Constraint6DOF tempRelease = flippedRelease.StartRelease;
+                flippedRelease.StartRelease = flippedRelease.EndRelease;
+                flippedRelease.EndRelease = tempRelease;
+                flipped.Release = flippedRelease;
+            }
 
             return flipped;
         }


pawelbaran:
@peterjamesnugent @IsakNaslundBh when working on this one, I started thinking whether the orientation angle should not be negated too? Not entirely sure about it, but a part of me argued for that - happy for you to take over this and make an informed decision 👍

pawelbaran:
@BHoMBot check required

pawelbaran:
@BHoMBot check unit-tests

peterjamesnugent (changes requested):
Happy that the script works as intended, minor change for the code if you think it's worth changing.

I have added a check for a horizontal bar, happy this works for the `BarRelease`.

As discussed offline, the orientational angle should update too, this needs to be set dependent on whether `Bar` is vertical if not:
- If `Bar` is vertical, `newOrientationAngle = -orientationalAngle + pi()`
- If `Bar` is not vertical ,`newOrientationAngle = -orientationAngle` 

I have added a bit in the script to show this.

IsakNaslundBh:
There is a method here: https://github.com/BHoM/ETABS_Toolkit/blob/develop/Etabs_Adapter/CRUD/Create/Bar.cs that potentially could be migrated over. Covers most things I think, but still to be double checked that all things are covered. Mirroring of sections is one thing that might need a bit of attention I think, which is not handled in that method.

pawelbaran:
I feel like I am least informed of us three @IsakNaslundBh @peterjamesnugent - is there a chance you could agree between your duo on how to finalise this PR? To me anything would work 😉

Structure_Engine/Modify/Flip.cs
```diff
@@ -86,6 +120,87 @@ public static Stem Flip(this Stem stem)
...
+        }
+        private static IProfile FlipProfile(TaperFlangeChannelProfile oldProfile)
+        {
+            return Spatial.Create.TaperFlangeChannelProfile(oldProfile.Height, oldProfile.FlangeWidth, oldProfile.WebThickness, oldProfile.FlangeThickness, oldProfile.FlangeSlope,
+                oldProfile.RootRadius, oldProfile.ToeRadius, !oldProfile.MirrorAboutLocalZ);
+        }
+        private static IProfile FlipProfile(IProfile oldProfile)
+        {
+            Base.Compute.RecordError("The given shape profile does not have a FlipProfile method implemented.");
+            return null;
```
IsakNaslundBh, line 202: Would change this to a warning and return the old profile.

Structure_Engine/Modify/Flip.cs
```diff
@@ -86,6 +120,87 @@ public static Stem Flip(this Stem stem)
...
+            return Spatial.Create.FreeFormProfile(curves.Select(x => x.IMirror(Geometry.Create.Plane(new Point(), Vector.XAxis))));
+        }
+        private static IProfile FlipProfile(GeneralisedTSectionProfile oldProfile)
+        {
+            return Spatial.Create.GeneralisedTSectionProfile(oldProfile.Height, oldProfile.WebThickness, oldProfile.RightOutstandWidth, oldProfile.RightOutstandThickness,
+                oldProfile.LeftOutstandWidth, oldProfile.LeftOutstandThickness, oldProfile.MirrorAboutLocalY);
+        }
+        private static IProfile FlipProfile(KiteProfile oldProfile)
+        {
+            return oldProfile;
```
IsakNaslundBh, line 182: If this is problematic, might be good to add a warning about it, or let it fallback to the other method

Structure_Engine/Modify/Flip.cs
```diff
@@ -86,6 +120,87 @@ public static Stem Flip(this Stem stem)
 
             return flipped;
         }
+
```
IsakNaslundBh, line 123: Need method separator lines between all methods `/***************************************************/`

Structure_Engine/Modify/Flip.cs
```diff
@@ -47,10 +58,33 @@ public static Bar Flip(this Bar bar)
...
+                flipped.Release = flippedRelease;
+            }
+
+            // Flip section property
+            if (bar.SectionProperty != null)
+            {
+                ISectionProperty flippedSectionProperty = Flip(bar.SectionProperty);
+                flipped.SectionProperty = flippedSectionProperty;
+            }
+
```
IsakNaslundBh, line 87: Think the offset also need to be handled here.

IsakNaslundBh (changes requested):
See some comments below. Think general work on sections make sense.

One alternative approach to the general filtering for if a profile should be flipped or not, that would be broadening the scope of this though, is to add a new Quary method in spatial engine for IProfile, returning the level of symmetry for the profile. Would then also need to add an enum that is specifying this level of symmetry though, but then you ifstatement could be changed to something like if(profile.ISymmetry() = Symmetry.DoubleSymetric) => return oldprofile.

Not sure if that is widening the scope to much for this though, but just a thought, as I think that potentially could be useful elsewhere as well.

Geometry_Engine/Query/Symmetric.cs
```diff
@@ -0,0 +1,69 @@
...
+        /***************************************************/
+        /**** Public Methods - Curves                   ****/
+        /***************************************************/
+
+        [Description("Checks whether the list of points are symmetric about the given plane.")]
+        [Input("pts", "The list of points to check the symmetry against.")]
+        [Input("p", "The plane to check symmetry about.")]
+        [Input("tolerance", "Distance tolerance to be used in the method.", typeof(Length))]
+        [Output("b", "True if the point is symmetric within tolerance, false if not.")]
+        public static bool Symmetric(this List<Point> pts, Plane p, double tolerance = Tolerance.Distance)
```
IsakNaslundBh, line 46: Would change method name to IsSymmeric or AreSymmetric.

Only not sure, will let you as english primary speaker decide if the subject is the list or the points :)

Spatial_Engine/Query/Symmetric.cs
```diff
@@ -0,0 +1,167 @@
...
+    public static partial class Query
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Determines the symmetry of the given IProfile about the X-Axis and Y-Axis.")]
+        [Input("profile", "The layout object to query the points from.")]
+        [Output("s", "The level of symmetry..")]
+        public static Symmetry ISymmetric(this IProfile profile, double tolerance = Tolerance.Distance)
```
IsakNaslundBh, line 45: As discussed offline, would change method name to Symmetry

Spatial_Engine/Query/Symmetric.cs
```diff
@@ -0,0 +1,167 @@
...
+{
+    public static partial class Query
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Determines the symmetry of the given IProfile about the X-Axis and Y-Axis.")]
+        [Input("profile", "The layout object to query the points from.")]
+        [Output("s", "The level of symmetry..")]
```
IsakNaslundBh, line 44:
```suggestion
        [Output("s", "The level of symmetry.")]
```

Structure_Engine/Modify/Flip.cs
```diff
@@ -47,10 +59,54 @@ public static Bar Flip(this Bar bar)
...
+            // Flip offsets
+            if (bar.Offset != null)
+            {
+                Offset flippedOffset = new Offset()
```
IsakNaslundBh, line 92: Flipped offset is never assigned! 
```suggestion
                flipped.Offset = new Offset()
```

Structure_Engine/Modify/Flip.cs
```diff
@@ -86,6 +144,123 @@ public static Stem Flip(this Stem stem)
...
             return flipped;
         }
+
+        /***************************************************/
+
+        private static ISectionProperty Flip(this ISectionProperty section)
+        {
+            if (section is IGeometricalSection geometricalSection)
+            {
+                ISectionProperty tempSection = null;
```
IsakNaslundBh, line 154: Think we can add a pre-check here, and simply return the section if the profile is doubly symmetric. No need to re-create the same thing.

```suggestion
                if(geometricalSection.SectionProfile.ISymmetric() == Symmetry.DoublySymmetric)
                    return geometricalSection;

                ISectionProperty tempSection = null;
```

IsakNaslundBh (changes requested):
Generally working well, but a few comments.

Geometry_Engine/Query/Symmetric.cs
peterjamesnugent: Updated to `IsSymmetric` - I guess technically should be are because it's plural but `Is` fits in with the convention for query methods. Updated the description too for the output.

peterjamesnugent:
@BHoMBot check compliance

IsakNaslundBh:
@BHoMBot check core

IsakNaslundBh:
@BHoMBot check serialisation

Structure_Engine/Modify/Flip.cs
```diff
@@ -47,15 +58,89 @@ public static Bar Flip(this Bar bar)
...
+
+            }
+            else if (layout is OffsetCurveLayout)
+            {
+                return layout;
+            }
+            else
+            {
+                Engine.Base.Compute.RecordError("ICurveLayout type not recognised.");
+                return null;
```
IsakNaslundBh, line 138: Would jsut return the unflipped verion with the error. Better than null I think

```suggestion
                Engine.Base.Compute.RecordError("ICurveLayout type not recognised. Unflipped curve layout returned");
                return layout;
```

Structure_Engine/Modify/Flip.cs
```diff
@@ -47,15 +58,89 @@ public static Bar Flip(this Bar bar)
...
+            }
+        }
+
+        /***************************************************/
```
IsakNaslundBh, line 142: Method for flipping 2d layouts

```suggestion
        /***************************************************/

        [Description("Flips the ILayout2D for it to be used with a flipped bar. This means flipping the layout around the XZ plane.")]
        [Input("layout", "The ILayout2D to flip.")]
        [Output("layout", "The layout with a flipped.")]
        public static ILayout2D Flip(this ILayout2D layout)
        {
            if (layout.IsNull())
                return null;

            if (layout is ExplicitLayout explicitLayout)
            {
                return new ExplicitLayout(explicitLayout.Points.Select(x => x.Mirror(Plane.YZ)));
            }

            if (layout is LinearLayout linearLayout)
            {
                return new LinearLayout(linearLayout.NumberOfPoints, linearLayout.Direction.Mirror(Plane.XZ), linearLayout.Offset, linearLayout.ReferencePoint.FLip());
            }

            if (layout is MultiLinearLayout multiLayout)
            {
                return new MultiLinearLayout(multiLayout.NumberOfPoints, multiLayout.ParallelMinimumSpacing, multiLayout.PerpendicularMinimumSpacing, multiLayout.Direction.Mirror(Plane.XZ), multiLayout.Offset, multiLayout.ReferencePoint.FLip());
            }

            return layout; //TODO: Implement flipping for other layouts as needed. Might reuire flipping of direction vectors as well as offset points, but for the most common usecase of horizontal rebars, this is not a factor.
        }

        /***************************************************/
```

Structure_Engine/Modify/Flip.cs
```diff
@@ -86,6 +173,253 @@ public static Stem Flip(this Stem stem)
...
+                else if (barReinforcement is LongitudinalReinforcement)
+                {
+                    LongitudinalReinforcement flippedReinforcement = (LongitudinalReinforcement)barReinforcement;
+                    flippedReinforcement.StartLocation = 1 - barReinforcement.EndLocation;
+                    flippedReinforcement.EndLocation = 1 - barReinforcement.StartLocation;
+                    flippedReinforcements.Add(flippedReinforcement);
+                }
+                else
+                    flippedReinforcements.Add(barReinforcement);
+            }
```
IsakNaslundBh, line 236: Start/end handled for all types. Also , previous verison had issues with mutability, leading to bugs with start/end position of the longitudinal (mutates the start, meaning the end gets set in relation to the changed start). 

Also, calling the suggested additional method for Layout2d for flipping those as well.

```suggestion
            foreach (IBarReinforcement barReinforcement in concSection.RebarIntent.BarReinforcement)
            {
                IBarReinforcement flipped = barReinforcement.ShallowClone();
                flipped.StartLocation = 1 - barReinforcement.EndLocation;
                flipped.EndLocation = 1 - barReinforcement.StartLocation;

                if (flipped is TransverseReinforcement transverseReinforcement)
                {
                    transverseReinforcement.CenterlineLayout = transverseReinforcement.CenterlineLayout.Flip();
                    flippedReinforcements.Add(transverseReinforcement);
                }
                else if (flipped is LongitudinalReinforcement longitudinalReinforcement)
                {
                    longitudinalReinforcement.RebarLayout = longitudinalReinforcement.RebarLayout.Flip();
                    flippedReinforcements.Add(longitudinalReinforcement);
                }
                else
                    flippedReinforcements.Add(flipped);
            }
```

Structure_Engine/Modify/Flip.cs
```diff
@@ -86,6 +173,253 @@ public static Stem Flip(this Stem stem)
...
+        /***************************************************/
+
+        private static IProfile FlipProfile(IProfile oldProfile)
+        {
+            Base.Compute.RecordWarning($"The given shape profile {oldProfile.Name} of type {oldProfile.GetType()} " +
+                $"does not have a FlipProfile method implemented, the original profile is returned.");
+            return oldProfile;
+        }
+
+        /***************************************************/
```
IsakNaslundBh, line 421: Required for 2d layout flipping

```suggestion
        /***************************************************/

        private static ReferencePoint FLip(this ReferencePoint referencePoint)
        {
            switch (referencePoint)
            {
                case ReferencePoint.BottomLeft:
                    return ReferencePoint.BottomRight;
                case ReferencePoint.BottomRight:
                    return ReferencePoint.BottomLeft;
                case ReferencePoint.MiddleLeft:
                    return ReferencePoint.MiddleRight;
                case ReferencePoint.MiddleRight:
                    return ReferencePoint.MiddleLeft;
                case ReferencePoint.TopLeft:
                    return ReferencePoint.TopRight;
                case ReferencePoint.TopRight:
                    return ReferencePoint.TopLeft;
                case ReferencePoint.TopCenter:
                case ReferencePoint.MiddleCenter:
                case ReferencePoint.Centroid:
                case ReferencePoint.BottomCenter:
                default:
                return referencePoint;
            }
        }

        /***************************************************/
```

IsakNaslundBh (changes requested):
Some additioanl fixes for rebars suggested below. PLease do have a look.

Also, found a unrelated bug in the spatial engine here:

https://github.com/BHoM/BHoM_Engine/blob/76b8d2021ab7d514833419bb11a15d470802f1ad/Spatial_Engine/Query/PointLayout.cs#L353-L354


This should be 
```
                case oM.Spatial.Layouts.ReferencePoint.MiddleLeft:
                    return new Point { X = bounds.Min.X, Y = (bounds.Min.Y + bounds.Max.Y) / 2 };
```

WOnder if you could fix that as well?

Then, I think we are finally good to go with this monster.

Structure_Engine/Modify/Flip.cs
peterjamesnugent: Agreed - will close.

Structure_Engine/Modify/Flip.cs
peterjamesnugent: Added without modification

Structure_Engine/Modify/Flip.cs
peterjamesnugent: Added without modification - this all makes sense with regards to the `Flip`.

peterjamesnugent:
@BHoMBot check unit-tests

peterjamesnugent:
@BHoMBot check unit-tests

peterjamesnugent:
@BHoMBot check unit-tests

IsakNaslundBh:
@BHoMBot check ready-to-merge

PR #3463, opened 2025-02-19
Serialiser_Engine: Added condition that the serialised value must be null if the deserialised result was null.

### LINKED ISSUE #3462: Serialiser_Engine: Converting from string (or other System type) to IObject does not check that data has been lost
#### Description:

When a property of an object has been converted from a system type (e.g. `System.String`) to an `IObject`, when deserialising from an old version, the property in question (if the old property is not null) deserialises as null, which is a valid value for an `IObject`, so deserialisation carries on as normal. 

There is no check that the original value was not null. This leads to data loss as the deserialiser doesn't have mismatches on property types, so the upgrader for the object is not run, which means that explicit versioning upgrades are not applied unless there is some other issue with the object.

#### Steps to reproduce:

clone the [LadybugTools_Toolkit](https://github.com/BHoM/LadybugTools_Toolkit) repository, checkout `LadybugTools_Toolkit-#296-SeparateSimulationFromPlotting` and build, then clone [Versioning_Toolkit](https://github.com/BHoM/Versioning_Toolkit), checkout the same named branch there and build. Then open the provided grasshopper script.

The `FromJson` method should return a `SimulationResult`, but the `EpwFile` `FileSettings` property will be missing `FileName` and `Directory` information.

#### Expected behaviour:

`FileName` and `Directory` should be set.

#### Test file(s):

Unzip this and make sure that both the json and gh files are in the same folder, then run the gh script:
[serialiser.zip](https://github.com/user-attachments/files/18848105/serialiser.zip)

@IsakNaslundBh

Tom-Kingstone:

Closes #3462 

See the linked issue:
With help from @IsakNaslundBh in isolating the problem, added a condition to `CanSetValueToProperty` which ensures that if a deserialised property is null, then the serialised property must also have been null to pass. 

### Test files

See the linked issue for a script that demonstrated the original problem - There is still an error being recorded (as a value is still being deserialised to null) but the versioning upgrade is happening and the object contains the FileSettings object in a correct form (rather than null/empty)

There will be knock-on effects from this PR on historical versioning, which will likely snowball so testing for this will be ongoing. Running versioning will probably highlight some of these but as @IsakNaslundBh says, much will be missed and it is best to test thoroughly

### DIFF
--- a/Serialiser_Engine/Compute/Deserialise/IObject.cs
+++ b/Serialiser_Engine/Compute/Deserialise/IObject.cs
@@ -117,7 +117,7 @@ namespace BH.Engine.Serialiser
                         {
                             object propertyValue = item.Value.IDeserialise(prop.PropertyType, prop.GetValue(value), version, isUpgraded);
 
-                            if (CanSetValueToProperty(prop.PropertyType, propertyValue))
+                            if (CanSetValueToProperty(prop.PropertyType, propertyValue, item.Value))
                             {
                                 prop.SetValue(value, propertyValue);
                             }
@@ -152,10 +152,10 @@ namespace BH.Engine.Serialiser
 
         /*******************************************/
 
-        private static bool CanSetValueToProperty(Type propType, object value)
+        private static bool CanSetValueToProperty(Type propType, object value, BsonValue bsonValue)
         {
             if (value == null)
-                return !propType.IsValueType || Nullable.GetUnderlyingType(propType) != null;
+                return (!propType.IsValueType || Nullable.GetUnderlyingType(propType) != null) && (bsonValue == null || bsonValue.IsBsonNull);
             else
                 return propType.IsAssignableFrom(value.GetType());
         }
--- a/Serialiser_Engine/Compute/Deserialise/Immutable.cs
+++ b/Serialiser_Engine/Compute/Deserialise/Immutable.cs
@@ -70,8 +70,10 @@ namespace BH.Engine.Serialiser
                     List<object> arguments = new List<object>();
                     foreach (var match in matches)
                     {
-                        object propertyValue = IDeserialise(match.Properties.First().Value, match.Parameter.ParameterType, null, version, isUpgraded);
-                        if (CanSetValueToProperty(match.Parameter.ParameterType, propertyValue))
+                        BsonValue bsonValue = match.Properties.First().Value;
+                        object propertyValue = IDeserialise(bsonValue, match.Parameter.ParameterType, null, version, isUpgraded);
+
+                        if (CanSetValueToProperty(match.Parameter.ParameterType, propertyValue, bsonValue))
                             arguments.Add(propertyValue);
                         else
                         {


Tom-Kingstone:
@BHoMBot check core
@BHoMBot check versioning

IsakNaslundBh:
@BHoMBot check versioning

Tom-Kingstone:
@BHoMBot check unit-tests

Tom-Kingstone:
FYI 
image
for `BHoM_Engine\.ci\Datasets\Physical_Engine\Create\Material.json`

IsakNaslundBh:
To comment on this as well as I was part of the investigation and finding the solution.

By pure principle, I think this change is correct as getting a null back should in principle only be ok if the serialised. Need to investigate if this causes any fallout for various cases and seems like you have already found one case @Tom-Kingstone .

That to me highlights that we need to be quite stringent and really check this through as full as possible.

Given the failing UT-s are mine, I can look into them.

Tom-Kingstone:
> Given the failing UT-s are mine, I can look into them.

@IsakNaslundBh 
I found the cause of the one that I commented about - Basically there is a property in the serialised object "Type" which contains the following: 
```json
{ "_t" : "System.Type", "Name" : "" }
```
Which when run through `FromJson` deserialises as `null`, so if there is a property that contains this, then it is not bson null and therefore fails the check, even though type being null in this case is correct - seems like a fairly sticky one to solve generally, but for specific cases we could just write an upgrader for them.

full json of the object so you can test yourself:
```json
{
    "_t" : "BH.oM.Matter.Options.DensityExtractionOptions",
    "Type" : { "_t" : "System.Type", "Name" : "" },
    "AllowFallbackIfNoType" : false,
    "ExtractionType" : "AllMatching",
    "EqualTolerance" : 0.001,
    "IgnoreZeroValues" : false,
    "ZeroTolerance" : 9.9999999999999995E-07
}

```

IsakNaslundBh:
Did a bit of digging, and the problem with the types is coming from our old serialiser:

https://github.com/BHoM/BHoM_Engine/blob/43ee03db84a135d99ea6871cd813be4d2bc5c399/Serialiser_Engine/Objects/BsonSerializers/TypeSerializer.cs#L52-L68

So, one solution to this might be to add a IsValidNull method accepting a BsonValue, were we can add this and any other edgecases we might find. Hope not to many of them though, but think we at least need to check through the old serialiser to check how things like methods etc are handled in there...

A method, something like:

```
        private static bool IsValidNull(this BsonValue bsonValue)
        {
            if (bsonValue == null || bsonValue.IsBsonNull)
                return true;

            //Handle case for old type serialiser storing nulltypes by setting name to empty string
            if (bsonValue.IsBsonDocument)
            { 
                BsonDocument bsonDocument = (BsonDocument)bsonValue;
                if (bsonDocument.Contains("_t") && bsonDocument.Contains("Name"))
                {
                    BsonValue typeValue = bsonDocument["_t"];
                    BsonValue nameValue = bsonDocument["Name"];
                    if (typeValue.IsString && nameValue.IsString)
                    {
                        BsonString typeString = (BsonString)typeValue;
                        BsonString nameString = (BsonString)nameValue;
                        if (typeString.Value == "System.Type" && string.IsNullOrEmpty(nameString.Value))
                            return true;
                    }
                }
            }

            return false;
        }
```

Might be a start - to be expanded upon if need be!

IsakNaslundBh:
@Tom-Kingstone had a brief chat with @adecler about this, and think this might require a bit of a zoomout before we just patch with the current PR.

If ok with you, will try to tackle this next sprint, making sure we set it up in a way that hopefully covers even more cases.

Tom-Kingstone:
added parked status awaiting end of testing sprint

PR #3498, opened 2025-05-07
Verification_Engine: fixed TryGetValueFromSource for nested properties

michal-pekacki:

Closes #

Improve TryGetValueFromSource to work with specific cases of nested properties:
- if there's more than one property of the same name in the object, get one without indexing
- if property has no getter, check property from the base type

### DIFF
--- a/Verification_Engine/Compute/TryGetValueFromSource.cs
+++ b/Verification_Engine/Compute/TryGetValueFromSource.cs
@@ -250,7 +250,7 @@ namespace BH.Engine.Verification
 
         private static PropertyInfo GetProperty(object obj, Type type, string sourceName)
         {
-            var properties = type.GetProperties().Where(p => p.Name == sourceName).ToList();
+            var properties = type.GetProperties().Where(p => p.Name == sourceName);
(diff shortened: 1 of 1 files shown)


Verification_Engine/Compute/TryGetValueFromSource.cs
```diff
@@ -247,6 +248,19 @@ private static Output<bool, object> TryGetValueFromSource(this object obj, strin
 
         /***************************************************/
 
+        private static PropertyInfo GetProperty(object obj, Type type, string sourceName)
+        {
+            var properties = type.GetProperties().Where(p => p.Name == sourceName).ToList();
```
pawelbaran, line 253: I do not think you need to cast `ToList` here, it is just an overhead (which may be costly in case of reflection being called many times)

michal-pekacki: Fixed 👍

pawelbaran (approved):
Changes make sense to me, happy to approve

michal-pekacki:
@BHoMBot check ready-to-merge

PR #3504, opened 2025-05-22
Serialiser_Engine: Improve serialisation for dynamic objects

### LINKED ISSUE #3503: Improve serialisation for dynamic objects
#### Description:
Have the dynamic properties sitting at the top level of the json as for the regular properties.

adecler:

Closes #3503

The makes sure that the dynamic properties are elevated at the top level of the json alongside any other regular properties.

One thing I would pay extra attention to is the way enums are turned into text. Right now, I am using `IToText()` but I could see benefit to use `ToString()` instead`. It will be good to have your opinion from the perspective of the documentation and json schemas.

### Test files
I would recommend you use your own test you have for LCA toolkit. But if you want additional tests, you can use this:
[DynamicObjects_Test (3).zip](https://github.com/user-attachments/files/20389556/DynamicObjects_Test.3.zip)

### DIFF
--- a/Serialiser_Engine/Compute/Serialise/IBHoMObject.cs
+++ b/Serialiser_Engine/Compute/Serialise/IBHoMObject.cs
@@ -45,6 +45,11 @@ namespace BH.Engine.Serialiser
                 writer.WriteNull();
                 return;
             }
+            else if (value is IDynamicObject)
+            {
+                SerialiseDynamicObject(value as IDynamicObject, writer, targetType);
+                return;
+            }
 
             writer.WriteStartDocument();
 
--- a/Serialiser_Engine/Compute/Serialise/IDynamicObject.cs
+++ b/Serialiser_Engine/Compute/Serialise/IDynamicObject.cs
@@ -1,0 +1,112 @@
+/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2025, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 
+ * GNU Lesser General Public License for more details.                          
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using BH.Engine.Base;
+using BH.Engine.Reflection;
+using BH.oM.Base;
+using BH.oM.Base.Attributes;
+using MongoDB.Bson.IO;
+using System;
+using System.Collections;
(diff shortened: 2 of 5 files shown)


Serialiser_Engine/Compute/Serialise/IDynamicObject.cs
```diff
@@ -0,0 +1,112 @@
...
+                        {
+                            writer.WriteName(prop.Name);
+                            ISerialise(prop.GetValue(value), writer, prop.PropertyType);
+                            BH.Engine.Base.Compute.RecordWarning($"A dynamic property container of type {prop.PropertyType.ToString()} is not supported. It will be serialised as a regular property.");
+                        }
+                        else
+                        {
+                            Type valueType = genericArguments[1];
+                            IDictionary dic = prop.GetValue(value) as IDictionary;
+                            foreach (Enum key in dic.Keys.OfType<Enum>())
```
IsakNaslundBh, line 90: Detail that has 0 impact on functionality, but just could make the JSON look a bit better
```suggestion
                            foreach (Enum key in dic.Keys.OfType<Enum>().OrderBy(x => x))
```

A bit of a trade-off, as this is an additional sorting done for every dynamic object, so would slow it down ever so slightly, but given it effectively is sorting a short collection of integers, I presume it will be quick enough.

Not saying we have to do this, but might be worth in?

Serialiser_Engine/Compute/Serialise/IBHoMObject.cs
```diff
@@ -45,6 +45,11 @@ private static void Serialise(this IBHoMObject value, BsonDocumentWriter writer,
                 writer.WriteNull();
                 return;
             }
+            else if (value is IDynamicObject)
+            {
+                SerialiseDynamicObject(value as IDynamicObject, writer, targetType);
```
IsakNaslundBh, line 50: Dispatching off here makes dynamic IBHoMObjects not get the treatment of skipping the empty base BHoMObjet properties that all other BHoMObejcts get:

Example json for [AcidificationMetric](https://github.com/BHoM/BHoM/blob/LifeCycleAssessment_Toolkit-%23306-AddCalculatedMaterialLifeCycleEnvironmentalImpactFactors/LifeCycleAssessment_oM/MaterialFragments/EnvironmentalMetrics/AcidificationMetric.cs) below:

```
{
    "_t": "BH.oM.LifeCycleAssessment.MaterialFragments.AcidificationMetric",
    "A4": 31.0,
    "B3": 82.0,
    "BHoM_Guid": "6ad82160-ed7f-4be9-a0cf-8a5d112233aa",
    "Name": "",
    "Fragments": [],
    "Tags": [],
    "CustomData": {},
    "_bhomVersion": "8.2"
}

```

As can be seen here, `Fragments`, `Tags` and `CustomData` all present, while would not be for a non-dynamic object.

This is not really critical though, more nice if it actually functioned as other BHoMObjects.

Not sure how to best resolve it in a clean way though, without copy pasting the exception rules from here over to the dynamic object serialiser as well.

IsakNaslundBh (commented):
All now functioning really well in general from the tests I have run.

Also generally happy with the code. Think the approach for the deserialization with the `RecoverDynamicObject` is a great solution!

Have found 2 more nitty gritty things to comment on. Both are more nice to haves, rather than absolutely critical, but think both worth to actually fix. 

The issue about skipping over base properties for dynamic IBHoMObjects is the more annoying one as I am not sure what the cleanest way to do that is really.

IsakNaslundBh:
@BHoMBot check unit-tests

IsakNaslundBh (approved):
Happy with how this now is functioning. Tested with LCA_Toolkit objects, which all work as expected. Also tested with a dummy IDynamicPropertyProvider, which also all works as expected.

Code looks good, and strategy for handling both the regular dynamic objects as well as the property providers for serialisation as well as deserialisation makes sense.

Happy for this to be merged in.

IsakNaslundBh:
@BHoMBot check ready-to-merge

PR #3518, opened 2025-06-13
Versioning_Engine: Fix Issue with all dictionary entries throwing and catching exceptions

### LINKED ISSUE #3517: Versioning_Engine: Constant exception throwing for instances in Dictionaries in Upgrade loop
#### Description:

Lines:

https://github.com/BHoM/BHoM_Engine/blob/76b8d2021ab7d514833419bb11a15d470802f1ad/Versioning_Engine/Modify/Upgrade.cs#L208-L213

Always crash when a dictionary item is sent into the loop.

This was added in https://github.com/BHoM/Versioning_Toolkit/pull/208

THink this is better resolved by directly calling the UpgradeObejctProperties method for dicitonary instances, as that is all that will be called anyway, as that is the only method in here that does not rely on _t being set:

https://github.com/BHoM/BHoM_Engine/blob/76b8d2021ab7d514833419bb11a15d470802f1ad/Versioning_Engine/Modify/Upgrade.cs#L205-L238

#### Steps to reproduce:

Upgrade an object with a dictionary in it. Currently throws an exception that is being caught for every instance in a dictioanry.

#### Expected behaviour:

For this to be handled more gracefully.

IsakNaslundBh:

Closes #3517

Fixes problem with exceptions being thrown and caught for each entry in a non-string based dictionary going through the upgrading loop.

Could have fixed by checking if the document contains "_t" in the UpgradeObejct method, but this almost felt clearer in terms of data flow, as the UpgradeObejctProperties is the only method in UpgradeObejct that will be called for a non-existing _t (all other parts relies on the "oldType" property.

This PR should not really change anything it terms of functionality, just make it run quicker, and hopefully make the code flow cleaner

### DIFF
--- a/Versioning_Engine/Modify/Upgrade.cs
+++ b/Versioning_Engine/Modify/Upgrade.cs
@@ -60,7 +60,7 @@ namespace BH.Engine.Versioning
             }
             else if (newDoc.Contains("k") && newDoc.Contains("v"))
             {
-                result = UpgradeObjectProperties(newDoc, converter);    //Calling UpgradeObjectProperties directly here, as that is the only part of UpgradeObejct method that is relevant. All other parts require _t to be set
+                result = UpgradeObjectProperties(newDoc, converter);    //Calling UpgradeObjectProperties directly here, as that is the only part of UpgradeObject method that is relevant. All other parts require _t to be set
             }
 
             return result;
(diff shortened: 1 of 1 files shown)


adecler:
@BHoMBot check compliance
@BHoMBot check required
@BHoMBot check unit-tests

adecler (approved):
Happy with the code changes and with all checks are passing

adecler:
@BHoMBot check ready-to-merge

PR #3522, opened 2025-06-17
Default message for stack trace for verisoning

### LINKED ISSUE #3521: Versioning_Engine: Stop recording stack trace for VersioningEvents
#### Description:

For cases where there is a massive amoutn of versioning happening, the AllEvents log is clogged by versioning events. After investigation the StackTrace is found to be the biggest culprit. We do not really need the stack trace for versioning, as that is implicit from the event type. Hence, replacing the stack trace with "No stack trace for versioning" will help remedy this issue.

Positive side effect of this is a quite significant speedup in versioning, as each and every object being upgraded no longer mean the full stack being unwound.

IsakNaslundBh:

Closes #3521

Hardcoded string for stack trace for versioning events. This is to drastically reduce RAM usage for cases where a lot of versioning is done. Also gives a massive speedup for files with a lot of versioning.

### DIFF
--- a/Versioning_Engine/Convert/ToNewVersion.cs
+++ b/Versioning_Engine/Convert/ToNewVersion.cs
@@ -136,7 +136,8 @@ namespace BH.Engine.Versioning
                     NewDocument = newDocument,
                     OldVersion = oldVersion,
                     NewVersion = newVersion,
-                    Message = message
+                    Message = message,
+                    StackTrace = "No stack trace for versioning"
                 });
             }
             return document;


adecler (approved):
Very happy with the simplicity of that solution.
Tested with @IsakNaslundBh on a call and it significantly reduces the memory usage and computational time.

IsakNaslundBh:
Code search on all UIs show that this property is never extracted in any of them:

https://github.com/search?q=repo%3ABHoM%2FBHoM_UI+StackTrace&type=code
https://github.com/search?q=repo%3ABHoM%2FGrasshopper_UI+StackTrace&type=code
https://github.com/search?q=repo%3ABHoM%2FExcel_UI+StackTrace&type=code

The only place it is used in the Engine is here:

https://github.com/BHoM/BHoM_Engine/blob/3f2e6f382b5e34b3979168aae127e2ed1a33abe9/BHoM_Engine/Convert/ToText.cs#L238-L253

Where the current message will be appropriate

PR #3532, opened 2025-07-09
Geometry engine #3531 add functionality to fillet a polycurve

### LINKED ISSUE #3531: Geometry_Engine: Add functionality to fillet a polycurve
#### Description:

Add functionality in `BH.Engine.Geometry.Modify` to fillet a polycurve with a given radius at each internal vertex

Chrisshort92:

Add `FilletPolyline` method to Geometry_Engine with `Polyline` and `Radius` inputs

Closes #3531 

### Test files

https://burohappold.sharepoint.com/:f:/r/sites/BHoM/02_Current/12_Scripts/02_Pull%20Request/BHoM/BHoM_Engine/Geometry_Engine/%233532-Geometry%20engine%20%233531%20add%20functionality%20to%20fillet%20a%20polycurve?csf=1&web=1&e=uPCjqA

### DIFF
--- a/Geometry_Engine/Modify/FilletPolyline.cs
+++ b/Geometry_Engine/Modify/FilletPolyline.cs
@@ -91,7 +91,7 @@ namespace BH.Engine.Geometry
                 double lenNext = curr.Distance(next);
 
                 // Improved robust max radius: never allow tangent points to cross neighbor points
-                double maxRadius = Math.Min(lenPrev, lenNext) * Math.Tan(angle / 2.0);
+                double maxRadius = Math.Min(lenPrev/2, lenNext/2) * Math.Tan(angle / 2.0);
                 double usedRadius = Math.Min(radius, maxRadius);
 
                 if (usedRadius < 1e-8)
--- a/Geometry_Engine/Modify/FilletPolyline.cs
+++ b/Geometry_Engine/Modify/FilletPolyline.cs
@@ -1,0 +1,127 @@
+﻿/*
+ * This file is part of the Buildings and Habitats object Model (BHoM)
+ * Copyright (c) 2015 - 2025, the respective contributors. All rights reserved.
+ *
+ * Each contributor holds copyright over their respective contributions.
+ * The project versioning (Git) records all such contribution source information.
+ *                                           
+ *                                                                              
+ * The BHoM is free software: you can redistribute it and/or modify         
+ * it under the terms of the GNU Lesser General Public License as published by  
+ * the Free Software Foundation, either version 3.0 of the License, or          
+ * (at your option) any later version.                                          
+ *                                                                              
+ * The BHoM is distributed in the hope that it will be useful,              
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of               
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the                 
+ * GNU Lesser General Public License for more details.                          
+ *                                                                            
+ * You should have received a copy of the GNU Lesser General Public License     
+ * along with this code. If not, see <https://www.gnu.org/licenses/lgpl-3.0.html>.      
+ */
+
+using BH.Engine.Geometry;
+using BH.oM.Base.Attributes;
+using BH.oM.Geometry;
+using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Linq;
+
+namespace BH.Engine.Geometry
+{
+    public static partial class Modify
+    {
+        /***************************************************/
+        /**** Public Methods - Curves                  ****/
+        /***************************************************/
+
+        [Description("Returns a PolyCurve representing the input Polyline with fillets of the given radius at each internal vertex.")]
+        [Input("polyline", "The Polyline to fillet.")]
+        [Input("radius", "The radius of the fillet arc at each corner.")]
+        [Output("polyCurve", "A PolyCurve with fillets at each corner, or null if unsuccessful.")]
+
+        public static PolyCurve FilletWithTrim(this Polyline polyline, double radius)
+        {
+            if (polyline == null || polyline.ControlPoints == null || polyline.ControlPoints.Count < 3)
+                return null;
+
+            List<ICurve> curves = new List<ICurve>();
+            var pts = polyline.ControlPoints;
+            bool isClosed = polyline.IsClosed();
+            int n = pts.Count;
+
+            int start = isClosed ? 0 : 1;
+            int end = isClosed ? n : n - 1;
+
+            // For each corner
+            for (int i = start; i < end - 1; i++)
+            {
+                int prevIdx = (i - 1 + n) % n;
+                int currIdx = i;
+                int nextIdx = (i + 1) % n;
+
+                Point P0 = pts[prevIdx];
+                Point P1 = pts[currIdx];
+                Point P2 = pts[nextIdx];
+
+                // Directions
+                Vector v1 = (P1 - P0).Normalise();
+                Vector v2 = (P2 - P1).Normalise();
+
+                // Arc start/end and center (simple construction)
+                Point arcStart = P1 + v1 * radius;
+                Point arcEnd = P1 + v2 * radius;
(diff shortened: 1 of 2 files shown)


Geometry_Engine/Modify/FilletPolyline.cs
```diff
@@ -0,0 +1,164 @@
...
+        [Description("Returns a PolyCurve representing the input Polyline with fillets of the given radius at each internal vertex.")]
+        [Input("polyline", "The Polyline to fillet.")]
+        [Input("radius", "The radius of the fillet arc at each corner.")]
+        [Output("polyCurve", "A PolyCurve with fillets at each corner, or null if unsuccessful.")]
+
+        public static PolyCurve Fillet(this Polyline polyline, double radius)
+        {
+            if (polyline == null || polyline.ControlPoints == null || polyline.ControlPoints.Count < 3)
+            {
+                Base.Compute.RecordError("incorrect polyline for filleting");
```
peterjamesnugent: This error message needs to be more descriptive, have proper punctuation and SentenceCase for types e.g. Polyline.

Geometry_Engine/Modify/FilletPolyline.cs
```diff
@@ -0,0 +1,164 @@
...
+                return null;
```
peterjamesnugent: Return the original `Polyline` instead of `null`.

Geometry_Engine/Modify/FilletPolyline.cs
```diff
@@ -0,0 +1,164 @@
...
+{
+    public static partial class Modify
+    {
+        /***************************************************/
+        /**** Public Methods - Curves                  ****/
+        /***************************************************/
+
+        [Description("Returns a PolyCurve representing the input Polyline with fillets of the given radius at each internal vertex.")]
+        [Input("polyline", "The Polyline to fillet.")]
+        [Input("radius", "The radius of the fillet arc at each corner.")]
```
peterjamesnugent:
```suggestion
        [Input("radius", "The radius of the fillet arc at each corner.")]
```
What defines a corner? Does this need an angular input? I assume this initially is for `pi/2` but that's not stated anywhere.

Geometry_Engine/Modify/FilletPolyline.cs
```diff
@@ -0,0 +1,164 @@
...
+            if (double.IsNaN(radius) || radius == 0)
+            {
+                // Convert the input polyline to a PolyCurve
+                var originalCurves = new List<ICurve>();
+                for (int i = 1; i < polyline.ControlPoints.Count; i++)
+                    originalCurves.Add(BH.Engine.Geometry.Create.Line(polyline.ControlPoints[i - 1], polyline.ControlPoints[i]));
+                if (polyline.IsClosed())
+                    originalCurves.Add(BH.Engine.Geometry.Create.Line(polyline.ControlPoints.Last(), polyline.ControlPoints.First()));
+
+                return new PolyCurve { Curves = originalCurves };
```
peterjamesnugent:
```suggestion
                return new PolyCurve { Curves = polyline.SubParts()};
```

Geometry_Engine/Modify/FilletPolyline.cs
```diff
@@ -0,0 +1,164 @@
...
+            bool isClosed = polyline.IsClosed();
+            int ptCount = pts.Count;
+            int segCount = isClosed ? ptCount : ptCount - 1;
+
+            Point lastTrimmedFilletEnd = null;
+
+            for (int i = 0; i < segCount; i++)
+            {
+                int prevIdx = (i - 1 + ptCount) % ptCount;
+                int currIdx = i;
```
peterjamesnugent:
```suggestion
```
Why not just use i?

Geometry_Engine/Modify/FilletPolyline.cs
```diff
@@ -0,0 +1,164 @@
...
+
+            List<ICurve> resultCurves = new List<ICurve>();
+            IList<Point> pts = polyline.ControlPoints;
+            bool isClosed = polyline.IsClosed();
+            int ptCount = pts.Count;
+            int segCount = isClosed ? ptCount : ptCount - 1;
+
+            Point lastTrimmedFilletEnd = null;
+
+            for (int i = 0; i < segCount; i++)
```
peterjamesnugent: Is iterating over the points the best approach? Why not iterate over the `SubParts()`, check the angle with the next line, if it meets the angle threshold you can fillet, otherwise keep it moving.

peterjamesnugent (changes requested):
Few comments on code, haven't tested with the file.

staintono (commented):
Tested with the sample script and my own, functionality is as expected with the curve being filleted at each node. 
- A warning message should show when the radius is limited, to tell the user that the radius is invalid/too larger.
- The warning message: "x and y are not orthogonal. y will be made othogonal to x and z" always shows regardless of the geometry

Geometry_Engine/Modify/FilletPolyline.cs
Chrisshort92: A corner is just the vertex between two straight joined segments. This is why the input is a `Polyline` and not a `Polycurve`.

Chrisshort92:
> Tested with the sample script and my own, functionality is as expected with the curve being filleted at each node.
> 
> * A warning message should show when the radius is limited, to tell the user that the radius is invalid/too larger.
> * The warning message: "x and y are not orthogonal. y will be made othogonal to x and z" always shows regardless of the geometry

1. Warning message added to highlight that radius has been limited to half the shortest segment length
2. This is a warning message from the `DotProduct` function that would need to be changed in that file

Geometry_Engine/Modify/FilletPolyline.cs
Chrisshort92: I get your point, but I feel like having previous, current and next labelled helps to understand the code. Agree a few lines could be saved by removing `currIdx`

Geometry_Engine/Modify/FilletPolyline.cs
Chrisshort92: I guess its the same thing really. Iterating over the points allows prev/next segments to be tested and assists with the end conditions.

Chrisshort92:
@BHoMBot check required

Chrisshort92:
@staintono Can you re-check functionality with a closed polyline

@peterjamesnugent can you check code updates?

Geometry_Engine/Modify/FilletPolyline.cs
peterjamesnugent: So this creates a fillet at each vertex?

I would update the description to use consistent language of vertex rather than corner.

Geometry_Engine/Modify/FilletPolyline.cs
peterjamesnugent: Unless it's used more than once, it does not make sense to store it (from a quick look it's only used once).

Also you can just start your iteration from 1 rather than having the modulus in `prevIndx`.

Geometry_Engine/Modify/FilletPolyline.cs
peterjamesnugent: Disagree, because you're using the previous and next point to get the `Vector` where you can just use the `Direction` of the segment. You then use the `SubPart` to check the radius does not intersect with the previous `subpart` - you can just check the radius is not great than any length of a subpart (otherwise it will overshoot).

Geometry_Engine/Modify/FilletPolyline.cs
Chrisshort92: Corrected term "corner" to "vertex" to be more clear

Chrisshort92:
@BHoMBot check required

Geometry_Engine/Modify/FilletPolyline.cs
Chrisshort92: This has been updated now to iterate over segments and not vertex's as discussed. Large change so new review required.

Chrisshort92:
@BHoMBot check required

PR #3579, opened 2026-04-13
Spatial_Engine: Mirror implemented for IElements

### LINKED ISSUE #3578: Spatial_Engine: Add Mirror method for IElements
#### Description:

As in the title, missing methods as noticed by @ssiimmoonnee

pawelbaran:

Closes #3578

### Test files

On [SharePoint](https://burohappold.sharepoint.com/sites/BHoM/_layouts/15/SkySyncRedir.aspx?Type=2&ResourceId=42faf04c22944912b06a7fcd1bf47c81&View=0)

### Additional comments

The code has been built based on the assumption that all transformations (translation, rotation, mirror, shear) should default down to `Transform` methods being called from within. This is exactly what is being done in this PR, requirement for transform methods being loosened to allow pure reflection.

Following the above, there is no need for extensive tests because literally all spatial mirrors will call `Transform` - if the latter is fine, we're good.


Geometry_Engine/Create/TransformMatrix.cs
```diff
@@ -94,6 +94,74 @@ public static TransformMatrix IdentityMatrix()
...
+            double len = Math.Sqrt(n.X * n.X + n.Y * n.Y + n.Z * n.Z);
+            if (len == 0)
+            {
+                BH.Engine.Base.Compute.RecordError("Cannot create mirror matrix: plane normal is zero.");
+                return null;
+            }
+
+            double nx = n.X / len;
+            double ny = n.Y / len;
+            double nz = n.Z / len;
```
peterjamesnugent: Could we not use the `Modify.Normalise()` method? Would still need to split out the components.

https://github.com/BHoM/BHoM_Engine/blob/develop/Geometry_Engine/Modify/Normalise.cs

Structure_Engine/Modify/Transform.cs
peterjamesnugent, line 102: Think we need to deal with assymmetric sections - effectively they need to be flipped using their `MirrorAboutZ` property:

Image

peterjamesnugent (changes requested):
One minor comment on using the existing `Vector` methods and one larger one on making sure the `Mirror` follows through for asymmetric profiles (similar case for `Pile` and anything that has an `IProfile` property.

Also, are we not supporting `IElement2D`?

pawelbaran:
Thanks for the review @peterjamesnugent, I've addressed all comments, also expanded the test script - would be great if you could have another look and address two things:
- correctness of the current solution
- whether `Bar.Flip()` is equal to mirroring against the plane with origin at bar center and normal parallel to bar tangent

...once this is done, I will update `Flip` method (if 2nd question receives positive answer), then create/update the tests and (hopefully) we will be ready to proceed with merging. Thanks!

pawelbaran:
@BHoMBot check compliance

peterjamesnugent (dismissed):
- Script works as intended;
- We can raise a seperate issue if we want to update `Flip` to make use of `Transform`
- For the `transformMatrix` creation I think you had a reference for how that's derived, can we include that as `//comment` for future reference?

pawelbaran:
@BHoMBot check compliance
@BHoMBot check unit-tests

peterjamesnugent (dismissed):
Approving after compliance fixes.

pawelbaran:
@BHoMBot check unit-tests

pawelbaran:
@BHoMBot check unit-tests

pawelbaran:
@BHoMBot check required

peterjamesnugent (approved):
Rechecked using the test script and happy the mirror works for `Node`, `Bar`, `IProfile`, `Offset`, `BarRelease`, `Panel` (facade and structural) and their respective `CoordinateSystem`.

peterjamesnugent:
@BHoMBot check ready-to-merge

PR #3589, opened 2026-05-31
make the code more robust for deserialisation of Bitmaps and loading of settings

adecler:
### NOTE: used by
https://github.com/BHoM/BHoM_UI/pull/551
https://github.com/BHoM/Grasshopper_UI/pull/751
https://github.com/BHoM/Excel_UI/pull/435
   
 - The new custom ribbon functionality created in https://github.com/BHoM/BHoM_UI/pull/551 was a bit fragile due missing try-catches and null checks.
 - I also made sure the correct package for System.Drawing.Common was copied to the assembly folder since the one that is currently copied is a stub.

### Test files
Excel UI was the most affected by those changes so I recommend testing the loading of the  custom ribbon settings (which includes deseralisation of bitmaps too) in that UI.

### DIFF
--- a/Serialiser_Engine/Compute/Deserialise/Bitmap.cs
+++ b/Serialiser_Engine/Compute/Deserialise/Bitmap.cs
@@ -58,7 +58,15 @@ namespace BH.Engine.Serialiser
             }
 
             var stream = new MemoryStream(binaryData.Bytes);
-            return new Bitmap(stream);
+            try
+            {
+                return new Bitmap(stream);
+            }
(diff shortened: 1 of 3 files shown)


michaelhoehn (approved):
I have tested successfully in Rhino 8 and 7 and am very happy with the functionality. So from that perspective I'm happy to merge this. As noted in my review of the https://github.com/BHoM/Excel_UI/pull/435 I'm unable to view the ribbon within Excel, however I can still verify through other UIs.

PR #3604, opened 2026-08-27
Create RoundToStep.cs and update Ceiling/Floor methods

### LINKED ISSUE #3603: Add RoundToStep feature for `Double`
#### Description:

Add RoundToStep feature to support ConnectionForces modules in Orca Structural Toolbox. Round, Ceiling, Floor (away from zero) at a chosen step interval, not just the next integer. 

(Step = 50, use ceiling, 1329.99 = 1350, or -29 = -50 as ceiling is away from zero).

Chrisshort92:
### NOTE: Depends on 

https://github.com/BHoM/BHoM/pull/1730
   
Closes #3603 

Updates methods to Ceiling & Floor rounding. Rounds a value to the nearest multiple of a step, operating on the magnitude so the sign is preserved: Ceiling always grows the magnitude (away from zero) and Floor always shrinks it (towards zero), regardless of whether the value is positive or negative. Unit-agnostic — the caller may feed any consistent unit.

### Changelog

- `Query.RoundToCeiling()` Query method changed in the `BHoM_Engine` for `double` class
- `Query.RoundToFloor()` Query method changed in the `BHoM_Engine` for `double` class


BHoM_Engine/Query/RoundToCeiling.cs
```diff
@@ -57,6 +76,12 @@ public static double RoundToCeiling(this double number, double tolerance)
             if (number == 0 || tolerance == 0 || Double.IsNaN(tolerance) || Double.IsNaN(number) || Double.IsInfinity(number) || Double.IsInfinity(tolerance))
                 return number;
 
+            if (preserveMagnitude)
+            {
+                double sign = number < 0 ? -1.0 : 1.0;
+                return sign * tolerance * Math.Ceiling(Math.Abs(number) / tolerance);
+            }
+
```
peterjamesnugent, line 86: Rather than adding another override, could we not just check if the number is negative instead?

So could just be multiplying through by `Math.Sign(number)` or keeping that in an `if` statement.

This way we save an additional override. From reading the description I don't think it was clear whether negative numbers are rounded to positive infinity (it only gives a rounding down example from 0.012) so I don't think it changes the method.

```
            return Math.Sign(number)*tolerance * Math.Ceiling(number / tolerance);
```

or

```
            return number < 0 ? Math.Sign(number)*tolerance * Math.Ceiling(number / tolerance); : tolerance * Math.Ceiling(number / tolerance);
```

Chrisshort92:
@BHoMBot check required

BHoM_Engine/Compute/RoundToStep.cs
```diff
@@ -0,0 +1,62 @@
...
+                return value;
+
+            switch (mode)
+            {
+                case RoundingMode.Ceiling:
+                    return value.RoundToCeiling(step, true);
+                case RoundingMode.Floor:
+                    return value.RoundToFloor(step, true);
+                default:
+                    return Query.Round(value, step);
```
peterjamesnugent: I think it's clear the rounding being done from the method name - probably not needed in the BHoM_Engine but at the toolkit level if useful.

BHoM_Engine/Query/RoundToFloor.cs
```diff
@@ -57,6 +76,12 @@ public static double RoundToFloor(this double number, double tolerance)
             if (number == 0 || tolerance == 0 || Double.IsNaN(tolerance) || Double.IsNaN(number) || Double.IsInfinity(number) || Double.IsInfinity(tolerance))
                 return number;
 
+            if (preserveMagnitude)
+            {
+                double sign = number < 0 ? -1.0 : 1.0;
+                return sign * tolerance * Math.Floor(Math.Abs(number) / tolerance);
+            }
+
```
peterjamesnugent, line 86: See comment on `RoundToCeiling` for simplifying this.

peterjamesnugent (changes requested):
I would remove the two extension methods and just extend the `RoundToCeiling` to round towards negative infinity without adding an additional input (original description of the method doesn't clarify this). We should update the description too.

Similar change for `RoundToFloor`.

The `RoundToStep` can sit at a toolkit level if useful enough, but I think when implementing them the in the BHoM_Engine it's clear what the different methods are doing.

PR #3605, opened 2026-08-28
Matter_Engine: Add material classification class

IsakNaslundBh:
### NOTE: Depends on 

https://github.com/BHoM/BHoM/pull/1731
   
- Add method for extracting MaterialClassification
- Add methods for filtering by classification (or parts of it)
- Add method for extracting a distinct set of available classification properties.

### Test files

See https://github.com/BHoM/BHoM/pull/1731


IsakNaslundBh:
@BHoMBot check serialisation

Matter_Engine/Query/FilterByClassification.cs
```diff
@@ -0,0 +1,172 @@
...
+            if(classification == null || other == null)
+                return false;
+
+            //Ignore case when comparing classification properties
+            //This is to avoid issues with different casing in the classification properties, e.g. "Concrete" vs "concrete"
+            StringComparison comparison = StringComparison.OrdinalIgnoreCase;
+
+            if(!string.IsNullOrWhiteSpace(other.Category))
+            {
+                if(!string.Equals(classification.Category, other.Category, comparison))
```
peterjamesnugent: Is it worth trimming the strings here?

For example `CEM 1 ` returns nothing in test script whereas `CEM 1` does.

peterjamesnugent (changes requested):
Test script works, did some `null` checks and happy with those.

When returning the `AvailableXXXX` would it be worth ordering the list? Currently, it's just in the order the datasets are presented which is not always easy to use:

Image

Matter_Engine/Query/FilterByClassification.cs
IsakNaslundBh: Yeah, think you might be right. Was considering igonring whitespace completly as well.

Might even go for that so `CEM 1`,  ` CEM 1`, `CEM 1 `, `CEM1`,  (and all other permutations) all works.

Cannot really think of a case when it would be an issue, so might just do that.

IsakNaslundBh:
> Test script works, did some `null` checks and happy with those.
> 
> When returning the `AvailableXXXX` would it be worth ordering the list? Currently, it's just in the order the datasets are presented which is not always easy to use:
> 
> Image

Yeah, happy to sort all of them

IsakNaslundBh:
THanks @peterjamesnugent 
Have fixed both the issues you raised now. Ready for a re-review!

Matter_Engine/Query/FilterByClassification.cs
```diff
@@ -0,0 +1,173 @@
...
+                return false;
+
+            //Ignore case when comparing classification properties
+            //This is to avoid issues with different casing in the classification properties, e.g. "Concrete" vs "concrete"
+            //The comparisons below also ignore any whitespaces to make sure for example `CEM 1` matches to `CEM1` as well as `CEM 1 `
+            StringComparison comparison = StringComparison.OrdinalIgnoreCase;
+
+            if(!string.IsNullOrWhiteSpace(other.Category))
+            {
+                if(!string.Equals(classification.Category.Replace(" ", string.Empty), other.Category.Replace(" ", string.Empty), comparison))
```
peterjamesnugent:
```suggestion
                if(!string.Equals(new string(classification.Category.Where(c => !char.IsWhiteSpace(c)).ToArray()), new string(other.Category.Where(c => !char.IsWhiteSpace(c)).ToArray()), comparison))
```
The `Replace` doesn't capture any tab spaces - so this would cover all types of whitespace.

peterjamesnugent (commented):
One last change to the trimming to capture all whitespace - otherwise happy.

IsakNaslundBh:
> One last change to the trimming to capture all whitespace - otherwise happy.

Done!

Added a method to handle it, and made a slightly more efficient way for it (asking gpt to give a hgigh performance method for it, and using that instead)

IsakNaslundBh:
@BHoMBot check compliance

IsakNaslundBh:
@BHoMBot check required

IsakNaslundBh:
@BHoMBot check unit-tests

peterjamesnugent (approved):
Changes addressed and tested the tab spacing in the test script.

peterjamesnugent:
@BHoMBot check ready-to-merge

PR #3610, opened 2026-09-06
Geometry_Engine: Add Fillet method for Polylines

### LINKED ISSUE #3609: Add Fillet function for polylines
#### Description:

Add feature to apply fillet to polylines, returning a polycurves with fillets generated by 3-Point Arcs.

Chrisshort92:

Closes #3609 

Rounds every internal corner of a Polyline with a circular arc of the given radius, trimming each adjacent segment back to make room. Corners that can't take the requested radius (too sharp, too close to an end, or running straight through) are capped or left sharp rather than dropped, so the output is always a single contiguous `PolyCurve`.

### Test files

https://burohappold.sharepoint.com/:f:/r/sites/BHoM/02_Current/12_Scripts/02_Pull%20Request/BHoM/BHoM_Engine/Geometry_Engine/%233610-add-fillet-function-for-polylines?d=w4aa8a6b5b8874c5f89f2c8358676f604&csf=1&web=1&e=nMEp3F

image

### Changelog

-`Modify.Fillet()` added to `Geometry_Engine` to round corners of `polylines`, providing the new `polycurve`

### Additional comments

`Offset.cs` already has a private `Fillet(ICurve, ICurve, ...)` helper, used internally to reconnect two offset curve segments that have drifted apart. Despite the name, it doesn't round anything: it extends or trims the two existing curves (a Line stays a Line, an Arc keeps its original radius) until they meet at their mutual intersection, producing a sharp, mitred joint. No radius is involved and no new curve is created.

This PR's `Fillet(this Polyline, double radius, ...)` does something different: given one Polyline and a target radius, it trims each segment back from every corner and inserts a new circular Arc of that radius, tangent to both sides, producing an actually rounded corner and a single continuous curve.

The two share a name by coincidence — the private method carries a stale TODO: make it public comment that this PR does not fulfil as they are target different functions. They solve unrelated problems (sharp-joining two disconnected curves vs. rounding one polyline's corners with a chosen radius), so the private method is left in place rather than replaced.


Chrisshort92:
@BHoMBot check required

Chrisshort92:
The finding: the expected values in these datasets were generated under .NET Framework and are now being checked on .NET 6, and the harness stores and compares them in a way that can't absorb the difference. That bites twice. Where a test's output is a string containing a number, .NET Core 3.0 changed the default double.ToString() from 15 significant digits to shortest-round-trippable, so the identical double now prints as 6.000262860000012 instead of the stored 6.00026286000001 and the exact string comparison fails. Where the output is numeric, the harness compares with an absolute NumericTolerance = 1e-12 against mm-scale coordinates up to 7.3e6 — magnitudes at which one ULP of a double is already 9.3e-10, so the tolerance is finer than a single bit and the test effectively demands bit-exact reproduction across runtimes. 51 of the 72 numeric failures are 1–3 ULP apart, i.e. ordinary JIT/runtime variation that no code change can fix. Adding Fillet.cs is incidental: the failures reproduce byte-for-byte with it absent.

The one caveat: the runtime switch explains the string diffs and the sub-3-ULP numeric ones, but not the 21 larger differences (PlaneIntersections expecting -1022.77 and getting 1.50). Those are too big to be precision artefacts — likely list ordering — and are a genuine open question for the geometry owners, just not one this PR caused.

IsakNaslundBh:
> The finding: the expected values in these datasets were generated under .NET Framework and are now being checked on .NET 6, and the harness stores and compares them in a way that can't absorb the difference. That bites twice. Where a test's output is a string containing a number, .NET Core 3.0 changed the default double.ToString() from 15 significant digits to shortest-round-trippable, so the identical double now prints as 6.000262860000012 instead of the stored 6.00026286000001 and the exact string comparison fails. Where the output is numeric, the harness compares with an absolute NumericTolerance = 1e-12 against mm-scale coordinates up to 7.3e6 — magnitudes at which one ULP of a double is already 9.3e-10, so the tolerance is finer than a single bit and the test effectively demands bit-exact reproduction across runtimes. 51 of the 72 numeric failures are 1–3 ULP apart, i.e. ordinary JIT/runtime variation that no code change can fix. Adding Fillet.cs is incidental: the failures reproduce byte-for-byte with it absent.
> 
> The one caveat: the runtime switch explains the string diffs and the sub-3-ULP numeric ones, but not the 21 larger differences (PlaneIntersections expecting -1022.77 and getting 1.50). Those are too big to be precision artefacts — likely list ordering — and are a genuine open question for the geometry owners, just not one this PR caused.

quick comment on this:

It is a known issue, that ofc needs to be resolved properly. In the meantime, though, if you re-create the unittests in RHino 7 (or set GH for rhino 8 to run in framework) and reserialise them, then they should be passing on the bot. To get this PR merged that will be unfortunately be required.

peterjamesnugent:
Covered in the documentation here:
https://bhom.xyz/documentation/Guides-and-Tutorials/Coding-with-BHoM/Testing/Data-Driven-Tests/#storing-test-data-for-engine-methods

Chrisshort92:
@BHoMBot check required

Martian42 (dismissed):
Image
Successfully replicate the intended output using the test script. Happy to merge.

Geometry_Engine/Modify/Fillet.cs
```diff
@@ -0,0 +1,325 @@
...
+{
+    public static partial class Modify
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Fillets a Polyline by inserting a circular arc at each internal vertex; the end points of an open Polyline stay sharp. The trim taken from either side of a corner is capped at half the adjacent segment, so an arc never runs past a segment midpoint. A corner that cannot be rounded - a joint running straight through or doubling back, or one where the requested radius leaves no room - is left sharp rather than dropped, so the result is always a single contiguous curve.")]
+        [Input("polyline", "Polyline to fillet.")]
+        [Input("radius", "Target fillet radius (> 0). The radius achieved at a corner reduces where the trim-length cap bites.")]
```
peterjamesnugent: > reduces where the trim-length cap bites.

Can we reword this to be clearer and more verbose.

Geometry_Engine/Modify/Fillet.cs
```diff
@@ -0,0 +1,325 @@
...
+        [Input("distTol", "Distance tolerance.")]
+        [Input("angleTol", "Angle tolerance (radians). A joint is left sharp where it comes within this of running straight through, or of doubling back on itself.")]
```
peterjamesnugent:
```suggestion
        [Angle]
        [Input("angleTol", "Angle tolerance - a joint is left sharp where it comes within this tolerance of running straight through, or of doubling back on itself.")]
```
I think this could be made clearer too.

Chrisshort92:
@BHoMBot copyright-compliance

Chrisshort92:

Chrisshort92:

Chrisshort92:
@BHoMBot check unit-tests

Geometry_Engine/Modify/Fillet.cs
```diff
@@ -0,0 +1,325 @@
...
+            return FilletedCurve(vertices, trim, arcs, segCount, closed, distTol);
+        }
+
+        /***************************************************/
+        /**** Private Methods                           ****/
+        /***************************************************/
+
+        // Fraction of a segment's own length that its two trims may occupy, on a segment too short for
+        // the distTol margin to leave any room at all.
+        private const double ShortSegmentFallbackFraction = 0.99;
```
peterjamesnugent: Would this not be better as an input with a default value?

Geometry_Engine/Modify/Fillet.cs
```diff
@@ -0,0 +1,325 @@
...
+                // predecessor, a there-and-back jitter passes both tests and leaves two retained vertices
+                // closer together than distTol.
+                if (vertices.Count == 0 || pt.Distance(vertices[vertices.Count - 1]) > distTol)
+                    vertices.Add(pt);
+            }
+
+            // A closed polyline repeats its first point last. Drop that repeat, and any further vertex
+            // that has collapsed onto the first, so the closing segment is no shorter than the rest.
+            while (closed && vertices.Count > 1 && vertices[0].Distance(vertices[vertices.Count - 1]) <= distTol)
+                vertices.RemoveAt(vertices.Count - 1);
```
peterjamesnugent: Could you not just check for `closed` at the start and remove the last `pt`?

Your `for` then covers anything that is too close.

Geometry_Engine/Modify/Fillet.cs
```diff
@@ -0,0 +1,325 @@
...
+            bool closed = polyline.IsClosed(distTol);
+
+            List<Point> vertices = FilletVertices(pts, closed, distTol);
+            if (vertices.Count < 3)
+            {
+                Base.Compute.RecordError("The Polyline has fewer than three distinct vertices, so it has no corner to round.");
+                return null;
+            }
+
+            // A closed polyline carries a segment from its last vertex back to its first; an open one does not.
```
peterjamesnugent, line 85: Is this comment still relevant given what you do in `FilletVertices`?

Geometry_Engine/Modify/Fillet.cs
```diff
@@ -0,0 +1,325 @@
...
+        // vertex, and zero where no fillet applies. This owns every trim decision: the cap from the
+        // requested radius, the balancing of the two trims a segment carries, and the final check that
+        // a segment keeps some length between them.
+        private static double[] FilletTrimLengths(List<Point> vertices, int segCount, bool closed, double radius, double distTol, double angleTol)
+        {
+            int nVerts = vertices.Count;
+
+            double[] segLengths = new double[segCount];
+            for (int s = 0; s < segCount; s++)
+                segLengths[s] = vertices[s].Distance(vertices[(s + 1) % nVerts]);
```
peterjamesnugent, line 139: Mighth be worth a comment what is going on here, especially the `(s + 1) % nVerts`

Geometry_Engine/Modify/Fillet.cs
peterjamesnugent:
```suggestion
        [Input("radius", "Target fillet radius (> 0). The radius achieved at a corner reduces where the trim-length cap bites.", typeof(Length))]
        [Input("distTol", "Distance tolerance.", typeof(Length))]
        [Input("angleTol", "Angle tolerance. A joint is left sharp where it comes within this of running straight through, or of doubling back on itself.", typeof(Angle))]
```

Geometry_Engine/Modify/Fillet.cs
```diff
@@ -0,0 +1,325 @@
...
+                Vector v1 = vertices[(i - 1 + nVerts) % nVerts] - corner;
+                Vector v2 = vertices[(i + 1) % nVerts] - corner;
+
+                double len1 = v1.Length();
+                double len2 = v2.Length();
+                if (len1 < distTol || len2 < distTol)
+                    continue;
+
+                v1 /= len1;
+                v2 /= len2;
```
peterjamesnugent: You could just normalise here:
https://github.com/BHoM/BHoM_Engine/blob/acc84d2c835c23cdd24d515c9c2df55fb4dc1333/Geometry_Engine/Modify/Normalise.cs#L41

Geometry_Engine/Modify/Fillet.cs
```diff
@@ -0,0 +1,325 @@
...
+
+                // Clamped so that rounding on near-parallel vectors cannot push Acos outside its domain.
+                double d = Math.Max(-1.0, Math.Min(1.0, v1.DotProduct(v2)));
+                double theta = Math.Acos(d);
+
+                // There is nothing to round on a joint that runs straight through, or that doubles back
+                // on itself.
+                if (theta < angleTol || Math.Abs(Math.PI - theta) < angleTol)
+                    continue;
```
peterjamesnugent: I think this logic - along with the unitised vector above can be replaced with the `Query.Angle` method:
https://github.com/BHoM/BHoM_Engine/blob/acc84d2c835c23cdd24d515c9c2df55fb4dc1333/Geometry_Engine/Query/Angle.cs#L40-L52

Geometry_Engine/Modify/Fillet.cs
```diff
@@ -0,0 +1,325 @@
...
+        // The arc rounding a single corner, tangent to both adjacent segments at the trim distance from
+        // the corner, or null where no valid arc exists.
+        private static Arc FilletArc(Point prev, Point corner, Point next, double trim, double distTol)
+        {
+            Vector v1 = (prev - corner).Normalise();
+            Vector v2 = (next - corner).Normalise();
+
+            // Clamped so that rounding on near-parallel vectors cannot push Acos outside its domain.
+            double d = Math.Max(-1.0, Math.Min(1.0, v1.DotProduct(v2)));
+            double theta = Math.Acos(d);
```
peterjamesnugent: Similar to above, you can use `Query.Angle` here:
https://github.com/BHoM/BHoM_Engine/blob/acc84d2c835c23cdd24d515c9c2df55fb4dc1333/Geometry_Engine/Query/Angle.cs#L40-L52

Geometry_Engine/Modify/Fillet.cs
```diff
@@ -0,0 +1,325 @@
...
+            return trim;
+        }
+
+        /***************************************************/
+
+        // The fillet arc at each vertex, or null where the vertex is not filleted or its arc could not be
+        // built. Resolving every arc before any curve is assembled is what keeps the result contiguous:
+        // assembly treats a null arc as an untrimmed corner, so a corner can never be cut back and then
+        // left open by an arc that failed.
+        private static Arc[] FilletArcs(List<Point> vertices, double[] trim, double distTol)
```
peterjamesnugent: Is it necessary to have an additional method that just calls `FilletArc`? Can't you just implement that in a `for` loop instead?

Geometry_Engine/Modify/Fillet.cs
```diff
@@ -0,0 +1,325 @@
...
+
+        /***************************************************/
+
+        // The polyline's vertices with coincident points collapsed, so that every returned vertex is
+        // further than distTol from both its neighbours - the closing pair of a closed polyline included.
+        // Everything downstream relies on that: it is what makes every segment long enough to trim and
+        // build on, and so what keeps the assembled curve contiguous.
+        // Not Modify.RemoveShortSegments, which cannot offer that guarantee here: it restores the
+        // original end point of an open polyline, so the last segment may come back shorter than the
+        // tolerance, and it re-closes a closed one, whereas the wrap is handled by segCount below.
```
peterjamesnugent: Is this still valid?

Geometry_Engine/Modify/Fillet.cs
```diff
@@ -0,0 +1,325 @@
...
+
+namespace BH.Engine.Geometry
+{
+    public static partial class Modify
+    {
+        /***************************************************/
+        /**** Public Methods                            ****/
+        /***************************************************/
+
+        [Description("Fillets a Polyline by inserting a circular arc at each internal vertex; the end points of an open Polyline stay sharp. The trim taken from either side of a corner is capped at half the adjacent segment, so an arc never runs past a segment midpoint. A corner that cannot be rounded - a joint running straight through or doubling back, or one where the requested radius leaves no room - is left sharp rather than dropped, so the result is always a single contiguous curve.")]
```
peterjamesnugent: This can be broken with `\n` for readability.

Geometry_Engine/Modify/Fillet.cs
```diff
@@ -0,0 +1,325 @@
...
+
+            double[] trim = FilletTrimLengths(vertices, segCount, closed, radius, distTol, angleTol);
+            Arc[] arcs = FilletArcs(vertices, trim, distTol);
+
+            return FilletedCurve(vertices, trim, arcs, segCount, closed, distTol);
+        }
+
+        /***************************************************/
+        /**** Private Methods                           ****/
+        /***************************************************/
```
peterjamesnugent, line 100:
```suggestion
        /***************************************************/
        /**** Private Fields                             ****/
        /***************************************************/
```

Geometry_Engine/Modify/Fillet.cs
```diff
@@ -0,0 +1,325 @@
...
+            {
+                Point startPoint = output.First().IStartPoint();
+                Point endPoint = output.Last().IEndPoint();
+                if (startPoint.Distance(endPoint) > distTol)
+                    output.Add(BH.Engine.Geometry.Create.Line(endPoint, startPoint));
+            }
+
+            return new PolyCurve { Curves = output };
+        }
+
```
peterjamesnugent, line 319:
```suggestion
        }
/***************************************************/

```

Geometry_Engine/Modify/Fillet.cs
```diff
@@ -0,0 +1,325 @@
...
+            // in FilletTrimLengths only rules that out while the caller leaves angleTol at its default.
+            if (Math.PI - theta < Tolerance.Angle)
+                return null;
+
+            double R = trim * Math.Tan(theta / 2.0);
+            if (R <= distTol)
+                return null;
+
+            // Non-zero because the sweep check above holds theta away from pi.
+            Vector bis = (v1 + v2).Normalise();
```
peterjamesnugent:
```suggestion
            Vector bisector = (v1 + v2).Normalise();
```
Can we be verbose with the names to add context when reviewing later.

Geometry_Engine/Modify/Fillet.cs
```diff
@@ -0,0 +1,325 @@
...
+            double d = Math.Max(-1.0, Math.Min(1.0, v1.DotProduct(v2)));
+            double theta = Math.Acos(d);
+
+            // The arc sweeps pi - theta. Create.ArcByCentre throws out of CartesianCoordinateSystem
+            // rather than returning null when asked for a sweep below Tolerance.Angle, and the joint test
+            // in FilletTrimLengths only rules that out while the caller leaves angleTol at its default.
+            if (Math.PI - theta < Tolerance.Angle)
+                return null;
+
+            double R = trim * Math.Tan(theta / 2.0);
```
peterjamesnugent:
```suggestion
            double r = trim * Math.Tan(theta / 2.0);
```
Variable names should be lower case - what is r in this context?

Geometry_Engine/Modify/Fillet.cs
```diff
@@ -0,0 +1,325 @@
...
+                if (theta < angleTol || Math.Abs(Math.PI - theta) < angleTol)
+                    continue;
+
+                // The segment leaving vertex i is segment i, and the one arriving is segment i-1. Capping
+                // at half of each keeps an arc clear of the corner at the far end of either segment.
+                double tDesired = radius / Math.Tan(theta / 2.0);
+                double prevHalf = 0.5 * segLengths[(i - 1 + segCount) % segCount];
+                double nextHalf = 0.5 * segLengths[i];
+
+                double t = Math.Min(tDesired, Math.Min(prevHalf, nextHalf));
```
peterjamesnugent: What is t in this context? I would rename t to a clearer variable name or add a comment. Giving it a clearer variable name is preferable.

peterjamesnugent (changes requested):
Comments below.

In your test script this fillet is only 1.719 whilst the input is 1.83:
Image

Can you clean up the comments in the file please - some seem like reasoning and should be cleaned up for merging.

Also, does your test script/unit test cover most cases? Is there one for covering where the radius is greater than the line segment? What about short segments where the points are near co-incident?

Chrisshort92:
> Comments below.
> 
> In your test script this fillet is only 1.719 whilst the input is 1.83: Image
> 
> Can you clean up the comments in the file please - some seem like reasoning and should be cleaned up for merging.
> 
> Also, does your test script/unit test cover most cases? Is there one for covering where the radius is greater than the line segment? What about short segments where the points are near co-incident?

Method used to terminate at half a segments length, which the case you highlight is doing. I have now edited the method to terminate once two arc meet, you can see this play out below.

image

image

Geometry_Engine/Modify/Fillet.cs
Chrisshort92: Keeping it private. It's a fallback for a segment too short for the tolerance margin to leave any room, not a design parameter — there's no physical meaning a user could set it from. Same treatment as m_ToleranceMultiplier in Query/HashArray.cs. Expanded the comment to say so.

Geometry_Engine/Modify/Fillet.cs
Chrisshort92: The `foreach` only ever compares a point with its predecessor, never with `vertices[0]`. If `P(n-1)` is within `distTol` of `P0` but not of `P(n-2)`, it survives, and the closing segment comes back shorter than `distTol`. The `while` collapses that chain, not just the duplicate. Reworded the comment to make that the stated reason.

Geometry_Engine/Modify/Fillet.cs
Chrisshort92: Still relevant, but only because `FilletVertices` strips the repeated closing point. Reworded to say that explicitly.

Geometry_Engine/Modify/Fillet.cs
Chrisshort92: adopted, good point.

Geometry_Engine/Modify/Fillet.cs
Chrisshort92: superseded by comment below

Geometry_Engine/Modify/Fillet.cs
Chrisshort92: adopted

Geometry_Engine/Modify/Fillet.cs
Chrisshort92: Adopted, but only the angle part. `v1`/`v2` are still used below for the bisector and tangent points, so the two Normalise() calls stay.

Geometry_Engine/Modify/Fillet.cs
Chrisshort92: updated!

Geometry_Engine/Modify/Fillet.cs
Chrisshort92: adopted

Geometry_Engine/Modify/Fillet.cs
Chrisshort92: done

Geometry_Engine/Modify/Fillet.cs
Chrisshort92: updated section and t

Chrisshort92:
> Comments below.
> 
> In your test script this fillet is only 1.719 whilst the input is 1.83: Image
> 
> Can you clean up the comments in the file please - some seem like reasoning and should be cleaned up for merging.
> 
> Also, does your test script/unit test cover most cases? Is there one for covering where the radius is greater than the line segment? What about short segments where the points are near co-incident?

On the test cases, yes, we cover each case.

1. Each vertex has its redius
2. Some vertexes max out
3. Input is open polyline + self intersecting
4. Input is closed polyline + self intersecting
5. Input has 1 short segment, less than distTol, no radius there
6. Input has 1 shallow vertex, less than angleTol, no radius there
7. Input has 2 colinear segments, works but no radius as less than angleTol
8. Input as 2 vertex (line), fails
9. Input is Polycurve, fails
10. Input is Circle, fails

Chrisshort92:
@BHoMBot check all

Chrisshort92:
@BHoMBot check required

Chrisshort92:
@BHoMBot check unit-tests

Chrisshort92:
Method updated to only ever fillet a corner at the specified radius. If that radius cannot be achieved, the fillet at that corner does not take place and the vertex is listed in a warning. 

image

image

Chrisshort92:
@BHoMBot check required

Martian42 (dismissed):
The test script has been checked, and fillet is only created when given radius can be achieved, ignored when it is not possible. This is working as intended from the latest commit.

Image

Chrisshort92:
@BHoMBot check unit-tests

Chrisshort92:

Chrisshort92:

Geometry_Engine/Modify/Fillet.cs
```diff
@@ -0,0 +1,321 @@
...
+        public static PolyCurve Fillet(this Polyline polyline, double radius,
+            double distTol = Tolerance.Distance, double angleTol = Tolerance.Angle)
+        {
+            if (polyline == null)
+            {
+                Base.Compute.RecordError("Cannot fillet a null Polyline.");
+                return null;
+            }
+
+            IList<Point> pts = polyline.ControlPoints;
```
peterjamesnugent:
```suggestion
            List<Point> pts = polyline.ControlPoints;
```
This doesn't need to be an IList.

Geometry_Engine/Modify/Fillet.cs
```diff
@@ -0,0 +1,321 @@
...
+            // What the radius costs at each corner, and zero where the vertex is no candidate for filleting.
+            double[] required = new double[nVerts];
+
+            for (int i = 0; i < nVerts; i++)
+            {
+                if (!closed && (i == 0 || i == nVerts - 1))
+                    continue;
+
+                Point corner = vertices[i];
+                Vector v1 = vertices[(i - 1 + nVerts) % nVerts] - corner;
```
peterjamesnugent, line 152: Might be worth a comment that these go in opposite directions so the angle calculation is the internal angle.

peterjamesnugent (changes requested):
Minor changes - I've tested using the script as well as:
- Did a `Rotate3D` to test in different planes;
- Tested with a simple right angle/radius when looking at distance tolerance and angle tolerance;
- Tested with some other cases to test the tolerances and I think it works as expected

Chrisshort92:
@BHoMBot check required

peterjamesnugent (dismissed):
Changes addressed - happy to merge

Chrisshort92:
@BHoMBot check unit-tests

Chrisshort92:

Chrisshort92:

Chrisshort92:
@BHoMBot check ready-to-merge

Chrisshort92:
@BHoMBot check ready-to-merge

IsakNaslundBh (commented):
Reviewed this in a call with @peterjamesnugent  and think it seem to be working well.

Only thing I would change is the below.

Will just put as a comment review so that this can be re-approved without having to rely on me once this has been fixed.

Geometry_Engine/Modify/Fillet.cs
```diff
@@ -0,0 +1,322 @@
...
+                return null;
+            }
+
+            if (double.IsNaN(radius))
+            {
+                Base.Compute.RecordError("The fillet radius is not a number, so no corner has been rounded.");
+                return null;
+            }
+
+            if (radius <= 0)
```
IsakNaslundBh, line 68: Would simply return the polyline (as a polycurve) for the case of a 0 radius. The <0 check can stay as is (returning null) as that is an invalid input.

Especially in the light of returning the initial curve if a to large radius is given.

So, essentially, `if (radius == 0)` return the initial curve, and raise a warning that no filleting has been performed.

IsakNaslundBh (commented):
Missed the comment

Geometry_Engine/Modify/Fillet.cs
Chrisshort92: image
adopted this now.

Chrisshort92:
@BHoMBot check required

peterjamesnugent (changes requested):
Tested this, and the change has been implemented.

Unit tests needs to be added for 0 radii.

Chrisshort92:
@BHoMBot check required

Chrisshort92:
@BHoMBot unit-tests

Chrisshort92:

Chrisshort92:

Chrisshort92:
@BHoMBot check unit-tests

Chrisshort92:
@BHoMBot check ready-to-merge

peterjamesnugent (approved):
Changes addressed - happy to merge.

peterjamesnugent:
@BHoMBot check ready-to-merge

